PR #23, opened 2014-03-17
TLEN clarification to SAM spec

jrandall:
States clearly that TLEN should be length (i.e. pos.right-pos.left+1) rather than subtraction 
(i.e. pos.right-pos.left) -- this was somewhat ambiguous before. 

States that TLEN may be set to 0 whenever the information is unavailable, 
and gives examples as to when this would be the case.

Adds explicit statement that soft-clipped bases don't count as mapped bases.


yfarjoun:
superceded by #366

PR #27, opened 2014-03-26
Alternative TLEN changes: admit it is implementation-specific

jrandall:
Provides an alternative to TLEN by allowing it to be a signed integer with 
arbitrary semantics that are application-specific, which better captures the 
real-world situation (if one finds a BAM/SAM file, there is no way to know 
how exactly to interpret the TLEN field without knowing which application 
generated the data). 

Changes the name of the field from: "Template LENgth (TLEN)" 
to: "Implemenation-Specific Integer, Zero Else (ISIZE)" 
which has an added benefit of being consistent with the API/internal naming of most 
extant BAM parsing libraries (e.g. samtools uses 'isize', Picard uses 'ISIZE_COL'), 
which have never converted from ISIZE to TLEN. 

Defines the value `0' as the default if an application does not use ISIZE, which is 
backwards-compatible with the undefined/unknown value TLEN. 

Also adds a tag 'IS' that can be used by future implementations to specify the 
semantics of what the ISIZE field is used for, although it's open for discussion 
whether it would be more appropriate to define this in the header (e.g. in @RG 
or @HD).


nh13:
@jmarshall should we propose to add the "implementation specific" part of this pull request?

yfarjoun:
superceded by #366

Issue #48, opened 2014-10-22
SAM: Clarification of primary vs secondary reads


jkbonfield:
When we have secondary reads, do we need to have TLEN defined?

The recommended practices section under "multiple mappings" states: RNEXT and PNEXT point to the primary line of the next read in the template.

This doesn't cover what TLEN should be on secondary reads.  I assume it is computed from primary reads only, but mirrored over to secondary reads, or is it simply that the field is undefined for secondary reads?  The specification doesn't explicitly state that TLEN comes from only primary alignments either, although maybe it is implied in "observed template length" with the assumption that in our observations we'd not look at secondary mappings.

jkbonfield:
I'm trying to write some validation for SAM format and this is, yet again, tripping me up.

Technically I have to produce completely illogical and IMO *wrong* test data if we want to validate what the spec states, but I'm still adamant the spec is plain wrong on this front.

The most extreme example is from these snippets.  Firstly some definitions:

> \item[Segment] A contiguous sequence or subsequence.

> \item[Read] ... A read may consist of multiple segments.

> \item[Linear alignment] ... A linear alignment can be represented in a single SAM record.

> \item[Chimeric alignment] An alignment of a read that cannot be represented as a linear alignment. ...

> For each read/contig in a SAM file, it is required that one and only one line associated with the read satisfies \mbox{`{\sf FLAG} {\tt \& 0x900 == 0}'}. This line is called the \emph{primary line} of the read.

So it's clear here that "segment" is potentially a portion of a read and if we map a read in more than one place, either as supplementary or secondary, then only the non-supp non-secondary read is considered as primary.

However it gets muddied.  RNEXT/PNEXT are based on primary line (IMO wrong, but let's ignore for now).  TLEN is not.  It's "all segments":

> \item {\sf TLEN}: signed observed Template LENgth. If all segments are mapped to the same reference sequence, the absolute value of TLEN equals the distance between the mapped end of the template and the mapped start of the template, inclusively (i.e., $\mbox{end}-\mbox{start}+1$).%

Hence a strictly this means if we include a secondary alignment in another contig, then the alignment records for the primary non-secondary mappings must now have TLEN 0 as it's only valid is "all segments" (not all primary lines) map to the same reference.  Even worse if the secondary alignment happened to map to the same reference as the primary alignment then TLEN must include that secondary map location.

[...]

Issue #44, opened 2014-09-30
SAM: "next" vs "next primary".


jkbonfield:
For RNEXT and PNEXT fields we explicitly state the next primary alignment. It is a bit unclear what the interpretation of these fields on a secondary alignment, but that's not a big issue.

For FLAG field the 0x8 and 0x20 bits mention "next segment" with no qualification of it being the next primary segment.  Clearly we need primary data as if we are looking at the first read in a pair then we use FLAG, RNEXT and PNEXT jointly to obtain information about the "other end". If RNEXT/PNEXT refer to the other end but FLAG may potentially not then these fields lose their usefulness.

I believe the table of bits in 1.4 should be modified as s/next segment/next primary segment/;

PR #53, opened 2014-11-17
Changed rnext/pnext/tlen to operate per template alignment rather than for primary alignments only.

jkbonfield:
This for a discussion point at the next GA4GH conference call, rather than for acceptance immediately.

Also see discussions in issues #44 and #48.

Things still to resolve:

1) TLEN definition has not changed from left/right to 5'/3'. We need to come to a decision once and for all.

2) Should PNEXT/RNEXT/TLEN _always_ be linking together to form circles per template alignment, or should we permit the old method of linking to next (aka "other end" in a 2-segment world) primary alignment for cases where it makes more sense to the aligner.  Ie do we have to double up the other end in such cases?

3) There are no additional auxiliary tags defined here yet; specifically no SI (which secondary alignment number this is within), SC (count of secondary alignments) and AC (total count of all segment alignments in file for this template).  Field names subject to discussion and change of course.

4) No commentary of CC/CP.  These are already in use by novoalign, but do not form a circular linked list meaning that it is not possible to start at any point and do a full round-trip to find all segment alignments for all primary, secondary and supplementary alignment cases.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -118,6 +118,14 @@ r001  147 ref 37 30 9M         =  7 -39 CAGCGGCAT         * NM:i:1
   flags.  Typically the alignment designated primary is the best alignment, but
   the decision may be arbitrary.\footnotemark
 
+\item[Template alignment]
+  A set of read alignments for all reads in the template.  In the case
+  of multiple mappings for a template, multiple template alignments
+  may exist.  All read alignments (whether chimeric or linear) within
+  a template alignment share the same value of the 0x100 flag
+  (secondary alignment).  Template alignments consisting of more than
+  2 segments must have the TC auxiliary tag present.
+
 \item[1-based coordinate system] A coordinate system where the first
   base of a sequence is one. In this coordinate system, a region is
   specified by a closed interval. For example, the region between the
@@ -296,6 +304,7 @@ of the mandatory fields in the SAM format:
     0x40 and 0x80 are unset, the index of the read in the template
     is unknown. This may happen for a non-linear template or the index
     is lost in data processing.
+  \item Bit 0x20 defines the next segment as per the definition in RNEXT and PNEXT.
   \item If 0x1 is unset, no assumptions can be made about 0x2, 0x8,
     0x20, 0x40 and 0x80.
   \end{itemize}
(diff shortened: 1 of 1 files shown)


SAMv1.tex
```diff
@@ -296,6 +304,7 @@ \subsection{The alignment section: mandatory fields}
     0x40 and 0x80 are unset, the index of the read in the template
     is unknown. This may happen for a non-linear template or the index
     is lost in data processing.
+  \item Bit 0x20 defines the next segment as per the definition in RNEXT and PNEXT.
```
d-cameron: I'm not sure what this means. If the next segment is the third segment in a 5 segment template, what does having 0x20 set actually mean? Alignments for segments 2 and 3 are consistent? 3 and 4? All segment alignment are consistent with having no (large) SVs wrt the set of reference alignment?

SAMv1.tex
```diff
@@ -344,28 +353,30 @@ \subsection{The alignment section: mandatory fields}
...
-  read in the template. If {\tt @SQ} header lines are present, {\sf
-    RNEXT} (if not `*' or `=') must be present in one of the {\tt SQ-SN}
-  tag. This field is set as `*' when the information is unavailable, and
-  set as `=' if {\sf RNEXT} is identical {\sf RNAME}. If not `=' and the
-  next read in the template has one primary mapping (see also bit
-  0x100 in {\sf FLAG}), this field is identical to {\sf RNAME} at the primary line of the
-  next read.  If {\sf
-    RNEXT} is `*', no assumptions can be made on {\sf PNEXT} and bit
+\item {\sf RNEXT}: Reference sequence name of the NEXT read in this
+  template alignment, where NEXT is defined to be the next read in
```
d-cameron: So essentially this PR essentially redefines secondary alignments as secondary alignments for the entire template as opposed to the current definition that implicitly (by requiring RNEXT to point to the primary alignment) defines secondary alignments as secondary alignments for the segment.

For read pairs that have one read uniquely mappable and the other read multi-mapping to 100 different location this change in definition will require spec-compliant implementations to write out the uniquely aligned segment 100 times, each with a different RNEXT.

It is a major change that will break many tools as tools can no longer assume that a high mapq means that record is unique. This as simple as calculating read depth of reads with (e.g.) mapq>10 will now require read deduplication. Is this intended?

SAMv1.tex
```diff
@@ -344,28 +353,30 @@ \subsection{The alignment section: mandatory fields}
...
+  template coordinates rather than mapping coordinates.  For the last
+  read, the next read is the first read in this template alignment.
+  Multiple template alignments may exist, with RNEXT/PNEXT forming a
```
d-cameron: Where supp alignment fit in should be explicitly stated. This is an issue with the current specs as it doesn't explicitly state that RNEXT and PNEXT should be to the non-supp primary record.

We should explicitly state that RNEXT and PNEXT never point to supp alignment records.

SAMv1.tex
```diff
@@ -559,8 +570,8 @@ \section{Recommended Practice for the SAM Format}
   \begin{enumerate}[label=\arabic*]
   \item When one segment is present in multiple lines to represent a multiple
 	mapping of the segment, only one of these records should have the secondary
```
d-cameron: Again, the current specs are missing clarification as to what's allowed when the primary alignment is chimeric -
 you can have multiple non-secondary lines, but only one of these can be non-supplementary.

Issue #55, opened 2014-11-25
Request for clearer documentation to explain bitwise FLAG values, ideally with documentation


kbradnam:
If you are new to bioinformatics, and are asked to work with any SAM file, then you might reasonably turn to this documentation to help to better understand how the format works.

I feel that many people have trouble understanding what is meant by bitwise FLAG values. The documentation is very technical and not very transparent to people who may be new to bioinformatics.

Many people might be turning to the documentation after looking at their SAM output file. Maybe they see that their output file has a range of integer values in column 2 and are puzzled by the explanation in the documentation (this is very likely if you have no familiarity with bit patterns). 

I think this section would be greatly helped by the following:
1.  A reminder that the SAM file itself stores an integer value (this is mentioned in the overview section for what all the columns mean, but it is not as obvious as it could be)
2. An explicit description that a bitwise value of zero means that your read has mapped to the forward strand of the reference (t is not intuitive to work this out because the 'segment unmapped' bit has not been set)
3. Some specific examples that explain what various integer values correspond to.

PR #59, opened 2015-01-06
Numbers expanded to individual flags

karel-brinda:
Hello,

for readers of the SAM specification, it may be annoying to manually search binary expansions of integers to get the corresponding flags. So I added the expansions.

Karel

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -72,7 +72,7 @@ r001  147 ref 37 30 9M         =  7 -39 CAGCGGCAT         * NM:i:1
 \end{verbatim}
 \end{framed}
 Remark that
-$\text{99} = \text{0x63} = \text{0x400}|\text{0x20}|\text{0x2}|\text{0x1}$,
+$\text{99} = \text{0x63} = \text{0x40}|\text{0x20}|\text{0x2}|\text{0x1}$,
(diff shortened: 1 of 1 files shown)


nh13:
@jmarshall seems useful, I am ok with this.

karel-brinda:
Fixed.

jmarshall:
[I have some comments to make on this, so please don't merge it in the meantime.]

jmarshall:
IIRC we discussed something related to this at our last file formats meeting, and concluded that the spec was not in the business of teaching binary arithmetic.  Nonetheless, it would be useful to clarify these examples.

Please see e78064fc638475d7b85821d7fb36db64bcea458b — Karel, does that address the things you wanted to address?

For the initial example, I figure it's even more useful to tell the new reader what the flags are supposed to mean.  Conversely I just added hex equivalents for the 5'/middle/3' fragment stuff, as IMHO writing out the full expansion hides the information that the example is trying to convey.

karel-brinda:
Yes, now it is perfect.

jmarshall:
Thanks, I've merged that one now.  I suspect our “not in the business of teaching binary arithmetic” discussion was in response to #55.

PR #62, opened 2015-01-26
Fix typo in Genotype encoding example

corburn:
The NA00002 GQ in the VCF line does not match the encoded equivalent.


pd3:
Indeed, thank you!

Issue #39, opened 2014-08-22
SAM: PLatform field in @RG for Oxford Nanopore


jkbonfield:
The SAM @RG header has a PL attribute for the platform. This needs updating to support more instruments, eg Oxford Nanopore.

Should it be PL:ONT, NANPORE, OXFORD_NANOPORE?  I think I prefer ONT as the easiest yet descriptive one. (There could be other nanopore vendors in time.)

Issue #16, opened 2014-01-29
Add PM tag to @RG header


keiranmraine:
Hi,

As PL is a restricted list could PM, 'Platform Model', be added to the RG header definition?  This would be an unrestricted text field allowing for more detailed information about the machine generating the data.

Regards,
Keiran

Issue #15, opened 2014-01-20
Describe @HD's GO field


keiranmraine:
Could 'collated' be added as a valid sort order for reads?  In many cases it isn't important to fully sort by name only bring pairs together and it would be useful if this was easy to detect from the header.

mdshw5:
Could someone elaborate on what exactly GO stands for?

jmarshall:
From the email message linked above:

> [e.g.] `@HD GO:query`, which is accepted by both samtools and Picard, but I don't think either of them do anything in particular with it.  This was described in earlier versions of the SAM/BAM spec as:
>
>   `GO`    Group order (full sorting is not imposed in a group). Valid values are: none, query or reference.

It's a slight misnomer.  It's really _grouped-by_, akin to the SQL construct.  The spec change mentioned is still "upcoming"...

mdshw5:
Ah, thanks for clicking the link for me :). I hope to see this in the spec at some point, since I try to keep my work as close to the spec as possible.

PR #69, opened 2015-03-03
Collected header additions: HD-GO, RG-PL:ONT, and RG-PM

jmarshall:
1. Reinstate the previously-specified `@HD-GO` tag.  See #15.  The text in the pre-TeX specification was

    > **GO**  Group order (full sorting is not imposed in a group). Valid values: none (default), query, and reference.

    It would be possible to require that if both `SO` and `GO` are specified then they must agree with each other (e.g. `SO:queryname GO:query`), but it seems easier to discourage specifying both together.  (Also perhaps we could have a footnote noting that `GO:reference` is fairly pointless!)

2. Add `ONT` (Oxford Nanopore) as a valid value for `@RG-PL`.  See #39.

3. Add new free-form `@RG-PM` tag to allow more details about the platform to be given, e.g., machine models such as 'HiSeq 2500'.  See #16.

These additions mangle the formatting of the header table, but I propose to tidy that up later when some of the information in the table moves to a new `SAMtags.pdf` (filename tbd) document.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -172,6 +172,11 @@ The following table give the defined record types and tags. Tags with
   minor sort key is the {\sf POS} field.  For alignments with equal {\sf RNAME} and {\sf POS}, order is
   arbitrary.  All alignments with `{\tt *}' in {\sf RNAME} field follow alignments with some other
   value but otherwise are in arbitrary order.\\\cline{1-3}
+  & {\tt GO} & Grouping of alignments, indicating that similar alignment records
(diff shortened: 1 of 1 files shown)


nh13:
I support 1-2.   For 1 we would need to implement the best practice enforcement (ex. ```SO``` and ```GO``` need to match if both are specified).

For 3, this may work out of the box in htsjdk, but to include it into the set of standard tags in ```@RG``` would require some more work, including testing.  This seems like a reasonable addition though.

Issue #32, opened 2014-07-28
Unify VCF spec


jmarshall:
At the moment, there are separate VCFv4.1 and VCFv4.2 specs with very little difference between them -- and it is hard to figure out by reading them just what the difference between them is.

Having two near-duplicate documents also means that typos have to be fixed twice, etc.

It would be better if these were unified into a single document, so that the common things could be seen to be common.  And it would note where there are differences between the point-releases, e.g., it would say that `Number=R` is a 4.2 feature.

cyenyxe:
A changelog has been added at the end of the VCF v4.3 specification, and it has been agreed offline not to merge all the documents into one so I will close this ticket.

PR #72, opened 2015-03-06
Cram v3

vadimzalunin:
RC of CRAM3 specs


jkbonfield:
I think it would be preferable to have cram.tex with different tags on it to mark v2.0, v2.1, v3.0 etc. That way we can compare it easily to see what changed between one version and another.

vadimzalunin:
The other specs have version info in the file names, for example VCFv4.1.tex and VCFv4.2.tex
But I understand that it's a cleaner way, what do others think?

jmarshall:
SAM is currently `SAMv1.pdf` which allows for a complete rewrite in `SAMv2.pdf` should we ever bump the major version number in a seriously-incompatible way — though I find it difficult to imagine that ever happening!

Certainly `SAMv1.pdf` just has the major number in the filename, and (when we eventually record what SAM v1.3, 1.4, 1.5 etc actually represent) I envisage discussing the tiny differences between minor versions all in the one `SAMv1.pdf` document.

IMHO VCF v4.1, 4.2, and 4.3 are similar enough that unifying them into a single `VCFv4.pdf` akin to the SAM one would be clearer for readers, and this is issue #32.  However this was discussed at the September(?) phone meeting, and there was not a lot of support for this view.  Instead IIRC most mildly supported the status quo re VCF.

Currently there is a bare-bones `CSIv1.pdf` document, and the CSIv2 branch adds a `CSIv2.pdf` document with one line added.  I have a draft of a more complete `CSIv1.pdf` document, and when that lands I think the thing to do will be to change the CSIv2 proposal to rename the spec to `CSI.pdf` and add the one line as “_v2 only_” in the unified document.

So there is precedent for both views :smile:.  What I think is that when two versions are substantially the same, it is clearest for the reader for them to be described in a single unified spec that documents the small differences between levels.  I think this is similar to but a little more work (for us) than @jkbonfield's tagging preference above.

And it's up to you guys to determine whether CRAM 2.x and 3.0 are “substantially the same”…

jkbonfield:
On Fri, Mar 06, 2015 at 04:12:16AM -0800, John Marshall wrote:
> And it's up to you guys to determine whether CRAM 2.x and 3.0 are "substantially the same"...

I didn't spot the other formats having their version numbers in the
files, so maybe we should be consistent there.  The differences are
*that* huge, but consistency helps and it's not hard to download both
and run diff.

However if and when v3.1 comes out I would expect it in the same
filename for ease of comparison.


-- 

peterjc:
I would use a single file in version control (with tags), but would very much encourage posting the main different versions as separate PDF files on the website, e.g. SAM/BAM v1.0, 1.1, 1.2, 1.3, 1.5 etc.

vadimzalunin:
I have renamed CRAMv3.0 to CRAMv3

jmarshall:
So there was this pull request with this discussion on it... and now something else (#74) has been merged to master?

vadimzalunin:
yes, I merged the branch with James's commit in it instead of this one. Was it too soon? The issue discussed here is about the file naming which can be fixed in the master branch later.

jmarshall:
The renaming has been cherry-picked as eee31a1d98db04cf9b30d692d9086486ff476bbe and the merge of CRAMv3 into master redone tidily.

Issue #75, opened 2015-03-19
Feature Request: structured data in VCF:INFO (XML, JSON ?)


lindenb:
(cross posted on  http://sourceforge.net/p/vcftools/mailman/message/33615341/  )

I wish I could insert **structured data** in a VCF, like those produced by VEP+REST,  EVS+SOAP , etc... (see above). Would it be possible to store XML||JSON in addition to the already existing types: Integer, Float, Flag, Character, and String. Trying to store the data in a simple string is a nonsense to me

```
CSQ=G|ENSG00000223972|ENST00000456328|Transcript|non_coding_transcript_exon_variant&non_coding_transcript_variant|1413|||||||1|DDX11L1|HGNC|37102|||,G|ENSG00000227232|ENST00000541675|Transcript|downstream_gene_variant|||||||198|-1|WASH7P|HGNC|38034|||,G|653635|NR_024540.1|Transcript|downstream_gene_variant|||||||197|-1|WASH7P||38034|||,G|ENSG00000223972|ENST00000515242|Transcript|non_coding_transcript_exon_variant&non_coding_transcript_variant|1406|||||||1|DDX11L1|HGNC|37102|||,G|ENSESTG00000013896|ENSESTT00000034761|Transcript|downstream_gene_variant|||||||242|-1||||||,G|100287102|NR_046018.2|Transcript|non_coding_transcript_exon_variant&non_coding_transcript_variant|1408|||||||1|DDX11L1||37102|||,G|ENSG00000227232|ENST00000538476|Transcript|downstream_gene_variant|||||||246|-1|WASH7P|HGNC|38034|||,G|ENSG00000223972|ENST00000518655|Transcript|non_coding_transcript_exon_variant&non_coding_transcript_variant|1239|||||||1|DDX11L
```

A problem that would arise would be , how can we store those documents containing spaces.

Pierre

---

VEP+ XML:

``` xml
<?xml version="1.0"?>
<opt>
  <data id="rs148327885" allele_string="C/T" assembly_name="GRCh37" end="878331" input="1 878331 rs148327885 C T . . ." most_severe_consequence="missense_variant" seq_region_name="1" start="878331" strand="1">
[...]

PR #76, opened 2015-03-24
content-type in INFO field used to include complex structured data like JSON or XML

lindenb:
Suggestion for the following issue https://github.com/samtools/hts-specs/issues/75 _"Feature Request: structured data in VCF:INFO (XML, JSON ?)"_

Here I suggest to add an optional field **ContentType** in the INFO header to specify the type of data. By  default it's **text/plain**. Other allowed values are **text/xml** and **application/json** and the value should be URL decoded.

An example with a JSON formatted Ensembl prediction :

```
(...)
##INFO=<ID=PRED,Number=1,Type=String,Description="Ensembl Prediction",ContentType="application/json">
(...)
#CHROM POS ID REF ALT QUAL FILTER INFO
1 873 . C T . . PRED=%5B%7B%22input%22%3A%221%20873%20.%20C%20T%20.%20.%20.%22%2C%22assembly_name%22%3A%22GRCh37%22%2C%22end%22%3A873%2C%22seq_region_n(...)
```

Once URL-decoded, the JSON value of ‘PRED’ is:

``` json
[
{
"allele_string": "C/T",
"assembly_name": "GRCh37",
"end": 873,
"id": null,
"input": "1 873 . C T . . .",
"intergenic_consequences": [
{
"consequence_terms": [
"intergenic_variant"
],
"variant_allele": "T"
}
],
"most_severe_consequence": "intergenic_variant",
"seq_region_name": "1",
"start": 873,
"strand": 1
}
]
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+.PHONY: all pdf mostlyclean clean
 all: pdf
 
 PDFS =	BCFv1_qref.pdf \
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -105,9 +105,11 @@ to be aware of the possible presence of multi-byte UTF-8 characters.
 \subsubsection{Information field format}
 INFO fields should be described as follows (first four keys are required, source and version are recommended):
 
+{\scriptsize
 \begin{verbatim}
-##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="version">
+##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="version",ContentType="mime">
 \end{verbatim}
+}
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -1,3 +1,4 @@
+.PHONY: all pdf mostlyclean clean
```
jmarshall: There are `.PHONY` targets at the bottom of the makefile.  Putting them before the `all` target breaks makes that don't understand `.PHONY`.  The existing `.PHONY` line at the bottom suffices, no?

lindenb: ops I didn't notice.

pd3:
Same as https://github.com/samtools/hts-specs/issues/75, the pull request can be closed

PR #86, opened 2015-05-17
[DOC] fixed description of linear index according to source code

weese:


weese:
I stumbled across `Given a region [rbeg,rend), we only need to visit a chunk whose end file offset is larger than the file offset of the 16kbp window containing rbeg.` It doesn't seem to work if the linear index stores the smallest file offset of alignments that **start** in the window. It should read "**overlap** the window". The for-loop in the function `insert_to_l` in hts.c seems to confirm my assumption.

jmarshall:
Alas, this PR got forgotten about. The error was rediscovered in 2016 and the spec was finally updated in df3bcefaaa06467b375f98ac9da77bc83ffff340.

Issue #78, opened 2015-03-27
Enforcing a standard for reporting REF and ALT allele depths


arq5x:
The current VCF specification does a very admirable job of defining the standards by which variant callers should report genotypes (GT), genotype likelihoods (GL), and overall observed depth (DP) for each sample at a given variant site.  

However, in my opinion, the specification is lacking in that it fails to define a standard for reporting the sequencing depths observed for each sample and allele. This lack of specificity has allowed variant callers to each define their own convention. Consequently, downstream tools are left to write custom code that attempts to handle the rules defined by each variant caller.  

For instance, GATK uses `AD`, which reports allele depths as a comma-separated list (173 for the reference and 141 for the alternate in this example:

    chr1    873762  .       T   G   100 .    AC=1;AF=0.50;DP=315 GT:AD:DP:GQ:PL   0/1:173,141:282:99:255,0,255

Freebayes, however uses two different tags, `RO` and `AO` for reference and alternate depth, respectively. 

    GT:GQ:DP:RO:QR:AO:QA:GL 0/0:0:5:5:132:0:0:0,-0.32169,-4.34301

Other tools such as VARSCAN and Platypus employ yet other conventions.

Given that the inspection of allele depths are are crucial to quality control and data interpretation, I think it would be best to define a standard for this. In my opinion, the GATK convention makes most sense, as it  works for single or multiple alternate alleles, so long as the reference allele is always reported first.

The standard could allow support for either reporting the total depths for each allele:

    GT:AD  0/1:173,141

or optionally allow another delimiter (`;` ?) defining the depth on each strand (positive first) for each allele:

    GT:AD  0/1:100;73,70;71

It would be great if we could make progress towards a standard here, as it would prevent hideous code such as: https://github.com/arq5x/cyvcf/blob/master/cyvcf/parser.pyx#L264-L295

Best,
Aaron Quinlan
USTAR Center for Genetic Discovery
Department of Human Genetics and Biomedical Informatics
University of Utah

lh3:
I have to deal with this DP issue in [my review][1] last year. It was a big headache. I had to treat the VCF generated by each SNP caller as a different format and parses it conditionally (see [this code block][2]). This means that script won't work with new SNP callers or when the existing SNP callers change their tags.

[...]

Issue #54, opened 2014-11-24
Reference contains IUPAC ambiguity codes but not legal in REF or ALT for VCF


mp15:
Affects: VCF 4.2 and prior versions
The reference contains IUPAC ambiguity codes:
    $ samtools faidx ../../ref/GCA_000001405.15_GRCh38_full_analysis_set.fna chr3:16902883-16902883
    >chr3:16902883-16902883
    B

dbSNP contains IUPAC codes because the reference does for example entry rs56708014:

```
chr3    16902883    rs56708014  B   BGC .   .   RS=56708014;RSPOS=16902887;dbSNPBuildID=129;SSR=0;SAO=0;VP=0x050000080005000002000200;WGT=1;VC=DIV;INT;ASP;OTHERKG
```

However the VCF spec says: "REF - reference base(s): Each base must be one of A,C,G,T,N (case insensitive)" and for ALT "Options are base Strings made up of the bases A,C,G,T,N,*, (case insensitive) or an angle-bracketed ID String (“<ID>”) or a breakend replacement string as described in the section on breakends".  Unfortunately this conflict means that InDels or SNPs that contain ref sites with IUPAC codes are not legally representable in VCF.

eitanbanks:
I really don't like adding IUPAC ambiguity codes to the REF.
I'm fine with pretty much any other solution.

pd3:
For those who are not subscribed to the `vcftools-spec` mailing list, here are some of the relevant emails:
http://sourceforge.net/p/vcftools/mailman/message/34096206/
http://sourceforge.net/p/vcftools/mailman/message/34102529/
http://sourceforge.net/p/vcftools/mailman/message/34101323/

The options discussed offline and on the mailing list were these:

**How to treat ambiguity codes in the reference sequence**
1. allow IUPAC codes in REF
2. replace with N
3. replace with one of the bases, for example, R=A/G becomes A
4. both 1 or 2 are possible

**How to treat abiguity codes in ALT**
1. allow IUPAC codes in ALT
2. use symbolic alleles, such as `<R>`,`<S>`, etc

Can people +1/-1 their preferred options?

eitanbanks:
I like options 3 (for REF) and 2 (for ALT).

auton1:
I think allow ambiguity in the REF is asking for trouble.

I vote 3 for REF.

For ALT, I have a slight preference for 2, but don't feel very strongly
about it.


[...]

PR #88, opened 2015-06-02
VCFv4.3 - first batch of changes

pd3:
Please see also the list of open issues here:
https://github.com/samtools/hts-specs/labels/vcf

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,8 @@ PDFS =	BCFv1_qref.pdf \
 	SAMv1.pdf \
 	tabix.pdf \
 	VCFv4.1.pdf \
-	VCFv4.2.pdf
+	VCFv4.2.pdf \
+	VCFv4.3.pdf
 
 pdf: $(PDFS)
 
@@ -15,6 +16,7 @@ CRAMv2.1.pdf: CRAMv2.1.tex CRAMv2.1.ver
 SAMv1.pdf: SAMv1.tex SAMv1.ver
 VCFv4.1.pdf: VCFv4.1.tex VCFv4.1.ver
 VCFv4.2.pdf: VCFv4.2.tex VCFv4.2.ver
+VCFv4.3.pdf: VCFv4.3.tex VCFv4.3.ver
 
 
 .SUFFIXES: .tex .pdf .ver
--- a/VCFv4.1.tex
+++ b/VCFv4.1.tex
@@ -498,7 +498,7 @@ VCF STRUCTURAL VARIANT EXAMPLE
 ##FORMAT=<ID=CN,Number=1,Type=Integer,Description="Copy number genotype for imprecise events">
 ##FORMAT=<ID=CNQ,Number=1,Type=Float,Description="Copy number genotype quality for imprecise events">
 #CHROM POS     ID        REF              ALT          QUAL FILTER INFO                                                               FORMAT       NA00001
-1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;END=2827762;HOMLEN=1;HOMSEQ=G;SVLEN=-68                 GT:GQ        1/1:13.9
+1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;END=2827708;HOMLEN=1;HOMSEQ=G;SVLEN=-14                 GT:GQ        1/1:13.9
 2       321682 .         T                <DEL>        6    PASS   SVTYPE=DEL;END=321887;SVLEN=-205;CIPOS=-56,20;CIEND=-10,62         GT:GQ        0/1:12
 2     14477084 .         C                <DEL:ME:ALU> 12   PASS   SVTYPE=DEL;END=14477381;SVLEN=-297;CIPOS=-22,18;CIEND=-12,32       GT:GQ        0/1:12
 3      9425916 .         C                <INS:ME:L1>  23   PASS   SVTYPE=INS;END=9425916;SVLEN=6027;CIPOS=-16,22                     GT:GQ        1/1:15
--- a/VCFv4.1.tex
+++ b/VCFv4.1.tex
@@ -231,7 +231,7 @@ VCF records use a simple haplotype representation for REF and ALT alleles to des
 \section{INFO keys used for structural variants}
 When the INFO keys reserved for encoding structural variants are used for imprecise variants, the values should be best estimates. When a key reflects a property of a single alt allele (e.g. SVLEN), then when there are multiple alt alleles there will be multiple values for the key corresponding to each alelle (e.g. SVLEN=-100,-110 for a deletion with two distinct alt alleles).
 
-The following INFO keys are reserved for encoding structural variants. In general, when these keys are used by imprecise variants, the values should be best estimates. When a key reflects a property of a single alt allele (e.g. SVLEN), then when there are multiple alt alleles there will be multiple values for the key corresponding to each alelle (e.g. SVLEN=-100,-110 for a deletion with two distinct alt alleles).
+The following INFO keys are reserved for encoding structural variants.
 \footnotesize
 \begin{verbatim}
 ##INFO=<ID=IMPRECISE,Number=0,Type=Flag,Description="Imprecise structural variation">
@@ -261,7 +261,7 @@ One value for each ALT allele. Longer ALT alleles (e.g. insertions) have positiv
 ##INFO=<ID=BKPTID,Number=.,Type=String,Description="ID of the assembled alternate allele in the assembly file">
 \end{verbatim}
 \normalsize
-For precise variants, the consensus sequence the alternate allele assembly is derivable from the REF and ALT fields. However, the alternate allele assembly file may contain additional information about the characteristics of the alt allele contigs.
+For precise variants, the consensus sequence of the alternate allele assembly is derivable from the REF and ALT fields. However, the alternate allele assembly file may contain additional information about the characteristics of the alt allele contigs.
 \footnotesize
 \begin{verbatim}
 ##INFO=<ID=MEINFO,Number=4,Type=String,Description="Mobile element info of the form NAME,START,END,POLARITY">
--- a/VCFv4.2.tex
+++ b/VCFv4.2.tex
@@ -515,7 +515,7 @@ VCF STRUCTURAL VARIANT EXAMPLE
 ##FORMAT=<ID=CN,Number=1,Type=Integer,Description="Copy number genotype for imprecise events">
 ##FORMAT=<ID=CNQ,Number=1,Type=Float,Description="Copy number genotype quality for imprecise events">
 #CHROM POS     ID        REF              ALT          QUAL FILTER INFO                                                               FORMAT       NA00001
-1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;END=2827762;HOMLEN=1;HOMSEQ=G;SVLEN=-68                 GT:GQ        1/1:13.9
+1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;END=2827708;HOMLEN=1;HOMSEQ=G;SVLEN=-14                 GT:GQ        1/1:13.9
 2       321682 .         T                <DEL>        6    PASS   SVTYPE=DEL;END=321887;SVLEN=-205;CIPOS=-56,20;CIEND=-10,62         GT:GQ        0/1:12
 2     14477084 .         C                <DEL:ME:ALU> 12   PASS   SVTYPE=DEL;END=14477381;SVLEN=-297;CIPOS=-22,18;CIEND=-12,32       GT:GQ        0/1:12
 3      9425916 .         C                <INS:ME:L1>  23   PASS   SVTYPE=INS;END=9425916;SVLEN=6027;CIPOS=-16,22                     GT:GQ        1/1:15
--- a/VCFv4.2.tex
+++ b/VCFv4.2.tex
@@ -529,7 +529,7 @@ VCF STRUCTURAL VARIANT EXAMPLE
 The example shows in order:
 \begin{enumerate}
   \item A precise deletion with known breakpoint, a one base micro-homology, and a sample that is homozygous for the deletion.
-  \item An imprecise deletion of approximately 105 bp.
+  \item An imprecise deletion of approximately 205 bp.
   \item An imprecise deletion of an ALU element relative to the reference.
   \item An imprecise insertion of an L1 element relative to the reference.
   \item An imprecise duplication of approximately 21Kb. The sample genotype is copy number 3 (one extra copy of the duplicated sequence).
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -105,6 +105,11 @@ character encoding that is a strict superset of 7-bit ASCII and has the
 property that none of the bytes in any multi-byte characters are 7-bit ASCII 
 bytes. As a result, most software that processes VCF files does not have 
 to be aware of the possible presence of multi-byte UTF-8 characters. 
+
+\subsubsection{Data types}
+Data types supported by VCF are: Integer (32-bit, signed), Float (32-bit, formatted 
+to match the regular expression \texttt{\^{}[-+]?[0-9]*\textbackslash.?[0-9]+([eE][-+]?[0-9]+)?\$}, \texttt{NaN}, or \texttt{+/-Inf}), Flag, Character, and
+String. For the Integer type, the values from $-2^{31}$ to $-2^{31}+7$ cannot be stored in the binary version, see \ref{BcfTypeEncoding}.
 }
 
 \subsubsection{Information field format}
@@ -114,8 +119,8 @@ INFO fields should be described as follows (first four keys are required, source
 ##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="version">
 \end{verbatim}
 
-Possible Types for INFO fields are: Integer {\color{red} (32-bit, signed)}, Float {\color{red} (32-bit)}, Flag, Character, and
-String. {\color{red} For the Integer type, the values from $-2^{31}$ to $-2^{31}+7$ cannot be stored in binary version, see \ref{BcfTypeEncoding}.}
+Possible Types for INFO fields are: Integer, Float, Flag, Character, and
+String. 
 The Number entry is an Integer that describes the number of values that
 can be included with the INFO field. For example, if the INFO field contains a
 single number, then this value should be $1$; if the INFO field describes a
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -181,13 +181,18 @@ As with chromosomal sequences it is highly recommended (but not required) that t
 \end{verbatim}
 \subsubsection{Sample field format}
 It is possible to define sample to genome mappings as shown below:
-\small
 {\color{red}
+\scriptsize
 \begin{verbatim}
-##SAMPLE=<ID=S_ID,Genomes=G1_ID;G2_ID; ...;GK_ID,Mixture=N1;N2; ...;NK,Description=S1;S2; ...;SK,DOI=url>
+##META=<ID=Assay,Type=String,Number=.,Values=[WholeGenome, Exome]>
+##META=<ID=Disease,Type=String,Number=.,Values=[None, Cancer]>
+##META=<ID=Ethnicity,Type=String,Number=.,Values=[AFR, CEU, ASN, MEX]>
+##META=<ID=Tissue,Type=String,Number=.,Values=[Blood, Breast, Colon, Lung, ?]>
+##SAMPLE=<ID=Sample1,Assay=WholeGenome,Ethnicity=AFR,Disease=None,Description="Patient germline genome from unaffected",DOI=url>
+##SAMPLE=<ID=Sample2,Assay=Exome,Ethnicity=CEU,Disease=Cancer,Tissue=Breast,Description="European patient exome from breast cancer">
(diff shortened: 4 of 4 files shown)


droazen:
Please note that this pull request was opened at my request to provide a centralized location for discussion of the existing and proposed changes in VCF 4.3. It should NOT be merged yet, as there are still a few open issues under consideration.

droazen:
Here is a link to the PDF version of the current VCF 4.3 draft (built from this branch), to make it easier to see what's changed (changes are highlighted in red):

<https://gist.github.com/droazen/71d394379607674764f8/raw/d9e43db33fa2dbbf4f0db060aa57bf379526dc1a/VCFv4.3.draft.pdf>

pgrosu:
Thanks @droazen - much easier to read :)

droazen:
Comments on changes in the existing draft should go directly in this thread. Comments on open issues (https://github.com/samtools/hts-specs/labels/vcf) not yet incorporated into the draft should go on the page for each respective issue.

vruano:
Why GP is from 0 to 1? Does not seem practical when you are dealing with very small probs. and we already have a mixture of log10 and Phred scale annotations.

vruano:
In 2.1 VCF Tag naming conventions...

I think is a bit too absolutist to say that The "X" suffix means "Blah".  I think that the standard should be flexible and allow users to, for example, have some annotations that finish in "L", that are not likelihoods.

I would change the three points to: "If you mean 'Blah' the annotation name should (or must?) finish  with 'X'.

So from "The 'L' suffix means 'Likelihoods'" -> "The name of likelihood containing annotations should finish with the 'L' suffix"

vruano:
As with GP, why CNP is in 0 to 1 scale?

bhandsaker:
Section 5.4.11 needs to be updated so that the #PEDIGREE lines have IDs as required by the changes to section 1.2.10.

bhandsaker:
Section 1.4.1 (subsection 5 - ALT) seems inconsistent with new section 5.5 as to whether the "unspecified" allele is represented as "*" or as "<*>".

I would also suggest adding to section 1.4.1, subsection 5, that "The '*' allele is reserved to indicate that the allele is missing due to an overlapping deletion or some other variant incompatible with the listed alleles."

The overlapping deletion does not necessarily need to be upstream, although this is the easiest case to visualize.  Consider two offset deletions of several Kb each. Both the first and second deletion records may use '*' to represent a missing allele, since the situation is symmetrical, yet for the first record the conflicting allele is downstream (i.e. appears later in the file). It is also possible that the conflicting variants are not just deletions.  For example, two overlapping inversions might also use the '*' pseudo-allele to indicate that an allele is unspecified, missing, or described in a different VCF record.

bhandsaker:
Regarding GL and ploidy, one could for clarity include the following as well:

The GL ordering can also be expressed recursively with the following pseudo-code:
```
    GLOrdering (P, N, suffix="")
        for a in 0..N
            if (P == 1) println str(a) + suffix
            if (P > 1) GLOrdering(P-1, a, str(a) + suffix)
```

I think it might also help to provide three salient examples:
```
For P=2 and N=1, the ordering is 00,01,11
For P=2 and N=2, the ordering is 00,01,11,02,12,22
For P=3 and N=2, the ordering is 000,001,011,111,002,012,112,022,122,222
```

pd3:
@vruano @atks I updated the [draft](http://samtools.github.io/hts-specs/VCFv4.3.pdf) to d6b46df, it now includes the points discussed in #83. 

@droazen There are quite a few parallel topics already in this thread. Wouldn't it be easier to navigate if each was open as a separate issues?

pd3:
@bhandsaker The `*` was added already in 4.2 and its purpose seems to be different from `<*>`. Is `*` being used at all?

@bhandsaker Good point about the `ID` and `PEDIGREE`. It would be good to hear from someone who is actually using it.

droazen:
@pd3 If there's a suggestion here that is more than a simple/straightforward correction to the draft and deserves its own ticket, by all means let's spin it off into a separate issue for discussion. But I think it's also good to have "code review"-style comments/corrections to the draft as a whole here, as it's hard to see the big picture (eg., inconsistencies between different sections of the spec) in the threads for individual issues. My hope is that having people review the complete text should help improve its quality quite a bit, and cut down on vagueness/inconsistencies that lead to implementation issues down the road.

bhandsaker:
@pd3 I think `*` and `<*>` are doing the same job, and I would argue they are the same thing.
But this is perhaps in part because I think `*` should be used to represent the presence of any not-listed allele (an overlapping deletion allele being one common example).

`*` is starting to be used. The `*` syntax is being added to GATK (release 3.4 began using `<*:DEL>` instead of `*` but this was a mistake because no one realized that `*` has been added to the spec - this is being corrected). But I would like to see the spec use either `*` or `<*>` and not both.

d-cameron:
There are still a number of ambiguities within the VCF specifications, 
mostly due to missing character restrictions. Some examples of these are 
< and > are not being reserved contig characters thus <DEL> is a valid 
contig name, symbolic alt alleles able to contain numeric values hence 
<DEL:123> is valid, ID is unrestricted thus can contain reserved 
characters such as tabs, quotes or newlines, a list of INFO field 
"characters with special meaning" is not defined and INFO field URL 
encoding is optional thus %20 could be a space or could be the literal 
string "%20". A systematic approach to these issues would be the best 
way to deal with them. I can construct an EBNF formal grammar which can 
used to identify all ambiguities. Are these best presented as a pull 
request with all changes, or a list of problems and proposed resoltions?

RFC3629 states: "A protocol SHOULD forbid use of U+FEFF as a signature 
for those textual protocol elements that the protocol mandates to be 
always UTF-8"
As such, VCF should explicitly forbid encoding with byte order mark 
included.

Cheers
Daniel


On 4/06/2015 5:24 AM, droazen wrote:
>
> @pd3 <https://github.com/pd3> If there's a suggestion here that is 
> more than a simple/straightforward correction to the draft and 
> deserves its own ticket, by all means let's spin it off into a 
> separate issue for discussion. But I think it's also good to have 
> "code review"-style comments/corrections to the draft as a whole here, 
> as it's hard to see the big picture (eg., inconsistencies between 
> different sections of the spec) in the threads for individual issues. 
> My hope is that having people review the complete text should help 
> improve its quality quite a bit, and cut down on 
> vagueness/inconsistencies that lead to implementation issues down the 
> road.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/samtools/hts-specs/pull/88#issuecomment-108587452>.
>



______________________________________________________________________
The information in this email is confidential and intended solely for the addressee.
You must not disclose, forward, print or use it without the permission of the sender.
______________________________________________________________________

pd3:
@bhandsaker `<*>` was a compromise replacement of samtools' `X` and GATK's symbolic `<NON_REF>` allele, which were both introduced to represent likelihoods of alleles not present in ALT. This had nothing to do with overlapping deletions or other structural variation.

pd3:
@d-cameron +1 for resolving any remaining ambiguities, either as separate issues or a pull request, both are fine. I am not sure if EBNF is the most accessible style for the VCF audience, could regular expressions be used instead?

bhandsaker:
@pd3 I'll just go on record as saying that I think `<*>` and `*` are describing the same thing, the likelihood of alleles not present in ALT.  But I seem to be in the minority, so will let it drop.

pd3:
@bhandsaker My view may be too limited, please let's hear also from others before we close the issue. The difference between the two concepts seems to be that `<*>` refers to the likelihood of the assertion made in ALT in the same record, while `*` refers to a different record. I haven't used `*` myself and don't know if collapsing the two meanings could lead to problems.

pd3:
@vruano There was  a discussion about the scale of GP on the vcftools-spec mailing list ([link](http://sourceforge.net/p/vcftools/mailman/message/32139785/)) and the naming conventions as well ([link](https://github.com/samtools/hts-specs/wiki/Proposed-updates-to-VCF-and-BCF-specification)). I think the intention was to provide a rule of thumb principle rather than a rigid rule.

d-cameron:
@pd3 EBNF is more appropriate than regex since it can describe the entire file format, and there exist tools that can determine ambiguities directly from the formal grammar.

I've created [https://github.com/samtools/hts-specs/issues/89] with the problems I encountered when attempting to create a grammar.

pd3:
Is everyone happy with the changes proposed in #89 ?

cyenyxe:
@d-cameron Has there been any further work in the grammar definition? I wrote one for a VCF validator which has already been used against many VCF files (4.1 and 4.2), so if you consider it interesting, I could isolate it from the rest of the code and prepare a PR for discussion.

pd3:
Hi, please find the updated VCFv4.3 draft [here](http://samtools.github.io/hts-specs/VCFv4.3.pdf). It does not resolve all the issues mentioned here and elsewhere, for example it would be good to hear from someone who is actually using the PEDIGREE tag about the new requirement that all structured tags have to have an ID.

cyenyxe:
Some comments on points 1.2 to 1.4:

* The no-tab at the end of line sounds like a recommendation for the header line (1.3, "there should be no tab characters") but a must for the data lines (1.4, "tab-delimited with no tab character at the end").

* When point 1.4.1.4 states that IUPAC codes in the REF could be "reduced", does that mean that the tools should still consider them valid and run the transformation themselves? Otherwise, it could be something like: ```"IUPAC ambiguity codes in the reference sequence must be reduced to a concrete base by using the one that is first alphabetically (thus R=A/G as a reference base is converted to A in VCF.)"```

* IUPAC codes are mentioned regarding the ALT meta in 1.2, but not data lines in 1.4.1.5 ("Strings made up of the bases A,C,G,T,N,*, (case insensitive) or an angle-bracketed ID String <ID> or a breakend replacement string").

pd3:
@cyenyxe Thanks for your comments, Cristina. My wording was incorrect and I change 1.3 to "there must be no tab characters". Also in the #54 issue, the wording you proposed is what was actually meant. Apologies.

pd3:
If there are no more comments and everyone is happy, I'd like to merge the VCFv4.3 draft on the master branch in a few weeks' time. The latest version is available here, all changes are marked in red
http://samtools.github.io/hts-specs/VCFv4.3.pdf

wm75:
just came across this discussion now. A few remarks if it's not too late.

Together, section 1.2.2 and section 1.2.9 are confusing (to me at least).

1.2.2 lists the characters that have to be percent-encoded:
:;=% and CR, LF, TAB. As far as I'm understanding it this applies to all ## lines. 
1.2.9, on the other hand, says that ##contig ID values must follow the reference sequence names allowed by the SAM format, excluding the characters
"<>[]:*". Many of the characters from 1.2.2, however, are valid SN characters in SAM format so which statement takes precedence here (the explicit mention of ":" seems to suggest that 1.2.9 is supposed to overrule 1.2.2, but I'm not sure that's the intention)?


1.2.2 is ambiguous by itself also since it is not stating to which parts of the meta section exactly it applies. I do not think percent encoding is required inside quoted strings or the example header in 1.1 would be malformatted?


1.2.9 is hard to understand where it says: "the format is identical to that of a reference sequence, but with an additional URL tag ..." because the format of a "reference sequence" is not explained anywhere.


Section 1.2 talks about general line formatting, but does not say explicitly whether quoted strings would be allowed as an ID value (I think not, but the quotes would get parsed as part of the value?).


Not allowing : in CHROM (section 1.4.1) and contig (1.2.9) names solely on the basis that it would compromise breakends parsing seems overly strict to me since breakends do not become ambiguous just because of that. It's merely that the last colon (as opposed to just any) separates the chromosome name from the position.


Why isn't DPR listed as a reserved Genotype fields keyword in section 1.4.2 when AD, ADF and ADR are mentioned?


Hope some of these make sense,
Wolfgang

d-cameron:
>Not allowing : in CHROM (section 1.4.1) and contig (1.2.9) names 
solely on the basis that it would compromise breakends parsing seems 
overly strict to me since breakends do not become ambiguous just because 
of that. It's merely that the last colon (as opposed to just any) 
separates the chromosome name from the position.

There's a number of characters that have special meaning in VCF (both 
not SAM) that are still allowed in CHROM. "<DEL>" and "<*>" and 
"[<ctg1>[" are valid VCF contig names. With colon reserved, it is 
possible to at least disambiguate N[<ctg1>:1[ as a breakend to the named 
contig ctg1, and not a simple insertion of the reference contig 
"[<ctg1>:1[".
It's unclear as to whether C<ctg1><ctg2> or C<ctg1><ctg2>[<ctg1>:1[ are 
valid. s5.4.2 mentions a recommended shorthand for insertions and gives 
an example of "C<ctg1 >" but then it's not mentioned anywhere else in 
the spec at all. Currently breakends such as "C[[:0][]:-0e0[" or even 
N[<*>:1[ conform to the spec.

pd3:
Hi,

> Together, section 1.2.2 and section 1.2.9 are confusing (to me at
> least).
> 
> 1.2.2 lists the characters that have to be percent-encoded:
> :;=% and CR, LF, TAB. As far as I'm understanding it this applies to
> all ## lines. 
> 1.2.9, on the other hand, says that ##contig ID values must follow the
> reference sequence names allowed by the SAM format, excluding the
> characters
> "<>[]:*". Many of the characters from 1.2.2, however, are valid SN
> characters in SAM format so which statement takes precedence here (the
> explicit mention of ":" seems to suggest that 1.2.9 is supposed to
> overrule 1.2.2, but I'm not sure that's the intention)?
> 
> 1.2.2 is ambiguous by itself also since it is not stating to which
> parts of the meta section exactly it applies. I do not think percent
> encoding is required inside quoted strings or the example header in
> 1.1 would be malformatted?

The text was placed incorrectly in the meta-information lines section,
the comments are valid throughout the VCF. I moved the text up one level
and extended the text slightly, hopefully this will make it clearer:

"Characters with special meaning (such as field delimiters ';' in INFO
or ':' FORMAT fields) must be represented using the capitalized percent
encoding".



> 1.2.9 is hard to understand where it says: "the format is identical to
> that of a reference sequence, but with an additional URL tag ..."
> because the format of a "reference sequence" is not explained
> anywhere.

OK, expanded the example a bit and changed the text to:

"It is highly recommended (and required for BCF) that the header
includes tags describing the contigs referred to in the VCF file. The
structured \texttt{contig} field must include the ID attribute and
typically includes also sequence length, MD5 checksum, URL tag to
indicate where the sequence can be found, etc."



> Section 1.2 talks about general line formatting, but does not say
> explicitly whether quoted strings would be allowed as an ID value (I
> think not, but the quotes would get parsed as part of the value?).

The allowed IDs are described at the appropriate place for each type:
INFO, FORMAT, etc.


> Why isn't DPR listed as a reserved Genotype fields keyword in section
> 1.4.2 when AD, ADF and ADR are mentioned?

DPR has been replaced by AD as discussed in this thread
https://github.com/samtools/hts-specs/issues/78

An updated version of the draft is available
http://samtools.github.io/hts-specs/VCFv4.3.pdf




-- 

awz:
As discussed on today's file format's call, it would be wonderful to see this approved!  

Have the GATK, Platypus and Samtools variant callers committed to outputting homozygous reference blocks as described in this version of the 4.3 specification?

ekg:
Freebayes will as well, although some development is required.

awz:
Erik - that's great news!

pd3:
@awz samtools/bcftools does as well. There is an experimental version available for testing https://github.com/pd3/bcftools

reedacartwright:
I'm using the PEDIGREE tags and find that IDs are important as they allow me to specify where in a pedigree an mutation occurred:
```
DNT=AAxAA>AC;DNL=LB-M3;DNQ=40
```
These INFO tags describe that the predict de novo type is `AAxAA>AC`, which occurred in one of the branches leading to LB-M3

The only issue with the PEDIGREE tags is that I cannot store additional information.  According to the specs, everything other than ID describes an ancestor.  In order to support additional pedigree information it would be useful to have another tag, e.g:

```
##PEDIGREE=<ID=LB-N3,Father=GL-N1,Mother=GL-N2>
##PEDIGREEDATA=<NID=LB-N3,Type=float,Description="Mutation Rates",Father=1e-8,Mother=1e-9>
```
Where NID is used in PEDIGREEDATA so there can be multiple lines that contain data for the same node in the pedigree.

pd3:
Could the existing SAMPLE field be used for this perhaps?

reedacartwright:
I'm not sure if SAMPLE is the proper field to use.  My pedigree involves a mix of both germline meiotic events as well as somatic mitotic events.  Not every node in my pedigree represents a sample. Some represents libraries (since we support technical replicates of the same sample).  Others represent inferred ancestral states that occurred during family history or somatic development.

For me conceptually, the SAMPLE field provides information about the biological samples that were sequenced.  Whereas, the data I need to store is about the analysis that was done.  E.g. we had 2 libraries for this sample and we assumed that the error rate due to sequencing was 1e-8, etc.

I've already extended the "sample" columns in the vcf to output information about my nodes, including inferred ancestral states and library-specific depths.  I have no issues reusing the SAMPLE field as well, but would like to propose some changes to the wording of the VCF format to support a broader concept of what a sample is.  If that seems reasonable, I will work on a proposal.

pd3:
I see what you mean now, you are right that SAMPLE is not right for the job here. However, the specification does not explicitly say that everything other than ID refers to the ancestor, it only lists some examples. How about using the META field also for PEDIGREE and have something like this:

    ##META=<ID=MR,Type=Float,Number=.,Description="Mutation Rates">
    ##PEDIGREE=<ID=LB-N3,Father=GL-N1,Mother=GL-N2,MR=1e-8,1e-9>

The PEDIGREEDATA field is a valid solution as well, only the NID attribute would have to become ID.

reedacartwright:
I think the `###META` field makes the most sense, since it allows me to keep the meta data with the pedigree information.  Is its usage defined in the spec?

However, I would be concerned about the ordering of rates in your example in case some conversion decided to swap the order of Father and Mother without changing the MR pattern.  Is there a guarantee in the spec that that the attributes can be retrieved in the order written and that order will be maintained if converted between formats?

`##PEDIGREEDATA` might be a safer option if I can't rely on the order of attributes mattering in a VCF.  The reason why I decided to use NID instead of ID because ID's are supposed to be unique withing a field type.  For example, the following would not be allowed:

```
##PEDIGREE=<ID=LB-N3,Father=GL-N1,Mother=GL-N2>
##PEDIGREEDATA=<ID=LB-N3,Type=float,Description="Mutation Rates",Father=1e-8,Mother=1e-9>
##PEDIGREEDATA=<ID=LB-N3,Type=float,Description="Indel Rates",Father=1e-9,Mother=1e-9>
```

pd3:
You are making a good point about the order of attributes, the specification is silent about this. How about splitting the field into FatherMR and MotherMR? I am afraid it was agreed that ID is mandatory, so the original proposal with NID cannot be accepted as is.

reedacartwright:
Yes, we could use separate `##META` lines for mother and father mutation rate.

We could also use `ID` to uniquely define a `##PEDIGREEDATA` line and use `PedigreeID` to connect the `##PEDIGREEDATA` line to a line in the pedigree.  It is a little verbose, but perhaps less verbose than defining separate tags for each mother, father, and original value. 

```
##PEDIGREE=<ID=LB-N3,Father=GL-N1,Mother=GL-N2>
##PEDIGREEDATA=<ID=MR-LB-N3,PedigreeID=LB-N3,Type=float,Description="Mutation Rates",Father=1e-8,Mother=1e-9>
##PEDIGREEDATA=<ID=IR-LB-N3,PedigreeID=LB-N3,Type=float,Description="Indel Rates",Father=1e-9,Mother=1e-9>
```

pd3:
Yeah, both are possible. The degree of verbosity depends on the number of samples in the study. If there are many samples, the FatherMR and MotherMR are defined only once with the META solution. Also no new tag is required

    ##META=<ID=FatherMR,Type=Float,Number=1,Description="Paternal Mutation Rate">
    ##META=<ID=MotherMR,Type=Float,Number=1,Description="Maternal Mutation Rate">
    ##PEDIGREE=<ID=LB-N1,Father=GL-N1,Mother=GL-N2,FatherMR=1e-8,MotherMR=1e-9>    
    ##PEDIGREE=<ID=LB-N2,Father=GL-N3,Mother=GL-N4,FatherMR=1e-8,MotherMR=1e-8>
    ##PEDIGREE=<ID=LB-N3,Father=GL-N5,Mother=GL-N6,FatherMR=1e-9,MotherMR=1e-8>

reedacartwright:
I will give it a go using your META solution.

pd3:
OK, it seems we are ready to merge the changes. If there are no further objections, I'll do it tomorrow.

tk2:
@droazen as the GATK rep - are you in agreement with merging?

jmarshall:
In §1.3 (Data types),

> Data types supported by VCF are: Integer (32-bit, signed) […] For the Integer type, the values from −2<sup>31</sup> to −2<sup>31</sup> + 7 cannot be stored in the binary version, see 6.3.3.

It's not too good to have things that are representable in one format but not the other.  So we should disallow -2<sup>31</sup> to −2<sup>31</sup> + 7 in both VCF and BCF.

----
In §1.6.1 (Fixed fields, §1.4.1 in previous spec versions),

> There are 8 fixed fields per record. All data lines are tab-delimited with no tab character at the end of the line. The last data line should end with a line separator. In all cases, missing values are specified with a dot (‘.’).

The first sentence is about fixed fields.  The next two sentences (re tabs and lines) are about the whole line, so should be lifted to §1.6 (Data lines).

The final sentence (re missing value dots (‘.’)) is hopefully intended to apply to **all** fields.  However at present being in this section it only applies to the 8 fixed fields.  It should also be lifted to §1.6, or a similar sentence needs to be added to §1.6.2 (Genotype fields).

jmarshall:
Re missing value dots, see the conversation that motivated this: samtools/htsjdk#340.

PR #91, opened 2015-06-11
Be definitive in the ##fileformat section

jmarshall:
This section is our opportunity to **specify** `##fileformat`, not just give an example.


jmarshall:
@pd3: okay to add this to the **VCFv4.3** branch?

pd3:
@jmarshall Sorry, I missed this. Yes, it's OK with me, thanks

PR #92, opened 2015-06-23
Define mate flags for singleton reads

lbergelson:
Flags 0x2, 0x8, 0x20, 0x40, and 0x80 are meaningless if flag 0x1 is unset.  The current spec allows them to take any value in this case.  I propose to restrict these values to be unset instead of leaving them undefined.  

This would leave less degrees of freedom in implementing the spec and would make interoperability between different conforming implementations simpler.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -312,8 +312,7 @@ Each bit is explained in the following table:
     0x40 and 0x80 are unset, the index of the read in the template
     is unknown. This may happen for a non-linear template or the index
     is lost in data processing.
-  \item If 0x1 is unset, no assumptions can be made about 0x2, 0x8,
-    0x20, 0x40 and 0x80.
+  \item If 0x1 is unset, 0x2, 0x8, 0x20, 0x40 and 0x80 must be unset.
   \end{itemize}
 \item {\sf RNAME}: Reference sequence NAME of the alignment. If {\tt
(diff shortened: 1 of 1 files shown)


jmarshall:
This text previously changed in e6ca319516752ecdbc25b5880c41b1a4970dd0ec (rewording) and during TeXification (previously _Flag 0x02, 0x08, 0x20, 0x40 and 0x80 are only meaningful when flag 0x01 is present_).

Me, I am generally :+1: on specifying values for meaningless combinations of flags, as then the flag bits can be reused for other purposes for unpaired reads.  However in this case, these flags have been _don't care_ for many years, so it is a bit late to be tightening this.  Moreover, the suggested change pertains only to unpaired reads, and I suspect SAM/BAM files full of unpaired reads are uncommon in practice anyway.

So in this case, my opinion is :-1: this is best left as is in the spec.  HTSJDK should feel free to continue validating that these bits are unset (when 0x1 is unset) as a best practice, of course.

jmarshall:
Opinions, @nh13?

nh13:
I agree it is tough to enumerate all possibilities for the flag bit cases versus just stating the ones that have meaning (and by definitions others don't).   So :-1:  from me.

lh3:
I am -1 on this. I think it is safe to close this now. Thanks anyway.

PR #95, opened 2015-07-14
FAO @vadimzalunin: Fixed bad encoding example.  Fixed Feature Codes table.

jkbonfield:
See https://sourceforge.net/p/samtools/mailman/message/34293215/ for a
discussion on these.  The first of these is just a bad example which
appears to have been incorrect since version 1.1.  The second is an
error caused during the multi-base sequence and quality features.
They're listed as BA/QS data series but are implemented in both C and
Java as BB/QQ.

Additionally the 'A' "bases and scores" pair is not implemented in
either C nor Java.  I commented this out for now as we haven't found a
reason for it and if anyone else did start using it then it'd break
our implementations.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -235,15 +235,14 @@ Subexponential encoding example:
 
 The first byte ``0x7'' is the codec id. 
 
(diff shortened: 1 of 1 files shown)


vadimzalunin:
+1 looks good

Issue #85, opened 2015-05-08
Real-world usage of 0x200 failedVendorQualityChecks flag


jmarshall:
The SAM specification describes the 0x200 flag bit as _not passing quality controls_ (and the pre-LaTeX version of the spec described it as _the read fails platform/vendor quality checks_).

It also describes it as the _filtered_ bit in §2 (Recommended Practices) when introducing annotation dummy reads:

> Annotation dummy reads: These have SEQ set to `*`, FLAG bits 0x100 and 0x200 set (secondary and filtered), and a CT tag.

In the real world today, do we actually see bad-quality reads in BAM files with their _not passing quality controls_ flags set?

@lh3 [recently commented elsewhere](https://github.com/ga4gh/schemas/pull/292#issuecomment-97127759) that "almost no one uses failedVendorQualityChecks nowadays".  Probably the usual practice is for such quality-failed reads to be just omitted from BAM files rather than kept and marked with this flag, and probably this has been the case for quite some time.

Would it be worth repurposing this flag as a general _this-read-is-filtered-out_ flag?  This would be as simple as changing the description in the flag bit table, e.g.:

> not passing filters, such as platform/vendor quality controls

lh3:
I support this.

yfarjoun:
With hiseqX this is true since there are so many failed reads, but with the
non-patterned flowcell machines we keep the failing reads around. I
remember that David Jaffee said that his group looks at all reads,
including the PF failed reads for assembly....

lh3:
John just changed this flag to something more general. It would not cause any compatible issues as far as I see. Broad can still use this flag to mark PF'ed reads.

nh13:
I support this as well.

yfarjoun:
Hi,

I might be missing something but I couldn't find 0x800 bit definition from
the BAM spec in the v3 spec document for CRAM.

The flag is included in the code
(htsjdk/samtools/cram/structure/CramCompressionRecord.java:41), but not in
the document.

I presume that this is a typo, but I just wanted to make sure before
changing the document.

Yossi.

jmarshall:
@yfarjoun: This is #61, and I'm a bit surprised they still haven't made this seemingly obvious editorial fix…

yfarjoun:
sorry, I should have checked the issues.

Y.

PR #99, opened 2015-08-25
Various small SAM spec changes

jmarshall:
1. Generalise 0x200 flag bit as _filtered-out bit_, as discussed in #85.

2. Clarify that unspecified FLAG bits are reserved for future use, in response to a [samtools-devel question](http://sourceforge.net/p/samtools/mailman/message/34386981/).  Possibly this would be better added to §**Recommended Practice** rather than in the **FLAG** definition?

3. Change SAM maximum QNAME length to match the BAM limit.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -296,7 +296,7 @@ Each bit is explained in the following table:
    64 &   0x40 & the first segment in the template \\
   128 &   0x80 & the last segment in the template \\
   256 &  0x100 & secondary alignment \\
-  512 &  0x200 & not passing quality controls \\
+  512 &  0x200 & not passing filters, such as platform/vendor quality controls \\
  1024 &  0x400 & PCR or optical duplicate \\
  2048 &  0x800 & supplementary alignment \\
   \hline
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -257,7 +257,7 @@ of the mandatory fields in the SAM format:
   \hline
   {\bf Col} & {\bf Field} & {\bf Type} & {\bf Regexp/Range} & {\bf Brief description} \\
   \hline
-  1 & {\sf QNAME} & String & \verb:[!-?A-~]{1,255}: & Query template NAME\\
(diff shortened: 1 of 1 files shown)


jmarshall:
It turns out that BAM can represent a QNAME of up to 254 characters.  If you try to read a SAM record with a 255-character QNAME, samtools complains:

> [E::sam_parse1] query name too long
> [W::sam_read1] parse error at line 7

(and samtools-0.1.19 segfaults).  If you try to convert such a SAM record to BAM, Picard complains:

> java.lang.IllegalArgumentException: Value (256) to large to be written as ubyte.
>	at htsjdk.samtools.util.BinaryCodec.writeUByte(BinaryCodec.java:317)

(although ValidateSamFile is happy with it).  So since 255-character QNAMEs don't work in both implementations, I think we can just change the spec to 254 for SAM even though this technically is a new restriction.

karel-brinda:
Hi, just a quick question. How difficult would be to add support for QNAMEs of any length in SAMtools and Picard (theoretically)? Is it only a problem of some fixed-size buffers or is it a conceptual issue of BAM?

For RNF and RNFtools, this is a very limiting restriction. If read name exceeds length 255 (I will have to change it to 254), a special file must be created to represent those long read names out of SAM/BAM.

Is it clear now that future sequencing technologies will not need more than 254/255 characters (e.g., to save some additional information besides bases and qualities)?

jmarshall:
As noted, the limiting factor is `l_read_name`'s representation in BAM (§4.2).  It is a question of the format, not the implementations (i.e., samtools/Picard).

Current sequencing technologies encode tile coordinates etc in QNAME, and 50 or so characters appears to be more than adequate.  If future sequencing technologies have more voluminous additional data, they will do well to output unaligned-BAM rather than FASTQ or otherwise use a more flexible approach that hacking information into a rather constrained QNAME.

jmarshall:
@jacarey: Can I get a :+1: on each of 1–3?

jacarey:
:+1:  to all

Issue #100, opened 2015-09-10
suggestion: aliases in @SQ section


lindenb:
A suggestion for the Sam Sequence dictionary: would it be possible to add an optional list of aliases in the @SQ header ? softwares would use those aliases to fix the various nomenclatures (UCSC, ENSEMBL )

Something like:

```
@SQ SN:MT LN:12345 AL:chrMT|chrM
@SQ SN:X LN:1234567 AL:chrX|chr0X|0X
```

then

```
samtools view my.bam MT:1-100
```

would return the same ouput than

```
samtools view my.bam chrM:1-100
```

atks:
I think we should have this for the contigs in VCF too.

jmarshall:
This is IMHO a rather good idea, although we'd have to come up with a different separator as e.g. [NCBI](http://www.ncbi.nlm.nih.gov/nuccore/251831106?report=fasta) likes to use `|` characters within sequence names.  I might prefer `AN:` or so ("alternative/alias names") to parallel `SN:`.

We'd also have to think about what rules are necessary around distinctness, and to be explicit whether alignment records in SAM files would be allowed to use these aliases in RNAME/RNEXT (surely not!).

jmarshall:
It seems NCBI will be [phasing out their `|` characters in sequence names](http://www.ncbi.nlm.nih.gov/news/03-02-2016-phase-out-of-GI-numbers/), though we probably still can't use `|` here as people will still have older NCBI files lying around.

vadimzalunin:
are aliases known at write time? seems like aliases are read time decision to me.

dpryan79:
Most programs don't currently know about them when they make BAM files, but that could be changed.

I have to mention that dealing with different naming schemes is a huge annoyance, particularly on things like Galaxy where you have novice users who are often not aware of this issue and inevitably need to be walked through properly munging things.

jmarshall:
We discussed this at some length back at our September meeting (and I'm finally writing up my notes from then).  We continue to mostly like the idea in principle, but need to spell out rules around uniqueness and so on.

We noted that we can use this as an opportunity to define the alias regexp to be the excellently tight regexp (disallowing especially `:`, so `chr:beg-end` notation is unambiguous, and perhaps other punctuation too; cf #167) that we wish we'd always had for `SN`.  Doing so will solve the what-separator-to-use problem, and provide a path towards restricting `SN` to the same regexp.

[...]

PR #103, opened 2015-10-12
implementation of https://github.com/samtools/hts-specs/issues/100

lindenb:
Hi , this is a pull request to suggest a new Tag **AN**  for **@SEQ** in the sequence dictionary.

> Reference sequence Aternative Names. A semicolon separated list of alternative names for this sequence e.g:
> 1;01;CM000663;NC 000001.10. Tools are free to use this information to decode the users’ posi-
> tions.

see also issue id 100: https://github.com/samtools/hts-specs/issues/100


jmarshall:
Thanks.  Closing so we keep discussion in one place: issue #100.

PR #111, opened 2015-11-19
Fixed typo/formatting in SAM spec

laserson:


laserson:
Sorry, even in initial commit I made a similar error.  Should be good now.

jmarshall:
Thanks.  It's hard to believe that one was there for so long!

PR #113, opened 2015-12-04
Split out tags into separate document

jmarshall:


jkbonfield:
Good and obvious first change (thanks), but we need to go through with expanded explanations of the various fields, which probably means quite a bit of discussion (some aren't *clear*).

tfenne:
@jmarshall is there a plan for finalizing this split into a separate core spec and tags document?  It would be nice to get this finalized so that the process of adding new tags can be unblocked.  If there is work that needs to be done and can be described in issues or `todo`s in the new document I'm happy to pitch in.

nh13:
@jmarshall want someone to look over this before merging?

nh13:
Oops, looks like it got merged in!

jmarshall:
Merged as ratified at the the June file formats meeting.

As noted in @jkbonfield's comment and in the commit messages (“This facilitates expanding the tag explanations so that they become actually *clear*, but this remains a work-in-progress”) this is the basis of the document, but in our copious spare time it would be good to expand the explanations so they actually communicate the details necessary to use some of these tags.

PR #117, opened 2015-12-22
Added a new DS attribute for @SQ lines to create a standard for storing descriptions

tfenne:
[Review on Reviewable](https://reviewable.io/reviews/samtools/hts-specs/117)

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -248,6 +248,7 @@ or~{\sf RNEXT} fields.
 \emph{Regular expression}: \emph{name}{\tt (,}\emph{name}{\tt )*}
 where \emph{name} is {\tt [0-9A-Za-z][0-9A-Za-z*+.@\_|-]*}\\\cline{2-3}
   & {\tt AS} & Genome assembly identifier. \\\cline{2-3}
+  & {\tt DS} & Description.  UTF-8 encoding may be used.\\\cline{2-3}
   & {\tt M5} & MD5 checksum of the sequence.  See Section~\ref{sec:ref-md5}\\\cline{2-3}
   & {\tt SP} & Species.\\\cline{2-3}
   & {\tt UR} & URI of the sequence.  This value may start with one of the standard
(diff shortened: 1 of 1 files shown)


tfenne:
@jmarshall - a pretty trivial change here to define a standard attribute for the `@SQ` lines in the header, so that there is a standard attribute for storing a description.  It mirrors the same attribute found on the `@RG` lines.  I'm suggesting this because I'm routinely finding I want to put more information from fasta files into the sequence dictionary, but don't want to resort to non-standard attributes to do so.

tfenne:
@yfarjoun @jmarshall @jkbonfield : this has been open for three weeks.  If nobody objects in the next day or so I'm going to merge since it's such a simple change (and can always be undone).

jkbonfield:
I didn't spot this earlier, not sure why.

Anyway, looks OK to me.  Infact I thought we'd already added it.

jmarshall:
Your three weeks started two days before Christmas, so I think you will agree that this not an ideal time for people to be available to consider spec changes :smile:.

On balance probably a mild :+1: from me on the general idea, but editorial concerns around the wording: punctuation after _e.g._; _FASTA_ in capitals; and the example should be phrased other than in terms of whitespace due to confusion as in samtools/samtools#449.

The people with the authority to merge this are the SAM specification editors (i.e., @jacarey and myself), and I expect we will do so when we are satisfied with the proposal.

tfenne:
:disappointed: Until late last night I was able to merge PRs here, I didn't realize that was by accident.

@jmarshall I wasn't complaining about the three weeks - just pointing out that there had been ample opportunity for anyone who wanted to to object to the proposed change.  I'll take a shot at rewording per your comments.

jacarey:
This looks fine to me.

mp15:
+1 for concept, as long as the editorial changes are fixed I think we have consensus?

tfenne:
Sadly, I don't think I can push to this branch any more.  @jmarshall would you prefer to make any editorial changes you choose and merge, or that I close this PR and start a new one from a forked repo?

jmarshall:
Apologies, Tim, I guess this got dropped after my editorial quibbles. As @jkbonfield said previous, I thought we'd already added this…

I've updated this to just say “Description” like the other headers' DS tags do. Previously you had

> Description of the sequence (e.g. value after the first whitespace in fasta header line).

We could add a (non-normative) footnote talking about FASTA headers, but the other description fields don't talk about how you might populate them either — so perhaps being agnostic and just saying “description” is appropriate for the spec.

yfarjoun:
👍

jmarshall:
Thanks. Looks like that's a go for rebasing and merging this.

Issue #109, opened 2015-11-16
Suggestion: add RX and QX tags for molecular barcodes


yfarjoun:
I'd like to add standard tags RX and QX (to the SAM spec) for the bases of a molecular barcode and their qualities as read by the sequencer. 

To be clear, this is _not_ the sample barcode which can be used to demultiplex the sequence data, but rather a molecular barcode/index that can be used to identify the original molecule from which the sequence was generated. This can be used to disambiguate PCR from "Biological" duplicates (say in RNA-seq or other high depth technologies), or to identify the haplotype from whence many read pairs arose (as in 10X technology, e.g.)

There might be a need for other tags for the "corrected" barcode that is identified after comparing with a whitelist, but I'm not including that in this suggestion.

Some discussion of this proposal has started here:  (https://github.com/broadinstitute/picard/issues/287)  

I'd like to here what people think before submitting a PR here.

tfenne:
The discussion in the linked picard issue left it somewhat ambiguous whether or not `BC` and `QT` were also for this purpose.  In my own work I've been using `BC` to store the molecular index, not the sample index.  It would be good to understand if others are also doing this.

yfarjoun:
It seems that as these are different things, they should have different tags. Do you not agree?

tfenne:
I agree that sample index and molecular index are different things.  However the current spec doesn't say "sample" or "molecuar", it just says "barcode".  Since I don't see the utility of stamping the sample barcode on every read, I assumed `BC` was for molecular indices and have been using it that way.  My only real concern is that many others may have done the same, and we'll have some confusion/bifurcation.

yfarjoun:
The place in the code where I see the actual sample-barcode being used in
the BC tag is when there is no good match with the list of provided
barcodes:

// If it's a barcoded run and the read isn't assigned to a barcode, then
add the barcode
// that was read as an optional tag
if (unmatchedBarcode != null) {
    sam.setAttribute(SAMTag.BC.name(), unmatchedBarcode);
}

from picard.illumina.ClusterDataToSamConverter#createSamRecord



On Mon, Nov 16, 2015 at 2:47 PM, Tim Fennell <notifications@github.com>
wrote:
>
> I agree that sample index and molecular index are different things.
However the current spec doesn't say "sample" or "molecuar", it just says
"barcode". Since I don't see the utility of stamping the sample barcode on
every read, I assumed BC was for molecular indices and have been using it
that way. My only real concern is that many others may have done the same,
[...]

PR #119, opened 2016-01-08
Add new RX/QX/FT tags and clarify intent for existing BC/QT

yfarjoun:
Follows discussion https://github.com/samtools/hts-specs/issues/109

**BC**: This should be used for the  _sample_-barcode so that there is no conflict in case that both a sample and a molecular-barcode are used
**QT**: This hsould be the quality for the _sample_-barcode BC
**RX**: (new tag) the raw sequence bases of the _molecular_-barcode(s) of the read(s) can be placed in this tag. how/whether to concatenate multiple tags or place them on separate records is left for the implementation
**QX**: (new tag) the qualited (as QUAL) of the bases in RX
**FT**: (new tag) This is for the reason why the record has been unaligned or marked as not-passing filters. This is already used in two places:
1. Picard's MergeBamAlignments will unalign records it identifies as cross-species contamination and "Cross-species contamination" will be placed in the FT tag
2. GATK's --bamOut mechanism (for emitting the reassembled reads from the HaplotypeCaller's engine) will sometimes fail to reassmble some reads, the reason for the failure will be placed in the FT field and the original alignment will be retained (though the read will be filtered, or unaligned)


nh13:
1. Thank-you for pushing this through.
2. How about a footnote for the `BC` and `RX` tags indicating that how they are concatenated (if there are multiple places) is implementation specific?
3. Is `QX` the same length as `RX`?  If so, even for concatenation, then this should be stated (see `BQ` and `CQ` for two differing examples).

yfarjoun:
1. my pleasure.
2. I'm not sure what to say about BC as I've clarified that it is a sample-barcode, not molecular. I'm adding a note about RX and QX
3. agreed. see incoming commit.

nh13:
RE 2. You say for `RX`: "how/whether to concatenate multiple tags or place them on separate records is left for the implementation".  I would also say the same thing for `BC`, and so a footnote that addresses this for both.  Consider a sample-barcode that is "in-line" and on "both ends".  This is similar to having molecular ids specific for each end of read pair.  We could concatenate them and put them in `RX` and so the `RX` values would be the same for both records, or we could keep them separate and so the `RX` values for each end of a pair would be different.

yfarjoun:
Oh, I see. This is very natural. Thanks for suggesting this.

yfarjoun:
@nh13 does this latest version align with your thoughts?

nh13:
:+1: from me, but how about other folks on #109?

yfarjoun:
Thanks. I'll wait a week or so to give folks a chance to chime in.

nh13:
Perhaps also "bump" that thread so folks get the notification.

mp15:
As we agreed to separate out the tags document in the last file formats meeting might I suggest this is rebranched off #113?

yfarjoun:
if #113 will be merged soon, I can rebase from master after that process is
done, but if it's still being edited, I'd prefer not to.

unless you are suggesting that I rebase on that branch and then the editing
of #113 will continue with my changes...but that would be a little strange.

jmarshall:
This change adds some new tags and adds to the descriptions of several existing tags.  The aim of the reorganisation with a separate _SAMtags.pdf_ document in #113 is to give us space to write a paragraph or more when we need it to adequately describe tags' nuances.  I think we need it here!

So I think it would be beneficial to write this proposal as if the separate tags document already existed.  (In practice, it'll be a while yet before #113 is merged.)

The amount of discussion about barcode tags we've already had shows that this stuff is far from simple.  So how about writing this proposal as if you have as much space for discussion prose as you like:

* Brief one-line table entry per tag (as we have in the existing single document)
* Detailed paragraph for each tag; or as in this case, several paragraphs discussing the related tags together in depth (which will appear as a **Barcodes** subsection in #113's _SAMtags.pdf_)
* Don't worry about the formatting too much; I'll merge it into the new documents after #113 lands.

jmarshall:
[Note for next time: please be specific in the title, just about any SAM spec proposal could fit the description _"Adding a few Tags…"_!]

yfarjoun:
yes. sorry. you are very correct.

ablewhiskey:
Hi Yossi, thanks for making this happen! Just want to chime in from 10X's perspective.

- Explicitly defining BC/QT to distinguish sample barcodes from molecular, is great and should help avoid conflicting implementations.
- The RX/QX definitions look good too and are consistent with our use of them.

Processed Barcodes
- I know there's been previous discussion of this, but we still think it makes sense to also introduce a processed barcode tag that is the result of error-correction, confirmation against expected list, or other operations. This processed barcode is what most secondary analysis code (alignment, variant calling, etc) will want to work with rather than the raw versions. We are using BX for this. 
- We'd also suggest not restricting this tag to being sequence, and allowing any string content, so it can essentially have the same level of generality as RG, but permit extremely high plex without blowing up the header.

Other Barcoding Applications
- Something else to think about is the introduction of **cellular** barcodes for single-cell applications.
- At the moment our upcoming single-cell pipeline is using RX/QX/BX for cellular barcodes, but this overloads the meaning of the molecular barcode, and would be problematic if one had both cellular **and** molecular barcodes.
- We are considering adopting a separate set of tags CR/CQ/CB for the cellular barcodes, which leaves open the possibility of representing both cellular and molecular barcodes at the same time.
- It may even make sense to move to a more generalized hierarchical barcode system that would allow for any number of domains over which to barcode. This is maybe starting to sound like RG, but again you'd need a way to support very high plex. This is probably out of scope of this particular discussion.

yfarjoun:
why not use BC for cellular barcode?

tfenne:
@ablewhiskey - a couple of thoughts:
1. For single-cell work, why not create a read group per cell and both assign the barcode at the read group level and group the reads that way?  I don't know anything about how 10X is doing single-cell but this has made good sense to me working with single-cell RNA-seq data.
2. I'm not against a processed-molecular barcode tag, but would like to point out that this doesn't make sense for all technologies.  I'm working with multiple different technologies currently that use a molecular barcode to convey a unique molecular id, but the IDs are randomers across the full space (i.e. at length 6 there are 4^6 valid sequences), so there is no error correction or assignment process.

ablewhiskey:
@yfarjoun We are using BC for sample indexes to distinguish sequencing libraries as per normal usage. Each of those sequencing libraries can comprise reads originating from thousands or tens of thousands of cells. So we still need BC to sample multiplexing, but that means a separate set of barcode tags to distinguish all the cells within each of those samples.

@tfenne 
1. For the same reason RG isn't suitable for molecular barcodes where you're scaling to hundreds of thousands or millions of distinct barocdes (the RG header would be enormous). For existing low-plex single-cell experiments, RG are probably fine, but we are already starting to work with upwards of 100,000 cells at a time and so again you'd run into the same practical scaling issues with RG as you would with high-plex molecular barcoding.
2. Understood. We're not proposing BX (processed) instead of RX, but rather in addition to, or as an option, in applications like ours where it does make sense.

ablewhiskey:
We've been discussing this more internally and trying to balance generality, being forward-thinking, and practicality.

Now that we are dealing with barcoding over multiple domains simultaneously (cells and molecules), and contemplating the possibility of other future domains, we are thinking about a more consistent and extensible naming convention for the barcode tags. What do you guys think of the following:

For any given domain x, there would be 3 tags:
- xR: raw barcode
- xQ: raw barcode quality
- xB: processed barcode (if applicable)

For molecular barcoding, this would imply MR, MQ, and MB (as opposed to RX, QX, BX).
For cellular barcoding, you'd have CR, CQ, CB.
For other types of barcoding, you'd pick an appropriate letter to prepend to R, Q, and B.

We are happy to stick with RX, QX, BX, since that is what we are already using, but thought the parallelism in the above scheme might reduce cognitive burden when working with SAM or BAM files that contain multiple types of barcodes.

Thoughts?

dkj:
I'll propose an alternative: concatenate all the "technical" read (e.g. standard sample index reads, molecular barcode reads, etc) basecalls and qualities in the existing BC and QT auxtags with some recommendations:
- to use a separator (precedent set by ```-```?)
- order them by those used for determining RG, then cellular barcoding, then molecular, then any PCR dup detection (appropriate ranking team X10?)
- normally put them on the first primary, non-sup, read of the template

Just as the standard sample (often dual) index info in BC and QT is then used to set RG over all the reads in the template I imagine mb/MB and cb/CB tags would then also be set over all the reads in the template whilst reads in the template are still collated (is that correct team X10?). (Depending on dup detection algorithm PCR dup detection tags may need to be copied as well).

This might make "processed barcode" tag determination code slightly more complex. Hopefully, it would not affect existing code to deplex BC info to set the RG.

We'd avoid a proliferation of tags for "technical" reads for every new purpose they have.

I can imagine people wanting to save space by throwing away the reads used for standard deplexing would also want to discard the molecular and cellular barcode reads once the processed barcodes have been set.

Why do we need separate tags for the different types of technical reads? Is it for some significant programming convenience? Are there different patterns of use or retention requirements? Is it important to mark their meaning by tag type? Or something else?

tfenne:
This is an interesting idea @dkj .  Both @nh13 and myself are also working on applications with more than one technical read (excluding the sample barcode), and I agree that the proliferation of tags doesn't sound ideal.  It's also likely to create problems as there are only a limited number of letters to put before R, Q, and B.

To build on your suggestion, you could imagine having a limited number of tags (perhaps one each for raw sequence, assigned sequence and qualities) and a fourth tag that describes the structure of the data in those tags?  E.g. Picard has, for a long time, had the notion of a `read structure` to describe how to break apart a read (or an arbitrary sequence of bases) into logical reads, that is similar in format to CIGAR.  We could enumerate the read types like:

```
C -> cellular index
L -> long molecule index
U -> unique molecular index
....
```

And then write a structure tag like `6C6L4U` to mean a six base cellular index, then a six base long molecule index then a 4 base UMI.

mp15:
I would like to get this sorted at the next File Formats meeting @ablewhiskey are you happy that this adequately describes your "as implemented" output? In particular are you happy that if we archived a BAM with these tags to the EGA or DBGAP and then pulled it out and ran it through LongRanger the BAM would contain enough information to able to recreate the analysis as if we had done it from data taken off the sequencing machine?

mp15:
@yfarjoun to confirm 10X using same tags with Chromium.

ablewhiskey:
@mp15 Sorry Martin, I missed your previous post on this thread. Let me clarify what we currently have implemented in our released software (Long Ranger for genome/exome and Cell Ranger for single-cell). 

The ground truth documentation for our BAM tags are here:
**Genome/Exome - Long Ranger**
http://software.10xgenomics.com/genome-exome/pipelines/latest/output/bam
**Single Cell - Cell Ranger**
http://software.10xgenomics.com/single-cell/pipelines/latest/output/bam

To summarize, for Long Ranger, BX contains the processed barcode which has been error-corrected and filtered against a whitelist of known-good barcodes. Note that BX does not contain strictly sequence bases only. The value is suffixed with a "GEM (partition) group" identifier, which enables virtualization of barcodes to increase effective barcode diversity when combining libraries. The suffix typically is "-1" when there is only one partition group. But the BX value must be treated as an opaque string, and processing of it must not assume it comprises strictly [ATCGN].

Long Ranger BAM files also contain RX/QX, which is the raw barcode as read from the sequencer, and its read quality. RX is strictly [ATCGN].

As far as round-tripping, Long Ranger only accepts FASTQs as input, not BAM. One would need to process the BAM to convert it back to FASTQ, keeping in mind that in Chromium the barcode is inlined as the first 16bp in Read1, and is no longer on a separate index read. In order for this to work, the round-tripping process would have to preserve RX and QX. You'd also want to make sure to preserve the BX tag so that people wanting to consume the BAM directly (rather than feeding it back through alignment in Long Ranger) could take advantage of the processed barcode.

Cell Ranger for single-cell data has its own set of tags: CR/CQ for the raw cellular barcode read and its quality, CB for the processed cellular barcode, UR/UQ for the raw molecular (transcript) barcode, and UB for the processed molecular barcode.

yfarjoun:
At this point I've only included a tag to capture UMIs as they came of the sequencer (and their quality). It seems to me that the space is too active right now to put in all the various tags that folks come up with, many of which pertain to internal processing. I would like to be able to move this PR forward, but am not objecting to a future PR that will build on it. 

I've addressed comments in an add-on commit.

SAMv1.tex
```diff
@@ -559,7 +561,22 @@ \subsection{The alignment section: optional fields}\label{sec:alnaux}
...
+The {\tt FT} tag will be set to say "Cross-species contamination" (as this is what it perceives the read to be sourced from in this case). 
+GATK's bamOut mode of Haplotype caller will set 0x200 bit for reads that wouldn't count towards the model.  
+The {\tt FT} tag is used to explain why this is the case (e.g. Trimmed, Downsampled, etc.). 
+\subsection{Regarding the RX, and QX tags}
+The {\tt RX} tag should match in length with the {\tt QX} tag. 
+What one does with multiple molecular barcodes (e.g. one on each end of the template) is left to the implementation. 
+One current implementation concatenates all the barcodes and qualities with no separator between the different barcodes.
+\subsection{Regarding the BC, and QT tags}
+The {\tt BC} tag should match in length with the {\tt QT} tag. 
+What one does with multiple molecular barcodes (e.g. one on each end of the template) is left to the implementation.
```
nh13: `molecular barcodes` -> `sample barcodes`

nh13:
@ablewhiskey It looks like `CQ` and `UQ` are already in the spec, so this will be confusing to your users, especially since `UQ` is of a different type (`i` in the spec, `Z` in your definition).  

@yfarjoun I agree, lets keep it simple for now, and add these in.  There are so many ways to add in molecular barcodes, both in terms of where in the template and when during sample and library preparation, and sometimes they are intended to be random and sometimes they are not, and sometimes multiple barcodes in the template can be related in very interesting ways.   I think it warrants further discussion on how to represent them beyond the scope of the PR due to the explosion of possibilities we would be required to support.

yfarjoun:
thanks @nh13.

mp15:
@ablewhiskey I have to second @nh13's point about CQ and UQ. There's plenty of free two letter combinations and it's really not worth the problems you're going to cause your customers if they try and do some algorithm downstream that expects them to mean something else. Please fix this as a matter of urgency.

yfarjoun:
@jmarshall what is this PR waiting on?

ablewhiskey:
Sorry for delay in reply. Yes we agree, the use of CQ and UQ was an oversight and we will change them in the next release of our Cell Ranger software. We are considering using CY and UY. Does anyone have concerns about these?

yfarjoun:
@jmarshall I'm not sure of the process. is there anything else I need to do do get this PR merged?

nh13:
@yfarjoun if it's not too late, could we add the `MI` tag (integer) that is a unique identifier of an original source molecule?  For example, due to errors, reads may have different values for `RX` but come from the same `MI`.  Also consider where two reads may observe the opposite strands of the original source (double stranded) molecule.

yfarjoun:
@jmarshall I rebased onto your changes and moved the tag information to the new file (thanks!!) I also added the MI tag as per @nh13 's request. 

thanks for taking a look.

mp15:
@ablewhiskey Can you do a quick review on behalf of 10X please, check we've not missed anything?

ablewhiskey:
@mp15 Hi Martin, yes, we'll review and get back to you shortly.

SAMtags.tex
```diff
@@ -66,6 +66,7 @@ \section{Standard tags}
   {\tt CT} & Z & Complete read annotation tag, used for consensus annotation dummy features.\\
   {\tt E2} & Z & The 2nd most likely base calls \\
   {\tt FI} & i & The index of segment in the template \\
+  {\tt FT} & Z & Filter field, indicating the reason why this record has been modified\\
```
tfenne: I'm not sure what it means here `has been modified`.  Is that supposed to say `has been filtered`?

SAMtags.tex
```diff
@@ -80,6 +81,7 @@ \section{Standard tags}
   {\tt MC} & Z & CIGAR string for mate/next segment\\
   {\tt MD} & Z & String for mismatching positions \\
   {\tt MF} & ? & Reserved for backwards compatibility reasons \\
+  {\tt MI} & Z & Molecular-barcode (corrected) sequence or identifier. May contain non sequence bases to encode more information (e.g., strandedness of original molecule) \\
```
tfenne: @nh13 and I introduced this tag in [fgbio](https://github.com/fulcrumgenomics/fgbio) several months ago.  The intended purpose is _not_ to contain sequence bases at all, but instead to contain a compact identifier that uniquely identifies the source molecule within the entire BAM without reference to other fields.  In our usage we just use serial numbers as identifiers.

I'd suggest changing the language to something like: `Molecular Identifier.  A unique ID within the SAM file for the source molecule from which this read is derived.  All reads with the sam MI tag represent the group of reads derived from the same source molecule.`

SAMtags.tex
```diff
@@ -90,11 +92,13 @@ \section{Standard tags}
   {\tt PQ} & i & Phred likelihood of the template \\
   {\tt PT} & Z & Read annotations for parts of the padded read sequence \\
   {\tt PU} & Z & Platform unit \\
-  {\tt QT} & Z & Barcode ({\tt BC} or {\tt RT}) phred-scaled base qualities \\
-  {\tt Q2} & Z & Phred quality of the mate/next segment sequence in the {\tt R2} tag \\
+  {\tt QT} & Z & Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag. Same encoding as {\sf QUAL}. \\
```
tfenne: Could we be explicit instead of saying "same encoding as QUAL" especially since that's now in a different document.  E.g. `Phred+33 ascii representation of....`

SAMtags.tex
```diff
@@ -90,11 +92,13 @@ \section{Standard tags}
...
+  {\tt QX} & Z & Phred quality of the molecular-barcode sequence in the {\tt RX} tag. Same encoding as {\sf QUAL}. \\
+  {\tt Q2} & Z & Phred quality of the mate/next segment sequence in the {\tt R2} tag. Same encoding as {\sf QUAL}.\\
   {\tt R2} & Z & Sequence of the mate/next segment in the template \\
   {\tt RG} & Z & Read group \\
   {\tt RT} & Z & Barcode sequence (deprecated; use {\tt BC} instead) \\
+  {\tt RX} & Z & Raw (uncorrected) molecular-barcode bases, with any quality scores stored in the {\tt QX} tag. \\
```
tfenne: I think we should add verbiage here about what to do if you have multiple molecular barcode reads.  In practice I believe most (or at least > 1) or us are concatenating multiple molecular barcode reads with hyphen separators.

SAMtags.tex
```diff
@@ -208,6 +212,16 @@ \subsection{Additional Template and Mapping data}
 \subsection{Metadata}
 
 \begin{description}
+
+\item[FT:Z:\tagvalue{text}] Filter field, free-text indicating the reason why this record has been modified or filtered
```
tfenne: Again, "modified" doesn't make any sense to me in this context.

SAMtags.tex
```diff
@@ -208,6 +212,16 @@ \subsection{Additional Template and Mapping data}
...
+There are two ways in which reads can be filtered: their 0x200 bit can be set (not passing filters) or their 0x4 bit can be set (segment unmapped).
```
tfenne: Eeek!  This is going to cause a lot of confusion.  Setting "segment unmapped" is _not_ "filtering" a read.  It's marking it as unmapped.  We should not conflate the two.

SAMtags.tex
```diff
@@ -208,6 +212,16 @@ \subsection{Additional Template and Mapping data}
 \subsection{Metadata}
 
 \begin{description}
+
+\item[FT:Z:\tagvalue{text}] Filter field, free-text indicating the reason why this record has been modified or filtered
+There are two ways in which reads can be filtered: their 0x200 bit can be set (not passing filters) or their 0x4 bit can be set (segment unmapped). 
+In either case it may be important or interesting to inform the users that this action was done after the vendor created the record and/or the aligner had been run. 
+The {\tt FT} tag is a place to put such information. 
+For example, Picard's MergeBamAlignment will unmap (unset 0x4) a read (and its mate) if it finds that it had been clipped on both ends, and it has less than some number of bases that are \emph{not} clipped.
```
tfenne: I'm not sure including examples from specific tools in a spec document is a great idea.

SAMtags.tex
```diff
@@ -235,12 +249,30 @@ \subsection{Barcodes}
...
+
+The {\tt RX} tag should match the {\tt QX} tag in length. 
+What one does with multiple molecular barcodes (e.g., one on each end of the template) is left to the implementation. 
+One current implementation concatenates all the barcodes and qualities with no separator between the different barcodes.
 
 \item[RT:Z:\tagvalue{sequence}]
 Deprecated alternative to {\tt BC} tag originally used at Sanger.
+
+\item[RX:Z:\tagvalue{sequence}] 
+Raw (uncorrected) molecular-barcode bases, with any quality scores stored in the {\tt QX} tag.
```
tfenne: I don't see a lot of value in limiting `RX` to be the "raw (uncorrected)" bases.  Or at least, if so, then we should define the tag for the corrected bases.  I have pipelines that go from raw -> corrected -> source molecule assignment, and each is a different logical step.

ablewhiskey:
@mp15, we've huddled at 10x to put together additional feedback:

RX and QX look reasonable to us and are consistent with the semantics in the BAM files our software produces.

For MI, we agree with @tfenne 's comment that this should be described as "Molecule identifier" rather than "Molecule barcode". This is the semantics in our BAM files, where we distinguish between the barcodes as read raw from the sequencer (RX), the barcodes after correction (BX), and molecule identifiers (MI) which are molecules inferred with the help of reference mapping.

**The main request we still have at this point is the inclusion of the BX tag to represent the corrected barcode.**

Here's more color from @pmarks:
In our case the BX tag is more fundamental to the underlying data than the MI tag, which is derived from the reference.  For example -- say you have a 1MB deletion: the same set of BX tags should be attached to the reads on each side of the deletion, but because of the way molecule inference works, there will be different MI tags on each side of the deletion _even though they are actually the same molecule_.  So if the spec is only telling you about BX and not about MI, you might get really confused.  For doing a phasing algorithm you could get away with using just the MI tag, but for any SV or assembly algorithm it's totally incorrect to use the MI tag instead of the BX tag.

nh13:
@yfarjoun bump

SAMtags.tex
nh13: 👍 for making this hyphens.  It is really important when you have multiple barcode bases joined from both reads, or attached at different stages in the prep process.

SAMtags.tex
yfarjoun: I'm thinking that FT is perhaps a bad tag for this. perhaps CO (COmment) would be better as it is akin to the @CO tag in the header, and you can put whatever free text you want there to explain something you did to the read.

SAMtags.tex
yfarjoun: OK. what about the qualities? should they also have hyphens between them? note that the hyphen is a legal quality score, so the split will have to occur on the bases first, and then use the location of the hyphens in the bases to split the qualities. It's probably simpler with something in the quality score...personally I prefer `~` as that is the highest quality and likely not to be used in practice.

SAMtags.tex
yfarjoun: sure.

SAMtags.tex
yfarjoun: and of-course CO already exists, so I'm removing this.

yfarjoun:
@tfenne, @ablewhiskey: I responded to your comments.

jkbonfield:
Having taken a look at this and seeing 32 comments in the issue (#109) and 54 in this PR, I'm loathe to spend the next hour trawling 2+ years of debate, not to mention my unfamiliarity with 10X & PacBio data tags.

What's the status of this PR?  Are there any outstanding issues to resolve with it?

jmarshall:
@jkbonfield: Just from the text being added in the PR, do you understand what the new tags mean and what they are intended to be used for?  There's two things to be reviewed here:

* Are the new tags still what the proponents are using / want?  (This is for 10X & PacBio to review)

* Do the descriptions of the new tags explain them adequately, and do the tags fit well in the existing SAM facilities?  (This is for you :smile:)

tfenne:
@jmarshall There are two different things going on in this PR, and they both need to be given consideration.  1) Is trying to ensure that the set of tags used by 10X (represented by @ablewhiskey ) are well documented and entered into the spec, 2) Is trying to ensure that the set of tags used by approaches that build molecular consensus reads (no _long_ reads) are also well represented.  The two overlap, but are not the same.  Also, I don't think there's anything in this PR that is PacBio related or any PacBio representation in the commenters.

I care more about (2) above.  @nh13 and I have developed a set of tools that are seeing use across a number of both academic and commercial groups.  See our tools for [assigning molecular IDs](https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala), [generating consensus reads](https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala), [generating duplex consensus reads](https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala) and lastly [filtering of consensus reads](https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala).

I think the PR as it stands is close to representing what we need, but needs some word-smithing (e.g. the one-line and paragraph descriptions of the `MI` tag don't sync up perfectly).  I'm happy to work with @yfarjoun to make those changes.

yfarjoun:
@tfenne, if you have suggestions I'll be grateful if you comment on the PR
or, if you prefer submit another commit to it.

SAMtags.tex
```diff
@@ -57,6 +57,7 @@ \section{Standard tags}
   {\tt AS} & i & Alignment score generated by aligner \\
   {\tt BC} & Z & Barcode sequence \\
   {\tt BQ} & Z & Offset to base alignment quality (BAQ) \\
+  {\tt BX} & Z & Corrected sequence bases for the unique molecular identifier \\
```
tfenne: Could this reference the RX tag, i.e. "for the unique molecular identifier (`RX`).

SAMtags.tex
```diff
@@ -80,27 +81,30 @@ \section{Standard tags}
   {\tt MC} & Z & CIGAR string for mate/next segment\\
   {\tt MD} & Z & String for mismatching positions \\
   {\tt MF} & ? & Reserved for backwards compatibility reasons \\
+  {\tt MI} & Z & Molecular-barcode (corrected) sequence or identifier. \\
```
tfenne: I think this needs to change to something like `An identifier which uniquely represents the source molecule within the BAM file without reference to other fields or tags`.  The way we (and I think 10X) are using this in practice is by taking the UMIs and position info, to determine molecular identity, then stamping each molecule's reads with a file-wide unique id.  That could be `chr:pos:pos:BX`, but it generally cannot be just the corrected sequence.  We (fgbio) use a serial number.

SAMtags.tex
```diff
@@ -80,27 +81,30 @@ \section{Standard tags}
...
   {\tt MQ} & i & Mapping quality of the mate/next segment \\
   {\tt NH} & i & Number of reported alignments that contains the query in the current record\\
   {\tt NM} & i & Edit distance to the reference \\
   {\tt OC} & Z & Original CIGAR \\
   {\tt OP} & i & Original mapping position \\
-  {\tt OQ} & Z & Original base quality \\
+  {\tt OQ} & Z & Original base quality.\\
```
tfenne: There's no consistency in whether or not lines end in periods, so why change just this one?

SAMtags.tex
```diff
@@ -234,13 +239,36 @@ \subsection{Barcodes}
 
 \begin{description}
 \item[BC:Z:\tagvalue{sequence}]
-Barcode sequence, with any quality scores stored in the {\tt QT} tag.
+Barcode sequence (Identifying the sample/library), with any quality scores stored in the {\tt QT} tag.
```
tfenne: -> "quality scores _optionally_ stored in the"

SAMtags.tex
```diff
@@ -234,13 +239,36 @@ \subsection{Barcodes}
...
+The {\tt BC} tag should match the {\tt QT} tag in length. 
+What one does with multiple sample barcodes (e.g. one on each end of the template) is left to the implementation. 
+One current implementation concatenates all the barcodes and qualities and places a dash (`{\tt -}') between the barcodes from the same template.
```
tfenne: I would change this from "one current implementation" to "the recommended implementation is to" since that will promote consistency, and that's already how multiple toolkits are doing it.

SAMtags.tex
```diff
@@ -234,13 +239,36 @@ \subsection{Barcodes}
...
+Barcode sequence (Identifying the sample/library), with any quality scores stored in the {\tt QT} tag.
+The {\tt BC} tag should match the {\tt QT} tag in length. 
+What one does with multiple sample barcodes (e.g. one on each end of the template) is left to the implementation. 
+One current implementation concatenates all the barcodes and qualities and places a dash (`{\tt -}') between the barcodes from the same template. 
+
+\item[BX:Z:\tagvalue{sequence}]
+Corrected bases from the unique molecular identifier. unlike MI, this may be non-unique in the file. Should comprise of sequence base possibly separated by hyphens as {\tt RX}.
+
+\item[MI:Z:\tagvalue{str}]
+Molecular Identifier. A unique ID within the SAM file for the source molecule/cell from which this read is derived. All reads with the same MI tag represent the group of reads derived from the same source molecule/cell.
```
tfenne: I'm not sure why "cell" found it's way in here.  A source molecule can only exist in one cell at a time, so I think it's redundant, but also confusing in that it could be interpreted as "for the source cell", which I don't think is right.

SAMtags.tex
jmarshall: There is consistency on the **master** branch: no periods in this table (except for one typo on the CT line).  So a minor editorial fix for this PR is to not add periods in its changes to this table.

jkbonfield:
Not entirely related to this PR, but observed while reading it:

I see Phred quality, phred probability and phred likelihood all being used in this document.  I'm pretty sure they all mean the same thing, so which term should we adopt universally?  "Quality" seems to be the most commonly used.

jkbonfield:
Making use of my excellent ignorance, I took at look at the wording to see whether it makes sense to someone "not in the know" as to how all these fit together.

"MI" is confusing.  It claims to be sequence **or** identifier, which presumably means it can be non-sequence.  However BX is then defined as the bases from the MI tag which presumably means MI is better described as "sequence plus optional identifier"?  However this is at odds with @tfenne's description of just using a serial number.   The documentation could benefit with some worked examples in section 1.3.

"BC" tag description and the wording around multiple barcodes needs tightening up as per Tim's suggestion.  His wording seems fine, or if preferred a middle ground such as "A common and recommended implementation is to ...".  The QT tag needs tightening up here too.  If separators are used in BC then the same separators should be used in QT for consistency.  Hopefully obvious though, but again an example helps clarify this.

I'd also reorder section 1.3 so that we have sample barcodes together (BC, RT, QC) followed by molecular barcodes (MI, BX, QX, RX) rather than interspersing them.  While alphabetical order in the primary list, we can collate in a more logical order when describing the subsets, as is already done via the headings.  It may also help to have a sentence at the start of 1.3 explaining single vs multiple barcodes (hence the SEQ vs SEQ-SEQ notation) and making the distinction explicit between sample vs molecular barcodes.

Addressing Tim's comment above about disliking raw vs corrected barcodes, I can see your view point but don't personally think it's too hard to check multiple places.   Do you expect systems that have no correction capability to store twice, both as final and original barcodes, or simply to store final only with original being implicit?  If the latter, then what do we do about quality (QX)?  It could in theory be a different level to the final barcode if the barcode contains error-correction codes that can cope with indels as well as snps.

tfenne:
@jkbonfield definitely good to have another viewpoint!

My point about raw/corrected vs. final/original was not that there should be one vs. two places for things, but just how to think about them.  I agree that in general it's not hard to `if (exists(a)) a else b`, but that doing so can become cumbersome, and that to some degree that's necessary in either case.  My thought was instead, let those dealing most closely with the UMIs have to do that, and let downstream users have a single interface.  I come back to base qualities as an example.  I'd wager that almost every tool that uses base qualities from SAM/BAM accesses the `QUAL` field, and is happy to take either raw or recalibrated qualities, whichever are found.  But imagine if quality score calibration only wrote the new qualities to a separate field, and then every downstream tool had to know to check for recal'd qualities first, then fallback to non-recal'd.  

To answer your question, I'd expect systems with no correction capability to store the UMI bases and quals in the `final` fields, and systems _with_ correction to store the raw bases/quals in the `raw` fields, and the corrected values in the `final` fields.

jkbonfield:
Fair comment and I can see (and agree) with that sentiment too.  I'm just unsure how it works with qualities.

It seems that we'd need a second quality string then, for final quality as well as raw quality.  I'm a bit unsure how that works though when error correcting.  I think we'd need feedback from 10X on this.

nh13:
I'll lend my support to @tfenne's proposal as we have worked on 3+ technologies that have raw and corrected bases, including long read data and others, and this has seemed the most natural across those different applications.

SAMtags.tex
```diff
@@ -234,13 +239,39 @@ \subsection{Barcodes}
 
 \begin{description}
 \item[BC:Z:\tagvalue{sequence}]
-Barcode sequence, with any quality scores stored in the {\tt QT} tag.
-
-\item[QT:Z:\tagvalue{qualities}]
-Phred quality of the barcode sequence in the {\tt BC} (or {\tt RT}) tag. Same encoding as {\sf QUAL}.
+Barcode sequence (Identifying the sample/library), with any quality scores (optionally) stored in the {\tt QT} tag.
+The {\tt BC} tag should match the {\tt QT} tag in length. 
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and qualities and places a dash (`{\tt -}') between the barcodes from the same template.
```
nh13: This sentence talks about " unique molecular identifiers" in the barcode sequence tag description.  I think `unique molecular identifiers` -> `barcode sequences`.  Also, consider a read structure: R1: `4B2S4B100T` and R2: `10B100T`.  Should the tag be `NNNN-NNNN-NNNNNNNNNN` or `NNNNNNNNNN-NNNNNNNNNN`.  I.e. should we delimit the reads or the template barcode segments?

SAMtags.tex
```diff
@@ -234,13 +239,39 @@ \subsection{Barcodes}
...
+Molecular Identifier. 
+A unique ID within the SAM file for the source molecule from which this read is derived. 
+All reads with the same MI tag represent the group of reads derived from the same source molecule. 
+
+\item[QT:Z:\tagvalue{qualities}] 
+Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag. 
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+See {\tt BC} for more detail.
+
+\item[QX:Z:\tagvalue{qualities}]
```
nh13: Can you move the `QX` description to after the `BX` description?

SAMtags.tex
```diff
@@ -234,13 +239,39 @@ \subsection{Barcodes}
...
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+The {\tt RX} and {\tt BX} tags should match the {\tt QX} tag in length. 
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and qualities with no separator between the different barcodes.
 
 \item[RT:Z:\tagvalue{sequence}]
 Deprecated alternative to {\tt BC} tag originally used at Sanger.
+
+\item[RX:Z:\tagvalue{sequence}] 
+Raw (uncorrected) unique molecular identifier bases, with any quality scores (optionally) stored in the {\tt QX} tag. 
+See {\tt QX} for more detail.
```
nh13: I'd echo @jkbonfield's comment about having two tags storing bases (raw/corrected) but only one tag for qualities.

SAMtags.tex
```diff
@@ -234,13 +239,39 @@ \subsection{Barcodes}
...
-\item[QT:Z:\tagvalue{qualities}]
-Phred quality of the barcode sequence in the {\tt BC} (or {\tt RT}) tag. Same encoding as {\sf QUAL}.
+Barcode sequence (Identifying the sample/library), with any quality scores (optionally) stored in the {\tt QT} tag.
+The {\tt BC} tag should match the {\tt QT} tag in length. 
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and qualities and places a dash (`{\tt -}') between the barcodes from the same template. 
+
+\item[BX:Z:\tagvalue{sequence}]
+Sequence bases from the unique molecular identifier. 
+These could be either corrected or uncorrected. Unlike MI, this may be non-unique in the file. 
+Should comprise of sequence base possibly separated by hyphens as {\tt RX}.
```
nh13: `of sequence bases` -> `of a sequence of bases`?

SAMtags.tex
```diff
@@ -234,13 +239,39 @@ \subsection{Barcodes}
...
+\item[QT:Z:\tagvalue{qualities}] 
+Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag. 
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+See {\tt BC} for more detail.
+
+\item[QX:Z:\tagvalue{qualities}] 
+Phred quality of the unique molecular identifier sequence in the {\tt RX} or {\tt BX} tag. 
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+The {\tt RX} and {\tt BX} tags should match the {\tt QX} tag in length. 
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and qualities with no separator between the different barcodes.
```
nh13: Ditto as UMIs/MIDs don't have to be one per read.

SAMtags.tex
jkbonfield: This is ambiguous.    It states we "concatenate all the barcodes and qualities" and then "places a dash between the barcodes", but not explicitly "barcodes and qualities".    This could imply that our sequence has separators but not the qualities, but this is at odds with the sentence "the BC tag should match the QT tag in length", unless length here defines the number of nucleotides and not our specific textual encoding.

Also given there are already differences between implementations, perhaps it is better to say:

> The recommended implementation when writing these tags is to place a dash between the barcodes from the same template, and when reading to interpret non-nucleotides or ambiguity codes as a separator.

If we expect the separators to be present in the QT too then we need to explain that the BC tag must be used in conjunction with QT in order to identify which are quality values and which are separators.  If this isn't the case then it should be spelt out explicitly.

SAMtags.tex
jkbonfield: BX and RX are defined to be sequence bases possibly separated by dashes.  QX is defined to be concatenated without separators.  However we have a requirement that all are the same length.

Firstly, this wording needs correcting to say they contain the same number of base calls as "length" is ambiguous.

Secondly if BX is corrected then RX is the raw uncorrected sequence.  Requiring these to be the same length is stating that we will never support correction of tags on machines with higher indel rates than SNP rates.  I do not see the requirement for BX and RX to be the same length.

dkj: I don't think `BX` needs to be sequence bases: in the examples that I've seen it happens to be  the corrected sequence bases (of `RX`) with a `-1` suffix (where the 1 would be bumped to 2 for one library if you used two separately created libraries for the same sample in the same analysis).

I think `BX` is used to represent known uniqueness of the long molecule partitioning technology: reads from the same long molecule, will have the same partition, and so same `BX` identifier, but multiple long molecules may have the same `BX` identifier.

(Other information like alignments to a reference can be used to mark sets of reads which are very likely to have come from a particular molecule and such a set would be marked with `MI`: reads from a single molecule might have 0,1 or more `MI` values, but reads from two different long molecules should not have a common `MI`). 

[related comment above](#issuecomment-249290370)

nh13: Long molecule technology is not the only application for these tags.  In fact, there are a ton of others that will use these tags for sequencing with UMIs/MIDs.  `MI` may store some string unique to the source molecule, `BX` could be the corrected/consensus set of bases for reads determined to come from the same long molecule, and `RX` could be the raw set of bases for each read that identifies the source molecule.  See tools that already do this like [UMI-tools](https://github.com/CGATOxford/UMI-tools) and `GroupReadsByUmi/CallMolecularConsensusReads/CallDuplexConsensusReads` in [fgbio](https://github.com/fulcrumgenomics/fgbio).

SAMtags.tex
yfarjoun: @nh13 I was thinking between barcodes, but I have little experience with more than one barcode, so that's just mirroring how I implemented it in Picard. if you have a different suggestion, please let me know. 

Regarding concatenating qualities. @nh13 what is your opinion on this? I think for parsing it is easier to have a separator in the quality field. but I am reluctant to use a `-` since that's a valid quality character. We could use a character that is not legal for quality (for example a space ` `) that will make parsing a little simpler. 

Regarding length vis-a-vis indels. @jkbonfield I'm not sure I understand how indel have anything to do with it. Let's assume the barcode is truly AACG and the sequencer read AAGT (skipping the C and reading into the next nucleotide.) and assume further that the correcting algorithm is smart enough to figure this out, it emit AACG as the corrected barcode....still the same length.

SAMtags.tex
yfarjoun: should there be two tags for quality? there's only one quality. it came from the sequencer...
this is the opposite of BQSR's OQ, there are two places to look for quality but only one place to look for bases..

SAMtags.tex
yfarjoun: sure.

SAMtags.tex
yfarjoun: @dkj I know that this how 10X uses it, but IMHO the -1 -2 are redundant. since they inform the library which is known form the RG tag, we do not need it in the BX tag.

yfarjoun: @nh13 it isn't clear who that comment is for 😄

SAMtags.tex
nh13: @yfarjoun if you're asking me, I would argue for concatenating all bases within a read without a delimiter, and concatenating across reads with a delimiter.  But there's not clear answer.  Regarding qualities, I have no preference but I see the dilemma.

SAMtags.tex
nh13: I'd say it's up to the tool developer whether the corrected (or consensus called) barcode has updated qualities.  Perhaps you have more confidence in the corrected bases now that you have seen it many times.  It's the same reason to have updated qualities when consensus calling the template bases.

SAMtags.tex
jkbonfield: @yfarjoun I see your point that if you define the tag to be N called bases then if it's wrongly called you just chose a different N bases potentially missing part of the tag or incorporating bases beyond the tag. That way we correct for the indels via sequence alignment and assign the correct tag without changing length.  Good point!

Potentially we could have an instrument, eg some pyrosequencing method, that defined variable length tags in terms of a fixed number of "flows" (so AAA and A are both 1 flow), but this is hypothetical and even the 454 barcoding worked on fixed base count and not fixed flow count (I think).

So it's probably fine, although I'm curious what we gain by adding the restriction of being the same length.

yfarjoun: @jkbonfield if the qualities and the bases are not the same length, then which base does each quality refer to?

SAMtags.tex
```diff
@@ -55,9 +55,10 @@ \section{Standard tags}
   \hline
   {\tt AM} & i & The smallest template-independent mapping quality of segments in the rest \\
   {\tt AS} & i & Alignment score generated by aligner \\
-  {\tt BC} & Z & Barcode sequence \\
+  {\tt BC} & Z & Barcode sequence identifying the sample \\
   {\tt BQ} & Z & Offset to base alignment quality (BAQ) \\
-  {\tt BX} & Z & Corrected sequence bases for the unique molecular identifier \\
+  {\tt BX} & Z & Sequence bases of the unique molecular identifier \\
+  {\tt BZ} & Z & Quality score the unique molecular identifier. See {\tt BX} tag for qualities \\ \\
```
jkbonfield: Quality score **of** the unique ...

SAMtags.tex
jkbonfield: My point is that we state we string base barcodes together with dashes but not the qualities, and then also state the BC and QT must be the same length.   I understand the intention - that length here is in number of base calls and not length of string, but it is ambiguous.  I don't mean that we *should* have different lengths.

Specifically it is this sentence: "The {\tt BC} tag should match the {\tt QT} tag in length. In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and qualities and places a dash (`{\tt -}') between the barcodes from the same template."

There is no corresponding statement in QT that it will also have dashes between barcodes from the same template.  Adding them here may also be problematic as dash is a valid quality (unlike a valid base-call).  We need to be explicit - does QT also have the dashes?  If so state this.  If not we should rephrase the comment about BC and QT being the same length to something like "The number of nucleotides in the {\tt BC} tag should match the length of the {\tt QT} tag."

yfarjoun: I modified the text a little. Now the lengths will really be the same.

jmarshall:
Before merging, please squash and rebase into two commits:

1. The BC description change and the new tags (this is the meat of the PR)
2. The unrelated whitespace and typo changes to CM/CT/IH/etc (this is unrelated editorial and could just be committed separately directly to **master**)

SAMtags.tex
```diff
@@ -55,15 +55,17 @@ \section{Standard tags}
   \hline
   {\tt AM} & i & The smallest template-independent mapping quality of segments in the rest \\
   {\tt AS} & i & Alignment score generated by aligner \\
-  {\tt BC} & Z & Barcode sequence \\
+  {\tt BC} & Z & Barcode sequence identifying the sample \\
   {\tt BQ} & Z & Offset to base alignment quality (BAQ) \\
+  {\tt BX} & Z & Sequence bases of the unique molecular identifier \\
+  {\tt BZ} & Z & Quality score the unique molecular identifier. See {\tt BX} tag for qualities \\ \\
```
jmarshall: Too much `\\`

SAMtags.tex
jmarshall: “Hyphen”, not “dash”.

jkbonfield:
Thumbs up after the trivial typographical tweaks in the above couple comments are applied.

SAMtags.tex
```diff
@@ -234,13 +240,46 @@ \subsection{Barcodes}
...
+\item[QT:Z:\tagvalue{qualities}] 
+Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag. 
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+See {\tt BC} for more detail.
+
+\item[BX:Z:\tagvalue{sequence}]
+Sequence bases from the unique molecular identifier. 
+These could be either corrected or uncorrected. Unlike MI, this may be non-unique in the file. 
+Should comprise of a sequence of bases possibly separated by hyphens as {\tt RX}.
+In case that the bases represent the corrected bases, the original sequence can be stored in {\tt RX} (similar to OQ storing the original qualities of bases.)
```
jmarshall: `\tt` for MI and OQ, and also for MI down in the `MI` paragraph.

jmarshall:
Several typographic suggestions as noted, and also please `s/dash/hyphen/` × 2

yfarjoun:
I've replaced this PR with a separate one. I didn't want to overwrite all the commits here and also hide the comments...so I'll just close this. Please continues the discussion in the other PR...

#200 

Thanks.

jkbonfield:
Any of the original people involved in this discussion, we welcome your views on how to resolve this long standing issue over here:

https://github.com/samtools/hts-specs/pull/200#issuecomment-305512849

Issue #107, opened 2015-11-13
VCF INFO and genotype reserved fields


cyenyxe:
I was wondering how to handle the case when reserved info/genotype fields have a contradictory declaration in the meta section.

For instance, if the DP field (Integer) was declared in the meta section as Float, and the values in the body matched correctly the Float type, would it be considered correct or incorrect?

What about restrictions that are not covered in the specification, like multiplicity? Using DP as an example again, the description makes it seem like a **single** Integer, but what if the meta said it had 2 values and the body satisfied that condition?

If these reserved fields have to follow some strict rules, it would be nice to add them to the document.

pd3:
The reserved tags must be precisely as described by the specification, therefore both the type and the number of fields matters. For example, we require that INFO=DP is Type=Integer and Number=1. If this is not obvious, I guess we should state it explicitly.

cyenyxe:
I would be fine with doing it myself. Should I fork the branch VCFv4.3?

pd3:
That would be excellent if you could do that. Please use the master branch, the VCFv4.3 branch will be deleted.

cyenyxe:
I have structured the fields description in a table (see 5df6aa5a). Please check that I got all of them right, and if you like the tabular style I will do the same for the genotype fields. Some of those are more complex (GT, GL, PS) so they will require some thought about how to display the comments along with the table.

I was also wondering whether it would be interesting to place the reserved keys for structural variants in the same section too. Maybe in another sub-section, but still closer than current sections 3 and 4.

On a side note, I had to comment the line ```\input{VCFv4.3.ver}``` to be able to compile the tex file. After seeing the output, the ver file doesn't seem necessary, but it was present in older commits.

jmarshall:
> I had to comment the line \input{VCFv4.3.ver} to be able to compile the tex file.

Use the Makefile to generate this version information file and build _VCFv4.3.pdf_.

pd3:
Thank you Cristina. There seems to be a forgotten `\end{enumerate}` and CIGAR should be Number=A, otherwise it looks good, +1 from me.

cyenyxe:
What is the cardinality of the PQ and PS genotype subfields?

pd3:
Both should be `Number=1`. This is because the phase is transitive: if A-B are phased and B-C are phased, also A-C are phased.

cyenyxe:
[...]

PR #123, opened 2016-01-15
VCF INFO and SAMPLE reserved fields

cyenyxe:
New tables used instead of plain text to describe strictly the type and number for INFO and SAMPLE reserved fields.

Related to #107

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -344,7 +344,7 @@ There are 8 fixed fields per record.  Fixed fields are:
 	AF		& A		& Float		& Allele frequency for each ALT allele in the same order as listed (use this when estimated from primary data, not called genotypes) \\
 	AN		& 1		& Integer	& Total number of alleles in called genotypes \\
 	BQ   		& 1		& Float		& RMS base quality \\
-	CIGAR		& 1		& String	& Cigar string describing how to align an alternate allele to the reference allele \\
+	CIGAR		& A		& String	& Cigar string describing how to align an alternate allele to the reference allele \\
 	DB		& 0		& Flag		& dbSNP membership \\
 	DP		& 1		& Integer	& Combined depth across samples \\
 	END		& 1		& Integer	& End position (for use with symbolic alleles) \\
@@ -362,8 +362,10 @@ There are 8 fixed fields per record.  Fixed fields are:
     \label{table:reserved-info}
   \end{table}
   
-The exact format of each INFO sub-field should be specified in the meta-information (as described above).
-Example for an INFO field: DP=154;MQ=52;H2. Keys without corresponding values are allowed in order to indicate group membership (e.g. H2 indicates the SNP is found in HapMap 2). It is not necessary to list all the properties that a site does NOT have, by e.g. H2=0. See Section~\ref{sv-info-keys} for additional reserved INFO sub-fields used to encode structural variants.
+  The exact format of each INFO sub-field should be specified in the meta-information (as described above).
+  Example for an INFO field: DP=154;MQ=52;H2. Keys without corresponding values are allowed in order to indicate group membership (e.g. H2 indicates the SNP is found in HapMap 2). It is not necessary to list all the properties that a site does NOT have, by e.g. H2=0. See Section~\ref{sv-info-keys} for additional reserved INFO sub-fields used to encode structural variants.
+\end{enumerate}
+  
 \subsubsection{Genotype fields}
 If genotype information is present, then the same types of data must be present
 for all samples. First a FORMAT field is given specifying the data types and
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -390,8 +390,8 @@ See table~\ref{table:reserved-genotypes} for a list of common, reserved keywords
       HQ		& 2		& Integer	& Haplotype (phred) qualities \\
       MQ		& 1		& Integer	& RMS mapping quality \\
       PL		& G		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer (and otherwise defined precisely as the GL field) \\
-      PQ		& ?		& Integer	& Phasing quality, the phred-scaled probability that alleles are ordered incorrectly in a heterozygote (against all other members in the phase set). More details in Subsection~\ref{para:genotype-PQ} \\
-      PS		& ?		& Integer (non-negative)	& Phase set, set of phased genotypes to which this genotype belongs \\
+      PQ		& 1		& Integer	& Phasing quality, the phred-scaled probability that alleles are ordered incorrectly in a heterozygote (against all other members in the phase set). More details in Subsection~\ref{para:genotype-PQ} \\
+      PS		& 1		& Integer (non-negative)	& Phase set, set of phased genotypes to which this genotype belongs \\
   \end{tabularx}
   \caption{Reserved genotype fields}
   \label{table:reserved-genotypes}
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -385,7 +385,7 @@ See table~\ref{table:reserved-genotypes} for a list of common, reserved keywords
       FT		& 1		& String	& Filter indicating if this genotype was ``called''. More details in Subsection~\ref{para:genotype-FT} \\
       GL		& G		& Float		& $log_{10}$-scaled likelihoods for all possible genotypes given the set of alleles defined in the REF and ALT fields. More details in Subsection~\ref{para:genotype-GL} \\
       GP		& G		& Float		& Genotype posterior probabilities in the range 0 to 1 using the same ordering as the GL field (could be used to store imputed genotype probabilities) \\
-      GQ		& 1		& Float		& Conditional genotype quality, encoded as a phred quality $-10log_{10}$ p(genotype call is wrong, conditioned on the site's being variant) \\
+      GQ		& 1		& Integer	& Conditional genotype quality, encoded as a phred quality $-10log_{10}$ p(genotype call is wrong, conditioned on the site's being variant) \\
       GT   		& 1		& Integer	& Genotype, encoded as allele values separated by either of $/$ or $\mid$. \\
       HQ		& 2		& Integer	& Haplotype (phred) qualities \\
       MQ		& 1		& Integer	& RMS mapping quality \\
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -367,99 +367,107 @@ There are 8 fixed fields per record.  Fixed fields are:
 \end{enumerate}
   
 \subsubsection{Genotype fields}
-If genotype information is present, then the same types of data must be present
-for all samples. First a FORMAT field is given specifying the data types and
-order (colon-separated FORMAT ids matching the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, duplicate fields are not allowed). This is followed by one data block per
-sample, with the colon-separated data corresponding to the types
-specified in the format. The first sub-field must always be the genotype (GT)
-if it is present.  There are no required sub-fields.
+If genotype information is present, then the same types of data must be present for all samples. First a FORMAT field is given specifying the data types and order (colon-separated FORMAT ids matching the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, duplicate fields are not allowed). This is followed by one data block per sample, with the colon-separated data corresponding to the types specified in the format. The first sub-field must always be the genotype (GT) if it is present. There are no required sub-fields. Additional Genotype fields can be defined in the meta-information, however, software support for such fields is not guaranteed.
 
-As with the INFO field, there are several common, reserved keywords that are standards across the community:
+If any of the fields is missing, it is replaced with the missing value. For example if the FORMAT is GT:GQ:DP:HQ then $0\mid0:.:23:23,34$ indicates that GQ is missing. Trailing fields can be dropped, with the exception of the GT field, which should always be present if specified in the FORMAT field.
 
+See table~\ref{table:reserved-genotypes} for a list of common, reserved keywords that are standards across the community, and the following points for detailed explanations about the most complex among them. See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for structural variants.
+
+\begin{table}[htbp]
+  \centering
+  \begin{tabularx}{\textwidth}{ | l | l | l | X | }
+      Field		& Number	& Type		& Description \\ \hline
+      AD		& R		& Integer	& Per-sample read depth for each allele \\
+      ADF		& R		& Integer	& Per-sample read depth for each allele on the forward strand \\
+      ADR		& R		& Integer	& Per-sample read depth for each allele on the reverse strand \\
+      DP		& 1		& Integer	& Per-sample read depth \\
+      EC		& A		& Integer	& Expected alternate allele counts for each alternate allele, in the same order as listed in the ALT field (typically used in association analyses) \\
+      FT		& 1		& String	& Filter indicating if this genotype was ``called''. More details in Subsection~\ref{para:genotype-FT} \\
+      GL		& G		& Float		& $log_{10}$-scaled likelihoods for all possible genotypes given the set of alleles defined in the REF and ALT fields. More details in Subsection~\ref{para:genotype-GL} \\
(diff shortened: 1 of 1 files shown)


Lenbok:
Note that this PR looks to change the definition of GQ from being Integer to Float, which is being discussed separately at #129

yfarjoun:
given the description of this PR, there should be no change in content. it is somewhat misleading to put a change in definition in a PR that claims to be a formatting PR. 👎

cyenyxe:
These commits are months older than #129, so it was probably a mistake rather than an intentional change. I've pushed a new commit changing it back to Integer.

jmarshall:
I'm not totally convinced about the tables: they deemphasise the description and encourage not writing very much of it, but the description is the important part.  And the tables float quite a confusingly long way away from the text they relate to (though this goes better if we remove the `\newpage \tableofcontents \newpage` at the start or otherwise change the position on the page).

However a very big :+1: to specifying the expected Number and Type of these fields.

We have just had a bcftools bug (samtools/bcftools#423) wherein `bcftools norm -m-` produces incorrect results and `bcftools view -s` crashes, all due to an AC info field not having the expected `Number=A`… but the current spec doesn't tell you what Number AC should have!

cyenyxe:
Then maybe the table could serve as reference to construct a sensible meta-data section, with or without the "Description" field, which is usually quite short anyway.

But it's true that a paragraph per field would give a chance to clarify some fields that currently are just a bullet point / row in the table. I could see at least AA and SB benefiting from that, both in terms of purpose and accepted values.

jmarshall:
For comparison, jmarshall/hts-specs@935fd9762d31596f115a9442f7d2987b594db8f6 tidies this up using `\begin{description}` rather than a table.  Description vs. table is mostly a matter of taste and YMMV, and perhaps tables would be fine if you forced them to appear at their position within the text rather than floating to later.

pd3:
+1

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
-  \item $\ldots$ see Section~\ref{sv-info-keys} for a list of INFO keys reserved for structural variants.
-\end{itemize}
+  INFO fields are encoded as a semicolon-separated series of short keys with optional values in the format: $<$key$>$=$<$data$>$[,data]. 
+  INFO keys must match the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, duplicate fields are not allowed. Arbitrary keys are permitted, although the sub-fields listed in Table~\ref{table:reserved-info} are reserved (albeit optional).
+  
+  \begin{table}[htbp]
+    \centering
+    \begin{tabularx}{\textwidth}{ | l | l | l | X | }
+	Field		& Number	& Type		& Description \\ \hline
+	AA		& 1		& String	& Ancestral allele \\
```
yfarjoun: I"m not sure what this means, now that it has a Number=1..before it was not defined.

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
+	AC		& A		& Integer	& Allele count in genotypes, for each ALT allele, in the same order as listed  \\
+	AD		& R		& Integer	& Total read depth for each allele \\
+	ADF		& R		& Integer	& Read depth for each allele on the forward strand \\
+	ADR		& R		& Integer	& Read depth for each allele on the reverse strand \\
+	AF		& A		& Float		& Allele frequency for each ALT allele in the same order as listed (use this when estimated from primary data, not called genotypes) \\
+	AN		& 1		& Integer	& Total number of alleles in called genotypes \\
+	BQ   		& 1		& Float		& RMS base quality \\
```
yfarjoun: You shortened the description, but I wonder if it should actually be lengthened: of which reads? reference _and_ ALT? only _ALT_? should this be per-allele? is it allowed to be per-allele? you specify that it is a 1:float, but that wasn't in the original..

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
+	CIGAR		& A		& String	& Cigar string describing how to align an alternate allele to the reference allele \\
+	DB		& 0		& Flag		& dbSNP membership \\
```
yfarjoun: This originally was not number=0, and could have been an A-flag per-allele. The introduction of this table sets it to a particular use-case.

yfarjoun:
Given that this PR is supposed to not change content, I think that many (most?) of the values in "number" in table 1 should be `.` (that is unbound) since that's what the original has. I commented on several, but that got repetitive.

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
-for all samples. First a FORMAT field is given specifying the data types and
-order (colon-separated FORMAT ids matching the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, duplicate fields are not allowed). This is followed by one data block per
-sample, with the colon-separated data corresponding to the types
-specified in the format. The first sub-field must always be the genotype (GT)
-if it is present.  There are no required sub-fields.
+If genotype information is present, then the same types of data must be present for all samples. First a FORMAT field is given specifying the data types and order (colon-separated FORMAT ids matching the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, duplicate fields are not allowed). This is followed by one data block per sample, with the colon-separated data corresponding to the types specified in the format. The first sub-field must always be the genotype (GT) if it is present. There are no required sub-fields. Additional Genotype fields can be defined in the meta-information, however, software support for such fields is not guaranteed.
+
```
yfarjoun: I'm not sure what the purpose of "Additional Genotype fields can be defined in the meta-information, however, software support for such fields is not guaranteed." is.

yfarjoun: nm. I see it moved from below.

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
+      GP		& G		& Float		& Genotype posterior probabilities in the range 0 to 1 using the same ordering as the GL field (could be used to store imputed genotype probabilities) \\
+      GQ		& 1		& Integer	& Conditional genotype quality, encoded as a phred quality $-10log_{10}$ p(genotype call is wrong, conditioned on the site's being variant) \\
+      GT   		& 1		& Integer	& Genotype, encoded as allele values separated by either of $/$ or $\mid$. \\
+      HQ		& 2		& Integer	& Haplotype (phred) qualities \\
+      MQ		& 1		& Integer	& RMS mapping quality \\
+      PL		& G		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer (and otherwise defined precisely as the GL field) \\
+      PQ		& 1		& Integer	& Phasing quality, the phred-scaled probability that alleles are ordered incorrectly in a heterozygote (against all other members in the phase set). More details in Subsection~\ref{para:genotype-PQ} \\
+      PS		& 1		& Integer (non-negative)	& Phase set, set of phased genotypes to which this genotype belongs \\
```
yfarjoun: In the wild I am pretty sure that people use a string here and not an integer... but perhaps this isn't the PR to change it.

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
+      ADF		& R		& Integer	& Per-sample read depth for each allele on the forward strand \\
+      ADR		& R		& Integer	& Per-sample read depth for each allele on the reverse strand \\
+      DP		& 1		& Integer	& Per-sample read depth \\
+      EC		& A		& Integer	& Expected alternate allele counts for each alternate allele, in the same order as listed in the ALT field (typically used in association analyses) \\
+      FT		& 1		& String	& Filter indicating if this genotype was ``called''. More details in Subsection~\ref{para:genotype-FT} \\
+      GL		& G		& Float		& $log_{10}$-scaled likelihoods for all possible genotypes given the set of alleles defined in the REF and ALT fields. More details in Subsection~\ref{para:genotype-GL} \\
+      GP		& G		& Float		& Genotype posterior probabilities in the range 0 to 1 using the same ordering as the GL field (could be used to store imputed genotype probabilities) \\
+      GQ		& 1		& Integer	& Conditional genotype quality, encoded as a phred quality $-10log_{10}$ p(genotype call is wrong, conditioned on the site's being variant) \\
+      GT   		& 1		& Integer	& Genotype, encoded as allele values separated by either of $/$ or $\mid$. \\
```
yfarjoun: add comment "see more below" (or similar pointer to more information)

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
+      HQ		& 2		& Integer	& Haplotype (phred) qualities \\
+      MQ		& 1		& Integer	& RMS mapping quality \\
+      PL		& G		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer (and otherwise defined precisely as the GL field) \\
+      PQ		& 1		& Integer	& Phasing quality, the phred-scaled probability that alleles are ordered incorrectly in a heterozygote (against all other members in the phase set). More details in Subsection~\ref{para:genotype-PQ} \\
```
yfarjoun: given that the PQ isn't well defined, I hesitate to set this as number=1 here. while this should be true for diploid samples, one could think of phasing in higher ploidy organisms and there it would be a more complicated object. I think that there should be more freedom in the spec, especially when we are not sure of it's use. (same for PS)

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
+      AD		& R		& Integer	& Per-sample read depth for each allele \\
+      ADF		& R		& Integer	& Per-sample read depth for each allele on the forward strand \\
+      ADR		& R		& Integer	& Per-sample read depth for each allele on the reverse strand \\
+      DP		& 1		& Integer	& Per-sample read depth \\
+      EC		& A		& Integer	& Expected alternate allele counts for each alternate allele, in the same order as listed in the ALT field (typically used in association analyses) \\
+      FT		& 1		& String	& Filter indicating if this genotype was ``called''. More details in Subsection~\ref{para:genotype-FT} \\
+      GL		& G		& Float		& $log_{10}$-scaled likelihoods for all possible genotypes given the set of alleles defined in the REF and ALT fields. More details in Subsection~\ref{para:genotype-GL} \\
+      GP		& G		& Float		& Genotype posterior probabilities in the range 0 to 1 using the same ordering as the GL field (could be used to store imputed genotype probabilities) \\
+      GQ		& 1		& Integer	& Conditional genotype quality, encoded as a phred quality $-10log_{10}$ p(genotype call is wrong, conditioned on the site's being variant) \\
```
yfarjoun: can you fix the typo and remove the `'s` while we're here?

yfarjoun: This made me curious: Does any caller actually calculate this as written? GATK simply puts the difference of the two best PLs...which is more like the odds-ratio of the best two genotypes, not the P(error).

yfarjoun:
I reviewed. put my comments in. Sorry for the late review. @cyenyxe: back to you.

cyenyxe:
In general, the Number and Type are defined based on:

1. The examples already in the spec document (as in page 1)
2. The descriptions: if they refer to a field in singular then Number=1, when they mention "per alternate allele" then Number=A, etc.
3. Checking well-known datasets, particularly 1000 Genomes, since the format was initially developed for them

Then there was some discussion in #107 about some of the fields, although some still seem to be unclear (PQ).

I'm not sure keeping *everything* unbounded would be beneficial. Some fields have been historically used by most (if not all) studies in a consistent way following the examples in the spec and/or the rules established by 1kG. Making all of them unbounded could bring more chaos, because people wouldn't assume the "implicit" Number any more. I will comment on all those you highlighted, but in general I think it's preferable to try to set a definition, unless a field use-case is really unclear.

VCFv4.3.tex
cyenyxe: Extracted from example in page 1 of the spec, and also http://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase1/analysis_results/supporting/ancestral_alignments/human_ancestor_GRCh37_e59.README for mode details.

VCFv4.3.tex
cyenyxe: Extracted from example in page 1 of the spec.
I agree it wouldn't be the case for multi-allelic variants though. This seems to be a conflict between how the Flag type and the rest work: using Number=0 supports only one use-case, but Number=A hides the "no value needed" consideration...

yfarjoun:
I shouldn't have said "unbounded", what I meant to say is that this PR, as it is claiming to be formatting only, should leave the tags that are not specified as such. It will be much harder to change the specification once specified, so we should leave things as they were so that a proper discussion about each tag can happen.

VCFv4.3.tex
yfarjoun: for cases where Number=1 and type=flag, that consideration applies, but multiallelics are a large component of todays call-sets (see ExAC for example).

VCFv4.3.tex
yfarjoun: I see. OK. AA is fine then.

yfarjoun:
@cyenyxe you are right, not everything should be unbound, but (in this PR) we should err on the side of not specifying Number if it wasn't explicitly specified in the original document and we can think of a use-case where Number can be A or R in addition to 0 or 1. Many of the studies were done when multi-allelic variants were not very common. Now have the choice of using many of the tags as multi-allelic, or having to create new tags rendering the old ones un-used. I think that this PR is not being billed as the place for that discussion.

jmarshall:
> this PR, as it is claiming to be formatting only

This PR is **not** claiming to be formatting only.  It is claiming to be describing strictly the type and number for INFO and SAMPLE reserved fields, and incidentally it also proposes some formatting.

However I agree with you that there may be some of these that are less obvious than others and we need to be sure before going ahead and specifying them.  Perhaps we can merge the obvious ones and continue research for the problematic ones (which are…?)?

VCFv4.3.tex
pd3: The language of the specification thus far has been to mean all bases unless explicitly saying it's per-allele, or ref vs alt. So this is correct.

VCFv4.3.tex
pd3: I don't think a vector Flag is allowed, and is not supported by htslib as far as I am aware.

yfarjoun:
Indeed, the specification states

   1.

   The ‘Flag’ type indicates that the INFO field does not contain a Value
   entry, and hence the Number should be 0 in this case.


I'm not sure why one cannot have a boolean per allele or per non-reference
allele, or a unbounded list of booleans..seems rather restrictive, but that
can wait for when it's needed (if it's needed).

VCFv4.3.tex
```diff
@@ -326,132 +327,147 @@ \subsubsection{Fixed fields}
...
+	AF		& A		& Float		& Allele frequency for each ALT allele in the same order as listed (use this when estimated from primary data, not called genotypes) \\
+	AN		& 1		& Integer	& Total number of alleles in called genotypes \\
+	BQ   		& 1		& Float		& RMS base quality \\
+	CIGAR		& A		& String	& Cigar string describing how to align an alternate allele to the reference allele \\
+	DB		& 0		& Flag		& dbSNP membership \\
+	DP		& 1		& Integer	& Combined depth across samples \\
+	END		& 1		& Integer	& End position (for use with symbolic alleles) \\
+	H2		& 0		& Flag		& HapMap2 membership \\
+	H3		& 0		& Flag		& HapMap3 membership \\
+	MQ		& 1		& Integer	& RMS mapping quality \\
```
yfarjoun: MQ (RMS) should be a float, no?

jmarshall: Re _MQ (RMS)_, are you suggesting `float` based on what one would think an _RMS_ value would be, or based on what you have seen in VCF files in the wild?

yfarjoun: Both.

RMS is a floating point value, and also GATK emits a float rather than a
integer. See here, from a recent production BAM:
```

##INFO=<ID=MQ,Number=1,Type=Float,Description="RMS Mapping Quality">

```

cyenyxe: That is what I thought too, but then the spec contains these two lines:

* **INFO:** MQ : RMS mapping quality, e.g. MQ=52
* **FORMAT:** MQ : RMS mapping quality, similar to the version in the INFO field. (Integer)

yfarjoun: I haven't seen MQ in the format field...so I don't know how it's used.

cyenyxe:
Superseded by #185 and #192.

Issue #104, opened 2015-11-03
What is the utility of 'encouraged' and 'not required' language


BradHolmes:
After looking over the v4.2 specification, I have been unable to get clear direction for a couple of questions, and in both cases, it is because the specification contains language like
- encouraged
- strongly encouraged
- not required

What is the utility in allowing the user to choose, and thus produce varying files for the same data?  Could we have v4.3 come down one way or another on these, and use 'must'?

d-cameron:
If the wording is to be updated, I would RECOMMEND using RFC2119 (
http://www.ietf.org/rfc/rfc2119.txt) language to indicate the conformance
level required by implementations.

cyenyxe:
After reading the whole specification (v4.3), the document listing all the sentences that could be considered ambiguous or non-compliant with RFC2119 is available in: https://docs.google.com/document/d/1_hFSImtpi6mHaYfn3twprPP3XOJBXqV0eWwVUh-Ctig/edit?usp=sharing

Any comments welcome (and allowed in the doc itself).

dotdotdotpaul:
+1 for moving towards unambiguous language

cyenyxe:
After a talk with the specification team I filled the PR #130. Most of the points made in the Google Docs were accepted, but some will still need more thought.

Could any of you guys take a look at the PR and give it a +1 if you think it's okay to go, or comment any possible improvements?

cyenyxe:
Fixed in #130

PR #127, opened 2016-01-27
Updated verbs to RFC2119 wording

d-cameron:
Potential update to address feedback from #104


cyenyxe:
I have compared these changes against #130 and there is no real difference, only some subtle details that were agreed between the maintainers.

PR #130, opened 2016-02-16
More strict language in the VCF specification

cyenyxe:
Following the discussion in #104, this PR introduces the changes that have been flagged as "OK" in the Google Docs document.

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -159,10 +159,10 @@ pair of numbers, then this value must be $2$ and so on. There are also
 certain special characters used to define special cases:
 
 \begin{itemize}
-  \item If the field has one value per alternate allele then this value must be `A'.
-  \item If the field has one value for each possible allele (including the reference), then this value must be `R'.
-  \item If the field has one value for each possible genotype (more relevant to the FORMAT tags) then this value must be `G'.
-  \item If the number of possible values varies, is unknown, or is unbounded, then this value must be `.'.
+  \item A: The field has one value per alternate allele.
+  \item R: The field has one value for each possible allele, including the reference.
+  \item G: The field has one value for each possible genotype (more relevant to the FORMAT tags).
+  \item . (dot): The number of possible values varies, is unknown or unbounded.
 \end{itemize}
 
 The `Flag' type indicates that the INFO field does not contain a Value entry, and hence the Number must be $0$ in this case. The Description value must be surrounded by double-quotes. Double-quote character must be escaped with backslash $\backslash$ and backslash as $\backslash\backslash$. Source and Version values likewise must be surrounded by double-quotes and specify the annotation source (case-insensitive, e.g. ``dbsnp'') and exact version (e.g. ``138''), respectively for computational use.
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -1288,7 +1288,7 @@ extensions/restrictions:
 No record may refer to a contig not present in the header itself.
 
 \item All INFO and GENOTYPE fields must be fully typed in the BCF2 header to
-enable type-specific encoding of the fields in records.  An error should be
+enable type-specific encoding of the fields in records.  An error must be
 thrown when converting a VCF to BCF2 when an unknown or not fully specified
 field is encountered in the records.
 \end{itemize}
@@ -1376,8 +1376,8 @@ In BCF2, the original VCF records are converted to binary and encoded as BGZF
 blocks.  Each record is conceptually two parts.  First is the site information
 (chr, pos, INFO field).  Immediately after the sites data is the genotype data
 for every sample in the BCF2 file.  The genotype data may be omitted entirely
-from the record if there is no genotype data in the VCF file.  Note that it's
-acceptable to not BGZF compress a BCF2 file.
+from the record if there is no genotype data in the VCF file. 
+Compression of a BCF file is recommended but not required.
 
 \subsubsection{Site encoding}
 
@@ -1388,7 +1388,7 @@ l\_shared &	uint32\_t &	Data length from CHROM to the end of INFO \\ \hline
 l\_indiv  &	uint32\_t &	Data length of FORMAT and individual genotype fields \\ \hline
 CHROM	  & int32\_t  &	Given as an offset into the mandatory contig dictionary \\ \hline
 POS	      & int32\_t  &	0-based leftmost coordinate \\ \hline
-rlen      &	int32\_t  &	Length of the record as projected onto the reference sequence.  May be the actual length of the REF allele but for symbolic alleles should be the declared length respecting the END attribute \\ \hline
+rlen      &	int32\_t  &	Length of the record as projected onto the reference sequence. Must be the length of the REF allele or the declared length of a symbolic allele respecting the END attribute \\ \hline
 n\_allele\_info	& int32\_t	& n\_info, where n\_allele is the number of REF+ALT alleles in this record, and n\_info is the number of VCF INFO fields present in this record \\ \hline
 n\_fmt\_sample	& uint32\_t	& n\_sample, where n\_fmt is the number of format fields for genotypes in this record, and n\_samples is the number of samples present in this sample.  Note that the number of samples must be equal to the number of samples in the header \\ \hline
 QUAL	  & float	  & Variant quality; 0x7F800001 for a missing value \\ \hline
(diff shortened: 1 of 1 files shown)


VCFv4.3.tex
```diff
@@ -1277,7 +1288,7 @@ \subsection{Overall file organization}
 No record may refer to a contig not present in the header itself.
 
 \item All INFO and GENOTYPE fields must be fully typed in the BCF2 header to
-enable type-specific encoding of the fields in records.  An error should be
+enable type-specific encoding of the fields in records.  An error must be
```
d-cameron: If the field is defined by the standard, an implementation could potentially perform the conversion by interpreting the field using the standard definition. I don't think we should ban implementations from attempting this by making them throw an error.

VCFv4.3.tex
```diff
@@ -1423,7 +1434,7 @@ \subsubsection{Genotype encoding}
 
 The value is always implicitly a vector of N values, where N is the number of samples.  The type byte of the value field indicates the type of each value of the N length vector.  For atomic values this is straightforward (size = 1).  But if the type field indicates that the values are themselves vectors (as often occurs, such as with the PL field) then each of the N values in the outer vector is itself a vector of values.  This encoding is efficient when every value in the genotype field vector has the same length and type.
 
-Note that the specific order of fields isn't defined, but it's probably a good idea to respect the ordering as specified in the input VCF/BCF2 file.
+It is recommended to respect the ordering as specified in the input VCF/BCF2 file, but parsers should not rely on a specific ordering.
```
d-cameron: Does this refer to the header ordering, or the row ordering?

VCFv4.3.tex
pd3: This is not about the standard, really. BCF must know all fields at the time of writing the header.

VCFv4.3.tex
pd3: It refers to the FORMAT field ordering. So, for example, in the example just above, BCF could in principle encode the GT:GQ:DP fields as DP:GQ:GT.

jrambla:
+1

VCFv4.3.tex
```diff
@@ -123,10 +123,10 @@ \subsection{Meta-information lines}
 ##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="description",Version="128">
 \end{verbatim}
 In the above example, the extra fields of ``Source'' and ``Version'' are
-provided. Optional fields should be stored as strings even for numeric values.
+provided. Optional fields must be stored as strings even for numeric values.
 
-It is highly recommended (but not required) that the header
-include tags describing the reference and contigs backing the data contained in
+It is recommended in VCF and required in BCF that the header
```
yfarjoun: why removed the "highly" for vcf? I'd keep some adjective...perhaps "strongly"?

VCFv4.3.tex
```diff
@@ -154,28 +154,28 @@ \subsubsection{Information field format}
...
+single number, then this value must be $1$; if the INFO field describes a
+pair of numbers, then this value must be $2$ and so on. There are also
 certain special characters used to define special cases:
 
 \begin{itemize}
-  \item If the field has one value per alternate allele then this value should be `A'.
-  \item If the field has one value for each possible allele (including the reference), then this value should be `R'.
-  \item If the field has one value for each possible genotype (more relevant to the FORMAT tags) then this value should be `G'.
-  \item If the number of possible values varies, is unknown, or is unbounded, then this value should be `.'.
+  \item If the field has one value per alternate allele then this value must be `A'.
```
yfarjoun: To me it seems that the logic goes the other way...if the value is A then the field MUST have one value per alternate allele...etc. It is the number in the header line that defines how many values the field must have, not vice versa.

VCFv4.3.tex
```diff
@@ -230,8 +230,8 @@ \subsubsection{Assembly field format}
 
 \subsubsection{Contig field format}
 \label{sec-contig-field}
-It is highly recommended (and required for BCF) that the header includes tags
-describing the contigs referred to in the VCF file. The structured \texttt{contig}
+It is recommended for VCF, and required for BCF, that the header includes tags
```
yfarjoun: strongly recommended?

VCFv4.3.tex
```diff
@@ -292,16 +292,16 @@ \subsection{Header line syntax}
...
+with no tab character at the end of the line. The last data line must end with a line separator. In all cases,
 missing values are specified with a dot (`.').
 
 \subsubsection{Fixed fields}
 There are 8 fixed fields per record.  Fixed fields are:
 
 \begin{enumerate}
-  \item CHROM - chromosome: An identifier from the reference genome or an angle-bracketed ID String (``$<$ID$>$'') pointing to a contig in the assembly file (cf. the \#\#assembly line in the header). All entries for a specific CHROM should form a contiguous block within the VCF file. The colon symbol (:) must be absent from all chromosome names to avoid parsing errors when dealing with breakends. (String, no white-space permitted, Required).
+  \item CHROM - chromosome: An identifier from the reference genome or an angle-bracketed ID String (``$<$ID$>$'') pointing to a contig in the assembly file (cf. the \#\#assembly line in the header). All entries for a specific CHROM must form a contiguous block within the VCF file. The colon symbol (:) must be absent from all chromosome names to avoid parsing errors when dealing with breakends. (String, no white-space permitted, Required).
```
yfarjoun: I'm not sure I understand the part about "form a continguous block"...should the CHROMs be sorted according to the order in the reference? I don't see it here...but I assume that this is true! perhaps that should be specified instead of the contiguous block formulation?

yfarjoun:
done with my comments.

VCFv4.3.tex
cyenyxe: The point of the PR is to remove adjectives as much as possible, following http://www.ietf.org/rfc/rfc2119.txt. This way something will be simply mandatory or recommended/optional, without intermediate states that don't have practical implications.

VCFv4.3.tex
cyenyxe: Considering "special characters" introduced in the previous paragraph, that could make the flow clearer. What about making it even simpler, like the following?

> A: The field has one value per alternate allele
> R: The field has one value for each possible allele (including the reference)
> G: The field has one value for each possible genotype (more relevant to the FORMAT tags)
> . (dot): The number of possible values varies, is unknown or unbounded

VCFv4.3.tex
cyenyxe: It means that a chromosome shouldn't be split across a file. A file with positions listed as below wouldn't be valid:

> 1 1000
> 1 1001
> 2 1000
> 1 1002

I don't think anybody enforces the CHROMs to follow the reference order, but it would be nice if somebody else could comment on this.

pd3: The chromosomes do not have to follow the reference order.

VCFv4.3.tex
pd3: @cyenyxe +1

VCFv4.3.tex
yfarjoun: That's unfortunate...I think that many Htsjdk/Picard tools depend on that.

pd3: The first version of vcftools had the same requirement, but it turned out to be limiting.  In htslib/bcftools we require indexed files when the order of sequences is important.

nh13: @yfarjoun how about a way to specify in the VCF to specify that the chromosomes are sorted according to the external reference?  I agree that not knowing is a short-coming that seems like an easy fix.

yfarjoun: hmmm. how about if the reference-dictionary is included then the order of
the contigs has to honor it?

pd3: This would invalidate many valid VCFs. What is the motivation for the proposal? With indexed VCF and/or reference, we don't need to rely on any particular sequence ordering.

nh13: To know the reference order we then must always require a second file (index), which I deem as more burdensome than just being able to look at the header, like in a SAM/BAM file.

pd3: The requirement of a second file is not a strong enough argument to invalidate valid VCFs. The argument can go both ways by the way, as other operations would become more burdensome, a second file (reference) would be required for example when concatenating or merging VCFs.

nh13: We could always *add* something to a VCF that would indicate that is is sorted according to the original reference, similar to the sort order header tag in a SAM file.  That would not invalidate previous VCFs.  In many, many tools I write, I do not need the reference explicitly, even when merging VCFs.

cyenyxe:
Added commit with agreed changes. I would suggest leaving the discussion about the reference order for another PR.

yfarjoun:
I am not sure if this is waiting for a OK from me...but in the meantime, I found some more issues related to this PR: The `ID` string in both the INFO and the FORMAT fields is not defined. What characters are allowed, etc. I think they should (probably) both be "unique, alphanumeric + some punctuation (but not all since we should probably avoid `,` `\` and `=`.

yfarjoun:
👍 regarding the work done here. 

ID still needs to be defined, perhaps open an issue for that.

cyenyxe:
Rebased against master to remove any conflicts, should be ready to merge now.

cyenyxe:
Is anything preventing this from being merged? If we were to accept #175, it would be good to have this one in first.

pd3:
Is this waiting for me? Sorry, merging now.

PR #131, opened 2016-02-18
Indicate that CRAM v3 is now the canonical spec

mp15:
CRAM v3 has now superseded CRAM v2.1 as the canonical spec

### DIFF
--- a/index.html
+++ b/index.html
@@ -51,7 +51,7 @@
 
 <p><strong>SAMv1.tex</strong> is the canonical specification for the SAM (Sequence Alignment/Map) format, BAM (its binary equivalent), and the BAI format for indexing BAM files.</p>
 
-<p><strong>CRAMv2.1.tex</strong> is the canonical specification for the CRAM format.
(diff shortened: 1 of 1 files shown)


jmarshall:
Thanks for pointing out this needs updating.

This text describes each of the _*.tex_ files in the repositories.  So the CRAM paragraph needs to describe both _CRAM*.tex_ documents — cf the _VCFv4.**x**.tex_ paragraph.

mp15:
Would it be fair to say that CRAM 2.1 is obsolete?

jkbonfield:
Yes, although given the CRAM 3 specification is only CRAM 3 and doesn't include any details on how CRAM 2.1 differed (as you found out with EOF blocks) we clearly need to keep both easily accessable.

jmarshall:
That text was copied from README.md, so I was going to say you might update that too (actually I was just going to cherry-pick yours to there!), but it turns out I did update that one when CRAMv3 appeared.

So you could just copy the relevant line from cbd60fe7d40fc12531dfae32bda08b64f9bed2af…

mp15:
Amended

jmarshall:
Thanks, merged as a449e08e4f55087e5a672f56218e17ac13fd3fda and 025cf969f1962cb049ea6cdc11f3947d2c70ae7d.

PR #143, opened 2016-04-04
Initial commit of mVCF specifications (modified from VCFv4.2)

heathsc:
This is a draft specification for an augmented version of VCF (based on VCFv4.2) that contains DNA methylation specific information.  This format arose from discussions within the IHEC (International Human Epigenome Consortium) assay standards committee, and is intended as a standard for storing the results from WGBS or RRBS experiments.

### DIFF
--- a/mVCFv1.0.tex
+++ b/mVCFv1.0.tex
@@ -1,0 +1,1523 @@
+\documentclass[8pt]{article}
+\usepackage{enumerate}
+\usepackage{graphicx}
+\usepackage{lscape}
+\usepackage[margin=0.75in]{geometry}
+\usepackage[pdfborder={0 0 0}]{hyperref}
+
+\begin{document}
+\input{mVCFv1.0.ver}
+\title{The Methylation Variant Call Format (mVCF) Version 1.0 Specification}
+\date{\headdate}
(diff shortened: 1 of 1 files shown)


pd3:
From a cursory look at the specification it is not clear what are the key differences between VCF and mVCF. Is it really necessary to introduce a new file format and break compatibility with VCF? It seems that mVCF just adds some new tags.

By the way, there are inconsistencies in the text: the example seems to suggest that the REF allele is not required, but the text says otherwise. The presence of the REF allele is a practical requirement, it helps to catch user errors.

heathsc:
Thanks for the feedback. If it would be possible to just have the new tags described in the standard spec would be perfect for us.  The aim is not to break any VCF tools, but to have a standard way to represent methylation information within a VCF. 
Simon.

pd3:
Best would be to add the new reserved info/format tags to appropriate sections and discuss any specifics of the methylation data in a new section. Please make this against the latest version 4.3.

heathsc:
OK thank you - I will do this.

Simon

yfarjoun:
seems like the way forward here is to use regular VCF with new tags.

PR #147, opened 2016-04-28
Added the AH header tag to @SQ

lh3:
This PR adds the AH header tag to @SQ. It indicates whether a reference sequence is an alternate haplotype. Reads mapped to the primary assembly and alternate haplotypes usually receive different treatment in processing. BWA will generate this tag. I am also open to other options.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -200,6 +200,7 @@ meanings can be avoided.}
   & {\tt AS} & Genome assembly identifier. \\\cline{2-3}
   & {\tt M5} & MD5 checksum of the sequence in the uppercase, excluding spaces but including pads (as `*'s).\\\cline{2-3}
   & {\tt SP} & Species.\\\cline{2-3}
+  & {\tt AH} & Indicate if the reference sequence represents an alternate haplotype. \emph{Valid values}: {\tt Y} or {\tt N}\\\cline{2-3}
   & {\tt UR} & URI of the sequence.  This value may start with one of the standard
   protocols, e.g http: or ftp:.  If it does not start with one of these protocols, it is assumed to be a file-system path.\\\cline{1-3}
   \multicolumn{2}{|l}{\tt @RG} & Read group. Unordered multiple {\tt @RG} lines are allowed.\\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -197,10 +197,10 @@ meanings can be avoided.}
   field is used in the
   alignment records in {\sf RNAME} and {\sf RNEXT} fields. Regular expression: {\tt [!-)+-\char60\char62-\char126][!-\char126]*}\\\cline{2-3}
   & {\tt LN}* & Reference sequence length. \emph{Range}: {\tt [1,2$^{31}$-1]}\\\cline{2-3}
+  & {\tt AH} & Position of an alternate locus (see also \href{http://bit.ly/GRCterm}{http://bit.ly/GRCterm}) in the format of `\emph{chr}:\emph{start}-\emph{end}', `\emph{chr}' or `{\tt *}' (if unknown), where `\emph{chr}' is a contig in the primary assembly. Present \emph{only if} the sequence is an alternate locus.\\\cline{2-3}
   & {\tt AS} & Genome assembly identifier. \\\cline{2-3}
   & {\tt M5} & MD5 checksum of the sequence in the uppercase, excluding spaces but including pads (as `*'s).\\\cline{2-3}
   & {\tt SP} & Species.\\\cline{2-3}
-  & {\tt AH} & Position of an alternative haplotype in the format of `\emph{chr}:\emph{start}-\emph{end}' or `{\tt *}' (if unknown). Present \emph{only if} the sequence is an alternative haplotype.\\\cline{2-3}
   & {\tt UR} & URI of the sequence.  This value may start with one of the standard
   protocols, e.g http: or ftp:.  If it does not start with one of these protocols, it is assumed to be a file-system path.\\\cline{1-3}
   \multicolumn{2}{|l}{\tt @RG} & Read group. Unordered multiple {\tt @RG} lines are allowed.\\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -200,7 +200,7 @@ meanings can be avoided.}
   & {\tt AS} & Genome assembly identifier. \\\cline{2-3}
   & {\tt M5} & MD5 checksum of the sequence in the uppercase, excluding spaces but including pads (as `*'s).\\\cline{2-3}
   & {\tt SP} & Species.\\\cline{2-3}
-  & {\tt AH} & Indicate if the reference sequence represents an alternate haplotype. \emph{Valid values}: {\tt Y} or {\tt N}\\\cline{2-3}
+  & {\tt AH} & Position of an alternative haplotype in the format of `\emph{chr}:\emph{start}-\emph{end}' or `{\tt *}' (if unknown). Present \emph{only if} the sequence is an alternative haplotype.\\\cline{2-3}
   & {\tt UR} & URI of the sequence.  This value may start with one of the standard
   protocols, e.g http: or ftp:.  If it does not start with one of these protocols, it is assumed to be a file-system path.\\\cline{1-3}
   \multicolumn{2}{|l}{\tt @RG} & Read group. Unordered multiple {\tt @RG} lines are allowed.\\\cline{2-3}


tfenne:
I like having this indicated in the header.  @lh3 do you think there would be value in going one step further and having the tag define the region that the sequence is an alternative haplotype for, e.g. `AH:chr1:5000000-5010000`?

lh3:
BWA doesn't use the ALT positions, so can't generate a tag like `AH:chr1:5000000-5010000`. In addition, sometimes we know the contig is an ALT, but it does not have enough flanking regions for us to map to the primary assembly.

One option is to use `AH:*` when the contig is an ALT but we don't know where it maps, or use a region when we know. How about this?

yfarjoun:
This would be very helpful. One possible problem is that an `@SQ` line may be alt for two reasons:

1. The _aligner_ thinks that it's an alt (e.g. bwa-mem due to a .alt file) and thus while the original dictionary has no AH in this `@SQ` record, the BAM's dictionary has AH.
2. Once this becomes a tag that is used, one could imagine that the references will be also decorated with AH (in fact, that might be better than a .alt file) and thus the dictionary's @SQ line may have AH which could be transferred over to the BAM by an aligner that does _not_ understand alternate contigs...

- It would be good to be specific about what AH means in a .fasta and .dict file and what it means to transfer the AH tag from the .dict to the .bam (for the **next** reference.)

- Will bwa mem respect AH tags in the `@SQ` line in addition to .alt files?

lh3:
I have changed `AH` to: "Position of an alternative haplotype in the format of ‘chr:start-end’ or ‘*’ (if unknown). Present only if the sequence is an alternative haplotype" and implemented in BWA-MEM.

jmarshall:
It would be useful to carry this information in the BAM headers; for example, samtools has already wanted this information, and it would be much better to already have it than to try to look for a `.alt` file: cf PR samtools/samtools#469.)

So :+1: to the feature, but IMHO this needs more explanation in the spec.

* It would be useful to describe _alternative haplotype_ in §1.2 Terminologies and Concepts.
* The `@SQ-AH` text needs clarifying as to which of this refsequence and the refsequence named in `chr:start-end` is the **alternative** one.
* Is it likely that we would know which other sequence we're an alternative for, but not have enough flanking region to know the exact coordinates over there?  If so, would we want to represent this as `AH:chr` or similar?
* In the table, maintain the `AS`/`M5`/`SP`/`UR` alphabetical order.

ernfrid:
👍  to the feature too.

I second @yfarjoun's comments on clarifying how aligners should interact with AH tags. Are they intended to be added by aligners if sequences are considered alts during mapping? And thus removed if alts are not utilized? Or are they intended to indicate whether or not a sequence is an alt, regardless of the behavior of the aligner?

@jmarshall - I would agree with using `AH:chr` in the case where an alt is localized to a chromosome, but the location is unknown. However, I don't know if this is possible with the current definition of an alternate locus by the GRC. Currently, there don't appear to be any alts in GRCh38 that have this property.

mp15:
I am not convinced an alt should be allowed to exist without a primary. Unplaced contigs aren't really alternate haplotypes are they? You either pick the most prevalent alt as your unlocalised primary or choose one arbitrarily, either way you always have a primary even it it is unplaced.

yfarjoun:
@mp15, I think that the problem is that some tools might know that a contig is alternate (from a list), but not know what the primary is...

eitanbanks:
Looks like this has been open for 3 weeks and that all comments are positive.  I'm eager to get this merged in.  Can we do that today?

jmarshall:
There are a number of unaddressed comments and questions.  For example, mine from two weeks ago.

If these comments are addressed adequately in the next ten days or so, this proposal will be on track to be put on the agenda for the FIleFormats meeting on June 3rd with a view to agreeing on it at the meeting and it being merged by one of the spec maintainers thereafter.

mp15:
@yfarjoun Unless I'm missing something there is no point in having a tool that knows a contig is an alternate unless it knows what other sequence it is alternate to, otherwise it's just another unplaced config and needs no special handling anyway. For example with mapping tools, unless the mapping tool knows the primary location for the alt, it won't know which primary reads to fix the mapping score for when they map to the alternate as well.

mp15:
I've produced an additional commit to incorporate the @jmarshall's feedback and my own at mp15/hts-specs@47bfb7ce375a26964da1f4bff15f0184046d9370. Will be happy to update it if anyone comes up with further feedback (or a good use case for AH:*).

yfarjoun:
@mp15 I think that you are mistaken. 

@lh3 can correct me, but given a list of "alt" contigs (without their connection to the other contigs) bwa-mem will treat the "regular" contigs differently from the "alt" contigs. For example, if a read could get placed well on a regular contig and an alt contig, this _will not_ give the read a low mapping quality, but if it can be placed well on two regular contigs the read _will_ have a low mapping quality. The (original) point of having a `AH:* ` tag in the `@SQ` line is to capture exactly which contigs were considered to be alternative by the aligner, not to inform about some graph structure these might have.

mp15:
Hmm if it's actually simplified in that way then that's fine, I don't want to get hung up on irrelevant theoreticals but I do want to ensure we consider whether things may cause future problems or be open to misuse, it would mean you're overestimating mapping quality if there's any similarity between two ALTs with different primaries or ALTs vs the wrong bit of the Primary but perhaps I worry too much.

tfenne:
@mp15 I have tools that would still make use of the knowledge that a contig is an alternative haplotyope, even if it doesn't know what region of the core assembly it's an alternative for.  For example, when designing primers and hybridization probes I use various methods to align them back to the genome to find off-target hits.  I immediately discard all those that map to an alternative haplotype since I don't see those as being off-target (necessarily).  However, hits to unplaced (not alternative) contigs are problematic for me.

lh3:
I prefer to allow alternate contigs without mapping positions. Without long enough anchors, some novel insertions can't be mapped to the primary assembly. Pragmatically, we often don't care where an ALT contig is mapped to the primary assembly. It is not necessary to always require the mapping information.

I prefer not to explain alternate haplotypes in the Terminology section. If we explain it, we will have to explain primary assembly; if we explain primary assembly, we need to talk about chromosomal assembly, unlocalized contigs and unplaced contigs. GRC has a [whole page of definitions](http://www.ncbi.nlm.nih.gov/projects/genome/assembly/grc/info/definitions.shtml) related to these concepts. I think we should add a footnote referring to the link above, instead of coming up with our own definitions in Terminology.

mp15:
Summary of discussion at GA4GH file formats working group:
- Agreed to adopt GRC terminology of "Alternative locus"
- Agreed on possible values of *, chr, chr:start-*, chr:*-end, chr:start-end.
- Agreed to reference to GRC definitions rather than include our own.

yfarjoun:
Wait, does this mean that the **AH** tag needs to be changed to **AL**? Can we keep it AH?

jmarshall:
In the discussion in the meeting (perhaps you couldn't hear this part clearly over the phone) that covered the GRC “alternate locus” terminology, @richarddurbin with his GRC hat on expressed that he was comfortable enough with the `AH` tag name but that it would be good to deemphasise the word “haplotype” in the explanations.

It would be nice if the tag followed the terminology and, I suppose, if this conversation had happened earlier.  Unfortunately our hands are now somewhat tied as a version of **bwa** has been released that emits `AH:*`.

And I suppose that, GRC or no GRC, lots of people think of these things as alternate haplotypes. :smile:

richarddurbin:
Thanks John.  Good summary.  Richard

> On 12 Jun 2016, at 14:41, John Marshall <notifications@github.com> wrote:
> 
> In the discussion in the meeting (perhaps you couldn't hear this part clearly over the phone) that covered the GRC “alternate locus” terminology, @richarddurbin <https://github.com/richarddurbin> with his GRC hat on expressed that he was comfortable enough with the AH tag name but that it would be good to deemphasise the word “haplotype” in the explanations.
> 
> It would be nice if the tag followed the terminology and, I suppose, if this conversation had happened earlier. Unfortunately our hands are now somewhat tied as a version of bwa has been released that emits AH:*.
> 
> And I suppose that, GRC or no GRC, lots of people think of these things as alternate haplotypes. 😄
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub <https://github.com/samtools/hts-specs/pull/147#issuecomment-225435458>, or mute the thread <https://github.com/notifications/unsubscribe/ADRb5nNkCCIjQgFgvYhzVpBtN8IJdIoPks5qLAx5gaJpZM4ISQgD>.
> 




-- 

jmarshall:
@yfarjoun: The minutes from June suggest you were going to do the minor wordsmithing needed here?

yfarjoun:
Oh, I thought Heng was going to do that...I can do it.

lh3:
I have pushed a change to the description of the AH tag, addressing concerns mentioned above. Nonetheless, I am disallowing a format of *chr*:*start*- as *chr*:\*-*end*, as if we can get one end, we should get the other end as well.

jkbonfield:
Following discussion at the GA4GH fileformats meeting, I've read through this again and am happy with the change.  @yfarjoun, if you're happy too then we should get this merged at last.

yfarjoun:
hurray!

jmarshall:
I am sure the specification editors will wish to replace the _bit.ly_ URL with a footnote containing a brief summary and the full NCBI URL!

> Position of an alternate locus in the format of `\emph{chr}:\emph{start}-\emph{end}' […]

The text is still misleading as to which one is the alternate.  This would be clearer using the GRC terminology and phrased along the lines of “Indicates that this sequence is an alternate representation of the specified locus […]”.

yfarjoun:
oh. I misunderstood the previous comment as the final word...sorry if I jumped the gun. @jmarshall would you like me to revert, or submit another PR fixing your comment?

jmarshall:
You jumped the gun.  What was agreed at the meeting was that **all interested SAM parties** would review the text and :+1: as appropriate.  I am sure @tk2 would agree that editors should make purely editorial changes themselves, but that where more substantial changes are being made the file formats group has always operated more by consensus.

You should revert this in the `push -f` sense, fix the wording, and squash this PR on merging so that we don't need to see all the early mis-steps.  You may be loath to do that, in which case I would suggest further commits based on 2745d58e1f9fc43883e4f0d812b3954d8b93fee9 and merged, so in the history it looks like the same PR branch merged twice.

yfarjoun:
sorry. did push -f from head^1 of master.

jkbonfield:
I did look at the bit.ly and ummed and ahhed a bit.  However I found a statement from bit.ly CEO specifically stating that they use 301 redirects and will never expire them, so they're as robust as the company is itself.   Given that there is a non-zero chance that the destination url itself will change in a "revamp" (sadly all too common here!), it's possible that the bit.ly link may even be preferable.  Either way I feel it's not an issue, so don't mind if it is expanded fully too.

As far as the meeting went, my interpretation of it was that we were already agreed in pinciple and it just needed an official acceptance from both maintainers.

jkbonfield:
"Present \emph{only if} the sequence is an alternate locus" - that doesn't sound ambiguous as to which is the alternate.  What am I missing?  The presence of AH is what indicates *this* is the alternate.

yfarjoun:
how about 
   >...where `\emph{chr}' is some other sequence in the primary assembly. Present \emph{only if} the current sequence is an alternate locus.

jmarshall:
The first sentence says “Position of an alternate locus in the format of `chr:start-end`/etc” which is the exact opposite of what is intended.

yfarjoun:
for what it's worth, I think that the point of the bitly link was so that the speck would be able to follow changes in GRC website. I think that @lh3 can redirect that bit.ly if needed

yfarjoun:
hmmm. I think an example would be good here:
```
@SQ chr1 LN:10000
@SQ chr1_alt LN:100 AH:chr1:500-1500
```

So the AH tag (on `chr1_alt`) indicates the position (in `chr1`) where a alternate locus is located (namely `chr1_alt`). how is that the opposite of the first sentence?

jmarshall:
What the first sentence **says** is that the _position_ of the _alternate locus_ is `chr1:500-1500`, which **implies** that this region on `chr1` is the locus that is designated _alternate_.  As you know, this is not what is intended.

At best, it is unclear. There is a suggestion for reworking the first sentence in my comment above.

jkbonfield:
I cannot see the ambiguity, but clearly it is as we've interpreted it differently so I'm in agreement for changing it to your alternative wording as it doesn't harm to hammer the point home that **this** is the alternate.

SAMv1.tex
```diff
@@ -197,6 +197,7 @@ \subsection{The header section}
   field is used in the
   alignment records in {\sf RNAME} and {\sf RNEXT} fields. Regular expression: {\tt [!-)+-\char60\char62-\char126][!-\char126]*}\\\cline{2-3}
   & {\tt LN}* & Reference sequence length. \emph{Range}: {\tt [1,2$^{31}$-1]}\\\cline{2-3}
+  & {\tt AH} & Position of an alternate locus (see also \href{http://bit.ly/GRCterm}{http://bit.ly/GRCterm}) in the format of `\emph{chr}:\emph{start}-\emph{end}', `\emph{chr}' or `{\tt *}' (if unknown), where `\emph{chr}' is a contig in the primary assembly. Present \emph{only if} the sequence is an alternate locus.\\\cline{2-3}
```
jkbonfield: Rob's suggestion here is to say "present only if **this** sequence is an alternate locus", which is a tiny but more explicit change.

colinhercus:
Hi,

Sorry to but in at this stage but GRCh38 Goldenpath version has a few tags
on the FASTA header related to this

On alternate scaffolds you'll see..

>chr5_GL383530v1_alt  AC:GL383530.1  gi:302201756  LN:101241
rg:chr5:95169857-95254491  rl:alt-scaffold  AS:GRCh38
>chr5_KI270796v1_alt  AC:KI270796.1  gi:568335889  LN:172708
rg:chr5:125279466-125444151  rl:alt-scaffold  AS:GRCh38
>chr5_GL383531v1_alt  AC:GL383531.1  gi:302201757  LN:173459
rg:chr5:162373667-162541648  rl:alt-scaffold  AS:GRCh38
>chr5_KI270795v1_alt  AC:KI270795.1  gi:568335892  LN:131892
rg:chr5:177865465-177993409  rl:alt-scaffold  AS:GRCh38

So the rg: tag can be translated to your AH tag. Nice & simple :)

Some of the unlocalised sequences also have a rg: tag

>chr1_KI270706v1_random  AC:KI270706.1  gi:568335410  LN:175055  rg:chr1
rl:unlocalized  AS:GRCh38
>chr1_KI270707v1_random  AC:KI270707.1  gi:568335409  LN:32032  rg:chr1
rl:unlocalized  AS:GRCh38
>chr1_KI270708v1_random  AC:KI270708.1  gi:568335408  LN:127682  rg:chr1
rl:unlocalized  AS:GRCh38
>chr1_KI270709v1_random  AC:KI270709.1  gi:568335407  LN:66860  rg:chr1
rl:unlocalized  AS:GRCh38
>chr1_KI270710v1_random  AC:KI270710.1  gi:568335406  LN:40176  rg:chr1
rl:unlocalized  AS:GRCh38

NCBI, Ensembl & Gencode versions of the FASTA headers don't start with
chromsome name so the rg could be useful for these.

I know this is not needed for variant calling or for calculation of MAPQ
but I wonder why we don't just copy rl,rg&gi from the reference fasta
header to the @SQ record.

Just my 2 cents.

Kind Regards, Colin

jmarshall:
\[Here are parts of @colinhercus's comment with the FASTA headers I think formatted as intended.\]

On alternate scaffolds you'll see..

```
>chr5_GL383530v1_alt AC:GL383530.1 gi:302201756 LN:101241 rg:chr5:95169857-95254491 rl:alt-scaffold AS:GRCh38
>chr5_KI270796v1_alt AC:KI270796.1 gi:568335889 LN:172708 rg:chr5:125279466-125444151 rl:alt-scaffold AS:GRCh38
>chr5_GL383531v1_alt AC:GL383531.1 gi:302201757 LN:173459 rg:chr5:162373667-162541648 rl:alt-scaffold AS:GRCh38
>chr5_KI270795v1_alt AC:KI270795.1 gi:568335892 LN:131892 rg:chr5:177865465-177993409 rl:alt-scaffold AS:GRCh38
```

So the rg: tag can be translated to your AH tag. Nice & simple :)

Some of the unlocalised sequences also have a rg: tag

```
>chr1_KI270706v1_random AC:KI270706.1 gi:568335410 LN:175055 rg:chr1 rl:unlocalized AS:GRCh38
>chr1_KI270707v1_random AC:KI270707.1 gi:568335409 LN:32032 rg:chr1 rl:unlocalized AS:GRCh38
>chr1_KI270708v1_random AC:KI270708.1 gi:568335408 LN:127682 rg:chr1 rl:unlocalized AS:GRCh38
>chr1_KI270709v1_random AC:KI270709.1 gi:568335407 LN:66860 rg:chr1 rl:unlocalized AS:GRCh38
>chr1_KI270710v1_random AC:KI270710.1 gi:568335406 LN:40176 rg:chr1 rl:unlocalized AS:GRCh38
```

jkbonfield:
Thanks for the suggestions @colinhercus.  For now we want to get the spec in sync with actual usage of this tag (bwa has been generating them for a while) so we'll merge #184 soon.

Can you then please submit a new issue against that version, once it lands, so we can discuss the amendments to be more in line with GRC without a huge historical discussion.  Thanks

colinhercus:
Hi James,

I'm just adding these tags in lower case which is allowed in the specs and
also the AH tags for the alt_scaffolds. I'll likely leave it at that.

Cheers, Colin

PR #148, opened 2016-05-03
separated sample_VCFv4.3.vcf from VCFv4.3.tex

knausb:
I thought it would be nice to have an example VCF format file for people to try. For example, those of us who write software that works with VCF format files may like an example VCF file to test our software on. So I extracted the example as a separate file. My two cents.

Thank you!
Brian


knausb:
I've realized that the example is space delimited (presumably to improve presentation) instead of tab delimited. This means that it would require more than what I've asked for to make a nice example. So I'll close this. Thanks!

PR #149, opened 2016-05-18
Uppercased version of proposed PacBio tags (for discussion and reservation purposes only)

mp15:
This is NOT TO BE MERGED pull request detailing an uppercased version of the tags from the PacBio BAM specification at: http://pacbiofileformats.readthedocs.io/en/3.0/BAM.html

I have moved mq to MR as it collides with existing MQ tag.

The reason why we need these tags uppercased and standardised is that read archives such as ENA or EGA (SRA or DBGAP in the US I think?) tend to strip lowercase tags as they are site-local. If an instrument manufacturer's tags were stripped in this way, it would reduce reproducibility which is undesirable.

We should probably engage with @PacificBiosciences to fill in the details and clarify which of these are meant to be used by downstream tools and if any are internal/un-specable and thus should remain lowercase and never be stored in archives.


mp15:
Hi @dalexander
I've spoken to @vadimzalunin at EBI and the good news is that my knowledge of tag dropping is out of date (This was proposed in the old archive BAM specification but nobody ever implemented it as folk were scared about what might accidentally be dropped). Nothing is currently being dropped from EBI archives. However, I still want to avoid having multiple analysis tools thinking the same tag means different things. Could you comment on this proposal please?

mp15:
Manually rewrote this to move to use of SAMTags.tex.

dalexander:
Hi there,

I appreciate the sentiment but I'm not sure we really *want* to commit to having our tags spec'ed by the SAM committee.  While we are not *actively* changing our format at this time, we would like to reserve the right to do so, and I would prefer that PacBio's documents remain the canonical source of information about our use of tags.

If the "tag dropping" of lowercase no longer happens (which is good to hear), then I don't think we need to change any of our tags to play nice.

Dave

mp15:
Hi @dalexander,

The GA4GH file formats committee does not really specify the meaning of tags. Whilst there are a few core ones we maintain, for most of them we just register them so that nobody else uses the same tag (thus the lack of detail). If you were not a sequencing instrument manufacturer I would not press the point but I would like to ask you to consider at least reserving as uppercase the few tags you expect 3rd party software to rely on. Local tags are fine for information that stays within your pipelines or debug info but if downstream tools need to parse information to assess call accuracy etc you really should reserve them.

With my customer hat on non-commitment to a single meaning for a tag creates a rather annoying data maintainablity problem for me even if you stay lower case. The use case I would like you to consider is that when I fish the data out of the EGA/ENA/SRA/DbGAP to look at it 2-3 years from now I want to be able to know that the meaning of any tag in there is unambiguous and that by feeding it to a modern tool it will not accidentally misinterpret the data because you changed the meaning. There are enough letters in the alphabet that if you need to invent a new meaning for a tag you just change the tag's ID and that removes the danger of misinterpretation. Also for third party developers who are not writing in Python or using pbcore this makes writing parsers a lot simpler.

Martin

jkbonfield:
We're closing this as we don't wish to bless a long series of platform specific tags.

It is acknowledged that there is room for a better mechanism of documenting standard platform/vendor specific tags, ideally in private name space or maybe even in some reserved subset of this.  This is better dealt with by resurrecting https://github.com/samtools/hts-specs/issues/203.

PR #156, opened 2016-06-28
fix styles

bicycle1885:
I guess this is what you intended to be.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -195,7 +195,7 @@ The most significant bits of the first byte have special meaning and are called
 the number of bytes to follow. To accommodate 32 bits such representation requires 
 5 bytes with only 4 lower bits used in the last byte 5.
 
-\item[LTF-8 long or (ltf8)]\ \newline
+\item[LTF-8 long (ltf8)]\ \newline
 See ITF-8 for more details. The only difference between ITF-8 and LTF-8 is the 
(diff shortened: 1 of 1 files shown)


mp15:
@jkbonfield @vadimzalunin Could we get a +1 -1 for this so that we can either merge or close?

jkbonfield:
Apologies we did nothing with this.  I'll take a look at it.  It wouldn't surprise me to find a few issues as this TeX doc is the result of an automatic Word to TeX translation (!).

Thanks.

jkbonfield:
Looking at this, I'm unsure if the original text is intentionally done that way.  I suspect so because we have fundamental data types (itf8, arrays, etc) and then more conceptual things utilising those data types (encoding<type>).  Arguably string should be a fundamental type though.

@vadimzalunin - as the original author, I'll punt this to you to let you decide which style you prefer.  I'm happy with it either way.

I do note though that in a lot of places we using ```encoding<int>``` without ever defining ```int```.  I think this is because ```int``` here isn't an actual on-disk type but simply short for an integer value?

vadimzalunin:
Yes, 'int' is a logical type rather than on-disk. 
I think it's ok to merge this.

jkbonfield:
Ok then +1 from me.  Go ahead and merge please Vadim (or ask me to do it if you wish).

PR #158, opened 2016-07-13
Remove dittography

corburn:


jkbonfield:
Thanks for spotting and correcting this.

PR #175, opened 2016-10-16
modularize VCF 4.3 tex file

SHuang-Broad:
Currently, the tex file for VCF 4.3 spec is a long tex file, which makes modifying and updating the spec a bit difficult.
This is to modularize the different sections in the tex file so that each section may be reviewed and updated separately.


jmarshall:
This repository contains eleven specifications.  The VCFv4.3 source is a bit bigger than the SAM/BAM source and a bit smaller than CRAMv3.

I'm not sure that splitting the TeX source makes this repository as a whole any more wieldy.  In particular, the proposed `VCFv4_3` name for this subdirectory worsens tab-completion.

Could you explain the particular ways that splitting 1888 lines into 420+24+51+22+682+593+35 lines makes modifying and updating the spec easier?  In the _con_ column, it seems to me that it would make it easier to forget to update places where a small piece of text makes reference to something in a different section, e.g., examples.

SHuang-Broad:
Thanks @jmarshall for getting to this PR so quickly. I appreciate your comments.

To reply, 

0. The reason, which should have been explained better in the PR message, for this PR is so that changes to one particular section doesn't interfere with others, unless a reference (in the TeX sense) is renamed, which in the old format would require changes to those that refer to it anyway. And resolving conflicts would be easier.
1. This is a PR for organizing VCF4.3 only. If this change makes sense, other specifications could be changed to the agreed-upon format as well.
2. The name `VCFv4_3` was not a well-thought name, so that can definitely be changed to a better one.
3. I'm not quite understanding the comment on _con_ column, would you please explain again?

Thanks!

pd3:
Hi @SHuang-Broad , thank you for the pull request, I appreciate the work you put into this. However, I also do not see the clear benefits of splitting the file into multiple parts. Isn't this just a minor optimization which will be welcome by some but unwelcome by others?

cyenyxe:
If we were to keep the repository structure consistent, it should be applied to all the .tex documents. The largest of those currently in the repository is 2000 lines, which is quite a manageable number.

Given that previous comments are not totally convinced of the advantages of this change, and the amount of work that would involve, I will close the ticket.

PR #176, opened 2016-11-16
Add missing genome IDs

anderspitman:


pd3:
Thank you

jmarshall:
VCFv4.1.tex and VCFv4.3.tex have the same problem.

PR #184, opened 2017-02-06
Yf hl add ah tag

yfarjoun:
The resurrection of a branch that I mistakenly merged too soon.

@jmarshall @jkbonfield sorry about that.


yfarjoun:
https://github.com/samtools/hts-specs/pull/147 is the old PR

yfarjoun:
how about the following definition:
   >Indicates that sequence is an alternate locus (see also http://bit.ly/GRCterm). The value is the locus in the primary assembly for which this sequence is an alternative. The format is ‘chr:start-end’, ‘chr’ (if known) or ‘*’ (if unknown), where ‘chr’ is another sequence in the primary assembly. Must not be present on sequences in the primary assembly.

@jmarshall @jkbonfield ?

jkbonfield:
I was just attempting to push a change and got a conflict as in the 5 mins since I checked it out you updated it!  Good timing. :-)

I'd say "Indicates that **this** sequence is an..." to make it doubly explicit.

I'd also replace the bit.ly link with https://www.ncbi.nlm.nih.gov/grc/help/definitions instead (that was part of my commit).  While bit.ly gives us the ability to redirect if NCBI ever have a "rebranding & reshuffle" (not that's we'd ever do that... *cough*), it's probably easier for us to simply edit the spec again with the new URL than attempt to get whoever owns the bit.ly link to redirect it.

Otherwise it's fine.

yfarjoun:
done.

jkbonfield:
I'm happy.  In the absence of any other comments we should try and get this signed off on Thursday.  No comments => it goes in! :-)

jkbonfield:
1. Noted - easy change to make.

2. No links currently work for me in PDFs due to some bizarreness of Sanger's desktop configurations! However firefox is happy with leading spaces so it probably would have been fine.  It's fixed anyway (incoming).

3.  I'll change it.

4. Fixed.

5. Yes, but to be frank I dismissed it as it's taken a YEAR to get one sentence defined already and I didn't desire starting all over again.  My view is get it merged and if needed make minor revisions later.  This is something already in production and used by bwa so we can't afford to take another year going around in circles!  (See https://github.com/lh3/bwa/releases/tag/v0.7.15)

6. Obviously it needs rebasing, but it's helpful for people following the discussion to only do this on merge and not for each update.

yfarjoun:
thanks @jkbonfield I didn't realize you're taking this over...we were doing double work...:-)

jmarshall:
:+1: Me, I'm happy with the text now (even if the last sentence is IMHO redundant :smile:), and I think it's much clearer than the previous text.  Someone on the call was concerned about the `\footnote` (rather than `\footnotemark`) in a table; but fortunately this is in a `longtable`, which supports `\footnote`.

It would be nice to tidy up the line breaking in the newly-added TeX source…

----

> 6. Obviously it needs rebasing, but it's helpful for people following the discussion to only do this on merge and not for each update.

Indeed.  The point was the rest of the sentence discussing the infrastructure for updating the PDFs, which people appear to be unaware of and so perhaps needs adding to a _CONTRIBUTING.md_ or so.

jkbonfield:
Adjusted line breaks, squashed into Heng's original commit (with appropriate linkage to longwinded discussions) and merged as https://github.com/samtools/hts-specs/commit/494628aa466e477d603324b6e7f43d3747c3c2fd

PR #185, opened 2017-02-13
Table describing INFO reserved fields

cyenyxe:
Subset of changes listed in #123. These are restricted to INFO fields, reformatting their description as a table, so Number and Type can be made explicit.

In some cases, Number and Type were already listed in the field description, in others they can be inferred, and in others they were extracted from the sample at the beginning of the specification document.

Fields with issues:
* MQ: INFO Type not specified. FORMAT Type is Integer, but an RMS should be a Float?
* SB: Number or Type not specified and can't be directly inferred from description
* DB, H2 and rest of flags: Only Number=0 is allowed, but this doesn't work well for multi-allelic variants as discussed in https://github.com/samtools/hts-specs/pull/123#r66714347

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -1,6 +1,7 @@
 \documentclass[8pt]{article}
 \usepackage{enumerate}
 \usepackage{graphicx}
+\usepackage{tabularx}
 \usepackage{lscape}
 \usepackage[margin=0.75in]{geometry}
 \usepackage[pdfborder={0 0 0}]{hyperref}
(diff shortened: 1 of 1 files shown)


pd3:
Hi Cristina, 

I did not spot any issues, +1 from me.

* INFO/MQ: it would make sense to make it a Float, similar to INFO/BQ
* INFO/SB: I think GATK used to output it (or maybe it still does?), defined as Number=1,Type=Float
* I think flags were introduced to be able to write just "FLAG" in the INFO field rather than "FLAG=1". For multiple alleles this is problematic, so I'd rather not change the established notation. One can always use integer type instead. Maybe the interpretation of DB and other flags should be that it refers to a whole site, rather than to a specific combination of alleles.

cyenyxe:
* INFO/MQ: Are we fine to make it inconsistent across the document, considering that FORMAT explicitly defines it as Integer?
* INFO/SB: Could you please confirm @yfarjoun ?
* Flags: Yes, leaving them as they are right now is probably the safest option. I shall create another ticket to discuss that issue.

pd3:
It is OK for INFO/MQ be of different type from FMT/MQ. FMT/MQ is an integer because of size considerations - integers can packed into 8 bits and the size savings can be significant. With INFO fields it does not matter. But I am equally happy with an integer INFO/MQ.

Issue #186, opened 2017-02-16
Inappropriate HTTP status codes in htsget


jeromekelleher:
The [htsget spec](http://samtools.github.io/hts-specs/htsget.html) currently requires that the  server should return a HTTP status 409 when the client requests a format that the server doesn't support (i.e. ``format=something_weird``). [HTTP 409](https://httpstatuses.com/409) indicates that "the request could not be completed due to a conflict with the current state of the target resource", which seems like quite a stretch from "you've requested a format I don't support".  

While being concrete about the HTTP status codes is a great idea, I think it's probably a bad idea to co-opt these (I would have thought) obscure error codes for our own purposes for two reasons:

1. There will be clients that ignore the JSON payload returned in error conditions, and just raise the HTTP error directly. In this case, the user will be presented with very confusing errors. For example, using ``requests`` if I make a request for a bad file type I'll get back the error "409 Client Error: Conflict for url: ".  Googling for this sends me off in completely the wrong direction. A simple "400 Bad Request" would be much better in this case I think, because even if it is vague at least it's not falsely precise.

2. We should assume that there will be a tall stack of other HTTP processing entities between the client and the server. These will interpret the HTTP error codes according to the original specifications (which are quite precise about what clients should and should not do in the context of the various error codes), and may take actions based on them. Co-opting the error codes for our own specific needs could lead to very weird and difficult to debug behaviour.

So, I would argue that we should limit ourselves to a small number of widely used and well understood HTTP status codes (400, 403, 404, 500?). Certainly we should have some mechanism for passing back more specific information to the client, but this should be held in the JSON payload, not the HTTP status.

Thoughts? cc'ing @mlin and @dglazer here, as I'm not sure if they are watching this repo.

vadimzalunin:
I think this error is basically saying that the data does exists but in some other format(s). Clients may want to recover after this error unlike say "400 Bad Request".

jeromekelleher:
> Clients may want to recover after this error unlike say "400 Bad Request".

I totally agree, but I think we need to make a distinction between the HTTP client and the htsget client. Currently, for the htsget server to communicate with the htsget client, we are giving the HTTP client misleading information (HTTP 409, in this example).  Htsget is layered on top of HTTP, and by co-opting HTTP error codes for our own specific purposes we're breaking this layering. To preserve the layering, the error information specific to the htsget protocol should be contained in the JSON body of the response.

[...]

PR #180, opened 2016-12-13
Add GA4GH retrieval API spec

jmarshall:
This adds a markdownified version of the previous [Google Docs document](https://docs.google.com/document/d/1OSPfxdJ3uPoCfUVzMaekCOPF5sNEwqkJEUj-SjlECy0).

You can preview the markdown text in this PR, the basically-rendered source repo [here](https://github.com/jmarshall/hts-specs/blob/ga4gh/ga4gh-retrieval.md), and the fully-rendered gh-pages version [here](http://jmarshall.github.io/hts-specs/ga4gh-retrieval.html) or as linked from <http://jmarshall.github.io/hts-specs/> (use the link in the sidebar).

There are (currently) four annotations on the Google Docs document, which I've turned into footnotes here.  I imagine that we'll want to turn them into issues or so, and delete the markdown footnotes as we resolve them.

PR #191, opened 2017-02-23
Make error handling more concrete & remove some HTTP codes.

jeromekelleher:
This is an update of the htsget error handling with a few changes. The initial draft is really just for discussion, as it seemed more efficient to make the changes and discuss them concretely rather than in issues.

Changes:

1. Removed HTTP 409, 406 and 416 as discussed in #186.

2. Changed the text where it referred to HTTP codes to refer to the specific error 'classes' that we have defined. 

3. Tried to be more RFC2119 about the language used to describe when and what type of errors should be raised.

4. Changed the structure of the JSON error object to remove the embedded "error" field (seems superfluous), and added a table to document the fields more systematically.

5. Added an ``InvalidRange`` error for when start > end. (Superfluous?)

Apologies for rolling so many changes together, but I think the error handling needs a bit of work in various ways, and this seemed to be the best way of getting the ball rolling. Any feedback would be much appreciated.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -4,7 +4,7 @@ title: htsget protocol
 suppress_footer: true
 ---
 
-# Htsget retrieval API spec v0.1
+# Htsget retrieval API spec v0.2rc
 
 # Design principles
 
@@ -37,29 +37,45 @@ Requests to the retrieval API endpoint may be authenticated by means of an OAuth
 
 ## Errors
 
-Non-successful invocations of the API return an HTTP error code, and the response body contains a JSON object (content-type application/json) with the following structure:
+The server MUST respond with an appropriate HTTP status code (4xx or 5xx) when an error condition is detected.  In the case of transient server errors, (e.g., 503 and other 5xx status codes), the client SHOULD implement appropriate retry logic as discussed in [Reliability & performance considerations](#reliability--performance-considerations) below.
 
-```json
-{
-   "error": {
-       "type": "NotFound",
-       "message": "No such accession"
-   }
-}
-```
+For errors that are specific to the `htsget` protocol, the response body SHOULD be a JSON object (content-type `application/json`) providing machine-readable information about the nature of the error, along with a human-readable description. The structure of this JSON object is described as follows.
 
-The following error types are defined:
+### Error Response JSON fields
 
-Type			| HTTP status code | Description
+<table>
+<tr markdown="block"><td>
+`error`  
+_string_
+</td><td>
+The type of error. This SHOULD be chosen from the list below.
+</td></tr>
+<tr markdown="block"><td>
+`message`  
+_string_
+</td><td>
+A message specific to the error providing information on how to debug the problem. Clients MAY display this message to the user.
+</td></tr>
+</table>
+
+The following errors types are defined:
+
+Error type	| HTTP status code | Description
 |-----|:---:|-----|
 InvalidAuthentication	| 401	| Authorization provided is invalid
 PermissionDenied	| 403	| Authorization is required to access the resource
 NotFound		| 404	| The resource requested was not found
-Unable			| 406	| The server is unable to fulfill the request
-UnsupportedFormat	| 409	| The requested file format is not supported by the server
-InvalidInput		| 422	| The request parameters do not adhere to the specification
-InternalError		| 500	| Server error, clients should try later
-ServiceUnavailable	| 503	| Service is temporarily unavailable
+UnsupportedFormat	| 400	| The requested file format is not supported by the server
+InvalidInput		| 400	| The request parameters do not adhere to the specification
+InvalidRange		| 400	| The requested range cannot be satisfied
+
+The error type SHOULD be chosen from this table and be accompanied by the specified HTTP status code.  An example of a valid JSON error response is:
+```json
+{
+   "error": "NotFound",
+   "message": "No such accession 'ENS16232164'"
+}
+```
 
 ## CORS
 
@@ -100,7 +116,7 @@ _optional string_
 </td><td>
 Request read data in this format. Default: BAM. Allowed values: BAM,CRAM.
 
-Server replies with HTTP status 409 if the requested format is not supported.
+The server SHOULD reply with an `UnsupportedFormat` error if the requested format is not supported.
 [^a]
 </td></tr>
 <tr markdown="block"><td>
@@ -108,6 +124,8 @@ Server replies with HTTP status 409 if the requested format is not supported.
 _optional_
 </td><td>
 The reference sequence name, for example "chr1", "1", or "chrX". If unspecified, all reads (mapped and unmapped) are returned. [^b]
+
+The server SHOULD reply with a `NotFound` error if the requested reference does not exist.
 </td></tr>
 <tr markdown="block"><td>
 `referenceMD5`  
@@ -115,19 +133,33 @@ _optional_
 </td><td>
 The MD5 checksum uniquely representing the reference sequence as a lower-case hexadecimal string, calculated as the MD5 of the upper-case sequence excluding all whitespace characters (this is equivalent to SQ:M5 in SAM).
 
-Server replies with HTTP status 422 if `referenceName` and `referenceMD5` are both specified and are incompatible.
+The server SHOULD reply with a `NotFound` error if the requested reference does not exist.
+
+The server SHOULD reply with an `InvalidInput` if `referenceName` and `referenceMD5` are both specified and are incompatible.
 </td></tr>
 <tr markdown="block"><td>
 `start`  
 _optional 32-bit unsigned integer_
 </td><td>
-The start position of the range on the reference, 0-based, inclusive. If specified, `referenceName` or `referenceMD5` must also be specified. [^c]
+The start position of the range on the reference, 0-based, inclusive. 
+
+The server SHOULD respond with an `InvalidInput` error if `start` is specified and a reference is not specified
+(see `referenceName` and `referenceMD5`).
+
+The server SHOULD respond with an `InvalidRange` error if `start` and `end` are specified and `start` is greater
+than `end`.
 </td></tr>
 <tr markdown="block"><td>
 `end`  
 _optional 32-bit unsigned integer_
 </td><td>
-The end position of the range on the reference, 0-based exclusive. If specified, `referenceName` or `referenceMD5` must also be specified.
+The end position of the range on the reference, 0-based exclusive.
+
+The server SHOULD respond with an `InvalidInput` error if `end` is specified and a reference is not specified
+(see `referenceName` and `referenceMD5`).
+
+The server SHOULD respond with an `InvalidRange` error if `start` and `end` are specified and `start` is greater
+than `end`.
 </td></tr>
 <tr markdown="block"><td>
 `fields`  
@@ -140,13 +172,17 @@ Default: all
 `tags`  
 _optional_
 </td><td>
-A comma separated list of tags to include, default: all. If the empty string is specified (tags=) no tags are included. It is illegal for the values of `tags` and `notags` to intersect; the server may return HTTP status 400 in this case.
+A comma separated list of tags to include, default: all. If the empty string is specified (tags=) no tags are included. 
+
+The server SHOULD respond with an `InvalidInput` error if `tags` and `notags` intersect.
 </td></tr>
 <tr markdown="block"><td>
 `notags`  
 _optional_
 </td><td>
-A comma separated list of tags to exclude, default: none. It is illegal for the values of `tags` and `notags` to intersect; the server may return HTTP status 400 in this case.
+A comma separated list of tags to exclude, default: none. 
+
+The server SHOULD respond with an `InvalidInput` error if `tags` and `notags` intersect.
 </td></tr>
 </table>
 
@@ -262,10 +298,6 @@ Initial guidelines, which we expect to revise in light of future experience:
 
 The URL and headers might contain embedded authentication tokens; therefore, production clients and servers should not unnecessarily print them to console, write them to logs, embed them in error messages, etc.
 
-## Method-specific error interpretations
-
-* 406 Unable: may be returned if a genomic range is requested, but the server is unable to provide genomic range slicing for the particular dataset (e.g. if no index is available).
-
 
 # Possible future enhancements
 
@@ -279,8 +311,6 @@ The URL and headers might contain embedded authentication tokens; therefore, pro
 ## Existing clarification suggestions
 
 [^a]: This should probably be specified as a (comma separated?) list in preference order.  If the client can accept both BAM and CRAM it is useful for it to indicate this and let the server pick whichever format it is most comfortable with.
-[^b]: Define error code (404?, 416?) for queries to a reference that is not present in the header.  (Note this is not the same as present but having no data aligned to it - that should just be an empty reply.)
-[^c]: Define error response codes - suggest 416 (range not satisfiable).  Perhaps this is appropriate for all chr, start and end failures.
 [^d]: How will compression work in this case - can we benefit from columnar compression as does Parquet?
 
 
--- a/htsget.md
+++ b/htsget.md
@@ -50,9 +50,8 @@ Non-successful invocations of the API return an HTTP error code, and the respons
 
 The following error types are defined:
 
-|-
 Type			| HTTP status code | Description
-|-
+|-----|:---:|-----|
 InvalidAuthentication	| 401	| Authorization provided is invalid
 PermissionDenied	| 403	| Authorization is required to access the resource
 NotFound		| 404	| The resource requested was not found
@@ -61,7 +60,6 @@ UnsupportedFormat	| 409	| The requested file format is not supported by the serv
 InvalidInput		| 422	| The request parameters do not adhere to the specification
 InternalError		| 500	| Server error, clients should try later
 ServiceUnavailable	| 503	| Service is temporarily unavailable
-|-
 
 ## CORS
 
@@ -156,9 +154,8 @@ A comma separated list of tags to exclude, default: none. It is illegal for the 
 
 The list of fields is based on BAM fields:
 
-|-
 Field	| Description
-|-
+|-------|-------|
 QNAME	| Read names
 FLAG	| Read bit flags
 RNAME	| Reference sequence name
@@ -170,7 +167,6 @@ PNEXT	| Alignment position of the next fragment in the template
 TLEN	| Inferred template size
 SEQ	| Read bases
 QUAL	| Base quality scores
-|-
 
 Example: `fields=QNAME,FLAG,POS`.
 


vadimzalunin:
Is anyone reviewing this? I don't mind if there is no takers.

jmarshall:
Some comments:

Re 1: Um, we didn't discuss 406 or 416.  They seem fairly reasonable to me.

Re 4: Without that _error_ field, the JSON text doesn't directly indicate that its supposed to represent an error.  If we're changing it, perhaps we could change it to `{ "error": "NotFound", "message": "No such accession" }` or so?

Re 5: This is a perfect use case for 416 (with the caveat that the issue hasn't come from a `Range` header, but then it hasn't come from a WebDAV header either!).

(And trivially, what's up with the \`\`doubled backquotes\`\`?)

Otherwise basically :+1: although the shouty way of using of RFC2119-style verbs makes me twitch…

jeromekelleher:
Thanks for the feedback @jmarshall. I don't feel particularly strongly about any of this and I'm very happy to iterate it. I don't think we really discussed the error handling very much at any point, so I thought it would be good to at least air the various options.

The wider point I was  trying to make in #186 was that it might be a good idea to restrict our use of error codes down to a small handful. Since we are using JSON to communicate the real content of the error back to the client (which may take actions based on that), then what's the point of specifying the HTTP error code at all (beyond 4xx)? It just seems a bit pointless to me to spend ages agonising over whether a 406 or a 409 (or whatever) is more appropriate, when no-one is ever going to use this information anyway. If we use the HTTP codes like 416 there is a real danger of double-specification: do we mean the error definition in the htsget spec, or the original error HTTP error definition?

Having said all that, if the HTTP error code is apt, and there is nothing in the original definition that contradicts our usage, then there's probably no harm in it.

> Re 4: Without that error field, the JSON text doesn't directly indicate that it's supposed to represent an error. If we're changing it, perhaps we could change it to { "error": "NotFound", "message": "No such accession" } or so?

Yeah, good point. Anyone else like this form?

> (And trivially, what's up with the ``doubled backquotes``?)

I'm hopelessly confused between Markdown and RestructedText and end up writing things in a mishmash of the two! I'll fix this and bring it into line with the rest of the doc.

> ... though the shouty way of using of RFC2119-style verbs makes me twitch…

Happy to fix this if non-shouty is more appropriate.

delagoya:
+1  - Seems OK to me.

vadimzalunin:
> 1. Removed HTTP 409, 406 and 416 as discussed in #186.

Why not 422 then? I think for (client) API-specific errors we should do either all HTTP codes or all JSON errors to avoid confusion. The meaning of everything else (401, 403, 404, 500 and 503) is not overridden - any HTTP client should expect them (and all other HTTP codes but hey!). 

> 2. Changed the text where it referred to HTTP codes to refer to the specific error 'classes' that we have defined.

ok

> 3. Tried to be more RFC2119 about the language used to describe when and what type of errors should be raised.

ok

> 4. Changed the structure of the JSON error object to remove the embedded "error" field (seems superfluous), and added a table to document the fields more systematically.

I agree with @jmarshall, { "error": "NotFound", "message": "No such accession" } looks better.

> 5. Added an InvalidRange error for when start > end. (Superfluous?)

Why just this one?

Other comments:

Is there a way to fix the error code table? I know the problem was not introduced by this PR but the codes are being updated. 

Once client errors are all under 400 then servers wouldn't have to explain what's wrong. Because good clients would still work fine, so why should servers care about buggy clients? Instead of range errors etc. the spec should communicate back something useful like data formats and sequences/ranges available.

jeromekelleher:
Thanks for the comments, I've updated the PR with a few items (I'll squash the commits if/when we're done with review and ready to merge). Updates:

1. Changed the JSON structure to ```{ "error": "NotFound", "message": "No such accession" }``` as advocated by @jmarshall and @vadimzalunin.
2. Changed the HTTP 422s to 400 (more below on this).
3. Fixed the Error codes and Field filtering tables, and other minor formatting stuff.

The formatted(ish) spec is available [here](github.com/jeromekelleher/hts-specs/blob/32052708104e03f0295a60664cf446f14c419c89/htsget.md)

@vadimzalunin, some responses to your points:

> Why not 422 then? I think for (client) API-specific errors we should do either all HTTP codes or all JSON errors to avoid confusion. The meaning of everything else (401, 403, 404, 500 and 503) is not overridden - any HTTP client should expect them (and all other HTTP codes but hey!).

I agree, it does seem simpler to have everything return 400 and just use the JSON to communicate with the htsget client. I've changed the 422s to 400s in the spec above.

However, there are some non-400 error codes left, which we need to resolve. These are:

| Type                  | HTTP Status   | Description    |
| --------------------- |-------------- | ------------------------- |
| InvalidAuthentication | 401           | Authorization provided is invalid |
| PermissionDenied      | 403           | Authorization is required to access the resource |
| NotFound              | 404           | The resource requested was not found |
| InternalError         | 500           | Server error, clients should try later |
| ServiceUnavailable    | 503           | Service is temporarily unavailable |

- For  `InvalidAuthentication` and  `PermissionDenied`  I guess these happen when a client provides an invalid authentication token or authenticated client can ask for something they do not have authorisation for. Should we return 401 and 403 here?

- NotFound errors are 404, which is the most commonly used HTTP error status. Should we change this to 400 for consistency?

- For the 5xx errors, I'm not sure I see much point in specifying them. We can't guarantee that the client will get back a JSON response (one of the layers in between the client and server that know nothing about htsget may be sending back the error), so I'm not sure there's much value in specifying them. I think it would be simpler to remove these, and state that the server should handle server errors using he standard HTTP methods.

>> Added an InvalidRange error for when start > end. (Superfluous?)

> Why just this one?

I don't follow --- can you be more specific?

> Once client errors are all under 400 then servers wouldn't have to explain what's wrong. Because good clients would still work fine, so why should servers care about buggy clients? Instead of range errors etc. the spec should communicate back something useful like data formats and sequences/ranges available.

I'm not sure I follow here. Can you give an example?

> Shouldn't the version be changed to 0.2?

It seems a bit early to go for the next version already... Do we have a policy for tagging versions? Perhaps we should mark this as ``0.2-pre`` or something, and then remove the suffix when we're happy it's ready?

cyenyxe:
In those cases were the HTTP codes have a specific enough, very well known meaning, I don't see any benefit in not using them. For instance, 401, 403 and 404.

Other codes under discussion are, judging by the conversation, more open to interpretation:

* From my understanding of https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html and https://tools.ietf.org/html/rfc7231#section-6.5.6, it looks like 406 could be a better code for `UnsupportedFormat`.
* I agree with @jmarshall that 416 seems reasonable for `InvalidRange`, see https://tools.ietf.org/html/rfc7233#section-4.4

jeromekelleher:
> In those cases were the HTTP codes have a specific enough, very well known meaning, I don't see any benefit in not using them. For instance, 401, 403 and 404.

I tend to agree @cyenyxe.

However, except for these (401, 403 and 404) I think we should either use 400, or not specify the exact HTTP error code at all. I think that if there is something that contradicts our specific usage in the definition of the error code, we definitely should not use it. For example, in 416 we have  

> The 416 (Range Not Satisfiable) status code indicates that none of the ranges **in the request's Range header field** overlap 

We don't use the Range header field, so this code really isn't applicable.  By specifying both the JSON and the HTTP error code, we are giving two different definitions of the error condition, which may be contradictory.

I think it would be much simpler to make the JSON the definitive information channel for passing error data back to the client, and just use the HTTP status as "a 4xx error occured, check the JSON for more information" indicator.

delagoya:
👍 +1 for common codes, and 400 + JSON for the rest

vadimzalunin:
@jeromekelleher I'm perfectly fine with keeping 401, 403, 404, 500 and 503 - they are not specific to HTSGET, therefore should not be JSONed under 400.

> > > Added an InvalidRange error for when start > end. (Superfluous?)

> > Why just this one?

> I don't follow --- can you be more specific?

It should be under 400 as well. I see you've already changed it. 👍 

About the version, the changes are not compatible with existing implementations, therefore I would strongly recommend something like 0.2-RC or pre or whatever. Basically this PR is the only feature change in months, so unless we want to save on numbers I'd just bump the version.

jeromekelleher:
OK, thanks for the feedback @delagoya and @vadimzalunin. I'll update the PR when I get a chance to reflect this policy (400 + JSON for htsget specific stuff, and HTTP error codes for everything else).

> About the version, the changes are not compatible with existing implementations, therefore I would strongly recommend something like 0.2-RC or pre or whatever. Basically this PR is the only feature change in months, so unless we want to save on numbers I'd just bump the version.

Fair enough, I'll bump the version so.

jmarshall:
Until recently, in the HTSlib implementation there was no way to get the error response body back, only the HTTP status code.  HTSlib then translates that status code into EPERM, ENOENT, etc as appropriate.  (After samtools/htslib@e7bf06fa4828659e19dcdef0a2990661742ae11c and samtools/htslib@2448fd5629137e887948d0dc1186fe1998d2f696 and a one-line patch it's possible but annoying.)

In general, far more parts of the stack are capable of checking and interpreting an HTTP status code than can parse our locally-defined JSON.  So from an implementation point of view, it is important to also provide information in the HTTP status code.  Compare this [Amazon S3 API](http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html) for example.

Which status codes are “common” is of course a matter of opinion…

jeromekelleher:
> Until recently, in the HTSlib implementation there was no way to get the error response body back, only the HTTP status code. HTSlib then translates that status code (from libcurl used with CURLOPT_­FAILONERROR, which is a not-uncommon strategy) into EPERM, ENOENT, etc as appropriate. (After samtools/htslib@e7bf06f and samtools/htslib@2448fd5 and a one-line patch it's possible to see the error text, but annoying.)

There's no deep technical reason for this though, right? If all of the htsget specific errors have a 400 status, then the client can check to see if the status is 400, and then try to parse the JSON, and figure out what the  protocol specific error is from here. If the code is not 400,  invoke the standard error handling code from the HTTP library. (This is the advantage of making all htsget errors return HTTP 400 for me; we know when we need to parse JSON. I take the point about using 401, 403 and 404 as well though, so perhaps the client should try to parse JSON if any 4xx code is received.)

> In general, far more parts of the stack are capable of checking and interpreting an HTTP status code than can parse our locally-defined JSON. So from an implementation point of view, it is important to also provide information in the HTTP status code. Compare this Amazon S3 API for example.

I agree --- if it's a HTTP error condition, that an upstream/downstream HTTP entity can conceivably act on, then it should be handed in the standard HTTP way. If it's a htsget error condition (which, by definition, no HTTP processing entities can know anything about) then the error should be a HTTP [400,404, ...] plus JSON so that the htsget client can act on it.

Thanks for the S3 API example by the way, this was very useful. I'd imagine this error API is pretty well stress tested after the last few days!

jeromekelleher:
I've updated this to change the language a bit and remove the 5xx errors from the list. We now have:

| Type                  | HTTP Status   | Description    |
| --------------------- |-------------- | ------------------------- |
| InvalidAuthentication | 401           | Authorization provided is invalid |
| PermissionDenied      | 403           | Authorization is required to access the resource |
| NotFound              | 404           | The resource requested was not found |
| Unable                | 400           | The server is unable to fulfill the request |
| UnsupportedFormat     | 400           | The requested file format is not supported by the server |
| InvalidInput          | 400           | The request parameters do not adhere to the specification |
| InvalidRange          | 400           | The requested range cannot be satisfied. |

It's debatable whether we should keep 401 and 403 as having a required JSON body.

I've also bumped the version number to 0.2.

The current text is available [here](github.com/jeromekelleher/hts-specs/blob/a3099e3273e46450c1f8df09d545d5ae6b540582/htsget.md). I think we should probably try to merge this soon (unless someone objects of course) and try it out on the implementations.

mlin:
(on concall) it's useful to think ahead a little bit to request retry logic that clients will use for reliability at scale. this is one potentially useful distinction to draw based on HTTP status code like 500 or 503. just for reference, this is our retry logic we try to implement across client bindings for different languages: https://github.com/dnanexus/dx-toolkit/tree/master/src/api_wrappers#http-retry-logic

jeromekelleher:
I'll update the PR in a few days to include some text to this affect @mlin.

jeromekelleher:
Updates:

1. Changed version number to 0.2rc
2. Added  mention of retry on 5xx errors with reference to upstream RFC.
3. Minor tinkering with wording.

Comments/reviews? The marked up text is [here](https://github.com/jeromekelleher/hts-specs/blob/htsget-error-update/htsget.md#errors).

htsget.md
```diff
@@ -37,31 +37,56 @@ Requests to the retrieval API endpoint may be authenticated by means of an OAuth
...
+</table>
+
+The following errors  are defined:
+
+| Error			        | HTTP Status   | Description    |
+| --------------------- |-------------- | ------------------------- |
+| InvalidAuthentication	| 401	        | Authorization provided is invalid |
+| PermissionDenied	    | 403	        | Authorization is required to access the resource |
+| NotFound		        | 404	        | The resource requested was not found |
+| Unable			    | 400       	| The server is unable to fulfill the request |
```
calbach: I think the other errors you have here clearly map to the protocol, but I don't understand when a server would return "Unable" outside the context of these other 400 errors you've given. If there's no use case, I would say just drop it from here and the existing htsnexus server can continue to use the standard HTTP 406 "Unable" if they need it. If there is a use case, then could you change this description to be more specific?

jeromekelleher: Good point @calbach, I tend to agree with you. @mlin, what do you think?

jmtcsngr: Hi @jeromekelleher, I personally agree with you and @calbach, you think the change should be applied or we wait to hear back from @mlin?

jeromekelleher: I've made the change @jmtcsngr --- I think the objection clock times out after a week or so!

jeromekelleher:
I think we're pretty close to done here. Can we get some final reviews please? By my count we have one +1 from Google, so we need some more from other institutions (Sanger, EBI, ...). It would be good to get your final thoughts also @mlin.

jmtcsngr:
+1 from Sanger

jmarshall:
Has something been gained by reformatting the “Field filtering” table?  This section doesn't seem to be related to error handling and there was no note about it in the commit message.

jeromekelleher:
> Has something been gained by reformatting the “Field filtering” table? This section doesn't seem to be related to error handling and there was no note about it in the commit message.

I think this was broken before @jmarshall . I had to fix the formatting of the error table (as Vadim suggested), so it seemed best to fix the field filtering table as well for consistency. I agree these sort of incidental changes are a bad idea, so I'm happy to either drop the change or amend the commit message as you prefer.

jmarshall:
@jeromekelleher Okay, so the idea is to make these simple tables display as tables in the source repository view as well as the gh-pages web page.  Good idea, but it would be good to do this without making the text in the rows more annoying to edit.  Perhaps consider doing it as in jmarshall/hts-specs@ef87fab3c352739fff702a73585ae7f8fcaa5a6a rather than by editing all the lines in the table?

jeromekelleher:
@jmarshall annoying to edit, but easier to read (in text form)? It's a subjective one. I'll change it if you think it's important, but under (mild) protest. People who edit this should have the vim/emacs foo do this sort of alignment in their sleep!

jmarshall:
Not screwing up `git blame` casts the deciding vote in a subjective trivia question.

jeromekelleher:
OK @jmarshall, that should be it. This should be a nice clean diff now with a note in the commit history to explain why the header fiddling was undertaken. Look good to you?

vadimzalunin:
👍

AlexanderSenf:
+1

jmarshall:
> In the case of transient server errors, (e.g., 503 and other 5xx status codes), the client SHOULD implement appropriate retry logic as specified in [RFC2616&sect;10.5](https://tools.ietf.org/html/rfc2616#section-10.5).

Reading §10.5, I don't think it's clear what retry logic you're referring to.  Do you mean *Retry-After*?  I'm not sure that that header is widely used, so saying clients SHOULD be retrying via it is a bit unusual.  OTOH our protocol is in large part about retrying; how does this text interact with that in the “Reliability & performance considerations” section?

>     -Type			| HTTP status code | Description
>     +Error			| HTTP status | Description

The term in the HTTP documents is `status code`.

Do we anticipate that servers will put useful information in the JSON `message` such as “No such accession ENS16232164”?  (If so, aren't these `error types` as in the existing text, rather than `errors` which might suggest that description/message are to be written verbatim?)

> a note in the commit history to explain why the header fiddling was undertaken. Look good to you?

Re this trivia, the note is not really accurate is it: _so that they are recognised as valid Markdown_.  They were always valid Markdown for the primary purpose of this document (gh-pages); now they are also valid GFM Markdown, aiding a secondary purpose of this document (readability in GitHub repo source).  You might consider just rebasing on jmarshall/hts-specs@aa934ec3c93ae01bc5b147f8e3816190fd7fb923 rather than mixing this trivia into your error handling commit.  This would leave your commit message available for motivating the error handling changes :smile:.

It is helpful that it is now apparent that there are a few error types that are being left as is by this change.

jeromekelleher:
Thanks @jmarshall, very thorough! Unfortunately, I'm travelling at the moment and won't be able to comment/make the changes for a few days.

jeromekelleher:
Thanks again for the careful reading @jmarshall. Answers to your comments below:

> Reading §10.5, I don't think it's clear what retry logic you're referring to. Do you mean Retry-After? I'm not sure that that header is widely used, so saying clients SHOULD be retrying via it is a bit unusual. OTOH our protocol is in large part about retrying; how does this text interact with that in the “Reliability & performance considerations” section?

Yes, you're right,  §10.5 doesn't help much here does it? What I was trying to do is to give a reference to some standard definitive text on how to handle transient HTTP error conditions, rather than trying to duplicate this here in our document. Reading §10.5 again, I agree that this it's actually pretty unhelpful in this regard. Any ideas on how to do this better? 

> The term in the HTTP documents is status code.

Changed.

> Do we anticipate that servers will put useful information in the JSON message such as “No such accession ENS16232164”? (If so, aren't these error types as in the existing text, rather than errors which might suggest that description/message are to be written verbatim?)

Yes, I think this is the main idea of the error messages (i.e., useful errors to display to the user so they can figure out what they did wrong). I've changed back to "error types", and updated the example error to include your example accession to hopefully make this clearer.

> Re this trivia, the note is not really accurate is it: so that they are recognised as valid Markdown. They were always valid Markdown for the primary purpose of this document (gh-pages); now they are also valid GFM Markdown, aiding a secondary purpose of this document (readability in GitHub repo source). You might consider just rebasing on jmarshall/hts-specs@aa934ec rather than mixing this trivia into your error handling commit. This would leave your commit message available for motivating the error handling changes ￼.

Rebased.

htsget.md
```diff
@@ -37,31 +37,53 @@ Requests to the retrieval API endpoint may be authenticated by means of an OAuth
 
 ## Errors
 
-Non-successful invocations of the API return an HTTP error code, and the response body contains a JSON object (content-type application/json) with the following structure:
+The server MUST respond with an appropriate HTTP status code (4xx or 5xx) when
+an error condition is detected.  In the case of transient server errors, 
+(e.g., 503 and other 5xx status codes), the client SHOULD implement appropriate
+retry logic as specified in [RFC2616&sect;10.5](https://tools.ietf.org/html/rfc2616#section-10.5).
```
cyenyxe: We should find a homogeneous way to split paragraphs, see https://github.com/samtools/hts-specs/pull/192#discussion_r104958856 .

htsget.md
```diff
@@ -37,31 +37,53 @@ Requests to the retrieval API endpoint may be authenticated by means of an OAuth
...
+The type of error. This SHOULD be chosen from the list below.
+</td></tr>
+<tr markdown="block"><td>
+`message`  
+_string_
+</td><td>
+A message specific to the error providing information on how to debug the problem. Clients 
+MAY display this message to the user.
+</td></tr>
+</table>
```
cyenyxe: Could the style for table definitions be made more homogeneous? Either using plain markdown (my preferred one) or HTML tags, but not both in the same document.

htsget.md
```diff
@@ -37,31 +37,53 @@ Requests to the retrieval API endpoint may be authenticated by means of an OAuth
...
-UnsupportedFormat	| 409	| The requested file format is not supported by the server
-InvalidInput		| 422	| The request parameters do not adhere to the specification
-InternalError		| 500	| Server error, clients should try later
-ServiceUnavailable	| 503	| Service is temporarily unavailable
-|-
+UnsupportedFormat	| 400	| The requested file format is not supported by the server
+InvalidInput		| 400	| The request parameters do not adhere to the specification
+InvalidRange		| 400	| The requested range cannot be satisfied
+
+The error type SHOULD be chosen from this table and be accompanied by the specified HTTP
```
cyenyxe: I am not sure we should give flexibility to pick an arbitrary error type when a suitable one already exists. It will complicate error handling on the client side, as servers will be unpredictable.

htsget.md
```diff
@@ -102,34 +124,50 @@ _optional string_
 </td><td>
 Request read data in this format. Default: BAM. Allowed values: BAM,CRAM.
 
-Server replies with HTTP status 409 if the requested format is not supported.
+The server SHOULD reply with an `UnsupportedFormat` error if the requested format is not supported.
```
cyenyxe: The new version sounds more permissive than the old one.

htsget.md
jmarshall: If you can find a way to express nested tables with cells variously containing formatting and multiple paragraphs in any flavour of markdown at all, go for it.  [I don't believe it's possible¹](https://github.com/samtools/hts-specs/pull/180#issuecomment-266769555), so you'd need to either find a way to express this information without using tables or live with the use of HTML for complex tables.

After that, whether we should use markdown for simple tables (yay readable markdown) or use HTML for simple tables (yay homogeneity) is a subjective choice between yays.

[ ¹ Apologies for not realising at the time that there was a way to do the simple markdown tables in both **gh-pages** markdown and GitHub "view" markdown.]

htsget.md
jeromekelleher: For the sake of uniformity with the rest of the document, I've removed all non-paragraph line breaks. If we want to apply a different style to the document we should follow this up in another PR.

htsget.md
jeromekelleher: I agree it would be better if we could just use Markdown, but I'm going to take @jmarshall's word that this is hard to do if we're using formatting within tables. We should follow this up with another issue/PR though once we've agreed on the actual content for this PR.

htsget.md
jeromekelleher: The argument for SHOULD rather than MUST here is that servers may need to create their own specific errors that are not in the list. In this case I think it's good to have the flexibility, since servers are likely to ignore a MUST anyway, as this is hardly an exhaustive list of all the possible errors that may happen. Presumably, as the protocol is used, server vendors will identify useful error classes which may be standardised in the spec if they are more widely applicable.

htsget.md
jeromekelleher: Well, it's a choice between SHOULD or MUST here:

```
SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.
```

```
MUST   This word, or the terms "REQUIRED" or "SHALL", mean that the
   definition is an absolute requirement of the specification.
```
I don't think it's an absolute requirement of the specification here that the server reply with the right error. If they respond with a more generic error, then the client won't be able to take the appropriate corrective action.

I'm not sure though, this is just my interpretation of how these things are specified.

jeromekelleher:
Following up on outstanding issues:

-  @cyenyxe: Are you happy with the responses above particularly re. SHOULD vs MUST?
-  @jmarshall: Do you have some suggestions on how to improve on the existing text recommending how the client should go about retrying failed requests? We could just remove the reference to the RFC here.

delagoya:
I think the current looks pretty good. Clarification: this PR is for a Release Candidate, and **not** a version 1.0, correct?

tk2:
Yes, v0.2rc (release candidate towards a 1.0 release).

jmarshall:
> Reading §10.5 again, I agree that this it's actually pretty unhelpful in this regard. Any ideas on how to do this better?

Not really.  It's good to mention client retrying in this section on errors and specifically mentioning 5xx status codes.  Maybe just refer to the retrying section below?

> the client SHOULD implement appropriate retry logic as discussed in _Reliability & performance considerations_ below

jeromekelleher:
> Not really. It's good to mention client retrying in this section on errors and specifically mentioning 5xx status codes. Maybe just refer to the retrying section below?

OK, good suggestion, thanks @jmarshall. I've done this.

I would note that we're not really giving very specific instructions about how to handle retry logic in the "Reliability & performance considerations" section, but I think that we should discuss this in an issue and open a PR once there is basic agreement on how to proceed.

jeromekelleher:
I think that's @jmarshall's concern's resolved, which just leaves @cyenyxe's points about SHOULD vs MUST. Can you let us know if you are satisfied with the responses above please Cristina?

cyenyxe:
My main concern is that recovery on the client side could be much more complex to implement due to the lack of certainty.

Clients could end up needing to double (very rough estimation!) the amount of code, half of it to support servers that return a specific error (allowing proper recovery) and the other half for those that don't. The most straightforward solution for them would be not even try to recover from errors at all 😓

jeromekelleher:
> Clients could end up needing to double (very rough estimation!) the amount of code, half of it to support servers that return a specific error (allowing proper recovery) and the other half for those that don't. The most straightforward solution for them would be not even try to recover from errors at all 

I agree this is an issue, but I don't think changing the above SHOULDs to MUSTs will stop people writing non-conforming servers. To be clear, my interpretation is that the server does not conform to specification if it does not respond with the correct error under the current wording. This is something I intend to add to the testing script once the changes have been merged and we have some implementations of the updated protocol.

Let's look at the example you've highlighted above:

> The server SHOULD reply with an `UnsupportedFormat` error if the requested format is not supported.

If we say MUST here, then we are absolutely insisting that under all circumstances, the server has to respond with this error. Well, what if the server doesn't want the client to retry under a different format? What if there is a pending deletion operation on the object in question, so that the current request is valid (but for the wrong format) but any future requests will 404? Is it correct to insist that the response must be an `UnsupportedFormat` error then? I know this is a contrived example, but the point I'm trying to make is that there may well be good reasons for returning a different error condition when the server receives a request for a format it doesn't support.  Saying MUST here is just a false precision.

From a client perspective, we can say that IF it receives an UnsupportedFormat error (which should probably have a field telling the client what formats they can try in order of preference) then it MAY repeat the same request using one of the suggested formats. (Perhaps this extension could go in a follow-up PR though, this one has been dragging on for a long time now.)

jeromekelleher:
Does anyone else have an opinion on the SHOULD vs MUST issue here? If not (given @cyenyxe's comment above), I think we're clear to merge...

jeromekelleher:
OK, I think the comment clock has time out. Can someone with write access merge this PR please?

PR #193, opened 2017-03-02
New OA tag for storing Orignal Alignment information when modifying records

yfarjoun:
This could be used in GATK's IndelRealignment, but the main use I have for it is in Picard's MergeBamAlignment where it now unmaps reads that look like bacterial contamination that got mapped due to a match of 20 or so bases randomly. Since this unmaping is lossy (as one can't keep original alignment information in place) I would like to store the alignment information in a tag. I used the same format as the OA tag.

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -354,8 +354,8 @@ The resulting string is the result of joining the following six values with comm
 \item {\sf RNAME}, which must be stated explicitly; the use of `{\tt =}', which is allowed in {\sf RNEXT}, is not allowed here.
 \item POS is 1-based.
 \item Strand is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to bit 0x10 of FLAG (unset/set).
-\item CIGAR as a string.
-\item MAPQ as a numeric string.
+\item CIGAR encoded as a string.
+\item MAPQ as a string.
 \item NM, the value of the {\tt NM} tag. It may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
 \end{itemize}
 In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -92,7 +92,7 @@ or
   {\tt MQ} & i & Mapping quality of the mate/next segment \\
   {\tt NH} & i & Number of reported alignments that contain the query in the current record \\
   {\tt NM} & i & Edit distance to the reference \\
-  {\tt OA} & Z & Original Alignment \\
+  {\tt OA} & Z & Original alignment \\
   {\tt OC} & Z & Original CIGAR (deprecated; use {\tt OA} instead) \\
   {\tt OP} & i & Original mapping position (deprecated; use {\tt OA} instead) \\
   {\tt OQ} & Z & Original base quality \\
@@ -347,13 +347,13 @@ In the case of multiple unique molecular identifiers (e.g., one on each end of t
 \subsection{Original data}
 
 \begin{description}
-\item[OA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt,\,}\sf{POS}{\tt,\,}\emph{strand}{\tt,\,}\sf{CIGAR}{\tt,\,}\sf{MAPQ}{\tt,\,}\sf{NM}{\tt; )}+}]
-The tag encodes the original alignment information of the record prior to realignment or unalignment by subsequent tool. 
-The resulting string is the result of joining the following 6 values with commas `,' (no additional spaces):
+\item[OA:Z:\tagregex{(\metavar{RNAME},\metavar{POS},\metavar{strand},\metavar{CIGAR},\metavar{MAPQ},\metavar{NM};)+}]
+The original alignment information of the record prior to realignment or unalignment by a subsequent tool.
+The resulting string is the result of joining the following six values with commas `,' (no additional spaces):
 \begin{itemize}
-\item {\sf RNAME}, which must be stated explicitly; the use of \,`{\tt =}', which is allowed in {\sf RNEXT}, is not allowed here. 
+\item {\sf RNAME}, which must be stated explicitly; the use of `{\tt =}', which is allowed in {\sf RNEXT}, is not allowed here.
 \item POS is 1-based.
-\item Strand is either `{\tt +}' or `'{\tt -}`, indicating positive/negative strand, corresponding to bit 0x10 of FLAG (unset/set). 
+\item Strand is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to bit 0x10 of FLAG (unset/set).
 \item CIGAR as a string.
 \item MAPQ as a numeric string.
 \item NM, the value of the {\tt NM} tag. It may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
@@ -367,11 +367,11 @@ In particular, realignments resulting in the the removal of Secondary or Supplem
 
 \item[OC:Z:\tagvalue{cigar}]
 Original CIGAR, usually before realignment.
-Deprecated by the more general {\tt OA}.
+Deprecated in favour of the more general {\tt OA}.
 
 \item[OP:i:\tagvalue{pos}]
 Original {\sf POS}, usually before realignment.
-Deprecated by the more general {\tt OA}.
+Deprecated in favour of the more general {\tt OA}.
 
 \item[OQ:Z:\tagvalue{qualities}]
 Original base quality, usually before recalibration.
@@ -474,8 +474,7 @@ This appendix lists when standard tags were initially defined or significantly c
 \setlength{\parindent}{0pt}
 \newcommand*{\gap}{\vspace*{2ex}}
 
-
-\subsection*{September 2018}
+\subsubsection*{October 2018}
 Added the OA tag for recording original/previous alignment information. 
 
 Deprecated the OC and OP tags.
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -92,8 +92,9 @@ or
   {\tt MQ} & i & Mapping quality of the mate/next segment \\
   {\tt NH} & i & Number of reported alignments that contain the query in the current record \\
   {\tt NM} & i & Edit distance to the reference \\
-  {\tt OC} & Z & Original CIGAR \\
-  {\tt OP} & i & Original mapping position \\
+  {\tt OA} & Z & Original Alignment \\
+  {\tt OC} & Z & Original CIGAR (deprecated; use {\tt OA} instead) \\
+  {\tt OP} & i & Original mapping position (deprecated; use {\tt OA} instead) \\
   {\tt OQ} & Z & Original base quality \\
   {\tt OX} & Z & Original unique molecular barcode bases \\
   {\tt PG} & Z & Program \\
@@ -214,11 +215,13 @@ Same encoding as {\sf QUAL}.
 Sequence of the mate/next segment in the template.  See also {\tt Q2}
 for any associated quality values.
 
-\item[SA:Z:\tagregex{{\tt (}\emph{rname}{\tt ,}\emph{pos}{\tt ,}\emph{strand}{\tt ,}\emph{CIGAR}{\tt ,}\emph{mapQ}{\tt ,}\emph{NM}{\tt ;)}+}]
-Other canonical alignments in a chimeric alignment, formatted as a semicolon-delimited list.
-Each element in the list represents a part of the chimeric alignment. Conventionally, at a supplementary line, the first element points to the primary line.
-\emph{Strand} is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to FLAG bit 0x10.
-\emph{Pos} is a 1-based coordinate.
+\item[SA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt ,}\sf{POS}{\tt ,}\emph{strand}{\tt ,}\sf{CIGAR}{\tt ,}\sf{MAPQ}{\tt ,}\sf{NM}{\tt ;)}+}]
+The tag encodes the values of other canonical alignments in a chimeric alignment, formatted as a semicolon-delimited list. 
+Where \emph{strand} encodes the 0x10 bit of {\sf FLAG}---indicating positive/negative strand---such that it is '+' if the bit is unset and '--' if the bit is set.
+The other values are taken directly from the canonical read.
+Each element in the list represents a part of the chimeric alignment. 
+Conventionally, when present at a supplementary line, the first element points to the primary line.
+\emph{\sf POS} is a 1-based coordinate.
 
 \item[SM:i:\tagvalue{score}]
 Template-independent mapping quality, i.e., the mapping quality if the read were mapped as a single read rather than as part of a read pair or template.
@@ -346,11 +349,30 @@ In the case of multiple unique molecular identifiers (e.g., one on each end of t
 \subsection{Original data}
 
 \begin{description}
+\item[OA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt,\,}\sf{POS}{\tt,\,}\emph{strand}{\tt,\,}\sf{CIGAR}{\tt,\,}\sf{MAPQ}{\tt,\,}\sf{NM}{\tt; )}+}]
+The tag encodes the original alignment information of the record prior to realignment or unalignment by subsequent tool. 
+Here \emph{Strand} is either `{\tt +}' or `{\tt -}', indicating whether the read has been reverse-complemented, which corresponds to FLAG bit 0x10.
+That is, \emph{Strand} is `+' if the bit is unset and `-' if the bit is set.
+The resulting string is the result of joining the 6 values with commas `,' (no additional spaces).
+If  {\tt NM} is unknown it may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
+In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
+adding to---rather than replacing---the existing {\tt OA} information.
+
+Note 1: While commas are legal characters in sequence names, it is forbidden to have an {\tt OA} tag with such sequences.
+
+Note 2: {\sf RNAME} must be stated explicitly: the use of \,`=', which is allowed in {\sf RNEXT}, is not allowed here.
+
+Note 3: The {\tt OA} field is designed to provide record-level information that can be useful for understanding the provenance of the information in a record. 
+It is not designed to provide a complete history of the template alignment information. 
+In particular, realignments resulting in the the removal of Secondary or Supplementary records will cause the loss of all tags associated with those records, and may alternatively leave the {\tt SA} tag in an invalid state.
+
 \item[OC:Z:\tagvalue{cigar}]
 Original CIGAR, usually before realignment.
+Deprecated by the more general {\tt OA}.
 
 \item[OP:i:\tagvalue{pos}]
-Original 1-based mapping position, usually before realignment.
+Original {\sf POS}, usually before realignment.
+Deprecated by the more general {\tt OA}.
 
 \item[OQ:Z:\tagvalue{qualities}]
 Original base quality, usually before recalibration.
@@ -452,6 +474,10 @@ This appendix lists when standard tags were initially defined or significantly c
 
 \setlength{\parindent}{0pt}
 \newcommand*{\gap}{\vspace*{2ex}}
+
+
+\subsection*{September 2018}
+Add OA tag for recording original/previous alignment information. 
 
 \subsubsection*{July 2018}
 
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -355,7 +355,7 @@ The resulting string is the result of joining the following six values with comm
 \item POS is 1-based.
 \item Strand is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to bit 0x10 of FLAG (unset/set).
 \item CIGAR encoded as a string.
-\item MAPQ as a string.
+\item MAPQ as a numeric mapping quality.
 \item NM, the value of the {\tt NM} tag. It may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
 \end{itemize}
 In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -363,7 +363,7 @@ adding to---rather than replacing---the existing {\tt OA} information.
 
 The {\tt OA} field is designed to provide record-level information that can be useful for understanding the provenance of the information in a record. 
 It is not designed to provide a complete history of the template alignment information. 
-In particular, realignments resulting in the the removal of Secondary or Supplementary records will cause the loss of all tags associated with those records, and may alternatively leave the {\tt SA} tag in an invalid state.
+In particular, realignments resulting in the the removal of Secondary or Supplementary records will cause the loss of all tags associated with those records, and may also leave the {\tt SA} tag in an invalid state.
 
 \item[OC:Z:\tagvalue{cigar}]
 Original CIGAR, usually before realignment.
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -474,7 +474,7 @@ This appendix lists when standard tags were initially defined or significantly c
 \setlength{\parindent}{0pt}
 \newcommand*{\gap}{\vspace*{2ex}}
 
-\subsubsection*{October 2018}
+\subsubsection*{January 2019}
 Added the OA tag for recording original/previous alignment information. 
 
 Deprecated the OC and OP tags.
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -349,15 +349,14 @@ In the case of multiple unique molecular identifiers (e.g., one on each end of t
 \begin{description}
 \item[OA:Z:\tagregex{(\metavar{RNAME},\metavar{POS},\metavar{strand},\metavar{CIGAR},\metavar{MAPQ},\metavar{NM};)+}]
 The original alignment information of the record prior to realignment or unalignment by a subsequent tool.
-The resulting string is the result of joining the following six values with commas `,' (no additional spaces):
-\begin{itemize}
-\item {\sf RNAME}, which must be stated explicitly; the use of `{\tt =}', which is allowed in {\sf RNEXT}, is not allowed here.
-\item POS is 1-based.
-\item Strand is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to bit 0x10 of FLAG (unset/set).
-\item CIGAR encoded as a string.
-\item MAPQ as a numeric mapping quality.
-\item NM, the value of the {\tt NM} tag. It may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
-\end{itemize}
+Each original alignment entry contains the following six field values from the original record, generally in their textual SAM representations, separated by commas (`{\tt ,}') and terminated by a semicolon (`{\tt ;}'):
+{\sf RNAME}, which must be explicit (unlike {\sf RNEXT}, `{\tt =}' may not be used here);
+1-based {\sf POS};
+`{\tt +}' or `{\tt -}', indicating forward/reverse strand respectively (as per bit~0x10 of {\sf FLAG});
+{\sf CIGAR};
+{\sf MAPQ};
+{\tt NM} tag value, which may be omitted (though the preceding comma must be retained).
+
 In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
 adding to---rather than replacing---the existing {\tt OA} information.
 
@@ -370,7 +369,7 @@ Original CIGAR, usually before realignment.
 Deprecated in favour of the more general {\tt OA}.
 
 \item[OP:i:\tagvalue{pos}]
-Original {\sf POS}, usually before realignment.
+Original 1-based {\sf POS}, usually before realignment.
 Deprecated in favour of the more general {\tt OA}.
 
 \item[OQ:Z:\tagvalue{qualities}]
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -215,13 +215,11 @@ Same encoding as {\sf QUAL}.
 Sequence of the mate/next segment in the template.  See also {\tt Q2}
 for any associated quality values.
 
-\item[SA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt ,}\sf{POS}{\tt ,}\emph{strand}{\tt ,}\sf{CIGAR}{\tt ,}\sf{MAPQ}{\tt ,}\sf{NM}{\tt ;)}+}]
-The tag encodes the values of other canonical alignments in a chimeric alignment, formatted as a semicolon-delimited list. 
-Where \emph{strand} encodes the 0x10 bit of {\sf FLAG}---indicating positive/negative strand---such that it is '+' if the bit is unset and '--' if the bit is set.
-The other values are taken directly from the canonical read.
-Each element in the list represents a part of the chimeric alignment. 
-Conventionally, when present at a supplementary line, the first element points to the primary line.
-\emph{\sf POS} is a 1-based coordinate.
+\item[SA:Z:\tagregex{{\tt (}\emph{rname}{\tt ,}\emph{pos}{\tt ,}\emph{strand}{\tt ,}\emph{CIGAR}{\tt ,}\emph{mapQ}{\tt ,}\emph{NM}{\tt ;)}+}]
+Other canonical alignments in a chimeric alignment, formatted as a semicolon-delimited list.
+Each element in the list represents a part of the chimeric alignment. Conventionally, at a supplementary line, the first element points to the primary line.
+\emph{Strand} is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to FLAG bit 0x10.
+\emph{Pos} is a 1-based coordinate.
 
 \item[SM:i:\tagvalue{score}]
 Template-independent mapping quality, i.e., the mapping quality if the read were mapped as a single read rather than as part of a read pair or template.
@@ -351,14 +349,14 @@ In the case of multiple unique molecular identifiers (e.g., one on each end of t
 \begin{description}
 \item[OA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt,\,}\sf{POS}{\tt,\,}\emph{strand}{\tt,\,}\sf{CIGAR}{\tt,\,}\sf{MAPQ}{\tt,\,}\sf{NM}{\tt; )}+}]
 The tag encodes the original alignment information of the record prior to realignment or unalignment by subsequent tool. 
-The resulting string is the result of joining the following6 values with commas `,' (no additional spaces):
+The resulting string is the result of joining the following 6 values with commas `,' (no additional spaces):
 \begin{itemize}
-\item {\sf RNAME} must be stated explicitly: the use of \,`=', which is allowed in {\sf RNEXT}, is not allowed here. 
+\item {\sf RNAME}, which must be stated explicitly; the use of \,`{\tt =}', which is allowed in {\sf RNEXT}, is not allowed here. 
 \item POS is 1-based.
-\item Strand is either ?+? or ?-?, indicating positive/negative strand, corresponding to FLAG bit 0x10. 
+\item Strand is either `{\tt +}' or `'{\tt -}`, indicating positive/negative strand, corresponding to bit 0x10 of FLAG (unset/set). 
 \item CIGAR as a string.
-\item mapq as a numeric string.
-\item NM (as a numeric string). It may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
+\item MAPQ as a numeric string.
+\item NM, the value of the {\tt NM} tag. It may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
 \end{itemize}
 In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
 adding to---rather than replacing---the existing {\tt OA} information.
@@ -478,7 +476,9 @@ This appendix lists when standard tags were initially defined or significantly c
 
 
 \subsection*{September 2018}
-Add OA tag for recording original/previous alignment information. 
+Added the OA tag for recording original/previous alignment information. 
+
+Deprecated the OC and OP tags.
 
 \subsubsection*{July 2018}
 
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -351,18 +351,19 @@ In the case of multiple unique molecular identifiers (e.g., one on each end of t
 \begin{description}
 \item[OA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt,\,}\sf{POS}{\tt,\,}\emph{strand}{\tt,\,}\sf{CIGAR}{\tt,\,}\sf{MAPQ}{\tt,\,}\sf{NM}{\tt; )}+}]
 The tag encodes the original alignment information of the record prior to realignment or unalignment by subsequent tool. 
-Here \emph{Strand} is either `{\tt +}' or `{\tt -}', indicating whether the read has been reverse-complemented, which corresponds to FLAG bit 0x10.
-That is, \emph{Strand} is `+' if the bit is unset and `-' if the bit is set.
-The resulting string is the result of joining the 6 values with commas `,' (no additional spaces).
-If  {\tt NM} is unknown it may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
+The resulting string is the result of joining the following6 values with commas `,' (no additional spaces):
+\begin{itemize}
+\item {\sf RNAME} must be stated explicitly: the use of \,`=', which is allowed in {\sf RNEXT}, is not allowed here. 
+\item POS is 1-based.
+\item Strand is either ?+? or ?-?, indicating positive/negative strand, corresponding to FLAG bit 0x10. 
+\item CIGAR as a string.
+\item mapq as a numeric string.
+\item NM (as a numeric string). It may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
+\end{itemize}
 In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
 adding to---rather than replacing---the existing {\tt OA} information.
 
-Note 1: While commas are legal characters in sequence names, it is forbidden to have an {\tt OA} tag with such sequences.
-
-Note 2: {\sf RNAME} must be stated explicitly: the use of \,`=', which is allowed in {\sf RNEXT}, is not allowed here.
-
-Note 3: The {\tt OA} field is designed to provide record-level information that can be useful for understanding the provenance of the information in a record. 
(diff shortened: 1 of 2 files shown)


nh13:
I'd bit hesitant to include this tag in the spec, as it seems very new, and while potentially useful, may be better suited to be in a user-defined tag.  I am not sure it has "general interest" at the moment I'd welcome the opinions of the spec maintainers and the community.

yfarjoun:
@nh13 I'm confused about your reaction. You have a 👍 on the description and a "hesitation" in a comment.

In Picard's MergeBamAlignment the alignment information is removed when it suspects a read as arising from cross-species contamination (by looking at the number of mapped bases and the fact that it is soft-clipped on both ends

I can foresee other post alignment modifications to happen in pipelines (for example in 10X pipeline there are multiple alignment events and the PA tag could store the original alignment information, When looking for SV's one can reach conclusion about the alignment that are different from the original) 

We've been asked to add this tag so that the information about the previous alignment isn't lost when the data passes through our pipeline. I'm happy to put it in a user-defined tag for now, but we are generating thousands of WGS genomes that will become available for others to use, and the cost of re-tagging might be onerous....

magicDGS:
I'd like to have this tag for evaluate performance of realignment algorithms, and I think that it will be something useful also for other purposes.

nh13:
@yfarjoun hesitation is not a "no", so I am board given I can see multiple uses now.  While one *could* store the "previous alignment" in a secondary record, and so the use case here could be to store the same information as the "SA" tag for secondary alignments rather than supplementary alignments".  If you would agree, then could we call it "SR" or "S2" for secondary alignment?  This would allow it to be used more broadly than something that is a "previous" alignment.  What do you think?

yfarjoun:
I have no problem with making it a more general tag. I think that the secondary vs. supplementary nomenclature is bad enough already...so I'd like to avoid using these terms in the tag name and description

So....how about:

AA (Alternative/Additional Alignment)
XA (eXtra Alignment)  bonus points for using X?
DA (Different Alignment)

d-cameron:
I was under the impression that the proposed "previous alignment" meant "alignment that this record had at an earlier stage in the pipeline". What's wrong with the existing OC and OP tags? Adding OR ("original reference") for completeness would make more sense than adding a new field with almost the same meaning.

d-cameron:
As for AA/XA/DA on secondary/supplmentary, isn't that already covered by the CC, CP, HI, and IH tags?

d-cameron:
As it stands, the difference between secondary and supplementary records is sane if you assume that each record is part of at most one potential alignment. That is, if you have a 100bp read with the first 50 bases aligning to a single location, and the next 50 aligning to two possible locations, you would need to write 4 records (not 3): a chimeric alignment record pair for the first possible alignment, and a chimeric alignment record pair for the second possible alignment (even though the first record in those pairs differ only by their SA,HI,CC,CP tags).

yfarjoun:
In the proposed PA/whatever tag I store in the same tag the following data:
ref,pos:strand,cigar,NM (same as the SA tag). it's too bad that the OC and
OP can't be co-opted for this purpose as I would  much rather do that then
add another tag...

d-cameron:
In which case, wouldn't it make more sense to deprecate CC, CP, OC, OP and replace them with something like CA/OA?

At least that will partially resolve the outstanding issue of CC/CP secondary alignment ambiguity cause by record with the same starting position but different CIGARs.

jkbonfield:
@d-cameron points are good here regarding either deprecating OC/OP (CC/CP are a different beast I think) or adding OR and maybe OF (flags? which includes the strand).

I like the suggested syntax of PA better than a raft of new tags, but am unsure if it is easier to do that than deprecate existing ones.

Regarding Alternative Alignments (AA) - AA is much better than XA which is a private tag (and used!).  Making this general purpose seems nice, but I'll raise a query.  What happens if an aligner emits the best alignment with alternatives in AA, and then we want to remove this alignment via MergeBamAlignment?  I assume we add the current alignment to the front of the list of AAs with some notion that undoing this implicitly means picking the first of the alternatives provided?

If we use AA then I can see it being used instead of secondary alignments in some situations.  We then have 3 choices:  emit best only, emit best only but report others, emit all (one primary and many secondary).  The latter allows random access, but maybe we don't really need that and switching from AA to real records and back again is possible albeit requiring a sort stage.

yfarjoun:
I'm happy to deprecate OC and OP, if that's the preferred option. 

I think that our tag space is sufficiently small that having (rname ,pos ,strand ,CIGAR ,mapQ ,NM) encoded as OR, OP, OS, OC, OM, and ON seems like an excessive amount of tags to spend on this  usecase....

yfarjoun:
I'm going to change the name to OA (OriginalAlignment), allow missing elements as dots `.` and deprecate OC and OP since they can be encoded using OA. 

Does that satisfy folks?

jkbonfield:
+0.99 :)

I'm happy with the idea of deprecating tags.  We should keep the original text, but add a statement like "deprecated in favour of OA" so over time people can switch.  We'll just have to support both though.

As far as missing elements as dots, why not just have them absent as we already have a comma separated list.  eg ```OA:Z:foo,100,+,*,,,```.

Also note technically this expands on an (already existing) limitation on reference names - that they should not contain commas (we should have used space in SA or fixed ```@SQ``` lines back then, but this ship has sailed).  Right now we only exclude * = for the first character and accept any non-whitespace printable for subsequent characters, but this has been overly liberal for far too long and already breaks other tags.  We should perhaps update the main spec to disallow comma, or at very least strongly recommend against it with an indication of what will break if it is used.

yfarjoun:
Anyone care to chime in? I've responded to the concerns and would like this to be merged...

@jkbonfield 
@d-cameron 
@magicDGS

jkbonfield:
+1 for the text and substantive change.

I think there is a minor oddity in the formatting, but it's still valid.  The SA:Z tag uses

```
\tagregex{{\tt (}\emph{rname}{\tt ,}\emph{pos}{\tt ,}\emph{strand}{\tt ,}\emph{CIGAR}{\tt ,}\emph{mapQ}{\tt ,}\emph{NM}{\tt ;)}+}
```

while your PA:Z tag uses:

```
\tagregex{({\tt}\emph{rname}{\tt,}\emph{pos}{\tt,}\emph{strand}{\tt,}\emph{CIGAR}{\tt,}\emph{mapQ}{\tt,}\emph{NM}{\tt;})+}
```

Note the ( and ) are in \tt blocks for SA and not in PA.  I think this is an accident given the `{\tt}` nop.  Also minor differences in spacing before the commas.  We should be consistent between both for formatting.

yfarjoun:
Thanks @jkbonfield. fixed.

d-cameron:
Looks sane to me and I'm happy with the change although some clarification regarding the following should be made:

- Original record was unmapped. Can be resolved by explicitly stating that unmapped and unknown should both be treated as comma padding.
   - This ambiguity is potentially an issue. Does anyone actually ever revert realignments?

- What should be done to hard clipping in the original alignment
  - I am strongly in favour of realignment tools writing hard clipping CIGARs consistent with the original record (not exactly the same as the ends swap if the alignment strand is different).
  - GATK indel realignment breaks GRIDSS because it strips the hard clips from split read alignments and I am unable to reconstruct the full sequence from the CIGAR because I don't know which subsequence of the full read this alignment is for.

The above is really a special case of a larger issue:

- What should be done to other SAM records that refer to this record? It would be nice if the specifications required tools that rewrote alignments to also updated the following fields in record from the same template:
 - FLAG 0x2
 - FLAG 0x8
 - FLAG 0x20
 - RNEXT field
 - TNEXT field
 - TLEN field
 - CC tag
 - CP tag
 - MC tag
 - MD tag
 - MQ tag
 - NM tag
 - SA tag
 - NM tag

Ones that that should be rewritten but you could make an argument for them not be to are:
 - AS tag 
 - H0 tag
 - H1 tag
 - H2 tag
 - NH tag
 - PG tag
 - PT tag
 - SM tag
 - U2 tag
 - UQ tag

Inconsistent data in the SAM inputs files have been the most common type of issue raised with GRIDSS. Maybe I should refresh my proposal for a "SAM strict" sub-specifications subset that requires the file to actually make sense and not just be syntactically valid.

jkbonfield:
I think really we need some form of SAM header field to indicate which fields may be invalid, if "invalid" is even correct in some cases.

Eg if we realign a read then the RNEXT/PNEXT/TLEN field may become false.  How do we correct this?  The other end may be many GBs away!  We'd have to keep track while streaming the entire file, saving the modified read names to disk and then have a second pass to correct it (as the other end may have already been written out).   The upshot is that you turn a relatively simple task into a nightmarishly complex one, plus is adds a large time penalty where typically no one cares.


A similar issue is what happens after filtering.  If we remove a record with, eg, `samtools view -F`, or done a sub-region query, then it may still be listed in the pnext/rnext daisy chain.  Technically the other fields are still correct as we're looking at a subset of the data, but if we attempt to use them we may become stuck.

Most tags are a bit easier.  If we realign something then we can just discard MD/NM (faster than recomputing, and we may not even have been using the reference).  However some of them are still "next hit" and similar which also break on sub-setting.

The right solution here I think is to indicate that these fields maybe incorrect rather than to correct them.  If you actually care then you'll need to name collate, run fixmates, and sort back again (etc).

d-cameron:
> Eg if we realign a read then the RNEXT/PNEXT/TLEN field may become false

I've collated the full list of possibly invalidated spec-defined fields and realignment possibly breaks over half the fields in the sam tags spec.

> If you actually care then you'll need to name collate, run fixmates, and sort back again (etc).

GRIDSS does care and actually fixes most of the regenerable tags but it still dies when record adjustment results in loss of information that you can't just fix (such loss of hard clipping information, or alignment-related information that the realignment did something different with/to). In some cases, you actually do need the tool changing to make the adjustments to the other fields to prevent loss of information.


In any case, +1 for this PR from me.

jkbonfield:
Completely agree with the issue of hard clipping information being lost.  That's the sort of thing that this PR is aiming at solving, although a realigner could infact preserve hard clips even without filling out a PA tag.

yfarjoun:
As usual, I'm a little confused as to when I am allowed to hit the "merge" button...I'm hoping it will be clearer here than with other PR.. :-)

magicDGS:
👍 for me... I think that consistency problems may be addressed in a different issue/PR, because they are clearly important but not only related to this change.

jkbonfield:
@yfarjoun As far as I'm concerned, when there is agreement between spec maintainers then you can go ahead and merge.   You fixed the very minor formatting issues and I'd already thumbed up, so it's an OK from me.

The subsequent comments I think are best dealt with in a new issue if we wish to tackle it, but to be honest I'm not really sure what can be done without fundamentally breaking a lot of algorithms and making them complex.  As I said, about the only workable thing I can think of is a way to indicate when fields have been invalidated, but that would be a new PR if someone wished to champion the cause.

yfarjoun:
So I'm rebasing and merging...OK?

jmarshall:
Thanks for squashing, which makes review easier at this point.  This is **not ready for merge**, comments from me are incoming.

yfarjoun:
Got it. waiting.

jmarshall:
> Original alignment of template prior to realignment/unalignment by subsequent tool. Unknown values may be dropped, but commas should remain for parsing. In the presence of an `OA` tag, a subsequent tool should append the current information after the semicolon, not replace its content.

I think you mean _segment_, not _template_.  And in general, IMHO this text is excessively abbreviated and hence unclear.  The point of having the SAMtags document is that we have space to describe tags and don't have to abbreviate!  Hence: “prior to realignment or unalignment by a subsequent tool”. “If any of the original rname, NM, etc are unknown, they may be omitted but the comma delimiters are retained for ease of parsing.” (But can any of them really actually be unknown besides NM?) ”a subsequent tool should append another set of original alignment information after the semicolon, adding to rather than replacing the existing `OA` information.”  Or similar.

SAMtags.tex
```diff
@@ -83,8 +83,9 @@ \section{Standard tags}
   {\tt MQ} & i & Mapping quality of the mate/next segment \\
   {\tt NH} & i & Number of reported alignments that contains the query in the current record\\
   {\tt NM} & i & Edit distance to the reference \\
-  {\tt OC} & Z & Original CIGAR \\
-  {\tt OP} & i & Original mapping position \\
+  {\tt OA} & Z & Original Alignment \\
+  {\tt OC} & Z & (Deprecated in favor of {\tt OA}.) Original CIGAR \\
+  {\tt OP} & i & (Deprecated in favor of {\tt OA}.) Original mapping position \\
```
jmarshall: Write these the same way as `RT`'s deprecation note.

SAMtags.tex
```diff
@@ -246,13 +247,17 @@ \subsection{Barcodes}
 \end{description}
 
 \subsection{Original data}
-
 \begin{description}
```
jmarshall: This blank line is consistent with the formatting elsewhere.  No need to make it inconsistent...

SAMtags.tex
```diff
@@ -246,13 +247,17 @@ \subsection{Barcodes}
...
+\item[OA:Z:\tagregex{{\tt (}\emph{rname}{\tt ,}\emph{pos}{\tt ,}\emph{strand}{\tt ,}\emph{CIGAR}{\tt ,}\emph{mapQ}{\tt ,}\emph{NM}{\tt ;)}+}]
+Original alignment of template prior to realignment/unalignment by subsequent tool. 
+Unknown values may be dropped, but commas should remain for parsing. 
+In the presence of an {\tt OA} tag, a subsequent tool should append the current information after the semicolon, not replace its content.
+
 \item[OC:Z:\tagvalue{cigar}]
-Original CIGAR, usually before realignment.
+Deprecated, use {\tt OA instead}. Original CIGAR, usually before realignment.
```
jmarshall: Write these after the “Original CIGAR etc” text.  On another line would make looking at the diffs better.  Fix the formatting.

jmarshall:
Also in the trivia department: don't add trailing spaces; the commit message should be talking about `OA` rather than “New PA tag …”.

yfarjoun:
@jmarshall I responded to your comments, rebased onto master and added to the change-log.

Please see if we are good to go now, thanks!

yfarjoun:
@jmarshall ?

jkbonfield:
Going through some of the stagnated PRs, so a bump on this one.

I'm happy with it and Yossi obviously is.  @jmarshall are you good to go on this?

yfarjoun:
I say, no negative comments for a month and two maintainers is good to merge!

jkbonfield:
It's almost a year since the first PR and closing in on 6 months waiting for sign off.

Agreed with the sentiment - lack of negative comments means good to go.  Please merge it Yossi.

jmarshall:
There are three SAM maintainers and the convention is that we operate by consensus. So two maintainers does not suffice. I am re-reviewing this now.

jmarshall:
I assume you are planning to squash this before merging it? Please show us the squashed commit, with an accurate commit message as requested in August. Also as previously requested, please don't be adding trailing spaces.

The bulk of the change is this paragraph:

> Original alignment of segment prior to realignment or unalignment by subsequent tool. If any of the fields' values (say {\tt NM}, or {\tt MQ} etc.) are unknown they may be omitted but, comma delimiters must be retained for parsing. In the presence of an existing {\tt OA} tag, a subsequent tool should append another set of original alignment information after the semicolon, adding to---rather than replacing---the existing OA information.

What's MQ? Should be mapQ probably. If you have `etc.`, use `…etc.)\@ are…` to fix the spacing; but I'm not sure the parenthesised bit is helping. `but,` comma is misplaced; `for parsing` is overabbreviated — explain what we're trying to achieve, which is **unambiguous** parsing into the fields.

IMHO we should specify that `rname` in this tag must not be `=`. How is `strand` represented in this tag?

jkbonfield:
@jmarshall Yes there are three maintainers, but 2+silence out of 3 is a consensus in my book.  If you had comments, it would be _really_ helpful if they are added at the time the others agree rather than several months later.  You were prodded several times starting in September!

I appreciate your efforts and pointing out the flaws, but when all said and done, if we merge something and then later need to adjust for clarity / English then so be it.  It's better than excessive delays.

jmarshall:
@jkbonfield If we want to discuss the conventions under which we operate, let's do it offline.

jkbonfield:
I should also point out, that we did discuss this at the last meeting, it was in my email to the GA4GH file formats list on 11th Jan 2018, bumped again by Rishi on 14th Feb.  I really don't see why we should be having this conversation at all - either online or offline, but frankly I prefer somewhere that is public and can't be simply ignored!

Please either comment in a timely fashion or accept that we merge things and raise a second issue to fix them - no objections to that at all.

jkbonfield:
Discussion within the office this afternoon included that comma is a legal character in reference sequence names.  I'm sure we discussed this on a phone call at some point, and dismissed it as a problem for OA as we already have this "problem" with SA.

However, I would go one step further - *ban* comma from contig names in the next SAM release.  It's not used right now as far as we can see (bar one utterly broken local reference we found that had all sorts of dross erroneously put into the headers), so we limiting it before we get in this pickle is probably a good thing.

jkbonfield:
Good comments about rname not being = and strand being undefined, but I'd add that we should update SA tag at the same time regarding these as it also doesn't state what these are.

yfarjoun:
I think I responded to all your requests @jkbonfield

jkbonfield:
The comments were @jmarshall's rather than mine, but thanks for the updates.

I don't think "carefully" is sufficient for handling commas in names as I don't see how it's even possible to disambiguate all cases.  I'd love for us to be fixing SAM to simply disallow comma from 1.6(?) onwards and then we can remove such language.

Also for completeness, I think "MAPQ" instead of "mapQ" as this is the typographic term used in the SAM specification and hopefully it gives a hint to people that we're referring to a SAM field rather than a new thing.

daviesrob:
Would it be better to store `flag` instead of `strand`?  My reasoning for this is that if a read or its pair becomes unmapped during realignment then more of the bits in `FLAG` need to change than just the strand.  Without storing them it would be necessary to run a `fixmates`-type program to get everything straight again if you wanted to restore the original data.

Also what happens with secondary or supplementary reads, which could conceivably disappear completely during realignment?

jmarshall:
I have taken the liberty of pushing fixes for the obvious formatting problems (and addressing my review comments from last August). Did you check how this looked in PDF form before presenting it for review?

A description of the _strand_ subfield needs to talk about the FLAG bits it is derived from rather than use vague reference to positive/negative. But I agree with @daviesrob that maybe we need to store some of the other flag bits too. We shouldn't necessarily just have a +/- _strand_ character paralleling CT/PT/SA if this tag's use doesn't parallel those.

If we all agree that commas in reference sequence names are a bad thing that we'd like to get rid of, then we don't need to pander to them in this tag. We could just have a footnote saying something like

> It is not recommended to add an original alignment with an RNAME containing `,` or `;` characters to the **OA** tag, as parsing the resulting tag value will be difficult.

(Or _It is forbidden to_ if we prefer…)

d-cameron:
I'm in the processing of implementing some realignment processes in my software and I've hit an edge case: how to we record the original alignments for chimeric alignments?

For the realignment to be lossless, we need the ability to record all partial alignments in a chimeric alignment record.

yfarjoun:
are you asking how do we note, *in the main record, *the change to the SA
tag? or are you asking how to we note, *in the supplementary record,* its
previous alignment? (or perhaps something else...)

yfarjoun:
Alternatively, I could change the order of the fields, and put RNAME last, which would make parsing a RNAME with a comma in it possible....

yfarjoun:
(although then it might get confusing if there's a sequence name with a semi colon in it....)

jkbonfield:
Approved.

I'm in agreement with simply disallowing comma in names for this tag.  It already causes problems elsewhere (SA tag) and see no reason to try and pander to making it work here via reordering shenanigans.

The use of FLAGs ought to permit supplementaries to be stored as well.  If something was supplementary and after realignment is completely non-existant (folded back into the primary) then it could be stored as an unmapped supplementary alignment with OA tag to permit undo, although frankly I think we're pushing the use case a bit far at this stage.

d-cameron:
>are you asking how do we note, *in the main record, *the change to the SA
tag? or are you asking how to we note, *in the supplementary record,* its
previous alignment? (or perhaps something else...)

In my use case, I'm perform split read realignment and I want to know how many bases are in common between this alignment and the previous alignment. This is simple when the previous alignment is not chimeric but when it is, it appears we lose information.


>Would it be better to store flag instead of strand? My reasoning for this is that if a read or its pair becomes unmapped during realignment then more of the bits in FLAG need to change than just the strand. Without storing them it would be necessary to run a fixmates-type program to get everything straight again if you wanted to restore the original data.

I'm not sure how using FLAGs instead of strand helps here. By realigning we've broken FLAGs 0x2, 0x8, 0x20, but we've also broken RNEXT, TNEXT, TLEN and the CC, CP, MC, MD, MQ, NM, SA tags so a proper realignment that writes a SAM that is internally consistent have to do a fixmate-type operation. Restoring back would again naturally require another fixmate-type operation. I don't think we should be catering to programs that write SAM files that don't make sense.

Writing FLAG has the side effect that reverting an alignment will also undo subsequent non-alignment operations such as duplicate marking. I feel that strand is a better choice than FLAG.

> Also what happens with secondary or supplementary reads, which could conceivably disappear completely during realignment?

This depends on whether we are considering realignment as something that is done at the template, read, or alignment level. I've been implicitly thinking of it as an alignment-level operation but that's not necessarily the case. As an alignment-level operation, secondary or supplementary records should be removed when the chimeric alignment they are involved is replaced by a new alignment. For example, if a split read has 4 chimeric alignments, and you realign only one of these, then the the main (secondary) and additional (secondary+supplementary) records associated with the alignment (ie those sharing a common SA chimeric alignment) are removed and the new alignment record is written.

d-cameron:
Proposal: chimeric alignment can be handled by allowing multiple OA values by the following minor change to the regex `(RNAME,POS,FLAG,CIGAR,MAPQ,NM,)*;)+`

yfarjoun:
Do you mean `+` instead of `*`? also, is the intension that the "old" SA tags are put into the later values?

d-cameron:
> Do you mean + instead of *? also, is the intension that the "old" SA tags are put into the later values?

I meant `*`. I was under the impression that the semi-colon is the separator of multiple realignment operations. My proposal change merely generalises from a single alignment, to a chimeric alignment of zero or more values. `*` enables a lossless tracking of changes such as a read that goes from 2 alignments, then a realignment occurs and it is unmapped (hence `*` instead of `+`), then subsequently realigned.

Conceptually, the change is that each record (or just the primary if you like, recording OA in supp alignments would be redundant) records the full previous chimeric alignment (SA of all read alignment instead of just the other records).

> I think that this significantly complicates matters, one can get the old SA values by reverting all the reads to their old OA values

Not when the number of SAM records change because there are different numbers of alignments. If we don't make this change, then we need to specify what to do with supp alignments, what happens when they get realigned multiple times, which alignment is the 'previous' alignment for each chimeric alignment (e.g what goes in each OA when you go from a 2 record split alignment, to a 3 record split alignment, or from a 2 record split alignment that aligns only the first half of the read to a 2 record split alignment which aligns only the second half of the read?), and other such complications. The OA tag in it's currently form does not handle or address any of these edge cases.

If we want the OA tag to support chimeric alignments (and I certainly do), I believe my proposed change is the simplest way to do this. Dumping everything in the OA tag means it's fully backwards compatible, fully supports chimeric alignments and we don't pollute the SAM with weird record that will break tools that are not OA-aware. I'm not particularly fussed about what record separators are used, although I do favour matching the `SA` tag format. If we use `;` for the split read separators then it matches the `SA` tag which is nice but then we'll need a different separator for realignments. `;;` could work (as it isn't valid for the SA tag and doesn't introduce yet another reserved character), but so could many others. E.g. `((RNAME,POS,STRAND,CIGAR,MAPQ,NM;)*;;)+`

My proposed change is not purely academic: I've actually implemented a single-realignment version of the OA tag where I just write the previous read alignment in `SA` tag notation. My use case is for targeted assembly realignment where I perform a full realignment of an SV breakpoint assembly then check whether the new alignment overlaps the old one - for this I do need the previous alignment of the entire read. I found this worked well in practice as I could just drop all previous supplementary alignments, write the SA+primary alignment in the OA, write the new alignments (copying the record from the primary for all the supp records so all the non-alignment fields were set correctly), then just compare SA+current record to OA (using the same `SA` tag parsing logic for both) to compare the chimeric alignments. I found this simple to implement and since it's fully contained with the `OA` tags, it had zero side effects.

d-cameron:
As with the `SA` tag, the primary alignment would be the first record in each chimeric alignment record set.

d-cameron:
As I outlined previously, my use case is not the same as the GATK indel realigner use case. In my case, the realignment is being performed at the read level. For the GATK indel realignment use case, the realignment is being performed at the alignment level. Potentially these two use cases could require separate tags.

jkbonfield:
I think this is overcomplicating things.

I don't like the notion that the first item in an OA tag is always the primary alignment.  We get into a whole world of pain in defining what primary means (see the massive debacle of PNEXT,RNEXT,PNEXT when secondary alignments start to be used).  It's easier to simply say the first item in OA is for *this* read.  For primary alignments that have supplementaries now removed, they can be listed after that first record.  We wouldn't expect to be filling them out for each and every supplementary as it's just duplicate data.

Think about why these fields are here - it is to be able to get information about an earlier mapping, not to be able to query where all other reads are given any single starting point (which is what SA is for).

d-cameron:
> It's easier to simply say the first item in OA is for _this_ read.

(using Section 1.2 terminology):

All the alignments are for this `Read`. Did you mean to say this `Linear alignment`, or this `SAM record`?

jkbonfield:
Regarding your use cases, case A makes sense in this regard.  It was one of the original intentions stated in the first comment, along with an intelligent merger.

Case B is, frankly, barking up the wrong tree and an abuse of the format!  If you use a totally different aligner then everything is going to change.  New mapping qualities, whole new rafts of auxiliary tags, etc.  This is a huge undertaking and frankly why bother.  Do we really want the same BAM file containing two completely different sets of alignments? Isn't it easier in that scenario to simply have two files and be done with it.

I was being sloppy in my terminology - yes I meant record and not read.  Specifically in SA we the first entry being the primary following by all other supplementaries in the chimeric alignment, because the purpose of the tag is a quick way to identify all other places the read aligned to.  OA is totally different in use.   It's a backup of the previous alignment information and not something we use for trying to spot rearrangements, etc, so I don't follow the logic of attempting to define it in the same manner and requiring supplementaries to have OA tags with the alignment information for their primary.

Merging and splitting reads isn't something I imagine this tag to cover, but maybe @yfarjoun could comment was the intentions were.   It certainly complicates it hugely, especially reversing a merge (so splitting back in two) as that requires *ALL* information for the two original reads to be squashed into a single read.  It would be ghastly.   At best I think we would have a special OA tag to indicate "complex and not stored", to disambiguate it from as-before (not realigned), but I don't think we should be attempting to implement a fully-fledged BAM undo feature.

d-cameron:
@jkbonfield I extensively edited my post soon after posting it and my initial example was indeed a weak one. Better examples of case B are the second pass of a RNA-Seq aligner in which junction calling is performed (`Chimeric alignment`s converted to `Linear alignment`s with CIGAR N operators), GATK's SplitNCigarReads (`Linear alignment`s with CIGAR N operators to `Chimeric alignment`), and my gridss.SoftClipsToSplitReads (`Chimeric alignment` from soft clipped `Linear alignment`). These are all real-world examples of where an naive user/implementer might expect an OA tag to have the capability of reverting the incremental alignment changes made by these tools.

The UNMAPPED_READ_STRATEGY parameger of Picard tools MergeBamAlignment indicates that this tools is already using an OA-like tag to retain information about the previous alignment, although from the documentation it looks like it might be a case A like GATK IndelRealignment .

>It certainly complicates it hugely, especially reversing a merge (so splitting back in two) as that requires ALL information for the two original reads to be squashed into a single read. It would be ghastly. ... I don't think we should be attempting to implement a fully-fledged BAM undo feature.

Totally agree. My proposal for the OA tag to handle `Chimeric alignment`s is lossy as non-alignment information is lost when the number of `SAM records` for a given `Read` is reduced as only alignment information from the discarded records. For realignment operations that have a 1:1 between pre and post `SAM record`s, then both OA approaches are lossless and unproblematic.

Lossy rehydration of chimeric alignments relatively straight-forward to perform. The simplest approaches are to either populate a minimal record with read and alignment information (losing the extra tags), or clone the primary and updating the alignment information (duplicating the extra tags). I expect an implementing tool would allow some control over which tags are copied and which are dropped in a similar manner to how Picard MergeBamAlignments retains or drops tags.

> Merging and splitting reads isn't something I imagine this tag to cover, but maybe @yfarjoun could comment was the intentions were.

Merging and splitting `Reads` are definitely out of scope, the question is what do to with `Reads` with `Chimeric alignments` thus multiple constituent `SAM records`. I was under the impression that `Chimeric alignment`s weren't considered either way in the initial proposal. We can either restrict the OA tag to be only usable for operations that retain a 1:1 mapping between `Linear alignments`/`SAM records`, or make it more general and able to handle `Chimeric alignment`s in which case it'll look a lot like the SA tag. What's your take @yfarjoun?

d-cameron:
How about the OA tag include information about all previous alignments that were considered in the realignment process? For tools such as IndelRealigner, we have a 1:1 mapping. For tools in which the entire read is remapped, the OA tag will look SA-like.

d-cameron:
>it could either unmap that read, or, if we include FLAG rather than STRAND, it could filter the record.

I presume you mean the 0x200 filter flag? We'd need to expand that flag to explicitly state that this flag now has a double meaning and records with this flag set could also be OA placeholder records.

I'm a bit worried about how poorly OA interacts with secondary and supplementary alignments. Unmapping and/or filtering primary alignment records is unproblematic as the record still needs to be in the SAM file but if you unmap a secondary and supplementary alignment then it shouldn't exist in the file at all. Someone using htsjdk will get spammed with `Supplementary alignment flag should not be set for unmapped read.` validation failures (and `MAPQ should be 0 for unmapped read` if you don't wipe the MAPQ).

Essentially, unmapping of a secondary and supplementary alignment introduces a new type of sam record (zombie record?) that should be ignored for all downstream analysis except for an OA restoring program. This seems like a lot of technical debt to take on to enable support for a single operation. I would prefer data loss when unmapping secondary or supplementary alignments to having every tools that processes SAM records to have to deal with yet another special case.

Speaking of OA restoration, we need to explicitly define how we track when operations are performed.  If program PA realigns only record RA then program B realigns only record RB, and we want to revert the program B change, then need some way of tracking that we only need to revert record RB and record RA should be unchanged. Some possible solutions are writing OA for all records for every program that changes alignment (seems rather heavy handed), or another tag to track which operations were performed on this record (a list field containing the applicable @PG IDs would work).

jkbonfield:
Argh, does htsjdk really arbitrarily disallow unmapped and supplementary together?  There's nothing in the spec that forbids this.

This is all getting excessively complicated.

@yfarjoun - can you elaborate what your intended use for the tag is?  I don't mean what you want to write into it, but what tools you envisage reading it and what they'd do with the information?  Storing data without having a use isn't helpful, but we don't really have a feeling for what that use is.

d-cameron:
When a supplementary alignment is realigned and and becomes unmapped, the supplementary alignment record should be removed from the BAM file.

d-cameron:
If a realignment operation realigns an entire read, the OA tag should contain the previous primary alignment record. If a realignment operation realigns only a single record, the OA tag should contain the previous alignment of that record.

yfarjoun:
This has become so much more complicated than I had originally intended.... here are my use-cases:

1. Picard's MergeBamAlignments unmaps reads that look like bacterial contamination (only few bases actually align, soft-clips on both end of the read). To allow folks to examine what the records were prior to this operation, I want to store the original alignment information
2. GATK's --bamOut mode emits records as they have been realigned via local reassembly. I wanted to emit the original alignment to assist in debugging and other validation efforts

For both of these cases, my sense is that the operation is local to the record and so I do not need to worry about other records.

jkbonfield:
Agreed, and it's been my feeling for some time that overcomplicating this has lead to long delays and an inability to merge.

I think we just need to clarify the intended scope and make it explicit what the tag cannot do (handle merges and/or splits) and leave it at that.  It has definite use cases, but it's not a complete "undo" system.  As long as the limitations are explicit then I'm happy with this.

d-cameron:
I have the following use cases:

1. Undo GATK indel realignment. This is due to a) GATK indel realignment updating neither the mate read fields, nor the SA of the other parts of a chimeric alignment (thus resulting in an internally inconsistent BAM file), and b) it being bad for SV callers (worse SV calling in 7 or the 8 SV callers I tested).

2. Full realignment of assembly contigs. If the chimeric alignment of the contig does not overlap the originating location of the assembly, the assembly is considered a multi-mapping false positive and is filtered.


I'm in favour of dropping undo chaining support (we could add it in another field if needed support at a later date), and just having an OA tag that matches the SA tag format: the OA tag contains the alignment of whatever records were considered when the realignment was performed. For @yfarjoun's 1:1 scenarios, then it's just a single record. More complicated scenarios are lossy (e.g removed supplementary records), but I think we agree we don't want intro zombie records just for an undo scenario that we don't yet have any actual demand for.

yfarjoun:
I rebased (had to squash, sorry), but I didn't commit the PDF. 

Also, note that I modified the SA tag description a little to make it conform to our font standards.

yfarjoun:
@d-cameron how does this look to you?

SAMtags.tex
```diff
@@ -317,11 +320,29 @@ \subsection{Barcodes}
 \subsection{Original data}
 
 \begin{description}
+\item[OA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt,\,}\sf{POS}{\tt,\,}\emph{strand}{\tt,\,}\sf{CIGAR}{\tt,\,}\sf{MAPQ}{\tt,\,}\sf{NM}{\tt; )}+}]
+The tag encodes the original alignment information of the segment prior to realignment or unalignment by subsequent tool.
```
d-cameron: Either change `segment` to `record` if we only support a GATK indel realignment style single record events, or add something along the lines of:
`Realignment operations that realign an entire segment should report the full previous chimeric alignment. Realignment operations that realign a single record should report the previous alignment of only that record.`

SAMtags.tex
```diff
@@ -317,11 +320,29 @@ \subsection{Barcodes}
...
+In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
+adding to---rather than replacing---the existing {\tt OA} information.
+
+Note 1: While commas are legal characters in sequence names, it is forbidden to have an {\tt OA} tag with such sequences.
+
+Note 2: {\sf RNAME} must be stated explicitly: the use of \,`=', which is allowed in {\sf RNEXT}, is not allowed here.
+
+Note 3: The {\tt OA} field is designed to provide record-level information that can be useful for understanding the provenance of the information in a record. 
+It is not designed to provide a complete history of the template alignment information. 
+In particular, information regarding Secondary and Supplementary alignments may be lost.
```
d-cameron: +In particular, realignments resulting in the the removal of Secondary or Supplementary records will cause the loss of all tags associated with those records.

d-cameron:
@yfarjoun added some minor changes for clarity. Sorry about the delay - I'm currently on holidays and out of communications range most of the time.

SAMtags.tex
yfarjoun: I think I'll leave it on single records for now...so I'll change segment to record as you suggested.

SAMtags.tex
```diff
@@ -317,11 +320,29 @@ \subsection{Barcodes}
 \subsection{Original data}
 
 \begin{description}
+\item[OA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt,\,}\sf{POS}{\tt,\,}\emph{strand}{\tt,\,}\sf{CIGAR}{\tt,\,}\sf{MAPQ}{\tt,\,}\sf{NM}{\tt; )}+}]
+The tag encodes the original alignment information of the record prior to realignment or unalignment by subsequent tool. 
+Here {\sf strand} encodes the 0x10 bit of {\sf FLAG}---{\sf SEQ} being reverse complemented---such that it is `+' if the bit is unset and `-' if the bit is set,
```
jkbonfield: I don't know about the relative merits of \`\\{tt -}\' vs \`-' vs \`--' in TeX, but we need to be consistent.  With have both single and double minus used in this document for OA vs SA tag (with the \\tt being the old method).

SAMtags.tex
```diff
@@ -317,11 +320,29 @@ \subsection{Barcodes}
...
+Resulting string is the result of joining the 6 values with commas `,' (no additional spaces).
```
jkbonfield: "Resulting" here shouldn't start with a capital letter, or change the previous comma to a full-stop and use "The resulting string".

SAMtags.tex
jmarshall: The existing document (notably the barcodes section) uses \`{\tt +}'  \`{\tt -}'  etc. The (TeX!) quotes and \\tt emphasise that they're verbatim characters; and the \\tt makes `+` and `-` the same size. So the relative merits are let's stick with the existing convention.

jmarshall:
Somewhere along the line, the SA section has been completely rewritten. I'm not sure it is an improvement — e.g. _The other values are taken directly from the canonical read_ doesn't seem right. IMHO it would be best to remove any changes to SA from this PR and come back to them separately.

----

On PR #326 there is a commit with fixes to `\tagregex` that finally fix the horrible formatting of CT and PT thus:
CT
I think it would be clearer to cherry-pick that (i.e. the first hunk with definitions for \\tagvalue, \\tagregex, and \\metavar) and use it for OA:Z here rather than \\sf, and leave the regexp metacharacters in \\tt as they are in the rest of the document.

> Here Strand is either ‘+’ or ‘-’, indicating whether the read has been reverse-complemented, which corresponds to FLAG bit 0x10. That is, Strand is ‘+’ if the bit is unset and ‘-’ if the bit is set.

Rather than saying this twice, how about just reusing the existing text from SA? _Strand is either ‘+’ or ‘-’, indicating positive/negative strand, corresponding to FLAG bit 0x10._ —Is there anything unclear there?

Should describe each of the six subfields in order: the Note2 sentence; POS is 1-based; Strand sentence; CIGAR/mapq/NM as per SAM fields and NM may be omitted.

Remove Note1 as it's superseded by Action 10 for this meeting (on me, PR upcoming) to make commas invalid in RNAME everywhere.

Note3 can now just be a paragraph (i.e. delete the words “Note 3:”).

----

The history entry has not been merged into the current document (i.e., it conflicts with #312).

jmarshall:
Actually you have a point about the regular expressions: it's not great that both the verbatim characters and the *?+ metacharacters are in the same \\tt, and we can do better. Probably best not to introduce spaces that you have to explain away later though.

yfarjoun:
@jmarshall the unclear part, was the exact correspondence of the 0x10 bit and the +/- value since the bit is set when the read is reverse-complemented.

SAMtags.tex
```diff
@@ -317,11 +320,30 @@ \subsection{Barcodes}
...
+\item[OA:Z:\tagregex{{\tt (}\sf{RNAME}{\tt,\,}\sf{POS}{\tt,\,}\emph{strand}{\tt,\,}\sf{CIGAR}{\tt,\,}\sf{MAPQ}{\tt,\,}\sf{NM}{\tt; )}+}]
+The tag encodes the original alignment information of the record prior to realignment or unalignment by subsequent tool. 
+Here \emph{Strand} is either `{\tt +}' or `{\tt -}', indicating whether the read has been reverse-complemented, which corresponds to FLAG bit 0x10.
+That is, \emph{Strand} is `+' if the bit is unset and `-' if the bit is set.
+The resulting string is the result of joining the 6 values with commas `,' (no additional spaces).
+If  {\tt NM} is unknown it may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
+In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
+adding to---rather than replacing---the existing {\tt OA} information.
+
+Note 1: While commas are legal characters in sequence names, it is forbidden to have an {\tt OA} tag with such sequences.
```
magicDGS: What's about adding a recomendation or enforce an invalid character in the sequence name to encode the `,`? I prefer the second option, as it enforce some guidelines for encode/decode the tag - otherwise, it will be implementation dependent and will make incompatible OA tags in the wild. For example, a tool for regenerate the alignments of a BAM file where some sequence name has commas (`my,sequence`), might generate a record mapped to an unknown RNAME (`my_sequence`, if that was the workaround).

jkbonfield: If I recall there are plans to change the spec to make comma illegal ref names, given they're not currently used in any noticable amounts (see https://github.com/samtools/hts-specs/issues/291).  Id' rather go down this route, even if it's an adjustment to the spec version numbers (1.6 onwards means commas and some other chars are forbidden), than putting something in place which further promotes the idea that we can put anything we like in a name.

magicDGS: Fair enough - but that means that OA won't be accepted after #291 is addressed? Otherwise, what will happen with v1.5 files after this gets in, if there are a file in the wild with commas in the RNAME?

SAMtags.tex
```diff
@@ -346,11 +347,31 @@ \subsection{Barcodes}
...
+\item CIGAR as a string.
+\item MAPQ as a numeric string.
+\item NM, the value of the {\tt NM} tag. It may be omitted, but the comma delimiter must be retained so that the string can be easily parsed into the 6 subfields.
+\end{itemize}
+In the presence of an existing {\tt OA} tag, a subsequent tool may append another set of original alignment information after the semicolon,
+adding to---rather than replacing---the existing {\tt OA} information.
+
+The {\tt OA} field is designed to provide record-level information that can be useful for understanding the provenance of the information in a record. 
+It is not designed to provide a complete history of the template alignment information. 
+In particular, realignments resulting in the the removal of Secondary or Supplementary records will cause the loss of all tags associated with those records, and may alternatively leave the {\tt SA} tag in an invalid state.
```
jkbonfield: I prefer "and may *also* leave the SA tag in an invalid state".  Nitpicking, but something we can fix during merge.

yfarjoun:
done.

yfarjoun:
sorry about that....removed the erroneous change.

yfarjoun:
let's merge this!

yfarjoun:
👍

yfarjoun:
so many wordsmiths!!! :-)

jkbonfield:
On Wed, Dec 12, 2018 at 09:33:15AM -0800, Yossi Farjoun wrote:
> so many wordsmiths!!! :-) 

Capitalise that s in "so" please. ;-)
I'm not sure I like three exclaimation marks in a row either!!!!

-- 
James Bonfield (jkb@sanger.ac.uk)
The Sanger Institute, Hinxton, Cambs, CB10 1SA


-- 
 The Wellcome Sanger Institute is operated by Genome Research 

yfarjoun:
@jkbonfield OK to merge now?

jmarshall:
As discussed in the last couple of meetings (check the minutes), this one is with me for wording improvements. That's why you assigned it to me…

jmarshall:
While agreeing with this feature, I have two problems with the wording:

* It is true that it's not intuitive which value of bit 0x10 corresponds to forward/reverse, but this is not the place to fix that. The place to find that information is in the description of the FLAG field, and PR #371 spells it out there.

* The bullet list of fields takes a lot of space to say not very much. Saying “in their textual SAM representations” once makes this more succinct and IMHO clearer. I've pushed suggested wording in a new commit for your consideration.

yfarjoun:
I accept these changes. Thanks @jmarshall!

yfarjoun:
I hope this can be merged now....thumbs?

jmarshall:
Thanks Yossi. It's :+1: from me, happy to squash and merge along with its new buddy #371. @jkbonfield?

jmarshall:
Squashed and merged as 9bdcdbcb8a83c2eb3ee281de542cd406032cea25. :tada:

jmarshall:
@yfarjoun: Please delete the branch when it is no longer relevant.

yfarjoun:
sorry, I usually delete when I merge...but you merged..so I expected you to delete. I'll try to remember next year when I get a PR merged 😅

jmarshall:
Thanks. I tend to expect the PR author to delete their own branch… and the two expectations aren't so compatible :smile:.

Let's endeavour to merge the next one in less than a year!

PR #198, opened 2017-04-04
Fixed the value type for E2 tag.

jkbonfield:
The original intention was E2 as base calls.  See https://sourceforge.net/p/samtools/mailman/message/23947691/.  This looks to just be a typo in the original transcribing from the SAM spec.

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -128,8 +128,9 @@ Reference name of the next hit; `{\tt =}' for the same chromosome.
 \item[CP:i:\tagvalue{pos}]
 Leftmost coordinate of the next hit.
 
-\item[E2:Z:\tagvalue{qualities}]
-The 2nd most likely base calls. Same encoding and same length as {\sf QUAL}.
+\item[E2:Z:\tagvalue{bases}]
+The 2nd most likely base calls. Same encoding and same length as {\sf SEQ}.
+See also {\tt U2} for associated quality values.
 
 \item[FI:i:\tagvalue{int}]
 The index of segment in the template.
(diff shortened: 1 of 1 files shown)


jmarshall:
Not simply a SAMv1 → SAMtags transcription typo.  Actually this reverses a change made by @lh3 in ec8fe9550700fc1d727de52e277faebf4d2e59fc so it would be good to figure out whether that part of that 2010 change was typo or intended.

jkbonfield:
Good point.   It makes absolutely no sense as-is, but hopefully we can make a decision in the meeting today as clearly it mismatches the stated original intentions in the discussion thread.

Has anyone ever seen this tag in use in the wild?

SAMtags.tex
```diff
@@ -182,7 +183,7 @@ \subsection{Additional Template and Mapping data}
 
 \item[Q2:Z:\tagvalue{}]
```
yfarjoun: add `qualities` to the tagvalue

jkbonfield: Thanks.  Although there's a bigger howler than I noticed anyway.  The E2/U2 go together and R2/Q2, not E2/Q2 as my PR states.  I'll fix both of these.

jkbonfield:
Updated and force pushed new version for review.

jkbonfield:
Needs new review, but I noticed the U2 tag has the long-standing typo of "probilities".   We can fix on merge or leave it until #200 is merged which has this fixed.

jkbonfield:
Bump - please comment prior to meeting tomorrow.

yfarjoun:
👍 LGTM

PR #199, opened 2017-04-06
Improved wording of POS field.

### LINKED ISSUE #80: Interaction of POS and soft-clipping is unclear
When an alignment starts with some soft clipping, e.g. CIGAR is `3S7M`, the spec is unclear whether POS is the location of the first base in SEQ or the fourth (i.e., the first unclipped base).

Studying htslib's `bam_cigar2rlen()` and `bam_cigar_type` suggests that it is the first unclipped base, as only M/D/N/=/X operators consume reference bases.

This interpretation is implied by §2 4(3) (which discusses best practices for mappings that exceed the length of the reference sequence).  It can be inferred from the description of POS:

> POS: 1-based leftmost mapping POSition of the first **matching** base

but it would be helpful to expand on exactly what _matching_ implies here.

### LINKED ISSUE #195: SAM mapping position field
The SAM spec defines the pos field as "1-based leftmost mapping POSition of the first matching base" (section 1.4 in the current document). I assume this refers to the first M/=/X operation in the cigar string. Is this correct? If so, could that be stated explicitly in the document?

jkbonfield:
Fixes #80; fixes #195.

Second added sentence is optional, but personally I think it helps to clarify why specifically that list of CIGAR ops count and not the others.  Also wondering if "referenced by" should be "corresponding to".  Maybe change "first base" to "first reference base" (but ugly if we keep "referenced by").

Regardless it's a step in the right direction.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -5,6 +5,7 @@
 \usepackage{framed}
 \usepackage{enumitem}
 \usepackage{longtable}
+\usepackage{makecell}
 \usepackage[pdfborder={0 0 0},hyperfootnotes=false]{hyperref}
 
 \makeindex
@@ -350,8 +351,9 @@ Each bit is explained in the following table:
   also have an ordinary coordinate such that it can be placed at a
   desired position after sorting. If {\sf RNAME} is `*', no assumptions
   can be made about {\sf POS} and {\sf CIGAR}.
-\item {\sf POS}: 1-based leftmost mapping POSition of the first matching
-  base. The first base in a reference sequence has coordinate 1. {\sf
+\item {\sf POS}: 1-based leftmost mapping POSition of the first {\sf
+    CIGAR} operation that ``consumes'' a reference base (see table below).
+    The first base in a reference sequence has coordinate 1. {\sf
     POS} is set as 0 for an unmapped read without coordinate. If {\sf
     POS} is 0, no assumptions can be made about {\sf RNAME} and {\sf
     CIGAR}.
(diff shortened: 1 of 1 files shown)


jkbonfield:
Updated and force pushed new version for review.

jkbonfield:
Can I get a thumbs up or suggestion for edits on this @yfarjoun?

SAMv1.tex
```diff
@@ -362,23 +364,28 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
+  {\tt P} & 6 & no  & no  & padding (silent deletion from padded reference)\\
+  {\tt =} & 7 & yes & yes & sequence match \\
+  {\tt X} & 8 & yes & yes & sequence mismatch \\
   \hline
   \end{tabular}
   \end{center}
   \begin{itemize}
+  \item ``Consumes query'' and ``Consumes reference'' indicates
+    whether the CIGAR operation causes the alignment to step along the
+    query sequence and / or the reference sequence.
```
yfarjoun: I'd use "indicate" (plural) above and "sequence and the reference sequence, respectively." here, otherwise, 👍

jkbonfield: Agreed, will correct.

jkbonfield:
Made some grammar adjustments.  Thanks @yfarjoun

jmarshall:
Nice, but the extra columns make the `Ops` and their descriptions very far away from each other.  IMHO this looks better with the new columns added _after_ the Description column (and possibly therefore leader dots in the description column).

“…to step along the **query sequence**…” is very jargony.  Possibly something like ”this alignment line's sequence” instead?

Remove the commented-out additions; should be _“Consumes query” and “consumes reference” indicate whether…_, i.e. second _consumes_ in lower case.

yfarjoun:
can you use something like here https://tex.stackexchange.com/questions/5363/how-to-create-alternating-rows-in-a-table to make it easier on the eyes to scan down a line?

jkbonfield:
I like that idea and we could do it, but I think it should be a separate commit that also adjusts the other similar style tables (4 of them in total I think).   Alternatively we could use a more boxed-in style, as in e.g. the table in 1.3.

For now I'm thinking of layout and wording primarily.

jkbonfield:
Given the lack of feedback I'll implement the middle of the 3 options (pics above).

jmarshall:
:+1: to this revision.

I think leader dots (as in the VCFv4.3.pdf table of contents) would be preferable to web-style alternating colours (which look fairly heavy to my eye in PDF), but it would be worth mocking up the two.

jkbonfield:
Thanks John.

Table formatting is a separate issue as it covers many more than this one table and I don't want the improvement to wording to become mired in general formatting styles.  Personally I'm happy with bi-colour or leader dots.

jkbonfield:
@yfarjoun Requesting review so if appropriate we can sign off and merge on 1st June meeting.

jkbonfield:
Merged as de0ab160e6cd6d8de4d15b009306103b43aae95e

PR #200, opened 2017-04-06
Adding UMI-related Tags to the SAM spec and clarifying intent for Barcode Tags

yfarjoun:
Adding a few Tags to the SAM spec and clarifying intent for older TAGs
Follows discussion #109
replaced PR #119 

Corrected Tags
**BC**: This should be used for the *sample*-barcode so that there is no conflict in case that both sample- and molecular-barcodes are used
**QT**: This should be the quality for the *sample*-barcode BC

New Tags
**RX**: Sequence bases of the unique molecular identifier, may be corrected or raw
**QX**: Quality score of the unique molecular identifier in RX. May be corrected or raw.

**OX**: Original unique molecular barcode base
**OZ**: Quality scores of the unique molecular barcode bases in the OX tag

**MI**: Molecular-barcode sequence or identifier, may include non-sequence bases


yfarjoun:
I'm not sure how the PDF gets updated....running make doesn't seem to help. 

is that done automagically for me? Please advise.

tfenne:
I think there is still an open issue here are "raw/corrected" vs. "final/original" barcode UMI bases.  We've been doing a ton of work over in `fgbio` that is in use by several of the companies in this space who sell kits that use UMIs, and are developing technologies using UMIs.

I understand that 10X's model is to have "raw" and "possibly corrected", but my strong preference is to have `RX` the UMI bases that should be used by downstream tools and `BX` be the original barcode bases, and present only if the barcode bases have been altered in some way post-sequencing.  This is far more in keeping with the way other fields in BAM are handled (QUAL vs. OQ, cigar vs. OC, etc.).

tfenne:
Commented a little quickly because I wanted to get in before any merge happened.  I see that the "final/original" split happened, but I think it's reversed.  `RX` is in widespread use as the tag where UMIs are expected to be found, and I think that should be the "raw or corrected" tag, with "BX" being the "original".

yfarjoun:
So if I swapped the meaning of RX and BX (and also the qual tags) it would work?

tfenne:
Yes, please!

jmarshall:
> I'm not sure how the PDF gets updated....running make doesn't seem to help.

There is no automagic.  Now that you've rebased to master post- 02df1e2e05f6ec140a15bd076384479b56cbb820, you probably want `make new/SAMtags.pdf`. Does that do the trick?  If not, what about `make clean; make new/SAMtags.pdf`?

SAMtags.tex
```diff
@@ -234,10 +239,42 @@ \subsection{Barcodes}
 
 \begin{description}
 \item[BC:Z:\tagvalue{sequence}]
-Barcode sequence, with any quality scores stored in the {\tt QT} tag.
-
-\item[QT:Z:\tagvalue{qualities}]
-Phred quality of the barcode sequence in the {\tt BC} (or {\tt RT}) tag. Same encoding as {\sf QUAL}.
+Barcode sequence (Identifying the sample/library), with any quality scores (optionally) stored in the {\tt QT} tag.
+The {\tt BC} tag should match the {\tt QT} tag in length.
```
magicDGS: So that means that the `QT` tag should be concatenated by hyphens too, which is a valid quality value?

dkj:
@yfarjoun and @tfenne, the -1 simply because we're already producing output from [10X's software](https://support.10xgenomics.com/genome-exome/software/pipelines/latest/output/bam) with `RX` as raw and `BX` as corrected.

jmarshall:
The expanded **§1.3 Barcodes** section has two different descriptions of **QX** and no description of **BZ**.  WTF?

nh13:
@dkj do you see any other organizations using the tags in the way you describe (`RX` is raw, `BX` is corrected)?  How about other long-read vendors?  Admittedly I am similarly biased as @dkj since I do work for a number organizations (including a different long-read vendor _and_ a non-trivial # of other vendors using UMIs for low frequency somatic variant detection).  Nonetheless, when they produce BAMs, they all produce it using the convention @tfenne describes: `BX` is the "raw"/"previous" tag, `RX` is the "corrected"/current tag, and `MI` is a unique molecular identifier used to group reads from the same source molecule.  I would further ask for the final agreed upon convention being included in the "recommended practices" section to avoid any future confusion.

So how does a decision get made here?  This has been dragging on for very long, since companies that produce data with UMIs (10x included) are generating data in the wild and have been for a while. To my knowledge 10x is the only one using `BX` as corrected, and I agree with @tfenne that using `BX` as raw is more in the line of how it was handled with other tags.

yfarjoun:
I think that the way forward is to use the most commonly used
**reasonable**

Where a "reasonable convention" is one where we would have been OK with
even if there was no data in the wild.

dkj:
@nh13 My exposure is particularly biased, so I'd be glad to see names (I've given one company disproportionate coverage) and links to public docs. (Apologies, I know there was a set of tools mentioned previously but I'm failing to find it now).

I'm a little frustrated with any tools which use unapproved uppercase tags - and wish they had used lowercase.  ( So kudos to [PacBio](http://pacbiofileformats.readthedocs.io/en/3.0/BAM.html#use-of-read-tags-for-per-read-information)  for using lowercase).

@yfarjoun in case it qualifies towards "reasonable": I like the R in `RX` reminding me of raw and that it was what was suggested back in [November 2015](https://github.com/samtools/hts-specs/issues/109#issue-117194333) - were the other tools, or data made public, using these tags the other way round before then?

nh13:
@dkj here are a few links to various public documents:
- http://www.idtdna.com/pages/docs/default-source/educational-resources/multiplexed-linear-amplification-for-rapid-targeted.pdf?sfvrsn=2https://www.neb.com/~/media/NebUs/Files/Usage%20Guidelines/NEBNext_Direct_UMI_Bioinformatics_note.pdf
- http://bcbio-nextgen.readthedocs.io/en/latest/contents/configuration.html?highlight=umi#umis
- https://swiftbiosci.com/wp-content/uploads/2016/10/MIDRunSetupAndBioinformaticAnalysis161007.pdf

I think also that Picard's `MarkDuplicates` supported duplicate marking with in Sept 2015, and interestingly, it lists to use the `BC` tag for one of the options (`BARCODE_TAG`)!  Also look at [fgbio](https://github.com/fulcrumgenomics/fgbio) that has included tools for UMI extraction, manipulation, and analysis for a while.  This does not include the private documents and software that I cannot share, as well as the many others that I don't know about.  Note that other tools, like [UMI-tools](https://github.com/CGATOxford/UMI-tools) places the molecular barcodes in the read names!

yfarjoun:
OK. I fixed the switcherroo and the double QX. 

How do we move past this last hurdle?

SAMtags.tex
magicDGS: @yfarjoun, could you please tell me something about the case of using hyphens for the QT tag? Thank you in advance!

SAMtags.tex
```diff
@@ -234,10 +239,43 @@ \subsection{Barcodes}
...
+Should comprise of a sequence of bases. 
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and qualities with a hyphen (`{\tt -}') between the different barcodes.
+If the bases represent corrected bases, the original sequence can be stored in {\tt BX} (similar to {\tt OQ} storing the original qualities of bases.)
+
+\item[QX:Z:\tagvalue{qualities+}] 
+Phred quality of the unique molecular identifier sequence in the {\tt BX} tag. 
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+The qualities here may have been corrected (Raw bases and qualities can be stored in {\tt BX} and {\tt BZ} respectively.)
+The lengths of the {\tt QX} and the {\tt RX} tags must match. 
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and qualities with a space (`{\tt \textvisiblespace}') between the different barcodes.
```
magicDGS: Why isn't this the same in BC/QT?

jkbonfield:
Sorry for the last minute review, but as I've been explicitly asked to sign off on this in the agenda I've been discussing it with the users of these tags in our production pipeline team.  So I have a few issues.

- We have BC/QT using hyphen for both as recommended separator, but BX/BZ and RX/QX as hyphen and space instead.  It's only a recommendation so not a big issue, but it would be nice to be consistent.

- Hyphen is a tragically bad choice for quality separator as it is a quality value itself.  Hence why space was chosen I guess.  If so, we should apply this to BC too.

- There is a good argument for using the *same* separator in base and associated quality string.  Maybe we just state that this must be non-alphanumeric with the separator character never occurring in the associated quality string; we recommend space.  If space is an issue (eg adding tags to fastq headers and wanting to work tokenise on whitespace rather than tab) then this also permits extreme high quality characters that never occur - eg tilde.

- Why have we switched raw vs corrected at the last minute?  This unfortunately has the side effect of making lots of existing data invalid, which means we'll have to do a lot of transcoding before submitting samples. (See https://support.10xgenomics.com/genome-exome/software/pipelines/latest/output/bam)  Yes technically it was already invalid as they were non-standard tags, but let's not make life hard for ourselves here.  IIRC  the argument was of wanting to check only one value rather than "if BX found then use it, otherwise use QX".

- Is there much meaning to corrected quality value?  I'm inclined to say until we actually have an algorithm wishing to generate it that we omit it.  It may be that this ends up never occurring.

Finally if we can agree on one part (eg BC/QT) but not on the rest, then we should split this in two and merge the BC/QT part first so we at least get there incrementally.

SAMtags.tex
yfarjoun: @magicDGS we do not expect two different sample-barcodes on a template..

yfarjoun: nm. I'll move everything to spaces.

yfarjoun:
- I think that changing BC separator to a space is not a good thing to do due to the vast quantities of data that is out and the many programs that may relay on it. 

- I'd like to use space for qualities (only because we need a non-printing character as all the printing ones are valid qualities) and a hyphen for barcodes (to remain consistent with BC) I can clean up the text to be less ambiguous. does this make sense?

- Regarding the argument of having the same character separating the barcodes and strings...I'm not sure I buy it. I thought so too in the beginning, but thinking about what that would actually enable, I'm not sure. willing to be persuaded, but I would like to see an actual use-case where it matters.

- The important part of the switch  (as far as I am concerned) is that the string to look at (whether uncorrected, or corrected) ends up in the same place, like corrected base qualities of a read. Currently (in this PR) that tag is RX as that is what I thought the majority of the data has. **However**  I'm happy to change it to BX if that's better for many folks. currently BX is for the uncorrected UMI, *if they were corrected*.

- Regarding (un)corrected quality. I'm happy to drop it if no-one is correcting qualities.

jkbonfield:
I'm happy to accept a recommendation of dash for bases and space for qualities.  It's only a recommendation so we're not painting ourselves into a corner.

Regarding the tag switching. Let's consider the scenarios.

1. A pipeline produces raw data only and has no correction. Emits BX (bases) and BZ (qualities).  Software processes BX and either works fine or occasionally fails to collate with other reads due to sequencing errors.  (In theory it could look for the absence of RX in order to know that BX is uncorrected, if that helps diagnostics in any way, but this isn't likely to be helpful.)

2. A pipeline (eg 10X) emits raw data as RX/QX and corrects the sequence (aka matched to a tag set) and stores the result in BX, but has no corrected qualities so should not emit BZ.  Software reads the BX tag as before in order to collate reads together.

Am I missing other scenarios here?

In terms of phrasing for the spec, we should talk about "current" bases / quality and "original" bases / quality, to avoid confusion over raw potentially being in either tag.  It is primarily thinking about it in this way that has lead me to conclude we have to have 2 quality tags - not because correcting qualities makes much sense, but because it is less confusing than getting scenarios where BX is corrected and BZ is uncorrected.  We need to make it explicit that if a process modifies the bases then *both* BX/BZ get migrated to RX/QX and the new BX is emitted (with BZ if and only if it is also updated).

(Ie I've gone back  on my BZ being pointless view due to a change in terminology from raw/corrected to original/current.)

jkbonfield:
Summary:  i.e. I'm in agreement with this PR once RX/QX and BX/BZ are switched back again so it's compatible with existing data.

nh13:
@jkbonfield there's a lot of existing data that has RX/X and BX/BZ as currently described.  A LOT of folks are using `RX` as the tag to find the UMIs (see my list above).  I definitely concur about having language that describes the _current_ bases / qualities.  Perhaps others can chime in who have produced non-10X data with UMIs in tags?

jkbonfield:
Are the RX tags from fulcrumgenomics error-corrected in any way, or just the raw sequence?

magicDGS:
I'm very interested in getting BC/QT in soon, and I agree with the option of hyphen for BC and space for qualities. As @jkbonfield suggested, I think that if we split this PR into BC/QT and the rest of tags it will be nice to get them in sooner. Thank in advance!

nh13:
@magicDGS I am also anxious to get this PR into the spec, since it is going to be a pain for a lot of people regardless of the decision (partly our own doing).

@jkbonfield it can be either, depending on the protocol and desired analysis.  Consider single-umis, dual-umis, and duplex-umis in from a fixed set (we _should_ error-correct) or a random set (we _can and do_ error-correct) of possible umis.  We have a large number of customers (some much bigger than `10x` or other companies listed above) that are using `RX`, so at best that argument (`10x` data in the wild) is a draw.  Regardless of the data in the wild, what is the right thing to do?  

What's the model for making a decision here?  I haven't been a file-formats GA4GH meeting for some time, so could someone help?

jkbonfield:
Argh!  So we have two competing issues here with different vendors using the same tags for totally different uses.  We clearly can't satisfy both outfits as their completely at odds with each other.  I was mistaken in my belief earlier that labelling BX as "current" and "RX" as original was a way to marry the desire of querying one tag with the existing data in the wild.

At this point I'm tempted to just say &^*( it and go with whoever was there first and who has actually published how their tags work - that is 10x.  Sorry.  @yfarjoun?

tfenne:
A third option, which I’m loathe to suggest, that would be painful for everyone but a bit less painful than being on the losing side of defining RX/BX would be to define RX/BX as legacy/deprecated tags with text that essentially says “may contain raw or corrected barcodes, proceed with caution".  And then to define new tags with the desired behaviour that all can migrate to.  That way nobody is left with a mass of data that is contradicted by the specification.

jkbonfield:
The may well be something to be said for just declaring them as "implementation defined UMI barcodes" and be done with the whole debacle. :)  Adding new tags is equally annoying to all existing vendors and users, but I guess fair!

yfarjoun:
I'm worried that by having the sam-spec define tags that are not actually
in use, it makes itself irrelevant and the vendors will continue using the
same tags they have been.

Could go with the high volume data option and add a caveat to the effect
that "While this is the recommended use of the RX & BX tags, some pipelines
switch their meaning. You should be familiar with your data and proceed
with caution"?

In a year or two consider removing the caveat.

jkbonfield:
How do we determine what the "high value data" option is?  We only have one vendor piping up here and it's inevitable they'll say "LOTS".  Personally I've only ever seen 10X data, but that is equally biased in the opposite direction.  We only really have a few choices that don't favour one over the other, and frankly this has been so much like pulling teeth I no longer have any desire to help vendors out who brought this whole mess on themselves by breaking the specification in the first place.   So can we maybe vote for one of two options here?

1.  Define the tags as
BX - barcode from UMI.  Maybe raw or error corrected.
RX - baecode from UMI.  Maybe raw or error corrected.

2. Define the tags as
BX - deprecated UMI barcode.
BZ - deprecated qualities for BX.
RX - deprecated UMI barcode.
QX - deprecated qualities for RX.
(e.g.) IX - raw barcode sequence from UMI.
(e.g.) IQ - quality values for MX.

The second situation is basically this: http://www.folgerdigitaltexts.org/html/Rom.html#ftln-1561

You made us angry and we'll equally ignore all the spec breakers!  For the future, what we need is better validators and for the archives (NCBI, EBI, etc) to start using them and rejecting non-compliant data.

jkbonfield:
Can we please get some thoughts on the way forward from this debacle before 1st June meeting?  I don't see a way to keep everyone happy as we have two completely incompatible uses of these tags already. 

Therefore it really is just a matter of stating them to be implementation defined or deprecating them and coming up with completely new tags.

If we can decide on that one small point on 1st June meeting then we can move on.  If the decision is new tags, then we can revisit the phrasing (current/previous vs corrected/uncorrected and which tag codes to use).  Also I'd say if we choose this latter route we should immediately merge the BC clarifications so we don't hold it up.

yfarjoun:
A another option that will only requiring reprocessing the header would be allowing a specific @CO record (e.g. "RX and BX meaning are swapped in this file" ) to mean just that. In the absence of this comment BX would be the raw bases and RX would be the raw/corrected bases.

Perhaps there's a better header section to put this into....

jkbonfield:
I'd be against adding meaning to the ```@CO``` tag, but a human readable description that is purposefully not expected to be parsed and acted on by software does add clarity and provenance at least.

Can we at least get some sense of agreement between spec maintainers as the route to go here?   Option 1 (with possible @CO tag explaining which it is), Option 2 (new tags and deprecate old) or something else (please not again!)?

jkbonfield:
If we go with specifying a preference, about the only fair route then is first come first served.    It still means people have to have configurable parts to their software indicating which tags to lookup, but we have this requirement already and I think most people have implemented it that way too.

Isn't this therefore just a case of reverting your commit which swapped the order of these two tags around? (Assuming we go with the original use of these tags by 10X?)

jkbonfield:
Can we enumerate them then please?  This has been my problem all along in understanding this PR - information gits dribbled out in a small fashion.  Originally I thought it was just 10X and then I hear about Fulcrumgenomics who have the tags flipped.  Is there another vendor you know about too?  As far as I'm aware this is solely 10X vs FG currently, in which case my inclination is to go with the 10X as they got there first.  I don't particularly like going with either as it's giving in to vendors who caused us huge headaches, but if I had to pick one it certainly wouldn't be who shouted last.

nh13:
@jkbonfield FG is not representing itself here, but instead stating that there are multiple other vendors, both bigger and smaller than 10x, that have been and are using the tags in the currently proposed manner.  Some of them include iGenomX (a long-read company), IDT (various sequencing kits), and a number of our own clients, which I can unfortunately not disclose without a significant delay to get permission.  Furthermore, we have vendors (Swift Bio and others) who are using our _tools_ and have generated data with the tags as currently described.  So far I have only heard of 10x as using them in the flipped manner.  So it's not apples-to-apples when it comes to # of vendors on one side versus the other.  Also, when the original PR was made, I engaged immediately (shouted first), as we had other vendors (even other long-read ones) wanting to use the tags as described.

As a non-maintainer, I don't see how governance works or the way that this gets resolved, which is very frustrating for all of us wanting to move forward.  I would encourage you to make a decision at tomorrow's file formats meeting so that the rest of us can standardize and move forward.

jkbonfield:
Regarding number of vendors, I was thinking originators rather than users, although it gets blurry when those users then sell services and become vendors themselves.

Agreed on wanting a decision.  I came to this late after being made a spec maintainer a couple YEARS after this was first discussed!  It's frustrating (and a little rediculous frankly) that it's taken this long as it helps no one.

If we still can't decide on this tomorrow then tbh I'd argue for tossing a coin and just settling the argument once and for all!  It's more productive than endless debate.  We can't really change existing data though as the ship has long since sailed.  So the only options are deprecating these tags or adjusting them to be ill defined.  I'm happy with either if it moves us forward, but I'm not happy with something that requires many TBs of existing data to need to be rewritten (whoever's it is).

jkbonfield:
OK, so we didn't toss a coin at the meeting (boo!), but as we had people not present we decided to have a quick vote here instead so it has a *fixed* deadline after which we know we can move forward at last.  We didn't discuss how this would work, but I'll put my thoughts here.

We have 1 week to thrash out what the choices should be and perhaps how we select which is best.  We then have 3-ish weeks to vote, or less if it's clear everyone has had their say already, so we know the policy before the next GA4GH conference call.

My thoughts on choices are from above:

1. Define the tags as per the current proposal, but with corrected vs raw as undefined.  Add a recommendation that the ```@CO``` tag should give human readable text explaining which is which.
BX - barcode from UMI. Maybe raw or error corrected.
RX - baecode from UMI. Maybe raw or error corrected.

2. As per 1, but with a computer readable tag somewhere.   Perhaps in ```@RG``` nomenclature somewhere?  We don't know what this would be yet.  This gives software a way to automatically determine whether the barcode is error corrected or not.

3. Deprecate BX, BZ, RX and QX (describe their approximate meaning so people know what they *were*, but clearly marked as do-not-use) and invent new tags.  (Offers on what new names?)

4.  Keep BX/BZ/RX/QX, with 10X definitions.

5. Keep BX/BZ/RX/QX, with FG definitions.

As to how to vote,  it's possible we may have more options than people who want to have a say, and so run the risk of 1 for each and a tie, ugh.  To make this less likely, I propose just assigning weights (say 0 to 10 inclusive) to all options and summing them at the end with the highest winning.  To avoid Gerrymandering we should probably only accept votes from people who have had input to this and the previous PRs. :-)

Comments?

yfarjoun:
I think that these options encapsulate the discussion we had and should enable us to move forward with the vote.

yfarjoun:
my votes are:

1. 8 
2. 8
3. 1
4. 5
5. 10

dkj:
1. 7
2. 5
3. 1
4. 8
5. 3

jkbonfield:
1. 10
2. 7
3. 3
4. 6
5. 2

tfenne:
1. 3
2. 1
3. 10
4. 5
5. 10

nh13:
1. 5
2. 1
3. 8
4. 1
5. 10

jkbonfield:
Vote summary:

1. 8 + 7 + 10 + 3 + 5 = 33
2. 8 + 5 + 7 + 1 + 1 = 22
3. 1 + 1 + 3 + 10 + 8 = 23
4. 5 + 8 + 6 + 5 + 1 = 25
5. 10 + 3 + 2 + 10 + 10 = **35**

yfarjoun:
So, if I'm not mistaken, the vote that passed indicates that the PR can be merged as is. 

@jmarshall @jkbonfield what are the next steps?

pmarks:
Pat from 10x here. Apologies for just noticing this thread.  If people can bear it we'd like to keep this conversation going a little longer:

* First off, our apologies for not being more proactive in requesting the standardization of the BX tag. We appreciate the work everyone here has done to keep the SAM spec in good shape. We'd reiterate that we've consistently documented and used BX for the linked-read data type since 2015: (https://web.archive.org/web/20150603064509/http://software.10xgenomics.com:80/docs/pipelines/output/bam)

* Existing uses of RX discussed in the thread (where there is data in the wild) are all 'traditional' UMIs where you're tagging a collection of reads that are PCR copies of the same 300-500bp library construct. 10x Linked-reads/BX has fundamentally different semantics, so we'd argue that these concepts should use different tags.

* At this point it's extremely difficult for us to justify switching off BX, now that we have substantial amount of data in SRA/ENA, IGV support (https://github.com/igvteam/igv/blob/a87260edb9eb53709f4467afa0232648eb4f4330/src/org/broad/igv/sam/AlignmentTrack.java#L2447), and an ecosystem of tools that expect BX tags (e.g. https://github.com/walaj/bxtools, https://github.com/abishara/athena_meta, https://github.com/bcgsc/arcs, https://github.com/bcgsc/tigmint, https://github.com/grocsvs/grocsvs)

* **We'd like to float a new proposal**: Define RX as a 'traditional' UMI (as currently decided), but define BX to denote a collection of short reads generated from an a limiting dilution subsample of long molecules from the sample genome. We can stop using RX in our BAM file fairly easily, because we've always told our users to focus on BX, and one tool for BAM->FASTQ conversion we have that does consume the RX tag is configured with a custom header that we can update.

Thoughts on this amended proposal?

yfarjoun:
bitmoji

I'm at a loss about what to do now. We've had over a year of discussion here, votes and everything. 

My sense is that we should merge this now. 

@jmarshall ? @jkbonfield ? @tfenne ?

jmarshall:
Alas indeed.

Could someone who's au fait with this summarise (or point once again at a previous comment describing) which of these tags are out there in existing files in the real world already?  And who/what they're produced by?

jkbonfield:
I really *really* wish @pmarks comments had arrived earlier, especially as they are more in line with my own personal choices and reflect how I voted.  Apologies from all of us involved that this all took so long, which is probably why people drifted away and stopped keeping track of the discussion.

However I don't really see how we can do a U-turn on our decision and start it all again.  The sad truth is that whatever we do someone is going to have to change their pipeline - actually many people rather than just one.  So this simply really comes down to a case of picking something - *anything* - is better than endless discussion.

As for merging, I don't know how the current PR reflects the decision as it flip-flopped around a lot.  If it's all in sync then +1 from me.

daviesrob:
@jmarshall I don't know about the other ones, but you can find some 10x data [here in the ENA](http://www.ebi.ac.uk/ena/data/view/PRJEB14173).  It was made using the 10x 'longranger' software.

An example of the tags included are:
```
QT:Z:AAFFFJJJ   BC:Z:CCGGGTAT   QX:Z:AAFFFJJJJJJJJJJJ   AM:A:0  XM:A:1  TR:Z:GAGCTAG    TQ:Z:JJJJJJJ    AS:f:-46        RG:Z:longranger_wgs_20476_4_b37_freebayes:LibraryNotSpecified:1:HW7TJCCXX:4     XS:f:-47.5      BX:Z:CTACACCGTCAACCGC-1 XT:i:0  RX:Z:CTACACCGTCAACCGC   OM:i:0
```

`RX`, `BX` and `QX` are the opposite way around to the current text (but were correct according to the original version).  `AS` has the wrong type.  `OM`, `TR` and `TQ` are non-standard.  It would very much have helped if @pmarks had followed the specification and used lower-case for the non-standard tags.

pmarks:
Are there _any_ public datasets or software using the BX tag as described in the current PR? The fgbio repo doesn't contain the string 'BX'](https://github.com/fulcrumgenomics/fgbio/search?utf8=%E2%9C%93&q=BX&type=).  If not, I'd propose that we just defer the specification of BX, and proceed with the specification of 'RX' as written. That way we can make some progress here, not break existing users of BX, and take more time to hash out BX.  10x will transition away from RX in the next release of Long Ranger.

tfenne:
FWIW @pmarks `fgbio` doesn't contain any reference to `BX` because a) when correcting UMIs we're simply re-writing the corrected value into the `RX` tag so that downstream applications use `RX` whether the UMI was corrected or not (because in the various technologies we've seen it's only possible to correct sometimes) and b) we're waiting for this PR to finally get merged to add code to store the un-corrected UMI in another tag post-correction.

@yfarjoun Perhaps to build on @pmarks suggestion, we could define RX as we've agreed (i.e. the possibly-corrected, possibly-uncorrected final UMI for use downstream), leave `BX` undefined and then define a new tag (RY?  OX?  something else) to store the "original UMI that was present before error correction?

As others have said, I'm more for getting this merged without more endless discussion, so if a compromise can be reached very quickly (say within a week?!) I'd prefer that, otherwise I think we should go with what was voted on.

pmarks:
@tfenne that makes sense. In that case it sounds like there's probably not any non-linked-read users of the BX tag. So we're very much in support of @tfenne's proposal.

Assuming we proceed with this proposal, we will make a new PR to specify 'BX' as the (possibly-corrected) barcode for a group of haplotype limiting dilution linked-reads.

SAMtags.tex
yfarjoun: since hyphen is  a legal character for QT, the recommended implementation puts a space between the different parts.

yfarjoun:
Fixed tags as discussed and rebased to resolve conflicts. 

Any final comments?

yfarjoun:
bitmoji

jkbonfield:
I confess I'm now confused as I thought the whole purpose of FG's proposal to swap around RX and BX was on the basis of having lots of data and pipelines out there that assumed they were that way around, yet now we're told BX doesn't occur anywhere in the data nor the pipeline?

I did previously feel we should just go ahead and merge (partly based on misinformation though - see above!), we now *appear* to actually have an agreement between both 10X and FG on RX at least, delaying BX for the time being.  Is this true?  **Can you please both state how you feel RX should be specified so we can be sure we have an agreement and not another misunderstanding.**

If we can get just one tag ticked off then we should add that one asap and punt the others to the next phase of discussion.

SAMtags.tex
```diff
@@ -75,32 +76,36 @@ \section{Standard tags}
...
   {\tt MF} & ? & Reserved for backwards compatibility reasons \\
+  {\tt MI} & Z & Molecular-barcode sequence or identifier \\
   {\tt MQ} & i & Mapping quality of the mate/next segment \\
-  {\tt NH} & i & Number of reported alignments that contains the query in the current record\\
+  {\tt NH} & i & Number of reported alignments that contains the query in the current record \\
   {\tt NM} & i & Edit distance to the reference \\
   {\tt OC} & Z & Original CIGAR \\
   {\tt OP} & i & Original mapping position \\
   {\tt OQ} & Z & Original base quality \\
+  {\tt OX} & Z & Raw (uncorrected) unique molecular barcode bases \\
```
tfenne: I would prefer for this to contain the word "original" just like the other `O?` tags, to make it clearer that this is populated only when the value in the `RX` tag has been corrected.

SAMtags.tex
```diff
@@ -75,32 +76,36 @@ \section{Standard tags}
...
   {\tt H2} & i & Number of 2-difference hits \\
   {\tt HI} & i & Query hit index \\
-  {\tt IH} & i & Number of stored alignments in SAM that contains the query in the current record\\
+  {\tt IH} & i & Number of stored alignments in SAM that contains the query in the current record \\
   {\tt LB} & Z & Library \\
-  {\tt MC} & Z & CIGAR string for mate/next segment\\
+  {\tt MC} & Z & CIGAR string for mate/next segment \\
   {\tt MD} & Z & String for mismatching positions \\
   {\tt MF} & ? & Reserved for backwards compatibility reasons \\
+  {\tt MI} & Z & Molecular-barcode sequence or identifier \\
```
tfenne: Can I suggest MI be defined with: "Molecular identifier; a string which uniquely identifies the molecule within the library".  The way we're using this tag, the value is file-wide or at least library-wide unique, which UMI sequenced tend not to be, as you often combine them with other info (read start/stop, proximity to other reads with the same UMI) to infer molecular identity.

SAMtags.tex
```diff
@@ -75,32 +76,36 @@ \section{Standard tags}
...
   {\tt PT} & Z & Read annotations for parts of the padded read sequence \\
   {\tt PU} & Z & Platform unit \\
-  {\tt QT} & Z & Barcode ({\tt BC} or {\tt RT}) phred-scaled base qualities \\
   {\tt Q2} & Z & Phred quality of the mate/next segment sequence in the {\tt R2} tag \\
+  {\tt QT} & Z & Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag \\
+  {\tt QX} & Z & Quality score of the unique molecular identifier in the {\tt RX} tag \\
   {\tt R2} & Z & Sequence of the mate/next segment in the template \\
   {\tt RG} & Z & Read group \\
   {\tt RT} & Z & Barcode sequence (deprecated; use {\tt BC} instead) \\
+  {\tt RX} & Z & Sequence bases of the unique molecular identifier \\
```
tfenne: Should this note `, possibly corrected` or something like that to capture the discussion in this thread?

SAMtags.tex
```diff
@@ -236,10 +241,44 @@ \subsection{Barcodes}
...
+
+\item[QT:Z:\tagvalue{qualities}] 
+Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag. 
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the quality strings with spaces (`{\tt \textvisiblespace}') between the different strings from the same template. 
+
+\item[RX:Z:\tagvalue{sequence+}]
+Sequence bases from the unique molecular identifier. 
+These could be either corrected or uncorrected. Unlike {\tt MI}, the value may be non-unique in the file. 
+Should comprise of a sequence of bases.
```
tfenne: "comprise of" -> "be comprised of"

pmarks:
10x is comfortable specifying RX as the 'to be used' UMI tag as currently written. This conflicts with our existing usage of RX, but it's not the 'final' barcode in our usage, so linked-read specific downstream tools we're aware of aren't consuming it.  We can reasonably stop using RX without impacting our user-base.

Our main goal is to specify BX as a 'to be used' 'limiting dilution linked-read barcode. Given the large semantic differences between a UMI and a limiting dilution linked-read barcode, we'd argue that a separate tag is justified. We're aware of ~10 third-party tools that look for BX and treat it as a linked-read barcode. I'll make a PR for that next week.

tfenne:
@pmarks I seem to recall seeing that `BX` often contains non-sequence characters (e.g. `BX:Z:CTACACCGTCAACCGC-1 ` from the example higher up the thread).  Does 10X still refer to that as a molecular barcode even though it's more than just the sequence?  In this PR we've used the name "molecular identifier" to mean a tag that uniquely identifies the source molecule (in our case we're seeing clonal copies of an original molecule, while I understand you're seeing many short reads from the same original long molecule).  I wonder if `BX` wouldn't be better characterized as "linked-read identifier" rather than "barcode"?

magicDGS:
Can you open a simpler PR with only the changes for the sample-barcode tags (BC and QT) @yfarjoun? I am waiting for this tags to be included for long time, and the discussion about molecular-barcodes is preventing the change to be included...

pmarks:
@tfenne - yes our `BX` tags look like `BX:Z:CTACACCGTCAACCGC-1`, with the `-1` added to disambiguate separate libraries created from the same sample. Agreed that it is better described as an "identifier" than a "barcode" so I'll use that term. Thanks for the suggestion.

yfarjoun:
@magicDGS I think that we are close enough to a final version, I don't want to stir the mud....

am I right @tfenne @pmarks ? a 👍  from both of you would be helpful!

jkbonfield:
I would recommend squashing to two commits; format changes (spaces, typos) and barcoding/ID definitions.  Also drop the PDF doc so we can add this in its own commit as per MAINTAINER.md.

Thanks for the all the cat-herding and hard work @yfarjoun

yfarjoun:
@jkbonfield I squashed to two commits and removed the pdf change. I presume that you would like me to merge using the "rebase and merge" option so that these two commits will exist independently on the master branch...correct? As discussed the PDF will be in a following commit.

jkbonfield:
I think that should be fine, although I usually do it on the command line.

yfarjoun:
bitmoji

ablewhiskey:
I have to ask, @yfarjoun are you drawing these cartoons yourself?

yfarjoun:
ah, no @ablewhiskey...that's my [bitmoji](https://www.bitmoji.com/)

ablewhiskey:
@yfarjoun Pretty good likeness. :)

tk2:
@yfarjoun did the pdf on the [website](http://samtools.github.io/hts-specs/) get updated?

yfarjoun:
it did!

pmarks:
All, please see #238 for a PR on the BX tag.

PR #205, opened 2017-05-10
Explicitly state character encoding, locale, and RE syntax.

### LINKED ISSUE #197: SAM encoding and C-style definitions
The SAM specifications do not define a file encoding. This is problematic as some "ASCII" encoding codepoints for punctuation that differ from "US-ASCII". This is extremely problematic as many of the Regexes use these characters as part of a regex interval.

Furthermore, the regex grammar used is not defined thus is ambiguous (see http://seqanswers.com/forums/showthread.php?p=205407#post205407 for context). Similarly, operators used in defining BAM headers are not explicitly defined.

Proposed resolution:

- Define the file format as "US-ASCII"
- Add introductory text that states that
   - All regexes are POSIX basic regular expressions
   - code and pseudo-code in the document should be interpreted as C

jkbonfield:
Still to do.
- We need to explicitly add which SAM header fields we permit use of UTF8.
- CRAM?  It's a binary format so I'm didn't change it as I don't think this applies.
- Wording choices: "encodes" maybe isn't the correct word.  Suggestions?

Fixes #197.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -40,6 +40,9 @@ This specification is for version 1.5 of the SAM and BAM formats.  Each SAM and
 BAM file may optionally specify the version being used via the
 {\tt @HD VN} tag. For full version history see Appendix~\ref{sec:history}. 
 
+Unless explicitly specified elsewhere, all fields are encoded using 7-bit US-ASCII \footnote{Charset ANSI\_X3.4-1968 as defined in RFC1345.} in using the POSIX / C locale.
+Regular expressions listed use the POSIX / IEEE Std 1003.1 extended syntax.
+
 \subsection{An example}\label{sec:example}
 Suppose we have the following alignment with bases in lower cases
 clipped from the alignment. Read {\tt r001/1} and {\tt r001/2}
@@ -230,7 +233,7 @@ where \emph{name} is {\tt [0-9A-Za-z][0-9A-Za-z*+.@\_|-]*}\\\cline{2-3}
   & {\tt ID}* & Read group identifier. Each {\tt @RG} line must have a unique {\tt ID}. The value of {\tt ID}
   is used in the RG tags of alignment records. Must be unique among all read groups in header section.  Read group IDs may be modified when merging SAM files in order to handle collisions.\\\cline{2-3}
   & {\tt CN} & Name of sequencing center producing the read.\\\cline{2-3}
-  & {\tt DS} & Description.\\\cline{2-3}
+  & {\tt DS} & Description.  UTF-8 encoding may be used.\\\cline{2-3}
   & {\tt DT} & Date the run was produced (ISO8601 date or date/time).\\\cline{2-3}
   & {\tt FO} & Flow order. The array of nucleotide bases that correspond to the nucleotides used for each flow of each read.
   	Multi-base flows are encoded in IUPAC format, and non-nucleotide flows by various other characters. \emph{Format}: {\tt /\char92*|[ACMGRSVTWYHKDBN]+/}\\\cline{2-3}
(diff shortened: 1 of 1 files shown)


pd3:
+1

daviesrob:
It might be good to have a footnote defining exactly what US ASCII is.  This would be something like ANSI X3.4-1986, ISO/IEC 646:1991-INV, [ECMA-6 IRV](http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-006.pdf) or [RFC1345](https://www.rfc-editor.org/rfc/rfc1345.txt) which defines US ASCII among its many options (with canonical name ANSI_X3.4-1968).  The latter two have the great advantage that it's possible to read them without having to shell out 88 Swiss francs.

For the full fun of different versions of ASCII, see http://www.aivosto.com/vbtips/charsets-7bit.html

I'm not sure if "using the POSIX / C locale" is necessary when defining the character set.  It is relevant for the regular expressions, though.

dkj:
Responding to @jkbonfield's request for header record fields where we might allow UTF8.  I'd appreciate allowing UTF8 in:
- Anything in a `@CO` record
- `DS:` in any record
- `CL:` in `@PG` records
- _maybe_ `SM:` in `@RG` records

jkbonfield:
@daviesrob Added a footnote to the RFC and canonical name, but the main text is the same as it's what the layman will recognise.  Locale is necessary mainly for floating point values (eg "." and not ",").

VCFv4.3.tex
```diff
@@ -43,6 +43,11 @@ \section{The VCF specification}
 In order to ensure interoperability across platforms, VCF compliant implementations must support
 both LF (\texttt{\textbackslash n}) and CR+LF (\texttt{\textbackslash r\textbackslash n}) newline conventions.  
 
+Unless explicitly specified elsewhere, all fields are encoded using 7-bit
```
lbergelson: As mentioned on the call, vcf is already explicitly [UTF-8](https://github.com/jkbonfield/hts-specs/blob/3b6d69a56509eeb1e75b8686d81b5b67043652c3/VCFv4.3.tex#L83-L93).

jkbonfield: Aye, I didn't read it fully before adding that.  My error - I'll drop it and force push a new update with SAM only.

jkbonfield:
Vcf change reverted.  Now SAM only. Comments please.

jkbonfield:
In the absence of any other feedback, I will assume the suggestion from @dkj is sufficient and will request acceptance of this during the meeting.  I'll then add a change to this effect to this PR.

yfarjoun:
I think that sample in PG should not be UTF8...

jkbonfield:
Sorry for letting this slip.  I'll work on documenting the CO, DS and CL tags with wording along the lines of all fields are 7-bit ascii unless explicitly specified, with the description of those fields containing a utf8 statement.

SAMv1.tex
```diff
@@ -40,6 +40,11 @@ \section{The SAM Format Specification}
 BAM file may optionally specify the version being used via the
 {\tt @HD VN} tag. For full version history see Appendix~\ref{sec:history}. 
 
+Unless explicitly specified elsewhere, all fields are encoded using 7-bit
+US-ASCII \footnote{Charset ANSI\_X3.4-1968 as defined in RFC1345.} in
+using the POSIX / C locale.  Regular expressions listed use the POSIX
+extended syntax.
```
yfarjoun: is there a reference for this "extended syntax"? I looked and found many dubious-looking sites...

jkbonfield:
The OpenGroup who define POSIX also define basic and extended regular expressions.  Eg http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap09.html

I'm unsure if they have a nice short stable URL though.  Maybe we should just quote a specification name.

jkbonfield:
I think "IEEE Std 1003.1" is probably sufficient for Extended Regular Expressions.  We could provide a url, but it's likely not worth it as I wouldn't trust it being stable compared to just a google search.

PR #209, opened 2017-05-18
Remove referenceMD5 query parameter.

jeromekelleher:
As discussed in yesterday's conference call, this PR removes the ``referenceMD5`` query parameter. The motivations for this are:

1. It has not been yet been implemented by enough servers, and we would like to hit v 1.0 of the spec soon.
2. It was not very well specified from a client perspective (i.e., should a client expect both referenceName and referenceMD5 to both be supported? Or one or the other? Is one or other considered mandatory for a conforming server?).
3. It is arguably redundant, since actual information discovery is done upstream of htsget where accurate reference sequence specification can be done. The ``referenceName`` part of this protocol can be seen as a local reference ID, and need not have any global interpretation.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -128,23 +128,13 @@ The reference sequence name, for example "chr1", "1", or "chrX". If unspecified,
 The server SHOULD reply with a `NotFound` error if the requested reference does not exist.
 </td></tr>
 <tr markdown="block"><td>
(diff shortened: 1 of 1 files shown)


vadimzalunin:
imo local ref ids should be avoided but for the sake of iteration2 here is a +1 from me

PR #210, opened 2017-05-18
htsget referenceMD5 query parameter

mlin:
This restores the referenceMD5 query parameter as an unambiguous way to identify the desired reference sequence in a range query, which was in the draft spec but removed in #209 for reasons discussed there. This PR can serve for further discussion, hashing-out, and eventual restoration of the feature.


brainstorm:
I don't know if it's too late for this, but perhaps on future revisions of the spec, I would switch that fieldname with "checksum" instead of "md5". More general and less prone to md5 collision attacks.

PR #213, opened 2017-05-24
Fix typo in BCF spec, 1 element string is 0x17, not 0x19

### LINKED ISSUE #211: Typo in BCF spec
When looking into the BCF spec I found a few typo's. 
From 6.4.4  the "1element string" is encoded as 0x**19**, while this should be encoded as 0x**17**
(search for 0x19 gives 9 hits, all of them incorrect) 

This is in all 3 versions of   VCFv4.x.tex

pd3:
Resolves #211


cyenyxe:
According to the bug report, shouldn't this be applied to all versions of the specification?

pd3:
It does change all three, I believe!

cyenyxe:
You are right and I must be blind 😅

PR #214, opened 2017-05-24
Add PDF generation instructions

jmarshall:
Will merge this later today, unless there's any suggestions about form or location.


cyenyxe:
My only suggestion would be to split it in 2 sections, about the how and the when, respectively.

jmarshall:
e.g. just by putting a “Rationale” subsubheading after step 3 and before “It is a little inconvenient having […]”?

PR #218, opened 2017-06-01
Add version history.

### LINKED ISSUE #194: Discussion: specification versioning
Following on from discussion here: https://github.com/samtools/hts-specs/issues/40#issuecomment-285224309.  This is SAM specific discussion, but the general thought processes apply to all specs. (Although SAM appears unique here in that it has multiple versions in one document instead of multiple documents?)

The SAM specification no longer contains an explicit string explaining the version number.  We have examples that contain HD lines, but there is more than one example in the specification and historically they haven't always agreed with each other on the version number.  Further more, a SAM file with ```@HD VN:1.4``` is still a valid file within the 1.5 specification so it shouldn't necessarily be interpreted that such an example is implying the entire rest of the document is for 1.4 only.

Commit 321f786cfed538cb3a8b6e38c4a9a2ec19fcd99a removed the version number string from the specification.  I propose adding this back.  It used to also have ```-r49``` etc attached. That's hard to keep up to date manually and also tricky for git describe as we have lots of commits in other specifications (and no tags).  However maybe it's still beneficial.  I find a commit hash of limited use as by itself it doesn't inform readers which of two documents are newest. 

I would also like a standard procedure for bumping major or minor version number including commit messages.   Commit 80bc68f7619da60eaa4f43a16878c4675e930570 changed SAM spec from 1.4 to 1.5, but there is no mention of this in the commit message.  It took some digging and binary searching with git blame to figure out when this version changed.  It would be wrong to go back and correct these historically, but for future we could require something like "Updated version to X.Y" somewhere in the commit message.

Therefore I'd also think about adding tags.  These aren't ideal for hts-specs as we have many formats, but it's possible to add say SAM:v1.4, SAM:v1.5, CRAM:v2.1, CRAM:v3.0 etc.  Yes a CRAM tag will be meaningless for the SAM spec and vice versa, but it does at least mean we can checkout the SAM v1.4 spec without lots of git blaming. 

Comments?

jkbonfield:
Fixes #194 (for SAM/BAM at least.  VCF and CRAM have the version in the document already).

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -6,6 +6,7 @@
 \usepackage{enumitem}
 \usepackage{longtable}
 \usepackage[pdfborder={0 0 0},hyperfootnotes=false]{hyperref}
+\usepackage[title]{appendix}
 
 \makeindex
 
@@ -33,6 +34,10 @@ alignments. Header lines start with `{\tt @}', while alignment lines do
 not. Each alignment line has 11 mandatory fields for essential alignment
 information such as mapping position, and variable number of optional
 fields for flexible or aligner specific information.
+
+This specification is for version 1.5 of the SAM and BAM formats.  Each SAM and
+BAM file may optionally specify the version being used via the
+{\tt @HD VN} tag. For full version history see Appendix~\ref{sec:history}. 
 
 \subsection{An example}\label{sec:example}
 Suppose we have the following alignment with bases in lower cases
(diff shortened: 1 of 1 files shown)


jmarshall:
I think there is too much detail here, and we will be hard-pressed to maintain this moving forward.

Most entries of the form “Tag XY added” are superfluous.  Are we saying that it's invalid for XY to be present in a file with a declared earlier format version?  There's no reason to (unless they use a newly-invented tag type, but then that would be the significant line item instead), and as many implementations just write `VN:1.3` everywhere, I'm sure there are plenty of files in the wild with a new tags that are newer than their declared format version.

jkbonfield:
I think it's useful to know what is technically valid even if in practice it is accepted.

For example the SUPPLEMENTARY flag appeared with 1.5, but if someone wrote out a sam header and slapped VN:1.3 on it because that's what their code has always done (such things exist) then would we actually notice?  Samtools wouldn't.   It'd blindly continue and interpret the supplementary flags as intended.  The same would be true of a 1.3 file containing a B type auxiliary tag, or a QNAME of "*".  All of these are invalid, but practically speaking the version number has less impact than the actual data itself as traditonally it's been optional and most people don't update it nor check it.

So under the principle of "are we saying thats it's invalid...", pretty much everything in the history could be thrown away.

Instead I view the history simply as a summarised document of what happened and when.  *Possibly* we'd want a validator that checks and offers to update the version number appropriately, but such a tool doesn't exist yet anyway.  I'm happy to update the introductory paragraph to the appendix if you suggest something clearer.

@yfarjoun Any comments of verbage?  Too much or too little?

jmarshall:
The issue is not up-to-date programs like current versions of Samtools.  Rather, the issue is previously-existing or elderly or non-updated programs.

The SUPPLEMENTARY flag introduces a new way for multiple alignment records with the same QNAME to appear in a SAM file.  A 1.3-era program that didn't know about supplementary records processing such a file might be confused to get a bunch of records for the same read.  This is admittedly tenuous, so SUPPLEMENTARY is a bit of a borderline case but it is one of very few defining differences between 1.4 and 1.5 in 80bc68f7619da60eaa4f43a16878c4675e930570.

The **B** array type auxiliary tag is a much more serious matter.  Along with CIGAR **N** it is the defining difference of format version 1.4, and it adds a completely new way of encoding BAM auxiliary data.  A 1.3-era program that didn't know about **B** aux arrays will crash and burn if it tries to decode any tagged item that happens to come after a **B**-typed field in an alignment record.  Such a program would be rightly aggrieved to find syntax or encoding that it didn't understand in a file declaring itself to be `VN:1.3` compatible.

jkbonfield:
Merged as be8e0643c2452c6375445cf2b5ed95eb4a76bb21

At the File Formats meeting we decided to merge as-is and will discuss the verbosity as a separate issue / PR if required.  For example potentially half of this could be punted to the SAMtags document, rather than documenting the date it split off in here.

yfarjoun:
also, I think that the few items regarding clarifications can be removed.

jkbonfield:
Now you say. :-)

Anyway it's an incremental improvement, albeit not perfect.  I'm happy to discuss another round of updates if someone wishes to make a proposal.

yfarjoun:
nothing like a merged PR to sharpen the senses... :shark:

PR #219, opened 2017-06-01
RFC: Add @TG header for alignment tag declarations

daviesrob:
Solve the problem of incompatible alignment tags with the same identifier by forcing them to be declared in the header.  This allows the small two-character namespace to be effectively increased by tying the ID
used to a much bigger unique identifier.  Software writing files can resolve clashes by changing the ID used while leaving the unique identifier unchanged.

The new header has the format shown below.  It's possible that more of the fields should be compulsory (e.g. TP, UR).

```
@TG Optional field tag declaration
  ID* Tag ID used in this file.  Format /^[A-Za-z][A-Za-z0-9]$/.
  UU* Unique identifier.
  TP  Content type.  Format /^[AifZHB]$/
  UR  URI giving a canonical definition of the tag.
  DS  Short description
  DI  Desired Tag ID.  Format /^[A-Za-z][A-Za-z0-9]$/.
```

UU could either be the string representation of a UUID as described in [RFC4122,](https://www.rfc-editor.org/rfc/rfc4122.txt) or in a similar manner to Java packages as described in http://www.oracle.com/technetwork/java/codeconventions-135099.html.
All software must only use tags with a given UU if they exactly match the semantics of all other software that uses the same UU.  I'm in favour of using UUIDs here, as I suspect vendors may be reluctant to use a different vendors' reversed domain even if the tag usage is in fact identical.

The way I would imagine this working is:

* Software that wants to write a tag registers it first in the header.
  This includes the ones currently in the SAMtags document, although libraries might want to provide a simpler API for them (e.g. `register_common_tags('BC', 'LB', 'MC', NULL);` as opposed to `register_tag(id, uu, type, uri, description);`)

* As all tags will need to be declared in the header, the header becomes compulsory, even for SAM files.  **This would require a version number change.**

* APIs would resolve clashes automatically.
  For ease of identification this could be done by playing with the ID case (e.g. try rX, Rx, rx for RX) before falling back to a counter-based solution (aa, ab, ac etc.).

* Software using tags can then refer to them by the expected ID and the API will take care of the translation.
  This should make for easier adoption, as the only change needed would be to register the tags in the header so that a translation table can be built.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -243,6 +243,13 @@ meanings can be avoided.}
         program in that {\tt PG} record, and the program(s) referred to via the {\tt PP} tag. \\\cline{2-3}
   & {\tt DS} & Description.\\\cline{2-3}
   & {\tt VN} & Program version \\\cline{1-3}
+  \multicolumn{2}{|l}{\tt @TG} & Optional alignment tag declaration. \\\cline{2-3}
+  & {\tt ID}* & Tag ID used in this file. \emph{Format} {\tt /\^{}[A-Za-z][A-Za-z0-9]\$/}. \\\cline{2-3}
+  & {\tt UU}* & Unique identifier. \\\cline{2-3}
+  & {\tt TP} & Content type.  \emph{Format} {\tt /\^{}[AifZHB]\$/}. \\\cline{2-3}
+  & {\tt UR} & URI giving a canonical definition of the tag. \\\cline{2-3}
+  & {\tt DS} & Short description \\\cline{2-3}
+  & {\tt DI} & Originally desired Tag ID, for information only. \emph{Format} { \tt /\^{}[A-Za-z][A-Za-z0-9]\$/}. \\\cline{1-3}
   \multicolumn{2}{|l}{\tt @CO} & One-line text comment. Unordered multiple {\tt @CO} lines are allowed.\\
   \cline{1-3}
 \end{longtable}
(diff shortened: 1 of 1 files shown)


jkbonfield:
I don't like this as an idea for existing tags.  We could (and should if adopted) define UUIDs for existing tags, but they shouldn't be mandatory or you lose all backwards compatibility.

For clarity of the above, which confused me until discussions, this is essentially a tag indirection table much like @RG values work now.  If we merge two files with clashing RG IDs referring to different header lines then we arbitrarily rename one of the IDs.  The same would be applied for tag types with this system.

daviesrob:
Allowing some tags to be omitted from the header would complicate validation.  Validators would have to agree on a list of known tags in order to work correctly.  The job is made simpler if we just say all used tags should be listed in the header.

There are issues with how to deal with old files that don't include the `@TG` headers.  When reading, we don't know which tags will have been used until they appear, and may not know the meaning of some of them.  We may need a way of relaxing the rules for files which are derived from ones written in earlier versions of the format.

nh13:
I would prefer that ID be the unique ID for the header record/line in the file, with no restriction, so that ID is consistent for all header lines. Then there would need to be a a field for the original tag name and the UUID or equivalent. 

I echo James' comment that making this compulsory seems quite onerous for tool authors as well as for the volumes of data already generated. I would further add that then the definition of tags is effectively removed from the spec maintainers and passed to the individual users themselves. I am not sure that this is the best idea.

jmarshall:
In my view, this is very complicated and essentially defines a new format.  One of the advantages of SAM's tags is that they are human-readable: you can look at an alignment line and understand what the tagged data fields are telling you.  (Okay, having contemplated your bullet points further, I accept that (if implementations followed your suggested policy) the tags wouldn't move around until there was an actual clash to separate.)

It remains very complicated and requires buy-in from implementations.  And the legibility of the resulting SAM files is dependent on how well those implementations implement the non-trivial policies you describe in your bullet points.  (And no guidance has as yet been suggested for the spec text.)

In summary IMHO this is a lot of machinery to have to add to a number of implementations for the purpose of dealing with an issue that shouldn't occur if our processes were up to scratch.

tfenne:
My views are mostly in line with @jmarshall.  I think if SAM/BAM were a format produced only by software engineers then this would be a great solution, as most implementors and generators would be directly concerned with compliance.  But what I would expect to happen is that most tools that generate SAM or BAM would not comply with this, or would do so in a half-hearted manner such that we'd end up with both a more complicated specification _and_ less compliance.

daviesrob:
Not popular, withdrawn.

Issue #167, opened 2016-09-02
New Restrictions on Contig Names not Strict Enough


jamesemery:
I see that in section 1.4.7 of VCFv4.3, it states that contig names can mostly follow restrictions placed upon them by the SAM spec except for a few explicitly disallowed characters. It would seem that the restricted characters are not enough to preserve parsing of VCF files. Specifically, the SAM format restrictions specify (”[!-)+-<>-~][!-~]_”) which translates into any printable character so long as it isn't in the first place. We place some restrictions on this, namely specifying that the characters ”<>[]:_” are additionally non-valid. Since the contig name has to appear in a header line this runs the risk of making it impossible to parse a header string. An example contig header line might be:

`##contig=<ID=ctg1,length=81195210,Description="Foo">`

Unfortunately both the `=` and the `,` characters are legal according to the new restrictions so technically the string "ctg1,length=81195210" COULD be a valid contig name, which means that parsing is ambiguous. Something needs to be done about these two characters at least, the way I see it there are a couple options:
1. Restrict the contig name to disallow `=` and `,` would fix this problem at the expense of further restricting contigs allowed by the SAM spec.
2. Escape the ID field in the contig header line somehow, for example with a second set of `<>` or `""` characters making the header line look something like this: 
   `##contig=<ID=<ctg1>,length=81195210,Description="Foo">` which would break forward compatibility and add some complexity into parsing.
3. Encode dangerous characters in the header somehow, perhaps by percent encoding of the header contig ID field. The VCFv4.3 spec only seems to explicitly specify percent encoding for INFO and FORMAT field values. Perhaps it could be expanded/clarified to also apply to header ID line values (except for INFO and FORMAT where a regex is already imposed)?

yfarjoun:
For what it's worth, I support the first option. and I would even strengthen that to support disallowing other "annoying" characters (annoying for command-line parsing purposes) such as: 
```
"`:*()?|\/!&
```
In fact, I would support disallowing all non-alphanumerics except for a select few: 
```
-_+
```

I know that this is incompatible with SAM, and we should find a way to deal with that.

tfenne:
I think that before making anything more restrictive it's worth doing a survey of what's actually being used.  E.g. disallowing anything that's in the broadly used hs38DH (including `:` and `*`) would cause a lot of people a lot of unnecessary pain.

yfarjoun:
[...]

Issue #124, opened 2016-01-16
Problems with allowed contig names in the SAM spec


yfarjoun:
We've recently come across an hg38 contig name of the form '**HLA-A*01:01:01:01**'  which caused problems when used on the commandline in the GATK, as the parser understood the suffix :01 to mean "at position 1" and then is complained that it cannot find **HLA-A*01:01:01** (which thankfully isn't in hg38..., otherwise we would have probably not noticed this for a while. Thus the argument **-L HLA-A*01:01:01:01** is not one that the GATK can handle. 

Its too late for hg38 to ask for any name changes, but I think it might make sense to revisit the allowed characters and remove a few. Currently there are a two characters that are not allowed as the _first_ character in the @SN  (`*` and `=`) but the remaining printable characters (`!` through `~`) are allowed, and they are all allowed as the following characters. I would propose that we add a few more characters to the "not allowed" list. In particular it would be good to avoid the various quotes (`',`,"`) anywhere in the @SN and I would like to suggest that the ending should not look like a genomic position, i.e. it should **not** match the regex '`:[0-9]*(-[0-9]*)?`'. The reason I say "suggest" is that given that hg38 already includes this ending, it would be quite difficult to reverse that and even if we only put it in a future "version" of the SAM Spec it isn't clear what that means since the old references will still be around...

This isn't a formal proposal...I mostly want to know what people think about this issue and see if we can come up with good solutions.

tfenne:
@yfarjoun I can totally empathize with your problems.  I've run into similar problems with other tools (a different variant caller would, internally, invoke shell commands with naked reference names, and the unquoted `*` caused problems where.

Unfortunately I think this ship has sailed.  Fwiw I suspect your issue isn't with the main hs38 assembly, but with the hs38DH reference created by `bwa.kit`.  Any change that would prohibit any new characters in reference names, or prohibit particular suffixes, would necessarily be backwards incompatible.  And that's likely to lead to tools, moving forward, not supporting previously valid reference names, and leaving a lot of data stranded and in need of re-writing.  I'd suggest that the best we can do is to add a _best practice_ not to name sequences like this.

Presumably the problem you're having can be fixed by checking to see if ``HLA-A*01:01:01:01` or HLA-A*01:01:01` or HLA-A*01:01` is a valid reference name before choosing how to deconstruct the string?  You're only really in trouble if two or more of them are valid reference names.

yfarjoun:
(markdown is messing with your post)

[...]

PR #220, opened 2017-06-01
Add @SQ-AN alternative reference sequence names

jmarshall:
Proposed text to implement #100, "suggestion: aliases in `@SQ` section".

Also take the opportunity to pre-emptively restrict the characters allowed in these alternative names.

As noted [in last September's meeting](https://github.com/samtools/hts-specs/issues/100#issuecomment-264209422), restricting the characters allowed in `AN` alternative names could be a first step towards applying the same restriction in `SN` canonical sequence names and finally being able to parse the colon in `chr:beg-end` unambiguously.

There is debate to be had about which punctuation characters to forbid; there are existing points for consideration in #100, #124, and #167.

The primary motivation here is to disallow `:` so `chr:beg-end` region syntax is unambiguous, so we have taken advantage of colon being available as the delimiter, inspired by `TAG:TYPE:VALUE` delimiters.

Other motivation might be to disallow `,` so `chr1,chr2,chr3` list syntax is unambiguous.  If disallowing comma has particular support, perhaps using comma as the `AN` delimiter would be preferable.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -194,14 +194,28 @@ meanings can be avoided.}
     grouped by {\sf QNAME}), and {\tt reference} (alignments are grouped by
     {\sf RNAME}/{\sf POS}).\\\cline{1-3}
   \multicolumn{2}{|l}{\tt @SQ} & Reference sequence dictionary. The order of {\tt @SQ} lines defines the alignment sorting order.\\\cline{2-3}
-  & {\tt SN}* & Reference sequence name. Each {\tt @SQ} line must have a unique {\tt SN} tag. The value of this
-  field is used in the
+  & {\tt SN}* & Reference sequence name.
+The {\tt SN} tags and all individual {\tt AN} names in all {\tt @SQ} lines
+must be distinct.
+  The value of this field is used in the
   alignment records in {\sf RNAME} and {\sf RNEXT} fields. Regular expression: {\tt [!-)+-\char60\char62-\char126][!-\char126]*}\\\cline{2-3}
   & {\tt LN}* & Reference sequence length. \emph{Range}: {\tt [1,2$^{31}$-1]}\\\cline{2-3}
   & {\tt AH} & Indicates that this sequence is an alternate locus.%
 \footnote{See \url{https://www.ncbi.nlm.nih.gov/grc/help/definitions} for descriptions of \emph{alternate locus} and \emph{primary assembly}.}
   The value is the locus in the primary assembly for which this sequence is an alternative, in the format `\emph{chr}{\tt :}\emph{start}{\tt -}\emph{end}', `\emph{chr}' (if known), or `{\tt *}' (if unknown), where `\emph{chr}' is a sequence in the primary assembly.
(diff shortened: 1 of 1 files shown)


daviesrob:
For information, I did a count of the number of punctuation characters in the entry names of our set of 292 reference fasta files.  I got:

```
 # 203
 % 203
 * 525
 + 1
 , 496
 - 154226
 . 1826561
 : 1577
 = 26
 _ 4961932
 | 1098333
```

The 496 commas all come from `GRCh38.rRNA.fa` which contains entries like this:
```
>ENSG00000276700|RNA, 5.8S ribosomal 5 [Source:HGNC Symbol;Acc:HGNC:37660]|RNA5-8S5|rRNA
>gi|189571632|ref|NR_023379.1| Homo sapiens RNA, 5S ribosomal 17 (RNA5S17), ribosomal RNA
>ENSG00000199334|RNA, 5S ribosomal 11 [Source:HGNC Symbol;Acc:HGNC:34372]|RNA5S11|rRNA
>ENSG00000201966|RNA, 5.8S ribosomal pseudogene 4 [Source:HGNC Symbol;Acc:HGNC:41958]|RNA5-8SP4|rRNA
```

Colons are in `Homo_sapiens.GRCh38_full_analysis_set_plus_decoy_hla.fa` (HLA regions) ...:
```
>HLA-A*01:01:01:01	HLA00001 3503 bp
>HLA-A*01:01:01:02N	HLA02169 3291 bp
>HLA-A*01:01:38L	HLA03587 3374 bp
>HLA-A*01:02	HLA00002 3374 bp
```

... and a few other files:
```
Plasmodium_berghei.rRNA.fa:
>berg10:rRNA:rfamscan:938985-939103

Plasmodium_falciparum.rRNA.fa:
>MAL14_5S_2|MAL14_5S_2:rRNA

Trichuris_muris_v0_1_281111.fa:
>NODE_48597_length_3376_cov_5.922986.1.1633:00009.7180000457210-00010.7180000828236

Trichuris_muris_11_03_2015.fa:
>U_TMUE_mito:1

haemonchus_V1.fa:
>scaffold2.1_size831208:1-30808
```

I guess GRCh38 is the biggest problem here.

jkbonfield:
Inspired by TAG:TYPE:VALUE is OK, but note we already have TAG:TYPE:VALUE,VALUE,VALUE for some tags and TAG:TYPE:VALUE;VALUE;VALUE for others when we're dealing with lists of values.  Frankly anything you do here will fit in given lack of consistency elsewhere!

Disallowing colon seems very logical and solves a lot of problems we've had, but I'd be inclined to still use comma as the separator.   I don't have any strong feelings though so would accept it as it is or with comma instead.

jmarshall:
Recreated using commas as a more natural list delimiter.  I hear there is either VCF or at least bcftools option notation using commas as refseq name delimiters, and, closer to home, SAM's **SA** tag already uses a comma to terminate a reference sequence name.

jmarshall:
More previous discussion of forbidding characters in #124 and samtools/samtools#149.

Issue #204, opened 2017-05-04
Representing custom sort and group orderings in SAM/BAM


tfenne:
I would really like a way to represent custom orderings in SAM/BAM/etc.  There are a number of tools, across multiple organizations and code bases that emit BAM files that have been sorted into specific orderings that are not representable in a standard way in the current header.

The only solution(s) currently are to add additional tags to the header to represent things, e.g.:
`@HD  SO=unsorted so=duplicate-ordered`.  I think this is a general enough issue that it would be worthwhile standardizing how to represent this information.  Though the orderings themselves are often esoteric enough that standardizing those would likely be a bad idea.  For example, I have one custom ordering I use that sorts all records by "the 5' unclipped position of the primary record for the read within the pair that is earlier on the genome"!

I can think of a few ways of specifying this, that range from total backwards compatible, to ones that are not backwards compatible (though would be easy to adjust to).  I'd like to gather feedback here, and if there is support, I will create a PR.

I'll also note, that the solutions I'm proposing could also be used to provide a nice solution to the samtools vs. Picard queryname sorting issue that has existed forever.

### Option 1: Totally Backwards Compatible
Add a new standard tag to the header that specifies "sub-sort-order", where the tag is standard but the values are unconstrained.  This would allow things like (using `SS` as the new tag for the examples):

```
@HD  SO:queryname  SS:lexical
@HD  SO:queryname  SS:numeric-lexical
@HD  SO:unsorted   SS:duplicate-order
```

Existing tools could largely ignore the `SS` tag in the header, and custom orderings could be defined with SO:unsorted and custom values for `SS`.  This could give rise to odd situations though when an existing tool resorts a BAM and doesn't know to remove/update the `SS` tag, giving rise to incorrectly labeling.

I'm also not super-keen on breaking the information across two fields personally.

### Option 2: Semi-backwards compatible
Allow "sub-typing" of sort orders.  Pick a delimiter character and then a sort order may be composed of a "major" and _optionally_ "minor" ordering, e.g.:

```
@HD  SO:queryname:lexical
@HD  SO:queryname:numeric-lexical
@HD  SO:unsorted:duplicate-order
```

Under this model `SO:queryname` would still be totally valid, so all existing files would be forward-compatible.  But tools would need to be updated to parse the slightly more complicated SO format.  The upside is that tools that resort BAMs, even before being updated, would not generate incorrect information.

### Option 3: Only fowards-compatible

[...]

PR #221, opened 2017-06-01
Added row on SS (subsort) to the SAM header line table.

tfenne:


jkbonfield:
Sorry for the slow review.  I like it, although I'm a bit skeptical about an "implementation-dependent" string.  The ability to have one is useful, but perhaps we could also define a few standard ones in addition to permitting implementation defined ones?

For example, instead of ```SS:coordinate:coordinate-then-queryname``` we could have standard of ```SS:```*primary*```:```*secondary* and permit any sam column to be listed as per the spec too. Examples:

```
SS:coordinate:queryname
SS:coordinate:flag
SS:queryname:coordinate
SS:queryname:mapq
```

We have queryname vs qname and coordinate vs pos regarding field defintiions, but it's probably good to at least start with a known bunch of potential sub-sorts in addition to permitting implementation defined ones that may have all sorts of extra complexities in there.

daviesrob:
Sorting has got more complicated in samtools since samtools/samtools#675 was merged.  It allows ordering by a given record auxiliary tag, then by name or position.  The actual comparison rules are quite complicated as it needs to deal with edge cases like the tag type differing between records.

Currently when using this option SO is set to 'unknown'.  Being able to describe it in an SS tag would be useful.  The tricky part is how to exactly define the comparison rules in use.

tfenne:
@jkbonfield I'm really wary of defining a whole bunch of extra orderings in the spec.  Part of the problem is that there's now way to represent this data today, so it's difficult to have any good idea of what auxiliary sort orders are in use.  The idea of defining them based on SAM fields doesn't require but does _imply_ that's how it should be used.

Take a look here for how I'm using it in practice: https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala

There are a bunch of extended sorts including `:random`, `:random-query` (randomized but query grouped), and `: template-coordinate` which sorts by the lower unpadded 5' coordinate of the two reads in a read pair.  None of those match the pattern you describe.

@daviesrob's usage is totally different and not something I would have anticipated either!

I'm happy to change the wording, but my preference is to leave the sub-sort piece open to unanticipated use cases, and then _if_ those uses become wide spread then to add those standardized sorts to the spec.

jkbonfield:
Agreed with only adding new sort orders as they become widespread, but maybe reserving keywords for likely standard orders may be preferable, or simply having guidance / best-practices on how to define your custom names.  Eg any sort order based on an auxiliary tag ought to be of the form "aux-XX" where "XX" is the tag key.

I did also think, after posting, that our nomenclature is extendable.  So if we wish to sort by primary X, secondary Y, tertiary Z then X:Y:Z is a good descriptor.  None of this necessarily needs to be hard fixed rules, but conventions for the sake of clarity.

jkbonfield:
This has just been raised again by a user wanting to know the sort order field for the aux BX tag: https://github.com/samtools/samtools/issues/811

We need to table this for discussion again.  Currently "Unknown" is the correct, but unhelpful, sort order.

jkbonfield:
Addendum: actually SS "unsorted:aux-BX" would still work for the above example as the primary sort isn't one of the recognised keywords and by stating it is "unsorted" we declare there is no primary sort and the secondary sort order becomes the primary.  It's a bit of a messy hack, but is one way of describing new unofficial sort orders.

SAMv1.tex
```diff
@@ -192,6 +192,12 @@ \subsection{The header section}
     \emph{Valid values}: {\tt none} (default), {\tt query} (alignments are
     grouped by {\sf QNAME}), and {\tt reference} (alignments are grouped by
     {\sf RNAME}/{\sf POS}).\\\cline{1-3}
+  & {\tt SS} & Sub-Sorting order of alignments. Valid values are of the form
+    {\tt <sort-order>:<sub-sort>}, where {\tt <sort-order>} is the same value stored in the
+    {\tt SO} attribute and {\tt <sub-sort>} is an implementation-dependent string further
```
yfarjoun: can you add a regex limiting the scope of the subsort string?

jkbonfield: Or possibly a combination of regex and textual description, eg colon separated list of sort orders, each of which must match the regex `[A-Za-z0-9_]+`? (ie alphanumeric plus underscore).

yfarjoun:
@tfenne please respond

tfenne:
@yfarjoun Sorry, trying to wrap my head back around where we left this.  So, I think then all that's really left is to add some kind of restriction on the sub-sort values?  I'm happy to amend the PR to do that.  Anything else that's necessary before merging?

sjackman:
I'd suggest changing the example like so
```diff
-@HD SO:coordinate SS:coordinate:coordinate-then-queryname
+@HD SO:coordinate SS:coordinate:queryname
```
and add one or both of these examples
```
`samtools sort -t BX` sorts by BX tag and then by coordinate.
@HD SO:none SS:none:BX:coordinate
`samtools sort -t BX -n` sorts by BX tag and then by query name.
@HD SO:none SS:none:BX:queryname
```

jkbonfield:
Sort order "none" isn't valid.  In my example above I used "unsorted", which isn't really correct, but is better than the alternative of "unknown".   We could perhaps add "none", but I'm wary of breaking too many tools.

We should be explicit about what to do when applying a sort order that isn't one of the accepted primary orders.  Eg.  If the primary sort order is not a recognised sort for the SO tag (ie queryname or coordinate) then the SO tag must be "unsorted" and the SS tag used to describe the sort order (with example, eg SS:unsorted:BX).

sjackman:
Thanks, Tim!

SAMv1.tex
```diff
@@ -224,6 +224,16 @@ \subsection{The header section}
...
+  & {\tt SS} & Sub-Sorting order of alignments. Valid values are of the form
+    {\tt <sort-order>:<sub-sort>}, where {\tt <sort-order>} is the same value stored in the
+    {\tt SO} attribute and {\tt <sub-sort>} is an implementation-dependent string further 
+    describing the sort order.  E.g. if an algorithm relies on a {\tt coordinate} sort that,
+    at each coordinate, is further sorted by queryname the header could contain 
+    {\tt @HD SO:coordinate SS:coordinate:queryname}.  If the primary sort is not one of the predefined
+    primary sort orders, then {\tt unsorted} should be used, and the subsort is effectively the major
+    sort.  E.g. if sorted by an auxillary tag {\tt MI} then by coordinate then the header could contain
+    {\tt @HD SO:unsorted SS:unsorted:MI:coordinate}.
+    Values must match regular expression {\tt [A-Za-z0-9\char92\char45\char95\char58]+}. \\\cline{1-3}
```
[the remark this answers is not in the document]
tfenne: @jkbonfield I am not good at latex!  When I generate the PDF I see the regex coming out as `[A-Za-z0-9\-_:]+`, which is what I think I meant.  I.e. alphanumeric plus `-`, `_` and `:`.  I'm happy to do whatever in terms of expressing it (i.e. English or Regex).  I do like the idea of making it clear that the colons are delimiters.  Though I would like to retain the use of `-` and `_` within the name parts.  Once there's consensus I'll update - or I'm also happy if someone else wants to take a shot.

jkbonfield:
I hope you don't mind @tfenne, but I added a new commit to change the regexp and also adjusted the white space a bit (removing trailing white space and a newline per sentence only, even though this isn't adherered to everywhere).

The regexp changed from permitting alpha-numeric, colon, \, ], ^ and _ to alpha-numeric, colon, _ and -.  This restricts the use of whackier things like close square brace and caret, but adds a hyphen.  I figured we don't care too much if someone names something tag-XX or tag_XX.

Any comments?

SAMv1.tex
```diff
@@ -224,6 +224,11 @@ \subsection{The header section}
     \emph{Valid values}: {\tt none} (default), {\tt query} (alignments are
     grouped by {\sf QNAME}), and {\tt reference} (alignments are grouped by
     {\sf RNAME}/{\sf POS}).\\\cline{1-3}
+  & {\tt SS} & Sub-Sorting order of alignments. Valid values are of the form {\tt <sort-order>:<sub-sort>}, where {\tt <sort-order>} is the same value stored in the {\tt SO} attribute and {\tt <sub-sort>} is an implementation-dependent string further describing the sort order.
+    E.g. if an algorithm relies on a {\tt coordinate} sort that, at each coordinate, is further sorted by queryname the header could contain {\tt @HD SO:coordinate SS:coordinate:queryname}.
+    If the primary sort is not one of the predefined primary sort orders, then {\tt unsorted} should be used, and the subsort is effectively the major sort.
+    E.g. if sorted by an auxillary tag {\tt MI} then by coordinate then the header could contain {\tt @HD SO:unsorted SS:unsorted:MI:coordinate}.
+    Values must match be colon separated alphanumerics with underscores and hyphens permitted, i.e. matching the regular expression {\tt [A-Za-z0-9:\_-]+}. \\\cline{1-3}
```
nh13: ` Values must match be colon` - I think there's a word missing here

nh13: Also, are colons allowed to be in the name of the sub-sort?  Or is a colon the delimiter for multiple sub-sorts?  Or is the example `SS:unsorted:MI:coordinate` interpreted as the `<sort-order>` is `unsorted`, and the `<sub-sort>` is `MI:coordinate`, where the latter can be interpreted independently?

jkbonfield: I fixed the "must match be colon" mis-edit.

Regarding colons in the name, no they are not allowed and I think this should be clear enough already by "colon separated".  By definition, any colon in a name is considered to be a separator so they cannot exist.

yfarjoun:
I'm confused about the need to have the SO value appear as the first element of the SS string. This is redundant and a source of confusion.  What does `SO:coordinate SS:queryname` mean according to the current version?

If SS only specifies the secondary sorting, then there's no need to put the SS value there, and if it doesn't then we need to explain what is the meaning of e.g. `SO:coordinate SS:queryname:coordinate`.

I think that `SO:coordinate SS:queryname` can mean "first sort by coordinate, then break ties by queryname".

jmarshall:
As discussed in last week's meeting, I've added a note with my recollection of @tfenne's reason for repeating the SO value. Also an entry in the History appendix. I think this is ready to be squashed and merged now…

jmarshall:
Footnote previously added with the reason for repeating the SO value, which in fact was @jkbonfield's idea in https://github.com/samtools/hts-specs/issues/204#issuecomment-299408421 and was discussed in that issue (#204). @jkbonfield @yfarjoun I think that addressed all outstanding comments and this is ready to be squashed and merged?

jmarshall:
@jkbonfield and @yfarjoun approved verbally during meeting.

SAMv1.tex
```diff
@@ -1227,6 +1229,7 @@ \section{SAM Version History}\label{sec:history}
 \subsection*{1.5: 23 May 2013 to current}
 
 \begin{itemize}
+\item Add {\tt @HD SS} sorting details header tag. (Sep 2018)
```
jmarshall: Will update this to “Oct”, of course.

PR #222, opened 2017-06-22
VCF v4.3 changelog

cyenyxe:
Summary of changes made to the VCF v4.3 specification document since it was first published.


cyenyxe:
Are you okay with merging this @droazen ?

PR #227, opened 2017-07-14
Support CIGARs with >65535 operations in BAM files

lh3:
This commit addresses samtools/hts-specs#40. It added optional tag `CG` and
explained the workaround to store alignments with >65535 CIGAR operations in
BAM files. The proposal is implemented in samtools/htslib#560.

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -60,6 +60,7 @@ or
   {\tt BQ} & Z & Offset to base alignment quality (BAQ) \\
   {\tt BZ} & Z & Phred quality of the unique molecular barcode bases in the {\tt OX} tag \\
   {\tt CC} & Z & Reference name of the next hit \\
+  {\tt CG} & B,I & BAM only: {\sf CIGAR} in BAM's binary encoding if (and only if) it consists of $>$65535 operators \\
   {\tt CM} & i & Edit distance between the color sequence and the color reference (see also {\tt NM}) \\
   {\tt CO} & Z & Free-text comments \\
   {\tt CP} & i & Leftmost coordinate of the next hit \\
@@ -130,6 +131,13 @@ At the $i$-th read base, ${\rm BAQ}_i=Q_i-({\rm BQ}_i-64)$ where $Q_i$ is the $i
 
 \item[CC:Z:\tagvalue{rname}]
 Reference name of the next hit; `{\tt =}' for the same chromosome.
+
+\item[CG:B:I,\tagvalue{encodedCigar}]
+Real CIGAR in its binary form if (and only if) it contains $>$65535 operations. This is
+a BAM file only tag as a workaround of BAM's incapability to store long CIGARs
+in the standard way. SAM and CRAM files created with updated tools aware of the
+workaround are not expected to contain this tag. See also the footnote in
+Section 4.2 of the SAM spec for details.
 
 \item[CP:i:\tagvalue{pos}]
 Leftmost coordinate of the next hit.
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -827,7 +827,7 @@ underlined word in uppercase denotes a field in the SAM format.
   & \multicolumn{2}{l|}{\sf refID} & Reference sequence ID, $-1\leq{\sf refID}<{\sf n\_ref}$; -1 for a read without a mapping position. & {\tt int32\_t} & [-1] \\\cline{2-6}
   & \multicolumn{2}{l|}{\sf pos} & 0-based leftmost coordinate ($=\underline{\sf POS}-1$)& {\tt int32\_t} & [-1]\\\cline{2-6}
   & \multicolumn{2}{l|}{\sf bin\_mq\_nl} & {\tt{\sf bin}\char60\char60 16\char124\underline{\sf MAPQ}\char60\char60 8\char124{\sf l\_read\_name}}; {\sf bin} is computed from the mapping position;\footnotemark\ {\sf l\_read\_name} is the length of {\sf read\_name} below ($={\sf length}(\underline{\sf QNAME})+1$). & {\tt uint32\_t} & \\\cline{2-6}
-  & \multicolumn{2}{l|}{\sf flag\_nc} & {\tt \underline{\sf FLAG}\char60\char60 16\char124{\sf n\_cigar\_op}};\footnotemark\ {\sf n\_cigar\_op} is the number of operations in \underline{\sf CIGAR}. & {\tt uint32\_t} & \\\cline{2-6}
+  & \multicolumn{2}{l|}{\sf flag\_nc} & {\tt \underline{\sf FLAG}\char60\char60 16\char124{\sf n\_cigar\_op}};\footnotemark\ {\sf n\_cigar\_op} is the number of operations in \underline{\sf CIGAR}\footnotemark. & {\tt uint32\_t} & \\\cline{2-6}
   & \multicolumn{2}{l|}{\sf l\_seq} & Length of \underline{\sf SEQ} & {\tt int32\_t} & \\\cline{2-6}
   & \multicolumn{2}{l|}{\sf next\_refID} & Ref-ID of the next segment ($-1\le{\sf mate\_refID}<{\sf n\_ref}$) & {\tt int32\_t} & [-1] \\\cline{2-6}
   & \multicolumn{2}{l|}{\sf next\_pos} & 0-based leftmost pos of the next segment ($=\underline{\sf PNEXT}-1$) & {\tt int32\_t} & [-1] \\\cline{2-6}
@@ -843,7 +843,7 @@ underlined word in uppercase denotes a field in the SAM format.
   \cline{1-6}
 \end{tabular}}
 \end{table}
-\addtocounter{footnote}{-4}
+\addtocounter{footnote}{-5}
 \footnotetext{{\sf BIN} is calculated using the {\sf reg2bin()} function
 in Section~\ref{sec:code}. For mapped reads this uses {\sf POS-1}
 (i.e.,~0-based left position) and the alignment end point using the
@@ -856,6 +856,19 @@ is $4680$.}
 \stepcounter{footnote}
 \footnotetext{As noted in Section~\ref{sec:alnrecord}, reserved {\sf FLAG} bits
 should be written as zero and ignored on reading by current software.}
+\stepcounter{footnote}
+\footnotetext{With 16 bits, {\sf n\_cigar\_op} can keep at most 65535 CIGAR
+operations in BAM files. For an alignment with more CIGAR operations, BAM
+stores the real {\sf CIGAR}, encoded the same way as the {\sf cigar} field in
+BAM, in the {\tt CG} optional tag of type `{\tt B,I}', and sets {\sf CIGAR} to
+`{\tt kSmN}' as a placeholder, where `{\tt k}' equals {\sf l\_seq}, `{\tt m}'
+is the reference sequence length in the alignment, and `{\tt S}' and `{\tt N}'
+are the soft-clipping and reference-clip CIGAR operators, respectively -- i.e.
+in the binary form, {\sf n\_cigar\_op}=2 and {\sf cigar}={\tt [k\char60\char60
+4\char124{4},m\char60\char60 4\char124{3}]}. If tag {\tt CG} is present and
+the first CIGAR operation clips the entire read, a BAM parsing library is
+expected to update {\sf n\_cigar\_op} and {\sf cigar} with the real CIGAR
+stored in the {\tt CG} tag and remove the now-redundant {\tt CG} tag.}
 \stepcounter{footnote}
 \footnotetext{For backward compatibility, a {\sf QNAME} `{\tt *}' is stored as a C string {\tt "*\char92 0"}.}
 \stepcounter{footnote}


SAMv1.tex
```diff
@@ -838,6 +838,18 @@ \subsection{The BAM format}
...
+CIGAR} operations in BAM files. For an alignment with more {\sf CIGAR}
+operations, BAM stores the real {\sf CIGAR}, in its binary form, to the {\tt
+CG} optional tag of type `{\tt B,I}', and sets the {\sf CIGAR} to `{\tt kS}' as
+a placeholder, where `{\tt k}' is the length of {\sf SEQ} and `{\tt S}' the
+soft-clipping {\sf CIGAR} operator (i.e. in the binary form, {\sf
+n\_cigar\_op}=1 and {\sf cigar}={\tt [k\char60\char60 4\char124{4}]}).
+This workaround is applied to BAM files \emph{only}. SAM and CRAM files are not
+affected. If tag {\tt CG} is present, BAM parsing libraries are expected to
+seamlessly update {\sf n\_cigar\_op} and {\sf cigar} with the real {\sf CIGAR}
+stored in the {\tt CG} tag.}
```
jkbonfield: Plus "and remove the now-redundant CG tag"?

SAMtags.tex
```diff
@@ -58,6 +58,7 @@ \section{Standard tags}
   {\tt BC} & Z & Barcode sequence \\
   {\tt BQ} & Z & Offset to base alignment quality (BAQ) \\
   {\tt CC} & Z & Reference name of the next hit \\
+  {\tt CG} & B,I & BAM-only tag to store the real {\sf CIGAR} if it contains $>$65535 operations\\
```
yfarjoun: I think that it should be allowed to leave the CG tag in a sam or a cram. the "recommended practice" can talk about the fact that only bam needs it, but there's no reason to complicate thing in the spec by making it illegal in anything but bam.

lh3: Sure, point taken. How about "Intended to store real CIGAR if it contains >65535 operations"? I would like to choose wording such that this tag won't be abused to frequently store normal CIGARs.

yfarjoun: 👍

jkbonfield: We intend not to use the real CIGAR *in bam only* if we have more than 64k, therefore the phrase "intended to store the real CIGAR..." is misleading.  We have no explicit intentions of deliberately using this for all formats.

However when all said and done we need to decide what we wish to do when we see a CG tag in SAM or in CRAM.  Do we ignore it and honour the real CIGAR, do we complain (it's an indication of using an out-dated piece of software that didn't correctly translate back), or do we try to patch the situation up and migrate CG to the cigar field once more?

My gut feeling is ignoring it is easiest and cleanest, but I could see a reason why issuing a warning may be useful.

lh3:
I have updated this PR to encode long cigar with fake cigar `<readLength>S<refLength>N`.

Do we also want to add a `CG:Y/N` tag at the `@HD` line with this PR? I am thinking to merge this and samtools/htslib#560 first and then to address the `CG` header tag with another pair of pull requests to both hts-specs and htslib.

yfarjoun:
I don't understand the CG:Y/N header....how can we tell in advance if there will be records with CG tags?

lh3:
You can't. I guess the intended use is that a user needs to ask the mapper to output a `@HD CG:Y` if he/she expects long-cigar alignment. This tag has no effect if there are no long cigars. However, if there is a long cigar and there is no CG:Y, samtools will throw an error.

One possible scenario, though, is that everyone might be outputting CG:Y regardless of long cigars.

yfarjoun:
that will be useless then...

jkbonfield:
Tools that don't know how to deal with long cigars won't be parsing and checking the CG:Y header.  Tools that do know how to deal with long cigars don't need a header to inform them of their presence.  I forget now why it came up, but apparently in response to something I said.  Hmm.

I'm tempted however to suggest that we ought to bump the SAM version number, even though this is technically a BAM only change.  Why? Because there is no BAM version number other than the sledgehammer complete format change and there is the risk that this BAM only change will leak into SAM anyway, hence having a fixed version where we know it can occur (or vice versa, where we know it cannot) acts in a similar way to CG header tag, but more useful IMO.  Plus, like it or not, the BAM and SAM formats have been inextricably linked in the same document since day 1.

sjackman:
I believe `CG:Y` or `CG:long` was initially my suggestion, though I'm having trouble finding the original post now. The idea was that tools should output either entirely short CG or long CG, but not a mixture of both, to prevent an old tool from silently ignoring just those reads, and producing incorrect results. The following suggestion allows the entire pipeline to use either short or long CG as necessary, without changing the API. The default behaviour of the tools should just work in most situations.

- A SAM writer, like a long-read aligner, should output either `CG:Y` or `CG:N`. The default behaviour is up to the tool author. The behaviour may be controlled by a command line option.
- A BAM writer should output all short CG or long CG, but not a mixture of both.
- A BAM reader and writer (like samtools view or sort) that does not see `CG:Y` in its input header should output exclusively short CG, and an error message if it encounters a long CIGAR.
- A BAM reader and writer that sees `CG:Y` in its input header should output exclusively long CG.

lh3:
I would defer the decision on CG:Y to another PR. It is not essential, but affects more users and is likely to require lengthy discussions on the precise behavior and implementation. We should merge this PR, the core component of the proposal first.

lh3:
I understand samtools/htslib#560 will take time to test. Any concerns with this PR?

SAMtags.tex
```diff
@@ -60,7 +60,8 @@ \section{Standard tags}
   {\tt BQ} & Z & Offset to base alignment quality (BAQ) \\
   {\tt BZ} & Z & Phred quality of the unique molecular barcode bases in the {\tt OX} tag \\
   {\tt CC} & Z & Reference name of the next hit \\
-  {\tt CM} & i & Edit distance between the color sequence and the color reference (see also {\tt NM}) \\
+  {\tt CG} & B,I & Intended to store the real {\sf CIGAR} if it contains $>$65535 operations\\
```
yfarjoun: all the other lines have a space before the `\\`...please keep with that style.

yfarjoun: Please mention here that it's for BAM format only

SAMv1.tex
```diff
@@ -857,6 +857,19 @@ \subsection{The BAM format}
 \footnotetext{As noted in Section~\ref{sec:alnrecord}, reserved {\sf FLAG} bits
 should be written as zero and ignored on reading by current software.}
 \stepcounter{footnote}
+\footnotetext{With 16 bits, {\sf n\_cigar\_op} can keep at most 65535 CIGAR
+operations in BAM files. For an alignment with more CIGAR operations, BAM
+stores the real {\sf CIGAR}, in its binary form, to the {\tt CG} optional tag
```
yfarjoun: to->in

SAMtags.tex
```diff
@@ -131,6 +132,13 @@ \subsection{Additional Template and Mapping data}
 \item[CC:Z:\tagvalue{rname}]
 Reference name of the next hit; `{\tt =}' for the same chromosome.
 
+\item[CG:B:I,\tagvalue{encodedCigar}]
+Real CIGAR in its binary form if it contains $>$65535 operations. This is
```
yfarjoun: If this is in binary form, we need to be more specific about which binary format and what parts of the binary format....would it be easier to just store it as a string? What would happen is an "old tool" took the bam and converted it to a sam? how would the tag be encoded?

SAMtags.tex
jkbonfield: I don't like "Intended" here as the specification isn't about what we'd like, but about what must happen.

How about "BAM only: stores the real {\sf CIGAR} field if it contains $>65535$ operations".

SAMtags.tex
```diff
@@ -131,6 +132,13 @@ \subsection{Additional Template and Mapping data}
...
+intended to be a BAM file only tag as a workaround of BAM's incapability to
```
jkbonfield: Please remove "intended to be".  Otherwise I'm happy with the wording.

jmarshall:
This PR is a useful description of the CG tag solution to this 64K-cigar-operations issue, a proof of concept of the spec changes necessary if you will.

There are various details that need to be in the spec: I was pleased to see that the text describes how an implementation should recognise that a CG tag is being used, but for example the spec needs to describe exactly how the CG tag array is laid out (e.g. what order are the operator and length in in the array?) and doesn't at all at the moment, as @yfarjoun has also noticed.

I do not think this PR should be merged into the spec as is. I am in the process of expanding and reorganising the added text in IMHO a better way and covering missing details.

SAMtags.tex
lh3: I have changed the wording a little bit, emphasizing that `CG` is encoded exactly the same way as the cigar field in BAM. This allows you to move the entire cigar around with memory copy/move, much more efficient and easier to implement than keeping text CIGAR in the CG tag.

jkbonfield:
@jmarshall If you're in the process of revising things like footnotes, please also look at the first commit of Rob's BAMv2 proposal.  This was a revamp of the BAM table to avoid the implementation specific bit-packing methods and to replace overly heavy footnotes with their own sections.

https://github.com/samtools/hts-specs/pull/259/commits/9ab9e3a40f36754a3b59991d9d7cf10332732f03

I was thinking of reviewing and likely merging this after we get Heng's CG tag done.

lh3:
> However please also remove the commited SAMv1.pdf to ease merging this pull request.

This was a mistake. How can I remove it from the pull request? Checkout an older version of the PDF and then commit again?

sjackman:
```sh
git checkout HEAD -- SAMv1.pdf
git commit --amend SAMv1.pdf
```

colinhercus:
Hi,

It  looks like you're deleting the CM tag! Why is this in 65535 CIGAR
thread?

We've always used CM as number of colour mismatches and NM as the number of
bases mismatches. A SNP will be one NM and 2 CMs and a single CM may not be
a even add to NM. There are still  people using Colour Space.

Best Regards, Colin

lh3:
I somehow deleted a space at the end of CM line, so you saw the line changed, but it is still there. I didn't delete it. I have just added the space back.

colinhercus:
Hi Heng,

Okay, no problem. I wondered if it was mistake after I posted. It did
clearly show as - CM line and + CG line.

Best Regards, Colin

jkbonfield:
Can we have a progress report on this.  @jmarshall are you revising the text? If so can we get a draft?

As far as I'm concerned, the text is fine but the placement and ever-increasing series of footnotes is not.  However that change is more wide ranging that just this PR so I propose to merge this and then look at the 1st commit of Rob's BAM2 PR which tidies up the excessive footnotes (with an amendment for the new one added here too).

Thoughts @yfarjoun?

lh3:
Can we merge this PR? I think all the concerns have been addressed so far.

jmarshall:
We will not be merging this PR before this week's meeting.

tk2:
@jmarshall can you clarify, are you preparing an amended pull request in time for the meeting?

jmarshall:
Yes. In any case, the topic is already on the agenda for the meeting — so no PR on this topic would be being merged before then.

SAMtags.tex
yfarjoun: if-> if (and only if)

lh3:
Changed "if" to "only if" as is requested by @yfarjoun. Rebased and squashed to a single commit.

sjackman:
> Real CIGAR in its binary form if it contains $>$65535 operations.

I disagree with `if and only if` and prefer `if`. A BAM file that uses exclusively either long CG tags or short CIGAR strings is preferable in my opinion to one that uses mixed encodings. Old software that reads a mixed encoding would ignore just the long alignments and produce inaccurate results that are not readily apparent. It's better to use CG tag for all alignments, so old software would ignore all alignments, and its failure would be more apparent. I'm not arguing that this be required behaviour. Changing the wording to `if and only if` however precludes using a file composed of only CG tags for all reads, which is in my opinion the better format.

See also https://github.com/samtools/hts-specs/pull/227#issuecomment-342230039

yfarjoun:
This raises an interesting point. 

Also, we would probably like to test the CG tags using shorter than 64K operations. 

My concerns is that folks will shift to using exclusive CG tags in BAM regardless of the length of their CIGARS (to be on the "safe side" in case a large cigar comes by)....

Thoughts?

sjackman:
> My concerns is that folks will shift to using exclusive CG tags in BAM regardless of the length of their CIGARS (to be on the "safe side" in case a large cigar comes by)

I don't think that's likely. People will likely continue to use old CIGAR for compatibility with older toolchains as long as they can. Only those motivated by extremely long reads will likely migrate to CG tags.

lh3:
I think we have reached a consensus during the call? We should merge this first. Create a new PR if you like and I will comment there.

yfarjoun:
I guess it is better to start with the stricter option and possibly relax
it than vice versa.

jkbonfield:
I'm fairly abivalent about it really and can see either way has merits.   I'm tempted to say keep it open and say "if" rather than "if and only if", but I'll go along with the latter if everyone else agrees it's best.

As for testing it with shorter CGs, I did that for the htslib implementation, including long "fake" cigars and very short real ones to stress test it (this found issues, now resolved).  However that is all implementation detail - it can and now does work with whatever we seem to throw at it so the spec is workable with is about the only implementation detail we need for the spec decision.

jkbonfield:
@jmarshall can we get an update on this? Above you said you'd post an updated PR before the meeting, but we're ~1 week after and still no feedback.

We need to get moving on this, but cannot do so if we don't know what the blocking issues are and nor can we review a revised PR.

jkbonfield:
Still waiting on an update.  What's the delay please?  Both myself and @yfarjoun are in agreement, as is the original PR author @lh3.  @jmarshall - you indicated you had concerns and an updated PR, but have so far been unable to tell us what these are nor offer any updated text.

I propose we merge this PR as it currently stands  and your updates can arrive as a subsequent PR if and when ready.

jkbonfield:
Merged.  Next up we need to look at https://github.com/samtools/hts-specs/pull/259/commits/9ab9e3a40f36754a3b59991d9d7cf10332732f03 (from https://github.com/samtools/hts-specs/pull/259) to see which bits of the tidy up we should merge in, but this is a separate issue.

Issue #207, opened 2017-05-17
htsget: add specificaton version to api/response


AlexanderSenf:
Now that there are more than one htsget (Streaming API) protocol specifications it will be possible that there will be live servers adhering to different spec versions. It would be helpful if the spec version would be available from the API itself; either as a request parameter, or as an extra field in the JSON response; or by specific request.

jmarshall:
[@mlin: There's already a `[transfer-protocol]` label for this.  Perhaps you should merge the labels.]

mlin:
@jmarshall thanks for pointing that out!

jmarshall:
I think the only thing that's changed between the two extant versions that could cause confusion for clients is the syntax of the error response JSON (`{"error": {"type": "foo", "message": "bar"}}` vs. `{"error": "foo", "message": "bar"}`).  So as yet there's probably not too much of a problem.

At the moment we only have one request in total (`…/reads/<id>`), so adding a specific request for versioning seems somewhat overkill.

[More conversation exists on the mailing list.]

daviesrob:
I'd still like to see something in the response to identify the protocol and version.  Currently htslib looks for the string `{"` in order to identify an htsget reply (as opposed to cram or html or jpeg or anything else the server may throw back at it), which hardly seems robust.  In particular, it would be nice to be able to reliably tell htsget JSON responses from any other form of JSON that arrives.

jmarshall:
Indeed, there's at least a couple of different facets here:

1. a way for clients to ask for a particular version of e.g. `/reads/` response that they would like to get back
2. a way to identify that some JSON is indeed of a particular response version
3. an identification of the server's supported version (whatever that means beyond the first two points; or is it just the `Server` response header?)

@daviesrob's is (2), which could just be something like `{…"htsget": "0.2"…}` in the response.

:+1: sure, why not, although it's not much good as a magic number if we keep having `{…}` as the top level of our JSON responses — as you surely can't force a particular ordering on people's JSON libraries in general.  So we might need to go to an artificial `[ "htsget-0.2", {…} ]` style of response JSON.

jmarshall:
Re (2), we could of course trigger on a response `Content-Type` like `application/vnd.ga4gh.htsget.v1+json` or so.  However it's non-trivial to get at the content-type in some clients (e.g. htslib, which unifies http streams and raw files etc, so can't easily access HTTP headers), so there is something to be gained from having the JSON be self-identifying too.

[...]

PR #226, opened 2017-07-05
Proposed Accept/Content-Type header-based versioning

jmarshall:
Partial proposal for versioning — other aspects are versioning/magic numbers within the JSON text, to be addressed separately.  See discussion in #207.

PR #230, opened 2017-08-01
Proposed htsget protocol identifier in JSON text

daviesrob:
Implements the proposal discussed in #207 to wrap the response data like this:
```
{ "htsget" : { ... } }
```

No version number is included here, but it would be trivial to add, either now or later.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -27,7 +27,7 @@ Explicitly this API does NOT:
 
 All API invocations are made to a configurable HTTP(S) endpoint, receive URL-encoded query string parameters, and return JSON output. Successful requests result with HTTP status code 200 and have UTF8-encoded JSON in the response body, with the content-type `application/json`. The server may provide responses with chunked transfer encoding. The client and server may mutually negotiate HTTP/2 upgrade using the standard mechanism.
 
-The JSON response SHOULD be an object with the single key `htsget`.  The value is an object as described in the [Response JSON fields](#response-json-fields) and [Error Response JSON fields](#error-response-json-fields) sections.  This ensures that, apart from whitespace differences, the message always starts with the same prefix.  The presence of this prefix can be used as part of a client's response validation.
+The JSON response is an object with the single key `htsget` as described in the [Response JSON fields](#response-json-fields) and [Error Response JSON fields](#error-response-json-fields) sections.  This ensures that, apart from whitespace differences, the message always starts with the same prefix.  The presence of this prefix can be used as part of a client's response validation.
 
 Any timestamps that appear in the response from an API method are given as [ISO 8601] date/time format.
 
--- a/htsget.md
+++ b/htsget.md
@@ -27,6 +27,8 @@ Explicitly this API does NOT:
 
 All API invocations are made to a configurable HTTP(S) endpoint, receive URL-encoded query string parameters, and return JSON output. Successful requests result with HTTP status code 200 and have UTF8-encoded JSON in the response body, with the content-type `application/json`. The server may provide responses with chunked transfer encoding. The client and server may mutually negotiate HTTP/2 upgrade using the standard mechanism.
 
+The JSON response SHOULD be an object with the single key `htsget`.  The value is an object as described in the [Response JSON fields](#response-json-fields) and [Error Response JSON fields](#error-response-json-fields) sections.  This ensures that, apart from whitespace differences, the message always starts with the same prefix.  The presence of this prefix can be used as part of a client's response validation.
+
 Any timestamps that appear in the response from an API method are given as [ISO 8601] date/time format.
 
 HTTP responses may be compressed using [RFC 2616] `transfer-coding`, not `content-coding`.
@@ -45,6 +47,12 @@ For errors that are specific to the `htsget` protocol, the response body SHOULD 
 
 <table>
 <tr markdown="block"><td>
+`htsget`
+_object_
+<td>
+Container for response object.
+<table>
+<tr markdown="block"><td>
 `error`  
 _string_
 </td><td>
@@ -55,6 +63,8 @@ The type of error. This SHOULD be chosen from the list below.
 _string_
 </td><td>
 A message specific to the error providing information on how to debug the problem. Clients MAY display this message to the user.
+</td></tr>
+</table>
 </td></tr>
 </table>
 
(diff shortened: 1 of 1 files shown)


jmarshall:
:+1: though I'm wondering why only _SHOULD_ rather than _MUST_ in

> The JSON response SHOULD be an object with the single key `htsget`

mlin:
I vote "-0" as I don't see it helping much in view of #226, but it's easy to do.

Probably this was discussed already, but what about merging e.g. `{"htsget": ".v0.2rc"}` into the existing top-level object instead of moving the existing stuff under a new envelope? Then we would not really need a synchronized client/server protocol change even at this early stage.

jmarshall:
@mlin: Indeed, that was discussed at length in #207.  The aim here is to get the text `htsget` as a magic number guaranteed to be at the start of the JSON stream.  If it was merely a field of the existing object, it could appear anywhere in the stream.

daviesrob:
@jmarshall SHOULD can be easily changed to MUST if that's the preferred wording.

calbach:
Sorry I hadn't reviewed this before the call. My 2 cents:

1. For the use case of "client wants to confirm they are talking to an htspec server", #226 seems to me to be sufficient.
2. Wrapping the entire JSON with a magic string seems heavy-handed, it's adding an extra layer to the application layer which does not have any semantic meaning.
3. It seems to me that the proposed client validation would come into effect under only exceedingly rare conditions. Namely:

    - I sent my htsget to a non-htsget server and it responded with a 200 HTTP status
    - The response it gave me was JSON, but not an htsget JSON response

    Given the frequency with which I would expect this to happen, taking a bit of extra time to parse the JSON or check the response header to validate seems like a negligible cost (especially given that your client needs to parse the JSON anyway if it intends to make use of it).

Perhaps I've missed something and this is solving a very specific problem that someone expects to occur frequently. If so, please provide this context.

jmarshall:
For a generic client, the client-server conversation looks rather different from what you describe. More like this:

* Given a random URL by the user, I make a request to a server. It responds successfully (e.g. HTTP 200).  _(At this point, the client software doesn't know that htsget or JSON might be involved; it might not even know what flavour of data — e.g., BAM vs. BCF — might be expected.)_

* Data comes back. If it was an http(s) URL, it merely SHOULD have a Content-Type with it that tells me it's htsget+JSON. If it was some other kind of URL (or there's a non-specific Content-Type), there will be no content type and I'll have to guess from the data stream. _(Half the #207 discussion was about whether to care about non-http(s) URLs, and the other half was about whether not propagating Content-Type to the top of your software stack is a valid lifestyle choice.)_

* So we're sniffing the first few tens or hundreds of bytes of the data stream. It starts with some `{` and `"` characters. Might be JSON or it might be next week's new textual bioinformatics format.  If it's JSON, it might be htsget or it might be some other perhaps yet to be invented JSON-based protocol.  We could see if those words in the quotes are `format`/`urls`/`url`/etc in some order and cross our fingers, but this proposal is to put something definitive at the start of the data stream.

As @daviesrob noted previously, this proposal is based on the JSON examples at <http://www.json.org/example.html>. Are there any JSON experts in the house who can tell us how typical these examples' layout is in this sort of use case?

Anyone who wants to can examine an implementation that works like this. Look at `hts_detect_format()` in HTSlib's _hts.c_.

calbach:
OK I see now, your intended use case is to probe an arbitrary URL and determine if it contains htsget content.

> it merely SHOULD have a Content-Type with it that tells me it's htsget+JSON

Alternatively, why not just make this a MUST? Or else only support automated type detection on a URL when the content type is returned?

jmarshall:
> Alternatively, why not just make this a MUST?

“SHOULD” there reflects the vagaries of the real world, which contains: misconfigured servers; servers that cannot practically be made more specific than `applicaton/json`; people experimenting with our protocol who do not have access to Apache `AddType` directives.  Compare [RFC 7231 §3.1.1.5](https://tools.ietf.org/html/rfc7231#section-3.1.1.5).

> Or else only support automated type detection on a URL when the content type is returned?

Very funny. “Dear developers, how come your software works with data at example.edu but fails with data at example.com? I checked what they both return and the JSON gobbledegook is identical…”

calbach:
I'm +0. I understand the use case now, but I'm not convinced that it's a hard requirement to support it. I don't know if I would even expect `samtools view` to work on an htsget URL without any other hints from the client (additional flags). For example, and maybe it's been changed in later versions, my version of samtools requires -S to indicate the input is SAM; it wouldn't be surprising if there were another flag to indicate the input is htsget.

Without fully knowing all of the intended usages, I'd classify this kind of automated content detection as nice-to-have and best-effort functionality (it works when the backend returns a content type, for example). If the developer wants better guarantees, they should be explicit about what protocol they intend to use.

jmarshall:
@calbach: `-S` has been unnecessary for 45% of the time that samtools has existed.

htsget.md
```diff
@@ -27,6 +27,8 @@ Explicitly this API does NOT:
 
 All API invocations are made to a configurable HTTP(S) endpoint, receive URL-encoded query string parameters, and return JSON output. Successful requests result with HTTP status code 200 and have UTF8-encoded JSON in the response body, with the content-type `application/json`. The server may provide responses with chunked transfer encoding. The client and server may mutually negotiate HTTP/2 upgrade using the standard mechanism.
 
+The JSON response SHOULD be an object with the single key `htsget`.  The value is an object as described in the [Response JSON fields](#response-json-fields) and [Error Response JSON fields](#error-response-json-fields) sections.  This ensures that, apart from whitespace differences, the message always starts with the same prefix.  The presence of this prefix can be used as part of a client's response validation.
```
jmarshall: The RFC 2119 shouty verbs are for quantifying implementation behaviour, while this section is probably clearest if it discusses the protocol as a Platonic ideal. The second sentence can be read as saying that the value attached to the `htsget` key is what is described in the tables, but the tables describe the whole thing and include the object containing the `htsget` key/value itself too.

So perhaps this would be clearer as “The JSON response is an object with the single key `htsget`, as described in the […]”.

daviesrob: Yes, probably better that way.  Will update...

mlin:
Whilst remaining "-0", aaa9bcf is a good adjustment IMO.

jeromekelleher:
While I agree with the sentiments of @mlin and @calbach, I'm +1 on this change. An 11 byte overhead here doesn't seem like much of a cost, it does make the JSON more self-describing and makes life a little easier for C programmers using libcurl.

tk2:
I think this is ready to merge @mlin

PR #233, opened 2017-08-18
Htsget Support for VCF Format

amilamanoj:
Extends htsget spec to support VCF format by adding a new method: Get Variants by ID.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -134,9 +134,9 @@ The client can request only records overlapping a given genomic range. The respo
 `id`  
 _required_
 </td><td>
-A string specifying which records to return.
+A string identifying which records to return.
 
-The format of the string is left to the discretion of the API provider, including allowing embedded "/" characters. Strings could be any format the API provider chooses, such as the following:
+The format of this identifier is left to the discretion of the API provider, including allowing embedded "/" characters. The following would be valid identifiers:
 
 * ReadGroupSetIds or VariantSetIds as defined by the GA4GH API
 * Studies: PRJEB4019 or /byStudy/PRJEB4019
--- a/htsget.md
+++ b/htsget.md
@@ -252,7 +252,7 @@ Study ids from which variants are to be returned.
 `format`  
 _optional string_
 </td><td>
-Request variant data in this format. Allowed values: VCF.
+Request variant data in this format. Allowed values: BCF, VCF.
 
 The server SHOULD reply with an `UnsupportedFormat` error if the requested format is not supported.
 [^a]
@@ -329,7 +329,7 @@ Container for response object.
 `format`  
 _string_
 </td><td>
-Read data format. Default: BAM. Allowed values: BAM,CRAM.
+Read data format. Default: BAM. Allowed values: BAM, CRAM.
 </td></tr>
 <tr markdown="block"><td>
 `urls`  
--- a/htsget.md
+++ b/htsget.md
@@ -230,9 +230,11 @@ Example: `fields=QNAME,FLAG,POS`.
 
     GET /variants/<id>
 
-The core mechanic for accessing specified variant data. The JSON response is a "ticket" similar to `reads` method.
+The core mechanic for accessing specified variant data. The JSON response is a "ticket" allowing the caller to obtain the desired data in the specified format, which may involve follow-on requests to other endpoints, as detailed below.
 
 The client can request only variants overlapping a given genomic range. The response may however contain a superset of the desired results, including all records overlapping the range, and potentially other records not overlapping the range; the client should filter out such extraneous records if necessary. Successful requests with empty result sets still produce a valid response in the requested format (e.g. including header and EOF marker).
+
+## URL parameters
 
 <table>
 <tr markdown="block"><td>
--- a/htsget.md
+++ b/htsget.md
@@ -252,7 +252,7 @@ Study ids from which variants are to be returned.
 `format`  
 _optional string_
 </td><td>
-Request read data in this format. Allowed values: VCF.
+Request variant data in this format. Allowed values: VCF.
 
 The server SHOULD reply with an `UnsupportedFormat` error if the requested format is not supported.
 [^a]
--- a/htsget.md
+++ b/htsget.md
@@ -247,7 +247,11 @@ Container for response object.
 `format`  
 _string_
 </td><td>
-Read data format. Default: BAM. Allowed values: BAM, CRAM.
+Response data in this format. The allowed values for each type of record are:
+
+* Reads: BAM (default), CRAM.
+* Variants: VCF (default), BCF.
+
 </td></tr>
 <tr markdown="block"><td>
 `urls`  
--- a/htsget.md
+++ b/htsget.md
@@ -239,8 +239,9 @@ The client can request only variants overlapping a given genomic range. The resp
 `id`  
 _required_
 </td><td>
-Study ids from which variants are to be returned.
+A string specifying which variants to return.
 
+The format of the string is left to the discretion of the API provider, including allowing embedded "/" characters. Strings could be VariantSetIds as defined by the GA4GH API, or any other format the API provider chooses (e.g. "/data/platinum/NA12878", "/byStudy/PRJEB4019").
 </td></tr>
 </table>
 
--- a/htsget.md
+++ b/htsget.md
@@ -225,6 +225,97 @@ QUAL	| Base quality scores
 
 Example: `fields=QNAME,FLAG,POS`.
 
+
+# Method: get variants by ID
+
+    GET /variants/<id>
+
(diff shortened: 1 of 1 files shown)


jeromekelleher:
This seems pretty sensible to me. Perhaps we should leave this feature out until after we've hit 1.0 though? The deadline is about about 6 weeks away and it would probably be better to harden up the rest of the spec before introducing a major new feature like this.

htsget.md
```diff
@@ -216,6 +216,97 @@ QUAL	| Base quality scores
...
+
+
+## Query parameters
+
+<table>
+<tr markdown="block"><td>
+`format`  
+_optional string_
+</td><td>
+Request read data in this format. Allowed values: VCF.
```
jeromekelleher: s/read/variant

Also: we should allow BCF as well as VCF.

htsget.md
```diff
@@ -216,6 +216,97 @@ QUAL	| Base quality scores
...
+The core mechanic for accessing specified variant data. The JSON response is a "ticket" similar to `reads` method.
+
+The client can request only variants overlapping a given genomic range. The response may however contain a superset of the desired results, including all records overlapping the range, and potentially other records not overlapping the range; the client should filter out such extraneous records if necessary. Successful requests with empty result sets still produce a valid response in the requested format (e.g. including header and EOF marker).
+
+<table>
+<tr markdown="block"><td>
+`id`  
+_required_
+</td><td>
+Study ids from which variants are to be returned.
```
AlexanderSenf: Should this be just 'Ids' instead of 'Study IDs' (For EGA this will initially have to be File IDs)

AlexanderSenf: Also: Would it be sufficient to use a single ID (in line with BAM/CRAM specs), and wait for a unified way to submit multiple ID via POST?

cyenyxe: I would be fine with changing it to "one or multiple IDs". Using only one wouldn't be sufficient for the EVA use case for instance as we support cross-study queries.

It could be useful to list examples of possible types of identifiers (studies, files, samples) for both reads and variants endpoints, just to make clear what could be supported. I helped to define this addition and we were a bit unsure at first.

tk2: On the last call, we agreed that we would add POST support to support mulitple IDs. Otherwise, the url could/will get truncated unless there is a hard bound on the total number of characters (maybe http already defines this).

jeromekelleher: This seems orthogonal to VCF support --- perhaps we should limit to using GET with a single ID here for now, and tackle multiple IDs/POST for both reads and variants in a separate PR (as @AlexanderSenf  suggests)?

htsget.md
```diff
@@ -216,6 +216,97 @@ QUAL	| Base quality scores
...
+
+
+## Query parameters
+
+<table>
+<tr markdown="block"><td>
+`format`  
+_optional string_
+</td><td>
+Request variant data in this format. Allowed values: VCF.
```
jeromekelleher: Why not BCF also?

jeromekelleher:
Looks good to me. I see no reason not to support this, since the protocol is identical. Perhaps we should think about abstracting a bit rather than copy-pasting? It might be better to get VCF support into a few clients and servers first though and see how it works before attempting this.

mgcam:
I agree with @jeromekelleher . Our implementation already supports VCF via 'format' param. We previously treated the 'reads/' part of URL as a suggestion; we are using 'sample/' since our IDs are sample accession IDs. Serving variants on the 'sample/' URL seems OK.

cyenyxe:
@jeromekelleher This was defined based on an implementation, although it is not public yet. We can try to deploy it in a test environment before the Christmas break.
I would be okay with having a more generic section like "URL structure" with a list of entity types (reads, variants, samples, etc) that must be used.

@mgcam I'm not completely sold about using 'sample' for variants. It can be enough for streaming those reported by a single sample, which I think it's a reasonable use case, by what if a VCF has multiple samples or it is aggregated and has none? In that case a study/file ID would make more sense.

mgcam:
On 22/11/2017 09:33, Cristina Yenyxe Gonzalez Garcia wrote:
> It can be enough for streaming those reported by a single sample, 
> which I think it's a reasonable use case, by what if a VCF has 
> multiple samples or it is aggregated and has none? In that case a 
> study/file ID would make more sense.
I agree. The sample/study/file URL is almost orthogonal to 
reads/variants. The former hints on the origin of the data, the latter 
on the nature of the data served.



-- 

tk2:
@cyenyxe it looks like this pull request will need some changes and is unlikely to be merged in current form. Will you or @amilamanoj be making the edits required or preparing a follow-up pull request?

cyenyxe:
I have made some changes to address the following comments:

* Generic ID support instead of "study ID"
* Limit GET requests to single value
* Support VCF and BCF
* Abstract to make more generic for reads and variants (less copy-pasting)

htsget.md
```diff
@@ -239,8 +239,9 @@ The client can request only variants overlapping a given genomic range. The resp
 `id`  
 _required_
 </td><td>
-Study ids from which variants are to be returned.
+A string specifying which variants to return.
```
AlexanderSenf: Should we use 'identifier'? That would be more general than 'Study ID' but more specific than 'string'?

htsget.md
```diff
@@ -114,14 +114,16 @@ If a request to the URL of an API method includes the `Origin` header, its conte
 The values of `Origin` and `Access-Control-Request-Headers` (if any) of the request will be propagated to `Access-Control-Allow-Origin` and `Access-Control-Allow-Headers` respectively in the preflight response.
 The `Access-Control-Max-Age` of the preflight response is set to the equivalent of 30 days.
 
+# Request
 
-# Method: get reads by ID
+## Methods
 
     GET /reads/<id>
```
AlexanderSenf: Should this actually be part of the specs? The EGA API doesn't use '/reads/' at all in the URL, because at the EGA I have to use file IDs; so I am using '/files/' instead. I think this should be more generic, so that different implementations can choose their own way. I do like the addition of '/variants/' for VCF/BCF files.

cyenyxe: If the URLs are completely arbitrary, how would a client know what endpoint to call to? Yet another one would be necessary to discover which is the URL to search for reads, variants, etc., in a particular server.

htsget.md
cyenyxe: Agree, it just replicated what was already there (this refers to an old commit btw).

htsget.md
jeromekelleher: I'm not sure there's much point in mandating the format of the URL either, since most people are actively ignoring the current 'reads' prefix.  In terms of the client know where to look for reads or variants, we have no idea how a client got the URL in the first place (explicitly out of band), so I think we can assume that the service that provided the URL will know whether it points to reads or variants.

htsget.md
```diff
@@ -162,29 +175,25 @@ _optional 32-bit unsigned integer_
 </td><td>
 The start position of the range on the reference, 0-based, inclusive. 
 
-The server SHOULD respond with an `InvalidInput` error if `start` is specified and a reference is not specified
-(see `referenceName`).
+The server SHOULD respond with an `InvalidInput` error if `start` is specified and a reference is not specified (see `referenceName`).
```
jeromekelleher: What's after changing here? I think we should keep the diff to the minimum.

jeromekelleher:
Looks good to me. Ther'e's a few 'drive-by' edits that should be removed in the interest of keeping the diff on topic, but other that I'd vote to squash and merge.

AlexanderSenf:
+1

htsget.md
cyenyxe: Just trying to make this section follow the same style than the rest of the document, where lines are not split.

jeromekelleher: Ah right, that's a good idea.  It might be better to do such housekeeping stuff separately to semantic changes like this though. I know @jmarshall likes a nice clean diff!

mlin:
:+1:

htsget.md
cyenyxe: I haven't been able to find any other formatting issues in the whole document. Is removing 3 line breaks worth rebasing this PR and creating a new one?

jeromekelleher: Definitely not worth another PR. Could split this into two commits (one housekeeping, one VCF) when squashing? It really makes no difference though. If nobody else complains, whatever you prefer is fine by me.

cyenyxe:
Is anything else needed (such as more +1) to get this merged?

mlin:
Let me make a final call for comments-- I'll merge this in a couple of days if no concerns are raised. I will try to do so with @jeromekelleher's suggestion to organize it into main and housekeeping diffs. Thanks @cyenyxe!

mlin:
I've reorganized the commits as suggested in https://github.com/samtools/hts-specs/pull/301

Closing this PR (the new one links back here for the record)

PR #234, opened 2017-08-21
Improve description of header @SQ M5 tag.

daviesrob:
Tightens up the description of how to calculate the sequence MD5, especially with regard to which characters should be included and which stripped out.  This pretty much follows what HTSlib does, which has the benefit of being easy to describe.

It side-steps the problem of 8-bit encodings by saying 7-bit US-ASCII must be used.  Is this enough?

It says the hexadecimal digest is lower-case.  Is this too strict?  Both HTSlib and htsjdk write lower-case digests.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -222,11 +222,7 @@ or~{\sf RNEXT} fields.
 \emph{Regular expression}: \emph{name}{\tt (,}\emph{name}{\tt )*}
 where \emph{name} is {\tt [0-9A-Za-z][0-9A-Za-z*+.@\_|-]*}\\\cline{2-3}
   & {\tt AS} & Genome assembly identifier. \\\cline{2-3}
-  & {\tt M5} & MD5 checksum of the sequence.
-The sequence must be in the 7-bit US-ASCII character set, with pads represented as `*'s if present.
-Before calculating the digest, all characters outside the inclusive range 33 (`!') to 126 (`\~{}')  are removed, and all lower-case characters are converted to upper-case\footnote{Equivalent to calling \href{http://pubs.opengroup.org/onlinepubs/9699919799/functions/toupper.html}{\sl toupper()} on the characters in the \href{http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1\_chap07.html\#tag\_07\_02}{\sl POSIX locale}}.
-The MD5 digest\footnote{See \href{https://tools.ietf.org/html/rfc1321}{\sl RFC 1321}} is calculated and presented as a 32 character lower-case hexadecimal number.
-\\\cline{2-3}
+  & {\tt M5} & MD5 checksum of the sequence.  See Section~\ref{sec:ref-md5}\\\cline{2-3}
   & {\tt SP} & Species.\\\cline{2-3}
   & {\tt UR} & URI of the sequence.  This value may start with one of the standard
   protocols, e.g http: or ftp:.  If it does not start with one of these protocols, it is assumed to be a file-system path.\\\cline{1-3}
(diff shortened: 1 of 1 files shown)


jkbonfield:
+1 as-is.

The original one had "excluded spaces", so the only files this would have a breaking change on are ones with unsupported binary data - eg control codes < 32 or 8-bit data, neither of which can be considered to be valid sequence.

The topic of what to do when meeting out of bound data is left undiscussed, which I personally feel is an implementation detail (we document how the checksum is computed, not how errors are reported), but if the consensus is to add a statement to that affect then I'm OK with it.

SAMv1.tex
```diff
@@ -217,7 +217,11 @@ \subsection{The header section}
 \emph{Regular expression}: \emph{name}{\tt (,}\emph{name}{\tt )*}
 where \emph{name} is {\tt [0-9A-Za-z][0-9A-Za-z*+.@\_|-]*}\\\cline{2-3}
   & {\tt AS} & Genome assembly identifier. \\\cline{2-3}
-  & {\tt M5} & MD5 checksum of the sequence in the uppercase, excluding spaces but including pads (as `*'s).\\\cline{2-3}
+  & {\tt M5} & MD5 checksum of the sequence.
+The sequence must be in the 7-bit US-ASCII character set, with pads represented as `*'s if present.
```
yfarjoun: The part about "pads" is unclear, I think. perhaps you can modify my example below to include "pads"?

SAMv1.tex
```diff
@@ -217,7 +217,11 @@ \subsection{The header section}
...
+Before calculating the digest, all characters outside the inclusive range 33 (`!') to 126 (`\~{}')  are removed, and all lower-case characters are converted to upper-case\footnote{Equivalent to calling \href{http://pubs.opengroup.org/onlinepubs/9699919799/functions/toupper.html}{\sl toupper()} on the characters in the \href{http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1\_chap07.html\#tag\_07\_02}{\sl POSIX locale}}.
+The MD5 digest\footnote{See \href{https://tools.ietf.org/html/rfc1321}{\sl RFC 1321}} is calculated and presented as a 32 character lower-case hexadecimal number.
```
yfarjoun: Should we specify that the result should not be preceded by `0x`? 

How about adding an illuminating example? 
>For example, if the content of the reference sequence is:
"ACGT ACGT ACGT
acgt acgt acgt
... 12345 !!!
"

the resulting md5 is that of the string "ACGTACGTACGTACGTACGTACGT...12345!!!", namely 
>M5:cc14900b11b876f79231ce3a5860796b

daviesrob:
I've pushed a revised version for you, including examples as requested.  The whole thing got too big for the table cell, so I moved it into a new subsubsection.

SAMv1.tex
```diff
@@ -267,6 +263,39 @@ \subsection{The header section}
...
+\end{itemize}
+
+As an example, if the reference contains the following characters (including spaces):
+\begin{verbatim}
+ACGT ACGT ACGT
+acgt acgt acgt
+... 12345 !!!
+\end{verbatim}
+then the digest is that of the string {\tt ACGTACGTACGTACGTACGTACGT...12345!!!}
+and the resulting tag would be {\tt M5:dfabdbb36e239a6da88957841f32b8e4}.
```
yfarjoun: I hope someone else checked this calculation. I ran the following code:

```
$ echo -n 'ACGTACGTACGTACGTACGTACGT...12345!!!' | md5
dfabdbb36e239a6da88957841f32b8e4
```

SAMv1.tex
```diff
@@ -267,6 +263,39 @@ \subsection{The header section}
...
+
+In padded SAM files, the padding bases should be inserted into the reference as `*' characters.
+Taking the example in Section~\ref{sec:padded-sam}, the padded version of the reference is
+\begin{verbatim}
+AGCATGTTAGATAA**GATAGCTGTGCTAGTAGGCAGTCAGCGCCAT
+\end{verbatim}
+and the corresponding tag is {\tt M5:caad65b937c4bc0b33c08f62a9fb5411}.
```
yfarjoun: (I concur on the md5)

SAMv1.tex
```diff
@@ -267,6 +263,39 @@ \subsection{The header section}
 \end{longtable}
 \end{center}
 
+\subsubsection{Reference MD5 calculation}\label{sec:ref-md5}
+The {\tt M5} tag on {\tt @SQ} lines allows reference sequences to be uniquely identified through the MD5 digest of the sequence itself.
+As the digest is based on the sequence and nothing else, it can help resolve ambiguities with reference naming - for example, it allows a quick way of checking that references named `1', `Chr1' and `chr1' in different files are in fact the same.
+
+The reference sequence must be in the 7-bit US-ASCII character set.
```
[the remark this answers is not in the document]
jkbonfield: It's not particularly narrowed down though, just ruling out 8-bit data for which I can see no justification at all, so I think it's appropriate.

daviesrob:
@yfarjoun, suggested changes have been applied, and I've also rebased this.

I decided on reflection to use a new sentence instead of the emdash.

Allowing 8-bit characters leads to complications like the many possible [unicode spaces](http://jkorpela.fi/chars/spaces.html).  It seems easier just to ban them, given they are not needed for the sequence.

The MD5 calculations were checked both using a program for calculating fasta MD5s and by using `echo -n '...' | md5sum`.  It looks like you agree with both too.

jkbonfield:
Approved.

yfarjoun:
👍

jkbonfield:
Squashed and merged as 7326e54db05075707fe133d003297681a5313c52

PR #238, opened 2017-09-06
specify BX as a linked-read style identifier

pmarks:
standardize the 'BX' tag as the 'to-be-used' identifier for a group of reads derived from a haplotype-limiting dilution pool of long molecules (aka 'Linked-Reads').  Are people OK with the clarifying note about example of technologies that produce this data type (even if 10x is the only actually using BX?)

For some history, see #200.


pmarks:
@jkbonfield @yfarjoun @jmarshall @tfenne @daviesrob - see here for proposal on standardizing BX

yfarjoun:
in light of the barcode tags that currently exist, can we have clarification of why this is needed and why they current tags cannot be used?

pmarks:
@yfarjoun this comment summarizes our position on `BX`: https://github.com/samtools/hts-specs/pull/200#issuecomment-318174517

yfarjoun:
@pmarks the only comment I could find in that long thread is "....define BX to denote a collection of short reads generated from an a limiting dilution subsample of long molecules from the sample genome."

To me that sounds very-much like a library, and so can be written into the BC tag in the @RG header field. can you explain why that is not enough?

pmarks:
@yfarjoun in a limiting dilution linked-read sequencing library (e.g. 10x Linked Reads) there will be roughly 1.5 million unique BX tags, each with a few hundred read pairs. Creating a header with that many @RG entries seemed likely to cause lots of issues with exsiting implementation.  The top-level 'library' is a distinct tube of DNA, and it is possible to make BAM files containing multiple of these libraries.  So for these reasons we'd didn't think it wise to the RG tag for this purpose.

yfarjoun:
I guess I'm still missing something....what is the biological thing that is being identified by the BX tag? and how is it that you need by BX and RX at the same time? I'm sorry for being so slow, but we want to make sure we are adding value to the spec and that it is not **too** technology specific.

pmarks:
In our usage both RX and BX identify a group of short reads that were generated from a random limiting dilution sample of long molecules.

RX is the raw barcode sequence from the sequencer.
BX is the 'identfier' which is the corrected barcode with a suffix like '-1' or '-2' known as the 'barcode group' that disambiguates distinct partitions that happen to have same barcode sequence. This is needed if you're combining data from multiple runs of the technology that were sequenced independently. 'Barcode group' separate concept from the read group, because a library derived from a single barcode group can be sequenced multiple times to generate multiple read groups.

nh13:
Use the read group to disambiguite libraries and sequencing runs as we have done since the beginning of SAM. Use BX to identify shirt reads that observe the same long molecule (is still argue MI is better than BX but that ship has sailed).  If two short reads have the same BX but different LB values in the RG, they’re from different libraries so should be treated as coming from different long molecules. 

I think the crux is flattening all the run/library/long-molecule information into one tag (BX) versus leveraging the existing header (RG.LB) and tags (RG and MI). For example, if I see two reads with the same MI, I still check the library value in the read group to see if the reads are from different libraries. The slippery slope is to assume the BX/MI tags are unique across sequencing runs, libraries, etc. end rant.

yfarjoun:
@pmarks I think I agree with @nh13 that you seem to be putting the library information into a tag. but this information should be represented in the LB field in the RG tag, so why not use that?

yfarjoun:
given the similarities between CB and BX would it be possible to use CB for identifiying either linked reads or cells. Do you envision a need for identifying linked reads _and_ cells for the same experiment?

yfarjoun:
After a priate discussion with @pmarks at AGBT 2019, we agreed that the way forward would be to consider adding a tag to the @RG header line so that the type of library can be specified. 

closing this in apprehension of that PR. (please re-open if I'm mistaken)

PR #242, opened 2017-09-24
htsget security section 2nd iteration

mlin:
incorporating detailed discussion on #235 and some further annealing by me

cc @bakerdb @tk2 @mgcam @kemp-ggh @jeromekelleher @jmarshall @dkj

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -329,10 +329,11 @@ While the blocks must be finally concatenated in the given order, the client may
 3. should provide CORS
 4. should allow multiple request retries, within reason
 5. should use HTTPS rather than plain HTTP except for testing or internal-only purposes (providing both security and robustness to data corruption in flight)
-6. Server must send the response with either the Content-Length header, or chunked transfer encoding, or both. Clients must detect premature response truncation.
-7. Client and URL endpoint may mutually negotiate HTTP/2 upgrade using the standard mechanism.
-8. Client must follow 3xx redirects from the URL, subject to typical fail-safe mechanisms (e.g. maximum number of redirects), always supplying the headers, if any.
-If a byte range HTTP header accompanies the URL, then the client MAY decompose this byte range into several sub-ranges and open multiple parallel, retryable requests to fetch them. (The URL and headers must be sufficient to authorize such behavior by the client, within reason.)
+6. need not use the same authentication scheme as the API server. URL and `headers` must include any temporary credentials necessary to access the data block. Client must not send the bearer token used for the API, if any, to the data block endpoint, unless copied in the required `headers`.
+7. Server must send the response with either the Content-Length header, or chunked transfer encoding, or both. Clients must detect premature response truncation.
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Looks good to me, +1.

I think the key sentence 
> Sensitive information transmitted on public networks, such as access tokens and human genomic data, MUST be protected using Transport Level Security (TLS) version 1.2 or later, as specified in [RFC 5246]

works well now, and allows for the reasonable cases where HTTPS isn't necessary.

tk2:
Thanks Mike, I think you've incorporated all of the concerns raised. +1 from me also.

kemp-ggh:
LGTM

bakerdb:
Line 101 renders authentication and authorization as “preconditions” of this specification – that is, any policy requirement for authentication and/or authorization has been fulfilled prior to use of this API.  But then lines 103 and 105 specify requirements for the precondition – creating ambiguity.  

Lines 103 and 105 seem to be saying that when calling this API, if the data holder requires client authentication, the client MUST present a bearer token that conforms to RFC 6750, and if the data holder requires authorization, it MAY present a bearer token that conforms to RFC 6750.   This makes no sense because if the API recognizes a bearer token confirming that an identity has been authenticated, it should also be able to recognize a bearer token confirming that the access has been authorized.   Both use the same RFC 6750 format:  “Authorization:  bearer xxx”

I suggest revising these lines 101-105 to read:

If this API is implemented by a data holder that requires client authentication and/or authorization, the API request MUST present the token in an Authorization Request Header Field with the "bearer" HTTP authorization scheme as defined in RFC6750. The  authorization header MUST identify the access_token as a "bearer" token:

    Authorization: Bearer [access_token]

The policies and processes used to perform authentication and authorization, and the means through which access tokens are issued, are beyond the scope of this API.

mlin:
@bakerdb thanks, this is great and quite a lot more concise and clear, I have updated the branch with approximately that (0663211). Does it look ok? Did you want to keep the gentle recommendation for OAuth 2.0 framework? (I included it in that commit)

I both appreciate and lament the difficulty in sorting out all the "auth*" terminology. There seems to be a missing bit of vocabulary to concisely distinguish (a) the user authenticating his/her identity using her primary credentials, and (b) API requests authenticating they're on behalf of this user by virtue of an access token. And the vehicle for such request authentication is, of course...the Authorization header!

mlin:
@bakerdb Agreed and done!

tk2:
@bakerdb are you happy with @mlin 's final set of changes? Keen to merge so we can tag 1.0!

bakerdb:
@mlin @tk2 Changes look good.  Thank you.

mlin:
Excellent, thank you!

I will merge this eod Friday if no further issues are raised

mlin:
Based on a hiccup encountered during one of the first implementation efforts, I have added one more commit f0f6ecb clarifying that the data block server may not use the same auth scheme than the API server (in which case the API server is responsible for telling the client exactly what to do).

jeromekelleher:
Looks like a good clarification to me.

PR #243, opened 2017-09-26
Yf file extensions

yfarjoun:
This isn't so much a proposal for inclusion as is, since it isn't polished at all. 

I hope that this can spark discussion about:

1. How should this information be included in the project (a single file as presented here, or broken up into the various components, or both)
2. What to do about all the missing information 
3. How to push the field into increased consistency and less fragmentation of specification and code pathways.


CRAMv2.1.tex
```diff
@@ -1214,6 +1216,26 @@ \subsubsection*{General notes}
 container headers, namely sequence id, alignment start and span, container start 
 byte offset and slice byte offset inside the container. 
 
+The currently supported index formats are listed in table \ref{tab:index}.
```
jkbonfield: The table arrives at a silly place, so maybe we can coerce it to appear after the Indexing chapter heading.

Also this needs to be applied to CRAMv3.tex too.

The description itself is fine.  Thanks.  Htslib doesn't actually support .cram.bai though, but I think it's fine to list the options in the spec and having .crai first is good.

FileExtensions.tex
```diff
@@ -0,0 +1,113 @@
...
+\hline
+SAM file &plain& {\tt .sam} & SAM \tabularnewline
+\hline
+CRAM file &cram& {\tt.cram} &CRAM\tabularnewline
+CRAM index &cram index&  {\tt .cram.crai} & CRAM+SAM  \\
+BAM index &bam index &{\tt .cram.bai} &CRAM+BAM\tabularnewline
+CRAM index & cram index&{\tt .crai}&CRAM\tabularnewline
+\hline
+BAM file & bgzip &{\tt .bam}&SAM  \\
+BAM index & bam index & {\tt .bai} &SAM  \\
```
jkbonfield: This misses .bam.bai and .bam.csi.  Note .bam.bai was the first index name and was the default written to by both Samtools and Picard up until 2010 (https://github.com/samtools/htsjdk/commit/7459fbacda9312b28eb4a22200ced530cb8a3297).  Therefore it should be the first one searched for.

It's also consistent with how cram works and with vcf.gz too (we get .vcf.gz.tbi from tabix, not .vcf.tbi or .tbi).

Finally, it's the only method that works correctly when we have foo.bam and foo.cram in the same directory both indexed in bai format.

FileExtensions.tex
```diff
@@ -0,0 +1,113 @@
...
+\begin{table}
+\begin{center}
+\begin{tabular}{|c|c|c|c|}
+\hline
+\textbf{File Type} & \textbf{Algorithm} &\textbf{File extension}& \textbf{Specified in} \tabularnewline
+\hline
+SAM file &plain& {\tt .sam} & SAM \tabularnewline
+\hline
+CRAM file &cram& {\tt.cram} &CRAM\tabularnewline
+CRAM index &cram index&  {\tt .cram.crai} & CRAM+SAM  \\
```
jkbonfield: I don't understand the distinction of CRAM+SAM vs CRAM+BAM.  The .crai is defined in CRAM only, while .bai is CRAM spec with SAM/BAM spec for indexing.  Hence the first one should be "CRAM" only instead of "CRAM+SAM".

FileExtensions.tex
```diff
@@ -0,0 +1,113 @@
...
+BAM index & csi & {\tt .csi} &CSI \\
+\hline
+VCF file & plain& {\tt .vcf}  & VCF \\
+VCF index & tribble & {\tt .vcf.idx} & CODE ONLY\\
+\hline
+gzipped VCF file & bgzip & {\tt .vcf.gz} &CODE ONLY \\
+gzipped VCF tabix index & tabix & {\tt .vcf.gz.tbi}  &tabix\\
+\hline
+BCF file (compressed or not) & bcf or bcf+bgzip& {\tt .bcf} &BCF  \\
+BCF tabix index & bam index& {\tt .bcf.idx}  &SAM\\
```
jkbonfield: Add .bcf.csi (and .bcf.bai, or has that been dead for so long we should ignore it?).  Bcftools since at least 1.3 and probably long before produces .bcf.csi as the index filename.

Also I'm not sure where .bcf.idx comes from.  It's definitely not in the SAM spec.

FileExtensions.tex
```diff
@@ -0,0 +1,113 @@
...
+\hline
+FASTA file &  plain &{\tt .fasta} & https://blast.ncbi.nlm.nih.gov/Blast.cgi \\
+FASTA file & plain & {\tt .fa} & https://blast.ncbi.nlm.nih.gov/Blast.cgi \\
+FASTA index & fai & {\tt .fai} & CODE ONLY  \\
```
jkbonfield: Samtools faidx creates _filename_.fai, so .fa.fai or .fasta.fai typically.   Furthermore, if we have foo.fa and foo.fai and use ```samtools faidx foo refID`` it won't read foo.fai; it'll index foo.fa as foo.fa.fai instead and use that.

FileExtensions.tex
```diff
@@ -0,0 +1,113 @@
...
+\hline
+gzipped FASTA file & bgzip & {\tt .fasta.gz}  & CODE ONLY\\
+gzipped FASTA index & fai+gzi & {\tt .fasta.gz.gzi} &currently in PR \#997 in htsjdk \\
+\hline
+gzipped FASTA file & bgzip & {\tt .fa.gz} & CODE ONLY \\
+gzipped FASTA index & fai + gzi & {\tt .fa.gz.gzi}  & currently in PR in htsjdk \\
+\hline
+FASTQ & ??? & {\tt .fastq} & "crappy paper" \\
```
jkbonfield: Harsh!  It's the ill defined format that is crappy, not the paper. :-)  We should add a footnote with the reference to the paper too, as it's external to the hts-specs site.

FileExtensions.tex
```diff
@@ -0,0 +1,113 @@
...
+remaining lines describe the possible index formats \& file extensions for said file. 
+The order of the index suffixes is the recommended order by which an implementation will search for an index file.
+
+\begin{table}
+\begin{center}
+\begin{tabular}{|c|c|c|c|}
+\hline
+\textbf{File Type} & \textbf{Algorithm} &\textbf{File extension}& \textbf{Specified in} \tabularnewline
+\hline
+SAM file &plain& {\tt .sam} & SAM \tabularnewline
```
jkbonfield: Technically tabix can index sam.gz to sam.gz.tbi, although I'm not sure what other than tabix itself can query the index.  Htslib can't (although I wonder if we ought to support it).

FileExtensions.tex
yfarjoun: I was quoting what I heard people say...perhaps I misheard... I was mostly trying to elicit a response...it worked!

Issue #208, opened 2017-05-18
fastq format?


colindaven:
Does this group aim to cover FASTQ format as well ? There was a recent case where developers used 5 lines (4 descriptive, followed by a blank line) to describe each read.

The format is defined in a publication
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2847217/
and in the MAQ webpages
http://maq.sourceforge.net/fastq.shtml

Is it planned to maintain this spec here ?
Colin

jkbonfield:
I don't believe it's in our remit,  although if the authors of that paper wished to maintain it then I'd be happy with them doing it here.   I think they published the paper out of frustration with no official spec existing, which was many years after fastq first came into existance.  (I don't suppose it was a desire to become a maintainer forever.)

vadimzalunin:
Too many dialects imo. A generic specification to suit them all would be too simple and hardly useful. 
Inevitably fastq specs would have to follow major datasets available, like SRA fastq dumps etc. So effectively this would be a set of publisher/tool/platform specific specs, which are probably already described somewhere. 
Having said that, some recommendation may be useful, not sure if a consensus can be easily reached though. Stability in roundtrip tests for example would be a good goal.

jmarshall:
That paper and the [Wikipedia page](https://en.wikipedia.org/wiki/FASTQ_format) are useful descriptions of the common parts of the various FASTQ dialects.  Extra conventions about subfields on the `@` line (such as the `1:Y:18:ATCACG` in Illumina FASTQ files) change over time and are described in various vendor documents and elsewhere (for `1:Y:18:ATCACG`, in the bcl2fastq documentation).

Rather than taking on FASTQ as an additional specification, this group might instead be inclined to recommend the use of unmapped-BAM or unmapped-CRAM in place of FASTQ.  Such files can do the job of a FASTQ file or a pair of paired-end FASTQ files, and have a well-defined way of adding tagged annotations that is more flexible than potentially-conflicting FASTQ `@` line conventions.

peterjc:
As one of the authors of the FASTQ paper, I personally would also recommend we move away from FASTQ to using unmapped-BAM (or nowadays, unmapped-CRAM) - here's what I wrote about 6 years ago on this (matching John's comments above):

https://blastedbio.blogspot.co.uk/2011/10/fastq-must-die-long-live-sambam.html

(I'd have commented back in May, but never saw the issue till now by chance via a Google search)

PR #247, opened 2017-10-06
Added FASTA and FASTQ external links

jkbonfield:
As per comments in #208 and the GA4GH conference call, this PR recommends use of unaligned {S,B,CR}AM instead of fastq.  External links to fasta/fastq provided for completeness.

See https://github.com/jkbonfield/hts-specs/tree/fastq for github rendering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,18 @@ CRAM discussions can also be found on the [samtools-devel mailing list][samdev-m
 
 The **[tabix.tex]** and **[CSIv1.tex]** quick references summarize more recent index formats: the tabix tool indexes generic textual genome position-sorted files, while CSI is [htslib]'s successor to the BAI index format.
 
+Unaligned sequence data files
+-----------------------------
+
+We do not define or endorse any dedicated unaligned sequence data format.
+Instead we recommend storing such data in one of the alignment formats (SAM, BAM or CRAM) with the unmapped flag set.
+However for completeness, we list the commonest formats below with external links.
+
+[FASTA] is an early sequence-only format originally defined by William Pearson's tool of the same name.
+
+[FASTQ] was designed as a replacement for FASTA, combining the sequence and quality information in the same file.
+It has no formal definition and several incompatible variants, but is described in a paper by Cock et al.
+
 Variant calling data files
 --------------------------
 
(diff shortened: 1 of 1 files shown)


jmarshall:
I don't think this is a section more important than _Variant calling data files_. I would have it either as a single paragraph at the end of the SAM/BAM/CRAM section or move it to the bottom.

Comments on the text: "do not define **or** endorse"; "unmapped **flag** set", but maybe something like "in alignment records with their unmapped flags set"; FASTA link doesn't work; "in the same **file**".

jkbonfield:
I don't think it's important either, but I put it there simply because it felt like it fitted (aligned sequence, unaligned sequence).  I can move it to the end if you wish.   There was agreement IIRC at a previous conference call that we'd add a section for these, but given the problems with finding adequate specs I'm also open to the idea of a simple sentence under BAM,SAM,CRAM instead.

Sorry about the bad English in places. I'll fix all those typographical things.

The fasta link... sigh!  This is the problem with trying to point people to external specifications. :-(  it worked when I wrote it and it's the top result on google still too.  Should we just replace it with the Wikipedia fasta page, or take their approach of mentioning fasta20.doc in any FASTA software distribution?

README.md
```diff
@@ -19,6 +19,18 @@ CRAM discussions can also be found on the [samtools-devel mailing list][samdev-m
...
+-----------------------------
+
+We do not define nor endorse any dedicated unaligned sequence data format.
+Instead we recommend storing such data in one of the alignment formats (SAM, BAM or CRAM) with the unmapped flagged set.
+However for completeness, we list the commonest formats below with external links.
+
+[FASTA] is an early sequence-only format originally defined by William Pearson's tool of the same name.
+
+[FASTQ] was designed as a replacement for FASTA, combining the sequence and quality information in the same.
+It has no formal definition and several incompatible variants, but is described in a paper by Cock et al.
```
yfarjoun: Could you add a year or title (or even better, a DOI) so that it's not so vague?

jkbonfield:
Updated and rebased on later master.

I kept the order the same though as it makes sense to have all sequence formats together without putting VCF inbetween them.

jmarshall:
Thanks for updating. Merged as 3eece9255a7ab93975677a03711949c1c1e387dc with some final tweaks (subsubheading for this minor section, the document uses Oxford commas, the relevant Wikipedia page is `FASTA_format`).

PR #248, opened 2017-10-06
CRAM: Fix the SAM header requirements.  Also clarify slice MD5 value zero.

### LINKED ISSUE #201: CRAM header
Section 8.3 of the specification has several errors, but also things that the C and Java implementations have drifted apart on.  We need to tighten this up.

1.  The SQ:MD5 checksum is required unless the reference sequence has been embedded into the file.  Maybe.  If I use scramble -x to produce a referenceless CRAM file then it also should operate without MD5sums surely?

This then segways into a totally different incompatibility.  I wasn't able to validate whether C/Java interop works without M5 tags for referenceless encoding because of my use of MD5 zero in *slice* headers.  The C code knows the slice reference is unused, therefore does not need validating.  It does this by setting MD5 to zero.  The Java code however checks this and fails:

```
Exception in thread "main" htsjdk.samtools.cram.CRAMException: Reference sequence MD5 mismatch for slice: sequence id 0, start 10001, span 33954, expected MD5 00000000000000000000000000000000
	at htsjdk.samtools.CRAMIterator.nextContainer(CRAMIterator.java:187)
	at htsjdk.samtools.CRAMIterator.hasNext(CRAMIterator.java:261)
	at htsjdk.samtools.SamReader$AssertingIterator.hasNext(SamReader.java:592)
	at picard.sam.SamFormatConverter.doWork(SamFormatConverter.java:83)
	at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:205)
	at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:94)
	at picard.cmdline.PicardCommandLine.main(PicardCommandLine.java:104)
```

This is perhaps a bug in my code.  I'm way out beyond the spec here (although I thought we discussed and changed it, but apparently that was just on my to-do list rather than a reality!) because there is no mention of zero in that situation.  We so have a specification that 0 is used for unmapped or unsorted data, so this is a minor tweak.  Is it a useful addition though?

What does picard.jar do when it creates referenceless CRAM files?

2. "At least one RG record is required."

This is not true.  C and Java implementations can read and write CRAM without RG tags.  (It was true for version 1.0 I believe.)

3. "The HD:SO sort order is always POS."

Firstly, POS is not a valid sort order; it is "coordinate" instead.
Secondly we all support unsorted and name sorted data too, so the statement is incorrect.

Although name sorted data in picard is "interesting".  With SO:queryname it *strictly* enforces one interpretation of name sorting, so A100 is before A99.  This is not the same name sort order than samtools uses, so the two are incompatible.  Perhaps the SAM spec needs tightening here as it has no definition of what sort actually means and whether it is an intelligent sort or not.

jkbonfield:
Fixes #201 

Also added a recommendation about referenceless encoding when storing e.g. name sorted data.  Reference encoding works, but to do so it ideally requires the entire genome to be held in memory and does a large amount of memory random access.  Typically we'll be more performant to just encode the sequence as-is instead.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -567,10 +567,6 @@ The following constraints apply to the BAM header:
 \begin{itemize}
 \item The SQ:MD5 checksum is required unless the reference sequence has been embedded 
 into the file.
-
-\item At least one RG record is required.
-
-\item The HD:SO sort order is always POS.
 \end{itemize}
 
 \subsection{\textbf{Compression header block}}
@@ -743,6 +739,12 @@ blocks associated with the slice. Mapped and unmapped reads can be stored within
 the same slice similarly to BAM file. Slices with unsorted reads must not contain 
 any other types of reads.
 
+A slice containing data that does not use the external reference in
+any sequence may set the reference MD5 sum to zero.  This can happen
+because the data is unmapped or the sequence has been stored verbatim
+instead of via reference-differencing.  This latter scenario is
+recommended for unsorted or non-coordinate sorted data.
+
 The slice header block contains the following fields.
 
 \begin{tabular}{|l|l|>{\raggedright}p{200pt}|}
@@ -770,7 +772,7 @@ itf8 & embedded reference bases block content id & block content id for the embe
 reference sequence bases or -1 for none\tabularnewline
 \hline
 byte[16] & reference md5 & MD5 checksum of the reference bases within the slice 
-boundaries or 16 \textbackslash{}0 bytes for unmapped or unsorted reads\tabularnewline
+boundaries or 16 \textbackslash{}0 bytes when unused\tabularnewline
 \hline
 byte[] & optional tags & a series of tag,type,value tuples encoded as
 per BAM auxiliary fields.\tabularnewline


jmthibault79:
The text looks good to me.  Do you know if there's an issue filed for this in HTSJDK?

jkbonfield:
Rebased.

It is linked to https://github.com/samtools/htsjdk/issues/1216

However comparing this PR and that issue I wonder whether the PR should be rephrased in terms of RR (reference required)?  Eg we could state that if RR is zero in the compression header, indicating no reference is required, then the MD5 value should be ignored by decoders and encoders should fill out this field with zeros.

jmthibault79:
SAM header part looks good.

I'm not too familiar with RR or referenceless mode, so to enumerate the possible Slice MD5 cases as I understand it...

1. External reference.  RR = 1.  MD5 must be present and verified.
2. Embedded reference: MD5 *may* be present.  If so, verify it.  Does the RR flag affect this?
3. Referenceless (currently not supported by HTSJDK): RR = 0.  MD5 should be set to all-zero.
4. Unmapped-Unplaced: MD5 should be set to all-zero.  RR is irrelevant here, and therefore ignored.
5. Multi-Ref: MD5 should be set to all-zero even though RR = 1, and the external references could be verified if these were single-ref slices.

I would like to see additional text clarifying this, and add "(see above)" or similar to the "unused" in the table.

Regarding our Thursday conversation about whether to (a) merge this PR separately or (b) include it in my forthcoming PR for #370: I would like to go with (a) here to get the MD5 text right as its own change, and leave that part out of my changes.

jkbonfield:
An additional complication is that there aren't any explicitly stated defaults for these values.  The lack of RR should probably mean it is required.  This was the way CRAM was originally written, and RR was added in order to permit referenceless encoding.

The table in 8.4 needs a new column showing the default interpretation.  I think it's 1 for all the booleans.

jkbonfield:
Merged as 9c8d2f21e4c3bdba2f5755e267a774c4f536757f so we can get the other CRAM PRs building on top of this.

jkbonfield:
Although we deferred the MD5 sum description to another PR, having seen your extra table and the improved clarification it brings to MD5s, I think the table below it needs more work too.

I commented on the PR with my thoughts, but I am rather ambivalent over whether the new table would bring much given the (much needed) updates to the table below.  I don't object to it, but I also think the other improvements to some extent remove the need for the clarification table.

Issue #249, opened 2017-10-09
Proposal: Read Group BC tag


magicDGS:
For each of the reads, the `BC` tag represents the barcode sequence usually detected from the read sequence. Nevertheless, this sequence is suppose to match to some extend the one used for identify the sample/library. Because this concept is tightly linked to the concept of read group, I suggest to add to the optional RG tags the BC indicating the expected sequence for the BC tag.

Like that, one can compare the expected value for the sample/library and the one coming from the sequencer to check if there is some mismatch and identify by barcode the samples.

yfarjoun:
This would be nice, I think. At Broad, we put the barcode in the PU field
of the ReadGroup record as <FLOWCELL>.<LANE>.<BARCODE> and use it together
as a unique identifier, but I agree that making it more "computable" seems
like a Good Thing.

jkbonfield:
Thumbs up in principle, but I think we need people more experienced with all the weird and wonderful combinations that exist to see how it works in practice.

For example, using multiple barcodes per library.  Do we simply state the limitations of this implementation (it doesn't cover it), or define a serialisation method such as BC:seq,seq,seq...?  Is this the sequence added, or the sequence to compare against when decoding?  With different sized barcodes, we may algorithmically want to compare all with the same length, which means adding a bit of adapter sequence while comparing.  Etc.   I'm not qualified to think about all these nuances.

magicDGS:
I was thinking about usign the `BC` for the sequence to compare against when decoding (the one used in the library prep). It does not have to match with the `BC` tag for the read (which is the raw barcode coming from the sequencer), so it can be compared to have statistics about mismatches or to discard not complete matches.

In the case of multiple barcodes, I guess that it should be similar to the definition of the `BC` tag on the read: left to the implementation, recommending hyphen `-` as separator. Maybe it should be stated that it is recommended to use the same separator in the header and the read tag. Like that, it can also be used for decoding as a full string matching without splitting into several sequences.

In my lab, I developed a tool to de-multiplex barcoded files in different formats; currently I am using the `BC` tag on the read against a file provided by the user, and then I assign a read group. It would be nice to be able to store the information on where the assignation come from to keep track that the file was already de-multiplexed and evaluating different algorithms to do it. I was thinking to use a custom `@RG` tag (probably `bc`), but as I believe that there is more people interested on this and it would be nice to have it defined in the specs...

PR #250, opened 2017-10-10
Add Read Group BC tag

magicDGS:
As described in #249

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -255,7 +255,8 @@ where \emph{name} is {\tt [0-9A-Za-z][0-9A-Za-z*+.@\_|-]*}\\\cline{2-3}
   \multicolumn{2}{|l}{\tt @RG} & Read group. Unordered multiple {\tt @RG} lines are allowed.\\\cline{2-3}
   & {\tt ID}* & Read group identifier. Each {\tt @RG} line must have a unique {\tt ID}. The value of {\tt ID}
   is used in the RG tags of alignment records. Must be unique among all read groups in header section.  Read group IDs may be modified when merging SAM files in order to handle collisions.\\\cline{2-3}
-  & {\tt BC} & Barcode sequence placed during library preparation identifying the sample/library.
+  & {\tt BC} & Barcode sequence identifying the sample/library.
+  The sequence is the expected bases read by the instrument in the absence of sequencing errors.
   In the case of multiple Sample Barcodes (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and places a hyphen (`{\tt -}') between the barcodes from the same template.\\\cline{2-3}
   & {\tt CN} & Name of sequencing center producing the read.\\\cline{2-3}
   & {\tt DS} & Description.  UTF-8 encoding may be used.\\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -257,7 +257,7 @@ where \emph{name} is {\tt [0-9A-Za-z][0-9A-Za-z*+.@\_|-]*}\\\cline{2-3}
   is used in the RG tags of alignment records. Must be unique among all read groups in header section.  Read group IDs may be modified when merging SAM files in order to handle collisions.\\\cline{2-3}
   & {\tt BC} & Barcode sequence identifying the sample/library.
   The sequence is the expected bases read by the instrument in the absence of sequencing errors.
-  In the case of multiple Sample Barcodes (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and places a hyphen (`{\tt -}') between the barcodes from the same template.\\\cline{2-3}
+  If there are several barcodes for the sample/library (e.g., one on each end of the template), the BC value is the barcode sequences concatenated together and separated by hyphens (`{\tt -}').\\\cline{2-3}
   & {\tt CN} & Name of sequencing center producing the read.\\\cline{2-3}
   & {\tt DS} & Description.  UTF-8 encoding may be used.\\\cline{2-3}
   & {\tt DT} & Date the run was produced (ISO8601 date or date/time).\\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -255,8 +255,8 @@ where \emph{name} is {\tt [0-9A-Za-z][0-9A-Za-z*+.@\_|-]*}\\\cline{2-3}
   \multicolumn{2}{|l}{\tt @RG} & Read group. Unordered multiple {\tt @RG} lines are allowed.\\\cline{2-3}
   & {\tt ID}* & Read group identifier. Each {\tt @RG} line must have a unique {\tt ID}. The value of {\tt ID}
   is used in the RG tags of alignment records. Must be unique among all read groups in header section.  Read group IDs may be modified when merging SAM files in order to handle collisions.\\\cline{2-3}
-  & {\tt BC} & Barcode sequence identifying the sample/library.
-  In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and places a hyphen (`{\tt -}') between the barcodes from the same template.\\\cline{2-3}
+  & {\tt BC} & Barcode sequence placed during library preparation identifying the sample/library.
+  In the case of multiple Sample Barcodes (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and places a hyphen (`{\tt -}') between the barcodes from the same template.\\\cline{2-3}
   & {\tt CN} & Name of sequencing center producing the read.\\\cline{2-3}
   & {\tt DS} & Description.  UTF-8 encoding may be used.\\\cline{2-3}
   & {\tt DT} & Date the run was produced (ISO8601 date or date/time).\\\cline{2-3}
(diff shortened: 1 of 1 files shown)


magicDGS:
Friendly ping here for maintainers of SAM/BAM - @jkbonfield @jmarshall @yfarjoun

jkbonfield:
Sorry this slipped off the radar.

Is the intention that by adding to RG it wouldn't be then added to read BC tags?  If so then the SAMtags.tex document should also be updated to state this in the description of the QT:Z tag.

magicDGS:
@jkbonfield - thanks for looking into this. I conceived this `@RG` tag for specifying the _**expected**_ barcode sequence for the sample.

The main purpose is to have a self-contained BAM file, where the `@RG` identify the sample and the barcode used on the library prep, and still maintain the record `BC` tag with the raw barcode, to being able to re-run a de-mulptiplexing algorithm matching the record tag to the `@RG` tag, or re-filter based on mismatches.

Thus, I believe that using the `BC` tag on the `@RG` does not necessarily mean that the `BC`/`QT` tag should be removed, neither that they need to match in length (e.g., I found 9bp `BC` tags where the used barcode in the library prep was 8bp).

Maybe we should just clarify the difference between the `@RG` and record-level tag? Or just have yet another name for the `@RG` one to avoid confussion?

SAMv1.tex
```diff
@@ -229,6 +229,8 @@ \subsection{The header section}
   \multicolumn{2}{|l}{\tt @RG} & Read group. Unordered multiple {\tt @RG} lines are allowed.\\\cline{2-3}
   & {\tt ID}* & Read group identifier. Each {\tt @RG} line must have a unique {\tt ID}. The value of {\tt ID}
   is used in the RG tags of alignment records. Must be unique among all read groups in header section.  Read group IDs may be modified when merging SAM files in order to handle collisions.\\\cline{2-3}
+  & {\tt BC} & Barcode sequence identifying the sample/library.
```
yfarjoun: Should be made clear that this barcode is the one actively placed during library preparation, rather than the measured value.

SAMv1.tex
```diff
@@ -229,6 +229,8 @@ \subsection{The header section}
...
+  In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and places a hyphen (`{\tt -}') between the barcodes from the same template.\\\cline{2-3}
```
yfarjoun: this isn't UMI...replace unique molecular identifiers with "Sample Barcode" (capitalized appropriatly)

magicDGS: This is copied from the `SAMTag.tex`, so probably it should be changed there too (out of the scope of this PR).

Changed here.

SAMv1.tex
magicDGS: Done.

magicDGS:
Back to you @yfarjoun (and @jkbonfield)

SAMv1.tex
jkbonfield: I think we noted this before and maybe it's in one of the other pending issues to resolve.  Anyway thanks for the change.

jkbonfield:
After office discussions we realise there is still some ambiguity.  Specifically for the orientation of paired-end barcodes as sequenced by the instruments may differ (some are reverse complemented, but not universally - this is a technology implementation thing).

For this to be useful I think we need to clarify how it is to be used, which I am assuming as a means for checking the BC aux tag on the reads.  In this case maybe something like: "The sequence is the expected bases read by the instrument in the absence of sequencing errors".

There is nuance regarding length, as mixed samples can have differing length barcodes and the instrument may read the end of the adapter for some (as obviously it has a fixed number of index cycles, not variable per cluster).  However I don't think we need to get bogged down with the complex and rather quirky implementation specific details as long as we don't forbid such uses.

yfarjoun:
So....I was thinking more about this PR and have some thoughts...

- The SampleBarcode is Library-level data. 
- There are other Library-level data that one might care about (Protocol, date generated, input sample aliquot, ...) 
- I'm personally interested in being able to put Lab Protocol and am looking for a place to put it into...

I see 2 possible paths forward and while a PR is probably not a good place to discuss, I think that it would be better to have the discussion before this PR goes through. happy to move this discussion to an issue if people prefer.

1. Add more library-related fields to the `@RG` header line
2. Create a new @LB header line that the @RG can point to, and put the library-level information there.

Thoughts?

yfarjoun:
I totally understand your position on `@LB`, but I wanted to be open about what the options are so that we can have a discussion...

Regarding DT and PL: Not quite. 
* DT is the date of _sequencing_, not the date of library construction. These can differ substantially.
* PL is the _sequencing_ platform (illumina, pacbio etc.) I'm looking for a open-ended string that describes the *lab* protocol for creating the library. An example for a string I had in mind would be "Somatic Human WES (Deep Coverage) v1" or "RNA-010400 Stranded, Long Insert Transcriptome v1.1" this might not mean much to a non-internal user, but in large studies it could be used to control for batch effect when you do not have databases informing you of the protocols that were used.

magicDGS:
Because a lot of information for library is already there, I think that it will be confusing to split into two...but I totally understand than other users might desire a more concrete header line for that kind of cases. For the lab protocol, I think that the `DS` (descriiption) tag might be sufficient if it is mostly for internal use.

If all that tags are optional, it might be useful to have them all in `@RG`: I always saw the read group as an identifier of sample and other metadata describing the origin of the read (sequencing technology, program, etc). My vote is for including them (in a different PR) at the `@RG` level. For the case of this PR, I believe that whatever decission is made, the `BC` should be in the read group.

yfarjoun:
I have never seen the DS field in use...do others use it? what for?

magicDGS:
I've only seen the `DS` field for a textual description of some data, but because is a free-text I think that it would be the best for internal usage (e.g., your lab protocol use case: `DT:"Somatic Human WES (Deep Coverage) v1"`.

jkbonfield:
Sanger use DS field.  For example one of ours states:

"DS:ERP002385: 4X coverage for the Mandinka from Gambia Western Division (GWD). This data is part of a pre-publication release. For information on the proper use of pre-publication data shared by the Wellcome Trust Sanger Institute (including details of any publication moratoria), please see http://www.sanger.ac.uk/datasharing/"

So boring blurb plus a description of the project and a link to the ERA accession.  That's quite an old run (modern ones are all so locked down I can't read any).

yfarjoun:
👍 from me

jkbonfield:
It still has the issue of not identifying the orientation - see my query above.

At the last conference call we agreed that it ought to be the bases we would expect to see as produced by the sequencing instrument; which in some experiments can be the reverse complements of the ones added during library preparation.  This is clearer than "sequence placed during library preparation".

magicDGS:
@yfarjoun / @jkbonfield - something like the current sentence is ok? Otherwise, I accept whatever suggestion...

jkbonfield:
Looks good. :+1: from me.

yfarjoun:
Oh, should modify the change log.

magicDGS:
@yfarjoun - I added the history line and rebase to avoid conflicts.

jmarshall:
At this point, I too am :+1: with adding a BC field to `@RG` with the meaning described.

One caveat: I don't think the way the final sentence talks about “the recommended implementation [does blah blah]” is appropriate for a specification. Are we explicitly leaving the door open for implementations to do something else? How does that help?

I'd quite like to rephrase that as

> If there are several barcodes for the sample/library (e.g., one on each end of the template), the BC value is the barcode sequences concatenated together and separated by hyphens ('-').

@magicDGS does that still reflect what you want to say there (in particular, I'm not sure what “from the same template” was implying in your text)? Agree/disagree/don't care from you and the other maintainers, and then I'm happy to squash and merge this.

jkbonfield:
That text (recommended blah) comes from the BC tag in SAMaux.tex, but likely it was phrased that way there because at the time of adding to the spec it was already in the wild with methods and we wouldn't be popular if all bar one became illegal.

Here it's a bit different as it's not in use yet (hurrah for doing things in the correct order!), so I agree we should tighten it up and not give wiggle room.

jmarshall:
@jkbonfield: Thanks — aargh, somehow I was looking at a copy of SAMtags.pdf from before PR #200 was merged. So that sort of thing is already over in that ancillary document…

magicDGS:
@jmarshall (and @jkbonfield) - I used the same as in the `BC` record tag to allow other implementations to use exact match, but I am ok with the enforcement to use hyphens as it is the only one that I use (unless the procedence of the read data is from outside).

magicDGS:
Friendly ping here for the maintainers: @jmarshall, @jkbonfield, @yfarjoun. Thanks!

jmarshall:
I have merged this with some further wordsmithing as 5b84bd3b33eaeb390e0c962d5ded5dd149b879d2.

I went back to the _recommended implementation_ wording as it seems important that even for these other implementations, the header's BC value matches the BC value on the records.

(Other notable wordsmithing: the spec uses _sequencing machine_ rather than _instrument_; _errors_ rather than _sequencing errors_ so there's no quibbling about PCR errors or similar.)

magicDGS:
Thanks @jmarshall!

magicDGS:
And thanks @yfarjoun and @jkbonfield for reviewing too!

Issue #251, opened 2017-10-16
Potentially ambiguous interpretation of FORMAT field duplicates in VCFv4.3.pdf


eyherabh:
I believe that section 1.6.1, item 8 of VCFv4.3.pdf can be potentially ambiguous. Specifically, that section says: "INFO fields are encoded as a semicolon-separated series of short keys with optional values in the format: <key>=<data>[,data]. INFO keys must match the regular expression ^[A-Za-z ][0-9A-Za-z .]*$, duplicate fields are not allowed."
As it is, the INFO fields are allowed to contain two fields with the same key but different values, because a field is a key-value pair.
Could you please clarify me whether this was indeed the intention, or whether the last sentence should actually say "duplicate keys are not allowed", as opposed to "duplicate fields...".
Thank you very much.
Best regards.

cyenyxe:
The intention is to disallow duplicate **keys**. Multiple values must be assigned to a single key in a comma-separated list.

eyherabh:
Thanks. That's what I thought but not what the text says. Then, I believe the following sentence:

"INFO  keys  must  match  the  regular  expression^[A-Za-z][0-9A-Za-z.]*$,  duplicate  **fields**  are  not  allowed."

which belongs to section 1.6.1 point 8 of VCFv4.3.pdf and incorrectly states that "duplicate fields are not allowed", should be modified as follows:

"INFO  keys  must  match  the  regular  expression^[A-Za-z][0-9A-Za-z.]*$,  duplicate  **keys**  are  not  allowed."

PR #256, opened 2017-10-24
Update description of INFO fields in VCFv4.3.tex to avoid confusion and improve consistency

eyherabh:
Changes between lines 331 and 400:
1) "$<$key$>$=$<$data$>$[,data]" turned into "key[=data[,data] ]", for the following two reasons. First, as opposed to the angle-bracketed IDs mentioned above, neither the key nor the data are required by the standard to be angle-bracketed, as the examples in the standard show. Second, the paragraph stated that the data is always optional, not just the second element in the list following the key.
2) As discussed in Issue #251, the intention was to prevent INFO keys from being duplicated, as opposed to fields. For this reason, and to avoid misinterpretations, keys in this new version are referred only as keys, and not also as fields and sub-fields as the previous version confusingly did. 
3) Additional changes may be required in the description of the genotype fields to improve consistency


VCFv4.3.tex
```diff
@@ -359,16 +359,16 @@ \subsubsection{Fixed fields}
...
+If genotype information is present, then the same types of data must be present for all samples. First a FORMAT field is given specifying the data types and order (colon-separated FORMAT ids matching the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, duplicate ids are not allowed). This is followed by one data block per sample, with the colon-separated data corresponding to the types specified in the format. The first sub-field must always be the genotype (GT) if it is present. There are no required sub-fields. Additional Genotype fields can be defined in the meta-information, however, software support for such fields is not guaranteed.
```
cyenyxe: I agree with renaming "FORMAT ids" to "FORMAT keys" for the sake of consistency.

PR #261, opened 2017-11-02
Inconsistent use of P suffix

eyherabh:
VCF4.3 states that the suffix P means probability. However, this is not the case for DP. Wouldn't it be nice that the standard does actually follow its own recommendations (or points out to the exceptions)?
With that in mind, I have added a note pointing out the exception.

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -493,7 +493,7 @@ VCF records use a simple haplotype representation for REF and ALT alleles to des
     represented in some cases as phred-scale in a separate tag (e.g.~PL).
 
     \item The "P" suffix means "probability" as linear-scale probability in the
-    posterior distribution, which is Pr(Model$|$Data).  Examples are GP, CNP.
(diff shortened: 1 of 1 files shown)


yfarjoun:
I don't think that the note should be taken to mean "every tag that ends in P means posterior probability"...if anything I'd change the note, rather than add DP as an exception.

pd3:
Indeed. The comment attempts to highlight the naming differences for qualities, probabilities and likelihoods. DP represents something completely different, I don't think there is a danger of confusion.

eyherabh:
Being it a standard, I took it as a rather strong suggestion, and without
owing its own exceptions, the standard in my view weakens. My change was
one option and its already done, changing the note might work too.

jmarshall:
There's space at the start of §2.1 for an introductory paragraph before the bullet list. Just a sentence something like “Some tag names follow conventions indicating how their values are represented numerically:” would clear this up.

And the `"` quotes need changing to TeX quotes and the maths should be in `$ math mode $`.

PR #262, opened 2017-11-06
CB, CR and CY tags.

nunofonseca:
Hi,
This pull request is to include the CR and CY tags in the spec and derives from the discussion in https://github.com/samtools/hts-specs/pull/239#issuecomment-34177715.
Cheers.


pmarks:
Thanks @nunofonseca.  10x definitely supports standardizing the raw barcode sequence & qual tags . This data allows reconstruction of the original FASTQ sequences used to run a single-cell pipeline & allows for alternative barcode correction schemes to be experimented with

SAMtags.tex
```diff
@@ -96,6 +97,7 @@ \section{Standard tags}
   {\tt PU} & Z & Platform unit \\
   {\tt Q2} & Z & Phred quality of the mate/next segment sequence in the {\tt R2} tag \\
   {\tt QT} & Z & Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag \\
+  {\tt CY} & Z & Phred quality of the cellular-barcode sequence in the {\tt CR} tag \\
```
nh13: Can you move this after `CT` so it's alphabetical?

SAMtags.tex
```diff
@@ -62,6 +62,7 @@ \section{Standard tags}
   {\tt CC} & Z & Reference name of the next hit \\
   {\tt CM} & i & Edit distance between the color sequence and the color reference (see also {\tt NM}) \\
   {\tt CO} & Z & Free-text comments \\
+  {\tt CR} & Z & Cellular barcode sequence as reported by the sequencer (uncorrected) with the respective quality scores stored in CY\\
```
nh13: Change `in CY\\` to `in the {\tt CY} tag`.

nh13: How about this:
`Sequence bases of the cellular barcode bases (uncorrected) with the respective quality scores stored in the {\tt CY} tag`?  I am trying to match the style of writing in the `RX` tag

SAMtags.tex
```diff
@@ -271,6 +273,15 @@ \subsection{Barcodes}
 A unique ID within the SAM file for the source molecule from which this read is derived. 
 All reads with the same {\tt MI} tag represent the group of reads derived from the same source molecule. 
 
+\item[CR:Z:\tagvalue{str}]
+Cellular barcode. 
+Cellular barcode sequence as reported by the sequencer (uncorrected) with the respective quality scores stored in {\tt CY}. All reads with the same {\tt CR} tag derive from the same cell.
```
nh13: Same issue as in the short description.  "sequence" what (bases)?

SAMtags.tex
nh13: Oh boy, the other tags are not alphabetically ordered.  Perhaps a second commit to fix that while your changing things?

SAMtags.tex
nunofonseca: All the above comments make sense. I sorted tags alphabetically in table 1 and in the Barcodes section. I also rephrased the description of the tags to make them more similar to the style of RX tag as suggested.

SAMtags.tex
jmarshall: FYI the intention in the SAMtags document was for the §1 table to be alphabetically ordered, but for the individual subsections' tags to be ordered in some semantically meaningful way for that particular subsection. For example, in §1.2 (Metadata) RG is fundamental while CO is trivial, and the subsection is ordered accordingly. YMMV as to what is semantically meaningful for a barcodes subsection.

nunofonseca: OK, I resorted the tags in the Barcodes section in a semantically meaningful way.

SAMtags.tex
```diff
@@ -59,13 +60,16 @@ \section{Standard tags}
   {\tt BC} & Z & Barcode sequence identifying the sample \\
   {\tt BQ} & Z & Offset to base alignment quality (BAQ) \\
   {\tt BZ} & Z & Phred quality of the unique molecular barcode bases in the {\tt OX} tag \\
+  {\tt CB} & Z & Sequence bases of the (possibly corrected) cellular-barcode sequence\\
```
jkbonfield: Looking at some recent cellranger data, I see CR and CB tags like "GCGCAGTTCTCGAGTA" and "GCGCAGTTCTCGAGTA-1".  Ie CB isn't just an error corrected version, but is a UMI-like identifier that contains more than just the sequence.

Note this isn't how 10x describe it on their web pages either: https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/output/bam

pmarks:
@nunofonseca  @jkbonfield - See our wording for `CB` in #239 -- we definitely want to use the term 'identifier' instead of 'sequence'. The `-1` suffix can be used to differentiate cells that were from distinct 'lanes' of e.g. microfluidics, and went into different libraries, but happened to get the same barcode.  

This PR is partially now with our PR #239.  Remaining issue is our proposed `UB` tag to encode an RNA-seq specific UMI. We propose `UB` as distinct from `RX` because in typical 3' RNA-seq protocol you do the following:
1. Introduce cell-barcode and UMI at 3' end during reverse-transcription
2. PCR Amp
3. Shearing, which introduced diverse 5' end positions _for the same UMI_
4. Sample Index Ligation & PCR

So you really need to think differently about how to 'dedup' / 'count UMIs' in the typical genomic DNA  (RX) and single-cell RNA-seq (UB) cases.

jkbonfield:
Yes I see what you mean.  We need to merge #239 and #262 together somehow as currently they have a clash, both defining CB and not with the same text.  @nunofonseca, @parks, @pryvkin10x can you work together on this to present a single completed PR for review?  Or failing that agree who is describing which tags?

Thanks.  Meanwhile I'll go through both again and make sure I understand it better.

jkbonfield:
Looking at this in more depth, listening to reports from the GA4GH conference, and looking at actual cellranger output files, I think I can see two requirements here.  Please correct me if I get any of this wrong!

1.  How to process data; this needs identifiers, which can be the raw uncorrected sequence if no correction is necessary or corrected sequence plus any additional disambiguation codes.  Analysis only needs a few bits of information then - specifically the sample barcode (BC), cell barcode (CB) and unique molecule barcode (UB).

2. To permit BAM to Fastq conversion, the original raw sequences are required, which is where CR and UR come in, possible TR too, along with their quality equivalents.  We also need to know how to stitch it all together, which changes between vendors, and even between chemistry revisions for a single vendor.

To solve issue 2, I see \@CO tags such as:

```
@CO     10x_bam_to_fastq:I1(BC:QT)
@CO     10x_bam_to_fastq:R1(CR:CY,UR:UY)
@CO     10x_bam_to_fastq:R2(SEQ:QUAL)
```

This is absolutely a good idea, but the definition of this isn't in any pull request yet.  The concept behind it though is definitely helpful.  If I recall the notion from the meeting is that the raw tags only serve a purpose for conversion back to fastq, so don't need to be blessed "official" tags and could be coded in lowercase.  All we need to a statement as to which tags to use.  I think that is a great solution that helps to reduce the complexity considerably for the end user who is trying to understand which tag to use for downstream analysis.

Is this a correct summary from the conference?  If so, is it still desired?  It is clearly not what this PR is about because it attempts to bless the fastq oriented tags, but does the community really want this or do they simply want "a way for bam2fastq to work" rather than prescribed naming for every component?

pmarks:
@jkbonfield - We're happy to close #239 and settle on this PR, if we can agree on the wording of `CB` and the inclusion of `UB`.  @nunofonseca - any thoughts on my previous comments.

At the GA4GH meeting, I committed to making a PR to formalize our `@CO` scheme for bam-to-fastq. I'm behind on this & will get it up ASAP -- glad to hear that you're in favor of the high-level approach.

SAMtags.tex
nunofonseca: I think that the CB tag description in the PR is more general and subsumes the 10x description (provided in the above link). Furthermore it is consistent with other tags that may be corrected.

nunofonseca:
@pmarks @jkbonfield Regarding the wording for the CB tag, namely the use of identifier vs sequence. I understand the motivation to combine two different types of information (cell+"lane") in a single tag to get an identifier - however it seems to me, correct me if I' m wrong, that this is 10x specific requirement. This could (and probably should) be accomplish by having  the "lane" in a different tag anyway, or/and have a ci tag with the cell identifier. Furthermore the CB tag as is in the current  PR is consistent with the other "corrected" tags - changing "sequence" to identifier would render it inconsistent.

nunofonseca:
@jkbonfield  I also tried to address the fastq->bam->fastq lossless conversion issue using the tags in the spec and the ones in this PR (prototype here: https://github.com/nunofonseca/fastq_utils#fastq2bam---lossless-fastq-to-bam-convertor).  An argument to have the raw tags as official would be the improvement of the BAM/CRAM compression by taking advantage of knowing what type of data is stored in a tag.

nunofonseca:
@pmarks I agree that the way you analyse the UMIs in RNA-seq is different from DNA-seq. However it is not clear to me why the need for a new tag instead of using RX?   If you know that the data is RNA-seq then  you should be able to perform the analysis of the values in RX accordingly, right?

pmarks:
@nunofonseca 
The motivation for 'lane' suffix isn't 10x specific -- it arises any time a user wants to combine data from multiple libraries (ie taking the same aliquot of cells & splitting across 2 'technical runs' to increase the cell count). If each of those technical runs are processed into a separate library, then barcode ACGT in library 1 references a distinct cell from barcode ACGT in library 2. I don't think there's any argument that it's critical to maintain this distinction. We believe that the usability is greatly enhanced by having the 'cell barcode' tag carry this 'technical run' disambiguation, rather than storing that in a different tag. 

Our arguments in favor of including the 'technical run' annotation are:
1) simplicity - it's putting more burden on downstream tools to figure out what reads came from the same cells if there's a multi-tag computation needed to make that decision
2) interoperability - for example [BAM sort-by-tag](https://github.com/samtools/samtools/pull/675) works with a single tag, but would not work with a multi-tag cell identification
3) continuity - there are now a very large number of BAM files from Cell Ranger in active use, that use definition for `CB`.

nh13:
@pmarks just for consideration, the platform unit tag in the read group header line is a good place to store the lane.  I can understand the desire to have it right along with the CB tag, but more food for thought.

pmarks:
@nh13 - here's where it gets complicated: imagine I make library A from aliquot A, and library B from aliqot B. I can potentially sequence library A multiple times, which would each have a distinct _sequencer_ `PU`, but the same _single cell_  'platform unit'. I'd be very nervous to start putting the _single cell_  information into the `PU` field, when to date it's been used to track the sequencer.

nunofonseca:
@pmarks @nh13 what about RG or/and LB?

pmarks:
@nunofonseca Agreed, RG.LB would be more appropriate.  Nevertheless, all my arguments above are still very much valid & should be debated. 
 
Re the arguments you made above:
1. The "-1" is 10x specific. I believe any technology that lets users to do multiple 'technical runs' to increase cell count will encounter this issue, so I don't think the basic motivation for the suffix is 10x specific.

2.  "[the current PR]  is consistent with the other "corrected" tags - changing "sequence" to identifier would render it inconsistent".  Our proposal is to specify `CB` as 'to-be-used' cellular identifier, not a sequence. Surely there's room in the spec to specify identifiers as tags?

3. "the current wording seems consistent to the description of the CB tag in the 10x website".  See a little further down on the page you link to: we document the "-1" suffix, and have been using it since the first release of Cell Ranger in 2015.

If there's desire on someones part to specify a 'corrected cell barcode _sequence_' tag, that's fine, but we'd ask that it use a different tag than `CB`, and `CB` be used according to the Cell Ranger definition, given the large number of BAM files with this convention already in the wild.

nunofonseca:
OK, so it seems that are discussing two different tags:
1) tagA - corrected cell barcode sequence (as in the PR)
2) tagB - a cell identifier tag (which combines tagA with LB or other tag)
and naming both of them CB.  If helps moving this PR forward I can change CB to CA in this PR so that we focus the discussion here on this tag.

@pmarks Honestly I don't know if there is room in the spec to include identifiers as tags that combine information from other tags (maybe @jkbonfield, @nh13  or other may comment on this?). I agree with the point that it potentially makes the consumption of the data simpler. The caveat is that would lead to data redundancy. What if htslib/samtools had an option to define a virtual tag (similar to a database view), as the combination of multiple tags? This would allow to define identifiers in a very flexible way and avoid redundancies.

jkbonfield:
I don't particularly wish to push for defining a tag name in the spec that conflicts with existing usage, and like it or not CB is used right now as an *identifier* and not as a pure *sequence*.  Yes it's wrong that these tags were all used in upper-case, but we're where we are and cannot wind back the clock.  (Realistically the way to prevent this mistake happening in future is for the archives to simply reject files that are non-spec compliant, but for some reason there has always been reluctance to do this.)

So I think the original PR definition of CB was better.  Possibly we can find a middle ground here by saying something like CB is a combination of sequence with optional separator (recommend dash) and additional non-sequence identifier.  This would permit tools to safely parse it and extract sequence if they wish to do so, without breaking the use of it as an arbitrary string for cell disambiguation.

jkbonfield:
@nh13 I dislike exploiting the PU header tag as arbitrarily restricts the way experiments can be designed and ran.  Possibly there is something in the library (LB) tag that may work, but I'm wary of going down this road and it also doesn't match the existing usage in the wild.

Again, I'm really not happy we're playing catch up here and considering existing non-compliant data in our definitions, but until we have AND USE validators, we're pretty much stuck in having to be pragmatic in making the spec match the data.

nunofonseca:
LB/RG tags are already used in bulk RNA-seq (for instance, last time that I checked,  BAMs in GDC portal contain these tags). Is there a reason not to use these tags for scRNA-seq too? A pragmatic compromise would be to have the CB tag together with a CA tag as I suggested above.

nunofonseca:
@pmarks @lh3 @jkbonfield Please check the new wording of the CB tag.

jkbonfield:
I think given we have barcode vs identifier intended uses, I'd like to tighten it as much as possible.

Rather than saying "the identifier may include non-nucleotide characters" we could be very specific and say something like "The sequence may optionally be followed by dash ('-') and one or more characters to form an identifier".  This way ACGT, ACGT-1, ACGT-TTCA would all considered as valid while "fred" would not.  This aids parsing and tools that look for the barcode can stop at the first non nucleotide position with safety.

nunofonseca:
@jkbonfield  I made the description more tighten.

SAMtags.tex
```diff
@@ -246,6 +250,39 @@ \subsection{Barcodes}
...
 In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and places a hyphen (`{\tt -}') between the barcodes from the same template. 
 
+\item[BZ:Z:\tagvalue{qualities+}] 
+Phred quality of the (uncorrected) unique molecular identifier sequence in the {\tt OX} tag.
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+The {\tt OX} tags should match the {\tt BZ} tag in length. 
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the quality strings with a space (`{\tt \textvisiblespace}') between the different strings.
+
+\item[CB:Z:\tagvalue{str}]
+
```
yfarjoun: remove extra nl here

SAMtags.tex
```diff
@@ -246,6 +250,39 @@ \subsection{Barcodes}
...
+  Cell identifier based on the (possibly corrected) cellular-barcode
+  sequence ({\tt CR}). The sequence may optionally be followed by a
```
yfarjoun: this is confusing as it seems to imply that CR has possibly corrected cellular-barcode sequence.  How about:

> Cell identifier based on the cellular-barcode sequence ({\tt CR}), with correction optionally applied.

yfarjoun: may optionally be -> may be optionally

SAMtags.tex
```diff
@@ -246,6 +250,39 @@ \subsection{Barcodes}
...
+  dash ('-') and one or more alphanumeric characters to form an
+  identifier. The respective quality scores are (optionally) stored in
```
yfarjoun: I don't think that there needs to be discussion of quality scores here. it's enough to have it in CR.

SAMtags.tex
```diff
@@ -246,6 +250,39 @@ \subsection{Barcodes}
...
+  the {\tt CY} tag. All reads with the same {\tt CB} tag putatively
+  derive from the same cell.
+
+\item[CR:Z:\tagvalue{str}]
+  Cellular barcode.
+  Sequence bases of the cellular-barcode as reported by the sequencer (uncorrected) with the respective quality scores stored in {\tt CY}. All reads with the same {\tt CR} tag putatively derive from the same cell.
```
yfarjoun: Is there room for discussion about what to do when there are multiple barcodes that together comprise of a barcode for a cell (as in UMIs)?

SAMtags.tex
```diff
@@ -246,6 +250,39 @@ \subsection{Barcodes}
...
+  identifier. The respective quality scores are (optionally) stored in
+  the {\tt CY} tag. All reads with the same {\tt CB} tag putatively
+  derive from the same cell.
+
+\item[CR:Z:\tagvalue{str}]
+  Cellular barcode.
+  Sequence bases of the cellular-barcode as reported by the sequencer (uncorrected) with the respective quality scores stored in {\tt CY}. All reads with the same {\tt CR} tag putatively derive from the same cell. 
+
+\item[CY:Z:\tagvalue{qualities+}] 
+Phred quality of the cellular-barcode sequence in the {\tt CB} (or {\tt CR}) tag.
```
yfarjoun: This should only refer to the CR tag, since the CB tag contains an identifier and possibly corrected sequence for which these are not the qualities....

SAMtags.tex
jkbonfield: I think I'd need to see some examples to judge whether this belongs here.  What we have now I think works, but it doesn't mean we can't extend it later on if necessary provided we can do so in a compatible manner.

jkbonfield:
Good in general, with a few trivial tweaks.

We should perhaps try and work out a sensible ordering.  This has changed it, eg moving BZ up after BC instead of being adjacent to OX which BZ is related to.    I think we want all the sample barcode tags together, all the cell barcodes together, followed by all the molecular barcode ones.  I believe this is the original order but with CB, CR and CY inserted after QT and before RX.   (I'd be happy to reorder it at time of merge if needed though if agreed.)

CB should be using `-' style instead of '-' to get proper TeX quotation.

I'm unsure about stating "The respective quality scores are (optionally) stored in the CY tag" for CB given it *may* potentially be error corrected.  I think it's best to just drop that sentence as it's implicit anyway given it's "based" on CR and CR has associated qualities in CY.  Stating it explicitly may cause more confusion than it's worth.

Being nit-picky there are also some white space differences in how CB/CR are defined, but frankly I'd just do a minor reformat to that at the same time as squashing so IMO that's in the remit of SAM maintainer standard "tidy before merge" task.

SAMtags.tex
```diff
@@ -250,29 +250,23 @@ \subsection{Barcodes}
...
 In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes and places a hyphen (`{\tt -}') between the barcodes from the same template. 
 
-\item[BZ:Z:\tagvalue{qualities+}] 
-Phred quality of the (uncorrected) unique molecular identifier sequence in the {\tt OX} tag.
+\item[QT:Z:\tagvalue{qualities}] 
+Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag. 
 Same encoding as {\sf QUAL}, i.e., Phred score + 33.
-The {\tt OX} tags should match the {\tt BZ} tag in length. 
-In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the quality strings with a space (`{\tt \textvisiblespace}') between the different strings.
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the quality strings with spaces (`{\tt \textvisiblespace}') between the different strings from the same template.
```
yfarjoun: should be sample-barcodes, no?

SAMtags.tex
```diff
@@ -250,29 +250,23 @@ \subsection{Barcodes}
...
 
 \item[CB:Z:\tagvalue{str}]
-  Cell identifier based on the (possibly corrected) cellular-barcode
-  sequence ({\tt CR}). The sequence may optionally be followed by a
-  dash ('-') and one or more alphanumeric characters to form an
-  identifier. The respective quality scores are (optionally) stored in
-  the {\tt CY} tag. All reads with the same {\tt CB} tag putatively
-  derive from the same cell.
+Cell identifier based on the cellular-barcode sequence ({\tt CR}), with a correction optionally applied. The sequence may optionally be followed by a dash (`{\tt -}') and one or more alphanumeric characters to form an identifier. All reads with the same {\tt CB} tag putatively derive from the same cell.
+In the case of the cell identifier being based on multiple cellular-barcode sequences the recommended implementation concatenates all the barcodes with a hyphen (`{\tt -}') between the different barcodes.
```
yfarjoun: shouldn't this sentence (about how to concatenate) be in the "Cellular Barcode" part, rather that the elular identifier part?

SAMtags.tex
yfarjoun: this comment still stands

nunofonseca:
@jkbonfield @yfarjoun I tried to address all the points raised by you. Following @yfarjoun  suggestion I included  some text to cover the case where the cell identifier is based on multiple barcodes (e.g., as described in Klein et al, 2015).

SAMtags.tex
```diff
@@ -274,15 +275,31 @@ \subsection{Barcodes}
...
+\item[QX:Z:\tagvalue{qualities+}] 
+Phred quality of the unique molecular identifier sequence in the {\tt RX} tag. 
+Same encoding as {\sf QUAL}, i.e., Phred score + 33.
+The qualities here may have been corrected (Raw bases and qualities can be stored in {\tt OX} and {\tt BZ} respectively.)
+The lengths of the {\tt QX} and the {\tt RX} tags must match. 
+In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the quality strings with a space (`{\tt \textvisiblespace}') between the different strings.
+
 \item[RT:Z:\tagvalue{sequence}]
-Deprecated alternative to {\tt BC} tag originally used at Sanger.
+  Deprecated alternative to {\tt BC} tag originally used at Sanger.
```
yfarjoun: why the extra space?

SAMtags.tex
```diff
@@ -251,21 +255,18 @@ \subsection{Barcodes}
...
 
-\item[RX:Z:\tagvalue{sequence+}]
-Sequence bases from the unique molecular identifier. 
-These could be either corrected or uncorrected. Unlike {\tt MI}, the value may be non-unique in the file. 
-Should be comprised of a sequence of bases. 
-In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes with a hyphen (`{\tt -}') between the different barcodes.
+\item[CB:Z:\tagvalue{str}]
+Cell identifier based on the cellular-barcode sequence ({\tt CR}), with a correction optionally applied. The sequence may optionally be followed by a dash (`{\tt -}') and one or more alphanumeric characters to form an identifier. All reads with the same {\tt CB} tag putatively derive from the same cell.
+In the case of the cell identifier being based on multiple cellular-barcode sequences the recommended implementation concatenates all the barcodes with a hyphen (`{\tt -}') between the different barcodes.
```
yfarjoun: shouldn't the concatenation sentence be replicate in CB?

SAMtags.tex
```diff
@@ -251,21 +255,18 @@ \subsection{Barcodes}
...
 
-If the bases represent corrected bases, the original sequence can be stored in {\tt OX} (similar to {\tt OQ} storing the original qualities of bases.)
+\item[CR:Z:\tagvalue{str}]
+Cellular barcode. Sequence bases of the cellular-barcode as reported by the sequencer (uncorrected) with the respective quality scores stored in {\tt CY}. All reads with the same {\tt CR} tag putatively derive from the same cell. 
 
-\item[QX:Z:\tagvalue{qualities+}] 
-Phred quality of the unique molecular identifier sequence in the {\tt RX} tag. 
+\item[CY:Z:\tagvalue{qualities+}] 
+Phred quality of the cellular-barcode sequence in the {\tt CR} tag.
```
yfarjoun: the barcode might be in the CB, so should end with "...{\tt CR} or {\tt CB} tag."

nunofonseca: As you well pointed out before, "This should only refer to the CR tag, since the CB tag contains an identifier and possibly corrected sequence for which these are not the qualities....", hence it makes sense that CB is not explicitly mentioned in CY.

SAMtags.tex
nunofonseca: Emacs...Fixed now.

SAMtags.tex
nunofonseca: Good point, makes more sense in the CR and CY. Changed the text accordingly.

SAMtags.tex
nunofonseca: Yes, I think so. The issue was already in the tex file before this PR. Shall I include the fix in this PR?

SAMtags.tex
nunofonseca: CY now only refers to the CR tag and CB does not refer CY.

jkbonfield:
The BC (sample barcode seq) concatenates with hyphens while QT (sample barcode qual) concatenates with spaces.  Presumably this was because space isn't a legal quality string so it's unambiguously multiple strings.  Therefore for consistency we should probably have space instead of hyphen for CY (cell barcode qual).  Other than this and minor formatting below I'm happy with the PR.

The order of non CB/CR/CY tags is still subtly different to the original, but I'm OK with that actually.  Eg MI moved up to be the first entry in bunch of molecular barcode tags, instead of half way down.  It's consistent with CB being first and I think I prefer that order anyway.  I just note the change is there incase it's accidental.

To other maintainers, and Nuno if you want to help streamline the process for us, before merging we'd want to squash and also fix some whitespace issues.  Git diff shows lots of trailing whitespace has been added ("[color]\nui=auto\n" in ~/.gitconfig shows this).   It may also be good practice to break multiple sentences on one line to separate lines, for ease of any diff-ing if we edit the text later.  Finally it'll need rebasing which will have a minor conflict to resolve I think (CG has been added since this PR, which will slot in the middle of the CB/CR/CY list).

Finally regarding the wording in BC/QT raised by @yfarjoun - yes it is incorrect and should be changed to sample barcode.  This is a mistake from an earlier commit (9f886dd8), but it's probably best for one of us to make the edit as a subsequent commit outside of this PR.

nunofonseca:
@jkbonfield  The change in the order was not accidental :)  - I tried to order the tags (sample, cell and molecular).  As suggested, for consistency I changed  the concatenation of multiple barcodes in CY to space instead of hyphen.   There is only one sentence per line now. I'll try to look at the trailing  white spaces later.

SAMtags.tex
jkbonfield: I agree with Nuno here - it should be CR only.  CB may also happen to be the same as CR, but let's not muddy the waters by linking CY to CR directly instead of via the correct indirect relationship.

jkbonfield:
Good to go from me.  There are still a couple tiny whitespace differences, but they can be fixed during merge by us anyway (if we deem them incorrect).

@yfarjoun, @jmarshall?

SAMtags.tex
```diff
@@ -251,20 +255,20 @@ \subsection{Barcodes}
...
-Should be comprised of a sequence of bases. 
-In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes with a hyphen (`{\tt -}') between the different barcodes.
+\item[CB:Z:\tagvalue{str}]
+Cell identifier based on the cellular-barcode sequence ({\tt CR}), with a correction optionally applied. The sequence may optionally be followed by a dash (`{\tt -}') and one or more alphanumeric characters to form an identifier.
+All reads with the same {\tt CB} tag putatively derive from the same cell.
 
-If the bases represent corrected bases, the original sequence can be stored in {\tt OX} (similar to {\tt OQ} storing the original qualities of bases.)
+\item[CR:Z:\tagvalue{str}]
+Cellular barcode. Sequence bases of the cellular-barcode as reported by the sequencer (uncorrected) with the respective quality scores stored in {\tt CY}.
+All reads with the same {\tt CR} tag putatively derive from the same cell.
```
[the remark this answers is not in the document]
nunofonseca: Hi! Thank you for the feedback. Can't the same observation about the sequencing errors be made about other barcodes?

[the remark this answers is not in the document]
jkbonfield: My interpretation of @yfarjoun's comments is that it isn't a 1:1 relationship, rather than it being due to errors.  Ie the same barcode can be used in multiple cells, but the reverse is not true (in each cell only one barcode will be used).

I'm not processing 10X data myself so I can't verify this, but having spoken with those that are right now the number of barcodes is typically much higher than the number of cells being used, but a barcode being shared is still possible (and likely given the birthday paradox).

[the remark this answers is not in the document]
pmarks: @jkbonfield  - correct. In the latest single-cell RNA-seq setup there are 737K possible barcodes, which are sample 'with replacement' into 100K droplets. Then a user-defined number of cells are populated into those droplets (we recommend < 10K cells). So you can have two cells sharing the same barcode because:
1) they ended up in the same droplet by chance
2) they ended up in different droplets, but the barcode in those droplets was the same by chance.
So the user really needs to understand those parameters to compute the probability that the CR tag is referencing a unique cell. Some platforms also have the ability to image the droplets and filter out barcodes that contain > 1 cell.  Various other schemes are being worked to identify doublets.  Here's a slightly more verbose proposal:
`All reads with the same {\tt CR} tag a likely to derive from the same cell. The probability that a given {\tt CR} value contains reads from multiple cells may could be substantial (>10%) and is technology and experiment dependent. All reads from any given cell are likely to have the same {\tt CR} tag, but not due to sequencing errors or technical artifacts in library preparation`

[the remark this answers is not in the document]
jkbonfield: Thanks @pmarks for looking over this PR and commenting.

I think the new description does rather muddy things a bit regarding more than one barcode for the same cell.  I understand the reporting of this barcode (ie sequencing) may cause differences, but it's still the same original barcode.   I'm not sure what the "technical artifacts" would be?  Are we just talking about things going wrong or mistakes?  If so it doesn't really belong in the spec description.

yfarjoun:
@nunofonseca, I"m not sure if you are still waiting for an answer. You might be right about the wording being inaccurate in other barcode sections. We will support a clarifying PR there as well, but let's not get ahead of ourselves. In this case, the only "wiggle room" in you current wording is the word "putatively", which doesn't help resolve the non one-to-one correspondence between barcode and cell.

SAMtags.tex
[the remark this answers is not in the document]
pmarks: @jkbonfield most protocols have some level of 'barcode exchange' that causes the cell-barcodes to 'hop' between library molecules (analogous to index-hopping on patterned flowcells). In a simple model, each read comes from the 'correct' cell with probability 1-e, and comes randomly from the whole pool of molecules with probability e.  This is what I was thinking about when I wrote 'technical artifacts'.  That's probably just a generic consideration that doesn't belong in the spec.

yfarjoun:
I still think it would be better to state that "Sequencing errors not withstanding, all reads from a single cell are expected to have the same barcode." which is _less_ dependent on the details of the protocol & experiment (how many barcodes there are, for example...and how many cells were sequenced...) Though it is still dependent on having **one** cell in each droplet, an assumption that could be false.

nunofonseca:
@yfarjoun I fully agree on your comment. Well, almost fully...I still don't understand why the description of this particular tag should include a remark about sequencing errors while the same remark is absent from other tags associated to barcodes.  Since sequencing errors may affect other barcodes, shouldn't  such a remark about sequencing errors also be included in the description of other tags? or, alternatively, only once in #293?

jkbonfield:
I think it's simply there for clarification.  Arguably other barcodes where they explain how they fit together also need similar text.

@yfarjoun I think you're wrong when you say "though it is still dependent on having one cell in each droplet, an assumption that could be false".  If we have two cells in a droplet, then all reads from a single cell do still have the same barcode, it's just that the reverse is not true - which is something which was already the case irrespective of this multi-cell droplet issue.

yfarjoun:
@jkbonfield I agree. I corrected my comment.

yfarjoun:
@nunofonseca, as I said perviously, if there is text elsewhere in the docs that can be made more accurate, I'm happy to consider a PR. 

In Cell Barcode description I think that either we state the full sentence, with the clarification, or we drop it entirely. The way it is currently stated is misleading.

jmarshall:
At this point, the PR consists of 3 one-line additions to the §1 table, the addition of CB, CR, and CY to §1.3 Barcodes, and a slight reordering of the entries in §1.3 (though the latter doesn't make it any easier to see the (non-)changes :smile:).

If the users of these new tags are happy, then this maintainer is happy too. Some questions arising from how the text differs from that of other tags in this section:

* Does CB not do the concatenate-multiple-items-with-a-hyphen thing? (It makes sense if it does not, I'm just confirming.)

* For CB, does “with a correction optionally applied” mean that the sequence part before the hyphen might be different from CR because errors have been corrected within it? This text could be clearer — at present it can be read as though the `-alphanumeric` suffix is the “correction”.

* For CR, the `\item` would usefully be explicit: `\item[CR:Z:\tagvalue{sequence+}]`.

* In the CR description, does “with the respective quality scores stored in CY” mean that if you have CR then you **must** have CY too? If not, let's phrase this the same way as the other barcode/quality tag pairs, or at least put the word “optionally” in there.

jmarshall:
> Finally regarding the wording in BC/QT raised by @yfarjoun - yes it is incorrect and should be changed to sample barcode.

There is indeed an existing error there. I will fix this separately after this and probably also PR #293 have landed, via a PR proposing lifting the “If multiple barcodes, the recommended implementation …” sentence to a single paragraph at the end of the §1.3 introduction instead of repeating it on every tag description.

pmarks:
Current state looks good to me. Re @jkbonfield's comments:
1.  If `CR` allows "-" concatenated subcomponents, then`CB` probably should as well, although we don't use that.
2. Yes, “with a correction optionally applied” means that the sequence before the dash has been changed. Any wording tweak there would be great

nunofonseca:
Thank you  @jkbonfield , @pmarks @jmarshall and @yfarjoun for the comments. I tried to address the comments in this latest commit.

SAMtags.tex
```diff
@@ -246,25 +250,26 @@ \subsection{Barcodes}
...
-In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the barcodes with a hyphen (`{\tt -}') between the different barcodes.
+\item[CB:Z:\tagvalue{str}]
+Cell identifier based on the cellular-barcode sequence ({\tt CR}) that may be optionally i) error corrected and ii) followed by a dash (`{\tt -}') and one or more alphanumeric characters to form an identifier. 
+In the case of the cellular barcode ({\tt CR}) being based on multiple barcode sequences the recommended implementation concatenates all the (corrected or uncorrected) barcodes with a hyphen (`{\tt -}') between the different barcodes. 
+Notwithstanding errors (e.g., sequencing), all reads from a single cell are expected to have the same {\tt CB} tag.
 
-If the bases represent corrected bases, the original sequence can be stored in {\tt OX} (similar to {\tt OQ} storing the original qualities of bases.)
+\item[CR:Z:\tagvalue{sequence+}]
+Cellular barcode. Sequence bases of the cellular-barcode as reported by the sequencer (uncorrected) with the respective quality scores (optionally) stored in {\tt CY}.
```
jkbonfield: Brackets are probably unnecessary.  Eg "Uncorrected sequence bases... quality scores optionally stored..".  However it's tiny nit-picking and I'm OK with it either way.

jmarshall:
That's answered all my questions, and I'm :+1: with this now (with some minor editorial stuff and wordsmithing that we can fix up when squashing and merging).

The existing order of tags in this section is

    BC QT   …   …    [RX QX] MI OX BZ RT

This PR changes it to

    BC QT [CB CR CY] MI OX BZ [RX QX] RT

So CB/CR/CY are inserted and RX QX are moved to the bottom of the UMI group. I agree with James that having MI first is good, but was there a strong reason to move the maybe-corrected ones below the raw ones (which is the opposite of your new tags)? I don't mind either way (and I didn't see a careful ordering mentioned in PR #200), but we should merge it as separate reordering and adding commits to keep what's happening clear.

jmarshall:
🎉 

Who wants to merge it? (I'm volunteering, as I have some tweaks in mind…)

yfarjoun:
go ahead!

jmarshall:
Now squashed, rebased, and merged as c8b9990d8e024d725eb0911a9e7bd1fb41e92f60.

I took the liberty of somewhat rewriting the CB description for clarity; see 1ddb7fad41f39d93f089a8e0c2655c40a809b247, which I believe retains @nunofonseca's intentions but is hopefully clearer.

Thanks to all for the protracted review and responses to it. I for one think I finally mostly understand what's going on with all these barcodes now! (And roll on #293 and then my followup [mentioned in this comment](https://github.com/samtools/hts-specs/pull/262#issuecomment-385677763).)

PR #265, opened 2017-12-05
Fixing typos in BCF genotype example table

### LINKED ISSUE #264: Typos in the example genotypes for BCF?
I recently read the example tables on page 32 of the vcf spec and was confused by some of the `0x04`s that encode the `0` genotype. Are these typos? (lines 3 & 9)

yfarjoun:
fixes #264 

Also took the opportunity to unify the way longish hex data is displayed in that table. splitting the data into nibbles and dropping the `0x` except in the very first nibble.

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -1525,8 +1525,10 @@ BCF2 site information encoding
 \begin{tabular}{ | p{2cm} | p{2.5cm} | p{9.5cm} | } \hline
 Field & Type & Notes \\ \hline
 fmt\_key & typed int & Format key as an offset into the dictionary \\ \hline
-fmt\_type & uint8\_t+ & Typing byte of each individual value, possibly followed by a typed int for the vector length.  In effect this is the same as the typing value for a single vector, but for genotype values it appears only once before the array of genotype field values \\ \hline
-fmt\_values	(by fmt type) & Array of values & The information of each individual is concatenated in the vector.  Every value is of the same fmt type.  Variable-length vectors are padded with END\_OF\_VECTOR values; a string is stored as a vector of char \\  \hline
+fmt\_type & uint8\_t+ & Typing byte of each individual value, possibly followed by a typed int for the vector length.  
+In effect this is the same as the typing value for a single vector, but for genotype values it appears only once before the array of genotype field values \\ \hline
+fmt\_values	(by fmt type) & Array of values & The information of each individual is concatenated in the vector.  Every value is of the same fmt type.  
+Variable-length vectors are padded with END\_OF\_VECTOR values; a string is stored as a vector of char \\  \hline
 \end{tabular}
 \normalsize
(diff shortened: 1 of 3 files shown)


yfarjoun:
@pd3 I'm not sure I follow your comment. Unless I'm misreading the spec, the missing value is the way to encode vectors of different length...see Vectors of mixed length under 6.3.3

pd3:
Ah, did not notice. That section then needs to be updated too, MISSING must be replaced with END-OF-VECTOR byte. 

Imagine a mixture of diploid and haploid genotypes, all missing. The end-of-vector byte is needed to distinguish between missing diploid `./.` and haploid `.` genotypes. This is what is implemented in htslib.

yfarjoun:
This is very different from what the spec says...and doesn't seem to fit into this PR. It sounds like htslib has implemented something that makes sense...but it's not what's in the spec. This PR was fixing a typo.... 

Specifically, the spec states:

> Note that this means that it’s illegal to encode a vector VCF field with missing values; the BCF2 codec should signal an error in this case.

Also, for genotypes, it states:

> ... here allele is set to -1 if the allele in GT is a dot ‘.’ ...

which seems to solve your usecase. 

Finally, I couldn't find the term END-OF-VECTOR anywhere in the text...

pd3:
I think it's that the specification was not updated when the end-of-vector byte was introduced later. 

The end-of-vector byte is introduced at the begining of 6.3.3 in the Integers and Floats paragraphs, together with the missing byte. 

The -1 works only for the special case of GTs, imagine other fields that depends on ploidy, such as PL.

pd3:
Yes, this appeared first in the VCFv4.3 version of the document.

pd3:
Yes, that makes sense. Re-read and did not spot any issues. +1

VCFv4.3.tex
```diff
@@ -1712,56 +1716,52 @@ \subsubsection{Type encoding}
...
-The BCF2 encoder / decoder must automatically add and remove these MISSING values from the vectors.
+\textbf{Vectors of mixed length} --- In some cases genotype fields may be vectors whose length differs among samples.  For example, some CNV call sets encode different numbers of genotype likelihoods for each sample, given the large number of potential copy number states, rather padding all samples to have the same number of fields.  For example, one sample could have CN0:0,CN1:10 and another CN0:0,CN1:10,CN2:10.  In the situation when a genotype field contain vector values of different lengths, these are represented in BCF2 by a vector of the maximum length per sample, with all values in the each vector aligned to the left, and END\_OF\_VECTOR values assigned to all values not present in the original vector.  The BCF2 encoder / decoder must automatically add and remove these END\_OF\_VECTOR values from the vectors. Note that the use of END\_OF\_VECTOR means that it is legal to encode a vector VCF field with MISSING values.
```
cyenyxe: This paragraph needs to be changed to the one-sentence-per-line style.

PR #266, opened 2017-12-06
SV Improvements - Stage 1

thefferon:
Improvements in support of structural variation representation. Stage 1: Minor improvements & edits that should not break any downstream analysis tools.


VCFv4.3.tex
```diff
@@ -513,28 +529,31 @@ \section{INFO keys used for structural variants}
...
 \end{itemize}
 
+
 \footnotesize
 \begin{verbatim}
 ##INFO=<ID=SVLEN,Number=.,Type=Integer,Description="Difference in length between REF and ALT alleles">
 \end{verbatim}
 \normalsize
-One value for each ALT allele. Longer ALT alleles (e.g. insertions) have positive values, shorter ALT alleles (e.g. deletions) have negative values.
+SVLEN indicates the relative size of the ALT allele compared to the REF allele. If ALT is longer than REF (e.g. insertions), SVLEN must be positive; if ALT is shorter than REF (e.g. deletions), SVLEN must be negative. One SVLEN value for each ALT allele.
```
d-cameron: Perfect inversions will have SVLEN=0. The size of the inversion can be determined by the POS field and the END attribute.

VCFv4.3.tex
```diff
@@ -33,15 +33,15 @@
 \newpage
 
 \section{The VCF specification}
-VCF is a text file format (most likely stored in a compressed manner). 
+VCF is a text file format (most likely stored in a compressed manner).
```
d-cameron: If it isn't too much trouble, would you be able to revert the whitespace changes in this PR so only actual content changes show up in the diff?

VCFv4.3.tex
```diff
@@ -185,31 +185,47 @@ \subsubsection{Individual format field format}
...
+DEL  &  Deletion relative to the reference \\
+INS  &  Insertion relative to the reference \\
+DUP  &  Region of elevated copy number relative to the reference \\
+INV  &  Inversion of reference sequence \\
+CNV  &  Copy number variable region (multiallelic) \\
+BND  &  Breakend \\
+\end{tabular}
+\newline
+
+
```
d-cameron: The list of symbolic alleles that represent structural variants is a closed set. Any symbolic alleles other than the first level types listed above are to considered to be non-structural symbolic alleles.

VCFv4.3.tex
```diff
@@ -185,31 +185,47 @@ \subsubsection{Individual format field format}
...
+INS:ME:ALU  &  Insertion of Alu element relative to the reference \\
+INS:ME:LINE1  &  Insertion of LINE1 element relative to the reference \\
+INS:ME:SVA  &  Insertion of SVA element relative to the reference \\
+INS:ME:HERV  &  Insertion of HERV element relative to the reference \\
+\\
+DUP:TANDEM  &  Tandem duplication, original orientation \\
+DUP:TANDEM:INV:LEFT  &  Tandem duplication to left of original, inverted orientation \\
+DUP:TANDEM:INV:RIGHT  &  Tandem duplication to right of original, inverted orientation \\
+\end{tabular}\newline
+
```
d-cameron: Variants should be written using the most precise type that can be determined by the variant caller. For example, a LINE1 insertion would be specified as a DUP of the originating LINE1 element if the insertion site cannot not be determined, but INS:ME:LINE1 at the insertion site if the site can be identified.

Note that the DUP type is not restricted to tandem duplications and that the DUP subtypes have different breakpoints than their parent type.

d-cameron: Variant subtypes are an open set and is it valid for a variant caller to define a subtype not specified in the above list. Such subtypes are considered to be instances of the first reserved subtype these types with breakpoint and copy number semantics matching the reversed type. For example, INS:ME:PELEMENT is a valid caller-defined subtype, but INV:FOLDBACK is not since foldback inversions do not result in the same breakpoint and copy number changes as simple inversion.

VCFv4.3.tex
```diff
@@ -185,31 +185,47 @@ \subsubsection{Individual format field format}
...
+\\
+INS:ME  &  Insertion of mobile element relative to the reference \\
+INS:ME:ALU  &  Insertion of Alu element relative to the reference \\
+INS:ME:LINE1  &  Insertion of LINE1 element relative to the reference \\
+INS:ME:SVA  &  Insertion of SVA element relative to the reference \\
+INS:ME:HERV  &  Insertion of HERV element relative to the reference \\
+\\
+DUP:TANDEM  &  Tandem duplication, original orientation \\
+DUP:TANDEM:INV:LEFT  &  Tandem duplication to left of original, inverted orientation \\
+DUP:TANDEM:INV:RIGHT  &  Tandem duplication to right of original, inverted orientation \\
```
d-cameron: Should we collapse INV:LEFT and INV:RIGHT to prevent a caller writing the ambiguous subtype of DUP:TANDEM:INV?

VCFv4.3.tex
Lenbok: Just FYI (and I don't know why it isn't exposed better), if you append ?w=1 to the github URL, it will hide whitespace-only differences.

thefferon: The whitespace changes were automatically introduced by my text editor - I did not introduce them intentionally. I will revert them, and reintroduce in a different PR.

thefferon: Whitespace diffs should be fixed now.

VCFv4.3.tex
thefferon: Added.

VCFv4.3.tex
thefferon: Changed to INV-LEFT and INV-RIGHT, respectively. OK?

VCFv4.3.tex
thefferon: Added.

thefferon: Added.

VCFv4.3.tex
thefferon: Added.

thefferon:
@d-cameron , please take a fresh look at changes, with '?w=1' appended to URL to hide whitespace differences. Preconfigured link:

https://github.com/samtools/hts-specs/pull/266/files?w=1


I am working on examples, including those you included in our Confluence discussion; please send me, or append, the DUP examples you mentioned (after reviewing latest changes to ALT in this PR).

Thanks,
Tim

VCFv4.3.tex
```diff
@@ -185,31 +185,53 @@ \subsubsection{Individual format field format}
...
-  \item CNV Copy number variable region (may be both deletion and duplication)
-\end{itemize}
-
-The CNV category should not be used when a more specific category can be applied. Reserved subtypes include:
-\begin{itemize}
-  \item DUP:TANDEM Tandem duplication
-  \item DEL:ME Deletion of mobile element relative to the reference
-  \item INS:ME Insertion of a mobile element relative to the reference
-\end{itemize}
+In symbolic alternate alleles for structural variants, the ID field indicates the type of structural variant, and can consist of a colon-separated list of types and subtypes. Every symbolic alternate allele that appears in a VCF file must be defined in the file pragmas. ID values are case sensitive strings and must not contain whitespace or angle brackets. The first level type must be one of the following:\newline
```
d-cameron: > and can consist of a colon-separated list of types and subtypes
 Replace with "and consists of a colon-separated list of a first level type and optional subtypes". Using "can" implies that it "can" also be something else that we haven't defined here.

d-cameron: >file pragmas
Replace with "VCF header" as pragma is a C specific term.

d-cameron:
Are we planning to fix the cardinality of SV fields in this PR? About half the SV fields assume one SV per loci, and the other half don't. See #133

d-cameron:
@thefferon inverted duplication example added to confluence

VCFv4.3.tex
```diff
@@ -536,29 +565,36 @@ \section{INFO keys used for structural variants}
 ##INFO=<ID=END,Number=1,Type=Integer,Description="End position of the variant described in this record">
 \end{verbatim}
 \normalsize
-For precise variants, END is POS + length of REF allele - 1, and the for imprecise variants the corresponding best estimate.
+For precise variants, END is POS + length of REF allele - 1, and the for imprecise variants, END is the corresponding best estimate.
```
cwhelan: I'm not sure if this PR is still active, but it might be good to further clarify the use of the `END` key. I've noticed a few cases where, for interchromosomal BNDs, tools are storing the second chromosome in the `CHR2` key and the position on the second chromosome in the `END` key. I think that Delly might do this, or did it in a previous version, as do VCFs produced by the current version of the Talkowski Lab's SV pipeline. This breaks other tools which want to produce the interval for the variant by looking at `END`. In particular htsjdk fails on these files. If the variant is an imprecise BND, I can see how this interpretation of the spec arose, but I think it's worth a clarification on whether this should be allowed or not. Personally I think it should not be, and the spec should clarify that the value in `END` should always be on the same contig as `CHR` and come after `POS`. 

If it's a topic worthy of its own issue, I could create a separate ticket/PR to try to address it.

Also, there's a typo in this sentence both the current version and the proposed change (`and the for`).

pd3: I support this. Ideally there should be only one meaning of the END tag and it makes sense to restrict it to the same chromosome. That was also the original intention.

d-cameron: The spec already has `CIPOS` for `BND` events. The `END` clarification was intended to indicate what happens to the RHS breakend for symbolic non-bnd events. Technically speaking, even with this change you still can't encode an imprecise `<DEL>` with different bounds intervals for the LHS and RHS breakends and have to use `BND`.

We should indeed be explicit in disallowing `END` for `BND` events.

VCFv4.3.tex
```diff
@@ -171,30 +171,58 @@ \subsubsection{Individual format field format}
...
+\end{tabular}
+\newline
+
+Variants should be written using the most precise type that can be determined by the variant caller.
+For example, if the insertion site of a new copy of a LINE1 element cannot be determined, it would be specified as a DUP of the originating LINE1 element.
+However if the new insertion site can be identified, the variant should be specified as INS:ME:LINE1 at the insertion site.
+
+Note that the DUP type is not restricted to tandem duplications and that the DUP subtypes have different breakpoints than their parent type.
+
+Variant subtypes are an open set and is it valid for a variant caller to define a subtype not specified in the above list. Such subtypes are considered to be instances of the first reserved subtype, with breakpoint and copy number semantics matching the reserved type. For example, INS:ME:PELEMENT is a valid caller-defined subtype, but INV:FOLDBACK is not since foldback inversions do not result in the same breakpoint and copy number changes as a simple inversion.
```
cyenyxe: I don't think we can make the last part of the paragraph work as part of the specification, because it is an open set with semantics involved. So we can mention what they are considered to be, but not provide any examples.

```suggestion
Variant subtypes are an open set and is it valid for a variant caller to define a subtype not specified in the above list. Such subtypes are considered to be instances of the first reserved subtype, with breakpoint and copy number semantics matching the reserved type.
```

d-cameron: The last paragraph is based on examples I've seen in the real world. Why wouldn't we provide examples of what not to do? Would a wording change be sufficient? For example, an explaination of why a caller should not create INV:FOLDBACK or DUP:INVERTED sub-types and should report these events using BND notation.

I definitely want something in the specs that explicitly states that a caller using the above notations will be non-compliant/be calling a different variant than what they think they're calling.

d-cameron:
Sorry I haven't been able to make the teleconference calls for some time. @cyenyxe is there movement on this PR/VCF 4.3?

PR #268, opened 2017-12-12
Fix RFC3986 encoding for "%"

nathanweeks:
RFC3986 encoding for "%" is "%25" (rather than "%2C", which encodes ",").


jkbonfield:
Good spot and totally non-contentious, so merging.

Issue #281, opened 2018-01-12
Developing a StrictSam standard


yfarjoun:
As discussed on the last and penultimate call, there's a desire to standardize a "Strict Sam" specification. 

There's already a stub of an implementation here: https://github.com/d-cameron/hts-specs/blob/strict/SAMstrict.tex

@d-cameron 

I didn't know how to respond to your request that we take a look at your initial branch...so I am opening this issue for a discussion...if you have a better idea, I'm game.

yfarjoun:
I think that overall the start is good. I think that it is overzealous at places (for example forbidding `@SO` and `@GO` together, and also the requirement that queryname be used as a `@GO` and not a `@SO`, among others...)  but that should be a discussion when a PR is prepared.


I think it would be good to agree on an identifier for each "rule" some thing like 

> Every SAM record with an non-* RNAME must have a corresponding @SQ header with matching SN. (NON_STAR_RNAME_MUST_HAVE_SN_IN_HEADER)

So that 
- validation programs can log violations in a way that is then searchable in the document, and 
- there's a relatively standard way to describe which standards are being checked by a validation program, and 
- There's the possiblity of allowing validation programs to overlook certain violations (by specifying which rules can be ignored for a given input) 

Also, in the call, we discussed tests. I'm wondering if we should have SAM files (as individual files!) that can be used by validation program authors to check their programs. It should be clear, for each SAM file which validations are expected to fail (if any)

jkbonfield:
I'd like to see levels of strictness.  Something like INVALID, NON_COMPLIANT and INCOMPLETE.

There are clearly some things which we can define as broken / invalid as it's basically nonsensical.

- Mismatching flags between pairs (mate is reversed vs not, etc)
- Mismatching sequence length and cigar length
- Mismatching sequence and quality lengths
- etc

Non-compliance is examples of auxiliary tags in the reserved address space.   Sadly this is *very* common, hence why it should be a separate category.   Similarly there is some looseness around the definition of TLEN; the specification is quite clear, but a *huge* portion of commonly used programs do something subtly different.    While technically these are failures, if we label all these as invalid then we lose discrimination power as most data in ENA is now labelled as invalid - not very useful in a diagnostic tool.

For incomplete data we have valid SAM, but one that may perhaps cause problems for fragile tools. For example

- A read is labelled as paired, but the pair is not in the file due to filtering.
- A read has no quality string ("*").
- Supplementary reads exist but the SA tags haven't been set to identify them.
[...]

PR #283, opened 2018-02-05
SAM/VCF Strict draft

d-cameron:
PR for discussing proposed SAM Strict rule changes. Work in progress continuing from #281 discussion.

### DIFF
--- a/SAMstrict.tex
+++ b/SAMstrict.tex
@@ -1,0 +1,397 @@
+\documentclass[10pt]{article}
+\usepackage[margin=1in]{geometry}
+\usepackage{longtable}
+\usepackage[pdfborder={0 0 0},hyperfootnotes=false]{hyperref}
+\usepackage[title]{appendix}
+
+\newcommand{\rulename}[1]{\tt #1}
+\newcommand{\rulecategory}[1]{\tt #1}
+\newcommand{\samrule}{\tt SAM}
+\newcommand{\vonefive}{\tt v1.5}
+\newcommand{\vonefivebestpractice}{\tt SAMv1.5 best practice}
+\newcommand{\vcffourthree}{\tt VCFv4.3}
+% #1: error message
+% #2: rule description
+% #3: categories
+\newcommand{\samstrictrule}[3]{
+	\paragraph{} #3
+	% error message formatting
+	{\tt #1}
+	% 
+	#2
+}
+% #1: header
+% #2: categories
+\newcommand{\headerrequired}[2]{
+	\samstrictrule{Missing #1 header}{A #1 header must be present.}{#2}
+}
+\newcommand{\headerunique}[2]{
+	\samstrictrule{Only one #1 header may be present}{Multiple #1 headers must not be present.}{#2}
+}
+% #1: header
+% #2: tag
+% #3: categories
+\newcommand{\headertagrequired}[3]{
+	\samstrictrule{Missing #1 header #2 tag}{The #1 header #2 must be present.}{#3}
+}
+% #1: header
+% #2: tag
+% #4: categories
+\newcommand{\headertagunique}[3]{
+	\samstrictrule{Duplicate #1 header #2 tags.}{Each #1 header #2 tags must be unique.}{#2}
+}
+% #1: header
+% #2: tag
+% #3: regex
+% #4: categories
+\newcommand{\headertagregex}[4]{
+	\samstrictrule{Malformed #1 header #2 tag}{The #1 header #2 tag must conform to the regex {\tt #3}}{#4}
+}
+% #1: header
+% #2: tag
+% #3: possible tag values
+% #4: categories
+\newcommand{\headertagvalues}[4]{
+	\samstrictrule{Malformed #1 header #2 tag}{The #1 header #2 tag must contain one of #3}{#4}
+}
+% #1: header
+% #2: tag
+% #3: categories
+\newcommand{\headertagmatchsamspecs}[3]{
+	\samstrictrule{Malformed #1 header #2 tag}{The #1 header #2 tag must match the format defined defined in the SAM specifications.}{#3}
+}
+
+\begin{document}
+
+\input{SAMstrict.ver}
+\title{SAM Strict Specification}
+\author{Daniel L Cameron}
+\date{\headdate}
+\maketitle
+\begin{quote}\small
+The master version of this document can be found at
+\url{https://github.com/samtools/hts-specs}.\\
+This printing is version~\commitdesc\ from that repository,
+last modified on the date shown above.
+\end{quote}
+\vspace*{1em}
+
+\noindent
+This document is a companion to the {\sl Sequence Alignment/Map Format
+Specification} that defines the SAM file format.\footnote{See
+\href{http://samtools.github.io/hts-specs/SAMv1.pdf}{\tt SAMv1.pdf} at \url{https://github.com/samtools/hts-specs}.}
+The SAM file format defines the syntax required for a file to be
+a valid SAM file. It does not require such files to be semantically
+valid and internally consistent.
+This document describes a set of additional semantic restrictions
+for which the subset of syntactically valid SAM files that comply
+with these restrictions can be described as \textit{SAM strict
+compliant}.
+
+\renewcommand{\abstractname}{Introduction}
+\begin{abstract}
+
+The SAM specifications have been instrumental in standardising
+the file formats used for sequence alignment. A large ecosystem of
+bioinformatics tools is now capable of reading and/or writing
+SAM files. Unfortunately, many tools that read SAM files are tightly
+coupled to a particular upstream tool 
+and fail to correctly execute on valid SAM files written by other
+tools. In part, this is due to the lack of semantic restrictions
+inherent in the SAM file format. A syntactically valid SAM file
+can be both internally inconsistent and semantically nonsensical.
+
+The purpose of this document is to provide a baseline of semantic
+validity for which tools should comply with when outputing SAM
+files, and tools which input SAM files can safely assume when
+they require input files to be \textit{SAM strict compliant}.
+
+\end{abstract}
+
+\section{Headers}
+
+{paragraph}
+
+The first segment is is considered to be the "next" segment of the final segment in a template as per the SAM specifications.
+
+
+\section{Headers}
+\samstrictrule{Undefined reserved header present}{Upper-case header record type codes are not defined in the SAM specifications must not be used.}
+\samstrictrule{Undefined header tag present}{Upper-case header tags not defined in the SAM specifications must not be used.}
+\samstrictrule{Tag present as both lowercase and uppercase}{A file should not contain the same tag in both upper-case and lowercase format. See the SAM specifications header tags best practice footnote.}
+\samstrictrule{Malformed header}{Header lines with conform to either the {\tt
+  /\char94@[A-Z][A-Z](\char92t[A-Za-z][A-Za-z0-9]:[
+  -\char126]+)+\$/} or {\tt /\char94@CO\char92t.*/} or {\tt /\char94@CO\char92t.*/} regex.}{\samrule}
+
+\subsection{HD}
+\headerrequired{HD}{\vonefivebestpractice}
+\headerunique{HD}
+\headertagrequired{HD}{VN}{\samrule}
+\samstrictrule{File does not start with HD header.}{The first header defined must be the HD header.}{\vonefivebestpractice}
+\headertagregex{HD}{VN}{/\char94[0-9]+\char92.[0-9]+\$/}{\samrule}
+\samstrictrule{Unknown SAM version}{The HD header VN tag version number must match a published version of the SAM specifications.}
+\headertagvalues{HD}{SO}{{\tt unknown}, {\tt unsorted}, {\tt queryname} and {\tt coordinate}}{\samrule}
+\headertagvalues{HD}{GO}{{\tt none}, {\tt query}, {\tt reference}}{\samrule}
+\samstrictrule{Inconsistent HD header SO and GO tags}{The record orderings defined in the HD header SO and GO tags must be consistent}
+
+\subsection{SQ}
+\samstrictrule{Missing SQ header}{The SQ header must be present if any reads have been mapped.}{\vonefivebestpractice}
+\headertagrequired{SQ}{SN}{\samrule}
+\headertagregex{SQ}{SN}{[!-)+-\char60\char62-\char126][!-\char126]*}{\samrule}
+\headertagunique{SQ}{SN}
+\headertagrequired{SQ}{LN}{\samrule}
+\samstrictrule{Malformed SQ header LN tag}{The SQ header LN tag value must be an integer.}{\samrule}
+\samstrictrule{Unsupported reference sequence length}{The SQ header LN tag value must greater than zero and less than 2147483648}{\samrule}
+\headertagrequired{SQ}{M5}{\samrule}
+\headertagmatchsamspecs{SQ}{AH}{\samrule}
+\samstrictrule{Alternate locus references unknown reference sequence name}{Sequence names in the SQ header AH tag must match a SQ header SN reference sequence name.}{\samrule}
+\headertagmatchsamspecs{SQ}{AN}{\samrule}
+\samstrictrule{Duplicate alternative reference sequence names.}{Alternative reference sequence names defined in SQ header AN tags must be unique. A single tag cannot define cannot define the same alternative reference sequence name multiple times and multiple SQ headers cannot define the same alternative reference sequence name.}{\samrule}
+\samstrictrule{Invalid alternative reference sequence names.}{Sequence names in the SQ header AN tag must not match any SQ header SN reference sequence names. }{\samrule}
+\headertagregex{SQ}{M5}{[a-f0-9]\{32\}}
+
+\subsection{RG}
+\headertagrequired{RQ}{ID}{\samrule}
+\headertagunique{RQ}{ID}{\samrule}
+\samstrictrule{RQ header DT tag is not ISO8601}{RQ header DT tag must contain a valid date in ISO8601 format}{\samrule}
+\headertagregex{RQ}{FO}{/\char92*|[ACMGRSVTWYHKDBN]+/}{\samrule}
+\samstrictrule{Malformed RQ header PI tag}{The RQ header PI tag value must be a floating point value.}{\samrule}
+\headertagvalues{RG}{PL}{{\tt CAPILLARY}, {\tt LS454}, {\tt ILLUMINA}, {\tt SOLID}, {\tt HELICOS}, {\tt IONTORRENT}, {\tt ONT}, and {\tt PACBIO}}{\samrule}
+\samstrictrule{Invalid RG program group}{The RG header PG tag must contain one of the program groups specified in an ID tag of a PG header.}
+
+\subsection{PG}
+\headertagrequired{PG}{ID}{\samrule}
+\headertagunique{PG}{ID}{\samrule}
+\samstrictrule{Invalid PG header PP tag}{The PG header PP tag must contain one of the program groups specified in an ID tag of a PG header.}
+
+\section{General}
+
+\subsection{File Format}
+
+\samstrictrule{File is not UTF-8}{The file must use UTF-8 encoding.}{\samrule}
+\samstrictrule{Inconsistent line terminators}{All lines must be separated with the same new line character\(s\).}
+\samstrictrule{Malformed floating point value}{All floating point values must conform to the regex {\tt [-+]?[0-9]*\char92.?[0-9]+([eE][-+]?[0-9]+)?}}
+\samstrictrule{Malformed integer value}{All integer values must conform to the regex {\tt [-+]?[0-9]+}}
+
+\subsection{Ordering}
+
+\samstrictrule{Record ordering does not match HD header SO tag}{The order of records must be consistent with the HD header SO tag}
+\samstrictrule{Record ordering does not match HD header GO tag}{The order of records must be consistent with the HD header GO tag}
+\samstrictrule{Orphaned unmapped read}{If a read is unmapped, RNAME and POS must either be * and 0, or the RNAME and POS of another read from the same template.}
+ 
+\section{Records}
+
+\subsection{QNAME}
+\samstrictrule{Empty QNAME}{QNAME fields must be at least one character in length}{\samrule}
+\samstrictrule{QNAME too long}{QNAME fields must be at less than 255 characters in length}{\samrule}
+\samstrictrule{Invalid character in QNAME}{QNAME fields must conform to the regex {TT \verb:[!-?A-~]}}{\samrule}
+
+\subsection{FLAG}
+\samstrictrule{Incorrect FLAG 0x1}{All templates with multiple segments must have FLAG 0x1 set}
+\samstrictrule{Incorrect FLAG 0x1}{All templates with a single segment must not have FLAG 0x1 set}
+\samstrictrule{Inconsistent FLAG 0x1}{All records with the same QNAME must have the same FLAG 0x1 value}
+\samstrictrule{Inconsistent FLAG 0x2}{All primary records with the same QNAME must have the same FLAG 0x2 value}
+\samstrictrule{Missing primary alignment record}{No supplementary or secondary alignments may exist for reads with an unmapped with primary alignment.}{\samrule}
+\samstrictrule{Inconsistent FLAGs 0x1 0x2}{The 0x2 FLAG must not be set if the 0x1 FLAG is not.}
+\samstrictrule{Inconsistent FLAGs 0x1 0x8}{The 0x8 FLAG must not be set if the 0x1 FLAG is not.}
+\samstrictrule{Inconsistent FLAGs 0x1 0x20}{The 0x20 FLAG must not be set if the 0x1 FLAG is not.}
+\samstrictrule{Inconsistent FLAGs 0x1 0x40}{The 0x40 FLAG must not be set if the 0x1 FLAG is not.}
+\samstrictrule{Inconsistent FLAGs 0x1 0x80}{The 0x80 FLAG must not be set if the 0x1 FLAG is not.}
+\samstrictrule{Inconsistent FLAGs 0x2 0x4}{The 0x2 FLAG must not be set if 0x4 is set in any primary alignments in the template}
+\samstrictrule{Inconsistent FLAGs 0x4 0x8}{The 0x8 FLAG for primary record for each segment must match the 0x4 FLAG of the primary record for the next segment in the template}
+\samstrictrule{Inconsistent FLAG 0x10 0x20}{The 0x20 FLAG must match the 0x10 FLAG for the primary alignment of the next segment in the template}
+\samstrictrule{FLAG 0x20 set on unmapped read}{The 0x10 FLAG must not be set if the 0x4 FLAG is is set.}
+\samstrictrule{Duplicate first segment primary records}{Of the primary records with the same QNAME, at most one can have FLAG 0x40 set}
+\samstrictrule{Missing first segment primary record}{Of the records with the same QNAME with 0x1 FLAG, at least one record must have FLAG 0x40 set.}
+\samstrictrule{Duplicate last segment primary records}{Of the primary records with the same QNAME, at most one can have FLAG 0x80 set}
+\samstrictrule{Missing last segment primary record}{Of the records with the same QNAME with 0x1 FLAG, at least one record must have FLAG 0x80 set.}
+\samstrictrule{Multiple primary alignment records}{Each segment must have at most one record with FLAG 0x100 and 0x800 not set.}{\samrule}
+\samstrictrule{Missing primary alignment record}{Each segment must have at least one record with FLAG 0x100 and 0x800 not set.}{\samrule}
+\samstrictrule{Unknown FLAG bit set}{FLAG bits higher than 0x800 must not be set}{\samrule}
+\samstrictrule{Unmapped reads should not have FLAG 0x10 set}{Unmapped reads should be stored in the orientation in which they came off the sequencing machine and have 0x10 unset.}{\vonefivebestpractice}
+
+\subsection{RNAME}
+\samstrictrule{Malformed RNAME}{RNAME must conform to the regex {\tt \char92*|[!-()+-\char60\char62-\char126][!-\char126]*}}{\samrule}
+\samstrictrule{RNAME not present in reference}{RNAME must be equal to the value of one of the SQ SN values defined in the header.}
+\samstrictrule{RNAME contains character not supported by VCFv4.3}{RNAME must not contain any of the following characters: {\tt \char60\char62\char91\char93\char58\char42}}{\vcffourthree}
+\samstrictrule{RNAME not supported by VCFv4.3}{RNAME must be not be one of {\tt DEL}, {\tt INS}, {\tt DUP}, {\tt INV}, {\tt CNV}, or {\tt BND}.}{\vcffourthree}
+\samstrictrule{RNAME does not match mate}{For a unmapped paired-end or mate-pair read whose mate is mapped, the unmapped read should have RNAME identical to its mate}{\vonefivebestpractice}
+\samstrictrule{RNAME specified for unmapped template}{If all segments in a template are unmapped, their RNAME should be set as *}{\vonefivebestpractice}
+
+
+\subsection{POS}
+\samstrictrule{Record placed outside of reference sequence}{If FLAG 0x4 is not set, POS must be between 0 and the length of the RNAME reference sequence inclusive. The length of the RNAME reference sequence can be found in the SQ header LN tag value for the SQ header with a SN tag matching the RNAME.}{\vonefivebestpractice}
+\samstrictrule{Invalid POS}{POS cannot be 0 if FLAG 0x4 is set.}
+\samstrictrule{Invalid POS}{POS cannot be negative.}{\samrule}
+\samstrictrule{Invalid POS}{POS cannot be greater than 2147483647.}{\samrule}
+\samstrictrule{POS specified without RNAME}{If RNAME is *, POS must be 0.}
+\samstrictrule{POS does not match mate.}{For a unmapped paired-end or mate-pair read whose mate is mapped, the unmapped read should have POS identical to its mate}{\vonefivebestpractice}
+\samstrictrule{POS specified for unmapped template}{If all segments in a template are unmapped, their POS should be set as 0.}{\vonefivebestpractice}
+
+\subsection{MAPQ}
+\samstrictrule{Invalid MAPQ}{MAPQ must be between 0 and 255 inclusive.}{\samrule}
+\samstrictrule{Missing MAPQ}{Aligned reads should not have 255 MAPQ.}{\vonefivebestpractice}
+
+\subsection{CIGAR}
+\samstrictrule{Invalid CIGAR}{All CIGAR strings must conform to the regex {\tt \char92*|([0-9]+[MIDNSHPX=])+}}{\samrule}
+\samstrictrule{Empty CIGAR}{All CIGAR strings must have at least one CIGAR operator}
+\samstrictrule{Zero length CIGAR operator}{All CIGAR operators must have a non-zero positive length}
+\samstrictrule{CIGAR contains operator repeat}{All adjacent CIGAR operators must be different.}{\vonefivebestpractice}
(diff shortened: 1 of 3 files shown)


d-cameron:
@jkbonfield I'd rather not expand this particular PR to include changes to the SAM specs themselves. My PR does appear to be highlighting some issues with SAMv1.5 that have avoided scrutiny, but I think the best course of action is to get a SAM v1.5 strict document and validator implementation based on the v1.5 specifications before updating the specs.

jkbonfield:
That's fine - I can accept that as an order in which things are done.  Yes I think there are problems with the spec which haven't been spotted as no one has really gone over it with such a fine tooth-comb before.  Thank you for doing this.

yfarjoun:
Could you do me a favor and include the pdf in these commits? when we get to the end you can remove the pdf and compile it "correctly" for the sake of having the PDF versioned.

yfarjoun:
I was wondering if we want the error "code" to not have spaces in it. This will help when we want to run a validation tool on the commandline and specify which codes to ignore. Spaces will mean that every code needs to be surrounded by quotes. COULD_WE_USE_THIS_FORMAT_FOR_CODES? (not shouting, just suggesting!)

jkbonfield:
The TeX is invalid at the moment, meaning a PDF cannot currently be produced.

I'm not a TeX expert, but it looks like you need eg `\vfifteen` instead of `\v15` commands and braces around the `{\tt text}` sections.  There are some other minor things like `\bam` not being defined too.  I tried correcting this and have a pdf that builds now, although it's formatted wrong and becomes pure fixed-width after a while.

jkbonfield:
The version in https://github.com/jkbonfield/hts-specs/tree/strict builds now, but needs a lot of formatting still.  I'm not planning on doing this myself - this was just a test to see how to get it in a readable format.

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{Incorrect CIGAR length}{Sum of lengths of the M/I/S/=/X operations must equal the length of SEQ when both CIGAR and SEQ are available.}{\samrule}
+\samstrictrule{Invalid CIGAR hard clip}{H must only be present as the first and/or last operation.}{\samrule}
+\samstrictrule{Invalid CIGAR soft clip}{S must only have H operations between them and the ends of the CIGAR string.}{\samrule}
+\samstrictrule{CIGAR overhangs reference sequence}{POS plus the number of reference bases consumed by the CIGAR must not exceed the length of the RNAME reference sequence.}{\v15bestpractice}
+\samstrictrule{Inconsistent CIGAR read lengths}{All mapped alignments for a given segment must have matching read lengths. That is, the sum of lengths of the M/I/S/=/X/H operations must be equal.}
+
+\samstrictrule{TODO: Unusual indel positioning}{TODO: should we disallow I/D operators at the ends of reads? There was some ambiguity as to how deletions interacted with POS but I think the spec has been updated in favour of the BWA interpretation since that discussion.}
+
+\subsection{RNEXT}
+\samstrictrule{Invalid RNEXT}{If the template contains one segment RNEXT must be *}
```
yfarjoun: missing \tt

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{RNEXT not present in reference}{RNEXT must be equal to the value of one of the SQ SN values defined in the header.}
+\samstrictrule{Invalid RNEXT}{If the primary alignment of the next read in the template is mapped, RNEXT must not be {\tt *}}{\samrule}
+\samstrictrule{RNEXT not using =}{If the primary alignment of the next read in the template is mapped to the same reference sequence, RNEXT must be {\tt =}}{\samrule}
+\samstrictrule{Incorrect RNEXT}{If this read is unmapped or the primary alignment of the next read in the template is mapped to the a different reference sequence, RNEXT must match the RNAME of the next read.}
```
yfarjoun: to ~the~ a different

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+
+\subsection{PNEXT}
+\samstrictrule{Invalid PNEXT}{If the template contains one segment PNEXT must be 0}
+\samstrictrule{Incorrect PNEXT}{If the primary alignment of the next read in the template is unmapped, PNEXT must be 0}
+\samstrictrule{Incorrect PNEXT}{If the primary alignment of the next read in the template is mapped, PNEXT must match the POS of that record.}
```
yfarjoun: error codes should be different

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+
+\subsection{RNEXT}
+\samstrictrule{Invalid RNEXT}{If the template contains one segment RNEXT must be *}
+\samstrictrule{RNEXT not present in reference}{RNEXT must be equal to the value of one of the SQ SN values defined in the header.}
+\samstrictrule{Invalid RNEXT}{If the primary alignment of the next read in the template is mapped, RNEXT must not be {\tt *}}{\samrule}
+\samstrictrule{RNEXT not using =}{If the primary alignment of the next read in the template is mapped to the same reference sequence, RNEXT must be {\tt =}}{\samrule}
+\samstrictrule{Incorrect RNEXT}{If this read is unmapped or the primary alignment of the next read in the template is mapped to the a different reference sequence, RNEXT must match the RNAME of the next read.}
+
+\subsection{PNEXT}
+\samstrictrule{Invalid PNEXT}{If the template contains one segment PNEXT must be 0}
```
yfarjoun: error codes should be different

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{Incorrect PNEXT}{If the primary alignment of the next read in the template is unmapped, PNEXT must be 0}
+\samstrictrule{Incorrect PNEXT}{If the primary alignment of the next read in the template is mapped, PNEXT must match the POS of that record.}
+
+\subsection{TLEN}
+\samstrictrule{TLEN out of range}{TLEN cannot be greater than 2147483647.}{\samrule}
```
yfarjoun: join these two errors?

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{TLEN out of range}{TLEN cannot be less than -2147483648.}{\samrule}
+\samstrictrule{Invalid TLEN}{TLEN must be 0 if flag 0x1 is not set.}{\samrule}
+
+\subsection{SEQ}
+\samstrictrule{Inconsistent SEQ read lengths}{All alignments of a given segment must have consistent SEQ lengths. That is, for all non-* SEQ, SEQ + length of CIGAR hard clip must be equal. }
+\samstrictrule{Inconsistent SEQ sequences}{All alignments of a given segment with non-* SEQ must have consistent base calls. A base cannot be called an A in one record, but a T in another. Note that to determine the read base, both the 0x10 FLAG, and any hard clipping CIGAR operators need to be taken into account.}
+\samstrictrule{SEQ of secondary alignments specified.}{SEQ of secondary alignments (0x100 FLAG set) should be set to * to reduce the file size.}{\v15bestpractice}
```
yfarjoun: remove period

yfarjoun: I think that the tone should be rules ("must") and the explanation need not appear. for explanation one can look for the best-practices or whatever...

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{Invalid sequence base}{Unless SEQ is "*", SEQ read bases must be one of the following characters: acmgrsvtwyhkdbnACMGRSVTWYHKDBN}
```
yfarjoun: can you put these characters in alpha-order please?

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{SEQ does not match reference when CIGAR indicates match.}{Unless SEQ is "*", read bases with CIGAR operator = must match the reference base. Bases are considered to match if overlap between the possible read and reference bases (based on their IUPAC codes) is non-zero.}
```
yfarjoun: period

yfarjoun: use \tt instead of "*"?

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{SEQ matches reference when CIGAR indicates mismatch.}{Unless SEQ is "*", read bases with CIGAR operator X must not match the reference base. Bases are considered to match on if, when ignoring case, the reference and read bases are the same character  and the character is one of the following characters: acgtACGT.}
```
yfarjoun: period

yfarjoun: match ~on~ if,

also, I'm not sure I agree with your definition of match...you have a different definition of "match" in the line above (I like that one better...).

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+
+\subsection{QUAL}
+\samstrictrule{QUAL specified without SEQ}{QUAL must be * if SEQ is *}{\samrule}
```
yfarjoun: \tt?

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{SEQ QUAL length mismatch.}{The length of a non-* QUAL must match the length of SEQ.}{\samrule}
```
yfarjoun: period

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{Invalid QUAL}{The ASCII value of all QUAL bases must be at least 33.}
```
yfarjoun: s/bases/characters/ ?

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{QUAL of secondary alignments specified.}{QUAL of secondary alignments (0x100 FLAG set) should be set to * to reduce the file size.}{\v15bestpractice}
```
yfarjoun: period

yfarjoun: should->must. remove explanation.

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{TODO: QUAL edge case}{What should we do when a read is length 1 and the QUAL encodes to "*" ?}{\samrule}
```
yfarjoun: ouch.

yfarjoun: I have to say that spending any time on this edge case seems to be a waste...

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{Malformed Z tag}{Z tags must conform to the the regex {\tt [\,\,\,!-\char126]*}}{\samrule}
+\samstrictrule{Malformed H tag}{H tags must conform to the the regex {\tt ([0-9A-F][0-9A-F])*}}{\samrule}
+\samstrictrule{Malformed B tag}{B tags must conform to the the regex {\tt [cCsSiIf](,[-+]?[0-9]*\char92.?[0-9]+([eE][-+]?[0-9]+)?)+}}{\samrule}
+\samstrictrule{Non-integer value in integer array}{Type B tags starting with one of "cCsSiI" must contain integer values.}
+\samstrictrule{Tag array value out of bounds}{Type B tags must not contain values that are greater than or less than the maximum or minimum value representable by the specified prefix.}
+\samstrictrule{Tag integer out of bounds for BAM representation}{Type i tag values must be within the range ~$[-2^{31},2^{32})$}{\bam}
+\samstrictrule{Unknown reserved tag}{
+	No record can include any reserved tags not defined in the
+	{\sl Sequence Alignment/Map Optional Fields Specification}.
+	Non-standard tags must start X, Y, Z or a lowercase letter as per the SAM specifications.
```
yfarjoun: ...must start **with**...

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+}
+\samstrictrule{Incorrect tag type}{The \textit{type} of all \textit{standard tags} must match the type
+defined in the {\sl Sequence Alignment/Map Optional Fields Specification}.}
+\samstrictrule{Duplicate tag}{}{\samrule}
```
yfarjoun: no text for "duplicate tag"?

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+TODO: Add validators for all standard tags (basic SAM validation done).
+
+
+
+
+
+
+
+\subsection{RG}
+3 When a RG tag appears anywhere in the alignment section, there should be a single corresponding
```
yfarjoun: remove 3 and 4 here and below

yfarjoun: should->must

yfarjoun: also, need error codes

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+@RG line with matching ID tag in the header.{\v15bestpractice}
+\subsection{RG}
+4 When a PG tag appears anywhere in the alignment section, there should be a single corresponding
```
yfarjoun: is this duplicated?

yfarjoun (commented):
Great work!! there seems to be a lo of work left at the bottom part of the document, but I like the rigor so far. 

I put some comments in the discussion...

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{Missing last segment primary record}{Of the records with the same QNAME with 0x1 FLAG, at least one record must have FLAG 0x80 set.}
+\samstrictrule{Multiple primary alignment records}{Each segment must have at most one record with FLAG 0x100 and 0x800 not set.}{\samrule}
+\samstrictrule{Missing primary alignment record}{Each segment must have at least one record with FLAG 0x100 and 0x800 not set.}{\samrule}
+\samstrictrule{Unknown FLAG bit set}{FLAG bits higher than 0x800 must not be set}{\samrule}
+\samstrictrule{Unmapped reads should not have FLAG 0x10 set}{Unmapped reads should be stored in the orientation in which they came off the sequencing machine and have 0x10 unset.}{\v15bestpractice}
+
+\subsection{RNAME}
+\samstrictrule{Malformed RNAME}{RNAME must conform to the regex {\tt \char92*|[!-()+-\char60\char62-\char126][!-\char126]*}}{\samrule}
+\samstrictrule{RNAME not present in reference}{RNAME must be equal to the value of one of the SQ SN values defined in the header.}
+\samstrictrule{RNAME contains character not supported by VCFv4.3}{RNAME must not contain any of the following characters: {\tt \char60\char62\char91\char93\char58\char42}}{\vcf43}
```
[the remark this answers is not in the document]
jkbonfield: True, but this rule comes directly from the specification.  Hg38 broke the spec!

If we wish to relax this rule, then we should also create a PR to amend the SAM spec.  Otherwise we just make it a rule here with a caveat that people may need to add an exception when running validation.

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{RNAME not supported by VCFv4.3}{RNAME must be not be one of {\tt DEL}, {\tt INS}, {\tt DUP}, {\tt INV}, {\tt CNV}, or {\tt BND}.}{\vcf43}
+\samstrictrule{RNAME does not match mate}{For a unmapped paired-end or mate-pair read whose mate is mapped, the unmapped read should have RNAME identical to its mate}{\v15bestpractice}
```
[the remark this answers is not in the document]
jkbonfield: It's terminology used for short-insert and long-insert libraries, but as far as the spec is concerned they're really both just paired reads.  Some mate-pair technologies were "outies" instead of "innies" for orientation, but the spec doesn't make any assertion on which is correct or not anyway.

If you wanted to generalise this it would need to be something like:  "unmapped reads should have the same RNAME of at least one mapped read, or '*' if all are unmapped".

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+\samstrictrule{Invalid POS}{POS cannot be 0 if FLAG 0x4 is set.}
+\samstrictrule{Invalid POS}{POS cannot be negative.}{\samrule}
+\samstrictrule{Invalid POS}{POS cannot be greater than 2147483647.}{\samrule}
+\samstrictrule{POS specified without RNAME}{If RNAME is *, POS must be 0.}
+\samstrictrule{POS does not match mate.}{For a unmapped paired-end or mate-pair read whose mate is mapped, the unmapped read should have POS identical to its mate}{\v15bestpractice}
+\samstrictrule{POS specified for unmapped template}{If all segments in a template are unmapped, their POS should be set as 0.}{\v15bestpractice}
+
+\subsection{MAPQ}
+\samstrictrule{Invalid MAPQ}{MAPQ must be between 0 and 255 inclusive.}{\samrule}
+\samstrictrule{Missing MAPQ}{Aligned reads should not have 255 MAPQ.}{\v15bestpractice}
```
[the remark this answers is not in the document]
jkbonfield: I have seen unaligned reads with MAPQ 255 before, but not sure about aligned.   The spec doesn't actually care - 255 is simply "unavailable" and has nothing to do with mapped or otherwise, although it does make more sense for mapped data than unmapped data.  Consider: I found an alignment and it's <here>, but I haven't bothered to do any of the maths for likelihood of it being correct nor have I looked for another alignment - hence 255.

SAMstrict.tex
```diff
@@ -0,0 +1,398 @@
...
+
+\subsection{CIGAR}
+\samstrictrule{Invalid CIGAR}{All CIGAR strings must conform to the regex {\tt \char92*|([0-9]+[MIDNSHPX=])+}}{\samrule}
+\samstrictrule{Empty CIGAR}{All CIGAR strings must have at least one CIGAR operator}
+\samstrictrule{Zero length CIGAR operator}{All CIGAR operators must have a non-zero positive length}
+\samstrictrule{CIGAR contains operator repeat}{All adjacent CIGAR operators must be different.}{\v15bestpractice}
+\samstrictrule{CIGAR does not contain any mapped bases}{All CIGARs must include a CIGAR operator that consumes a reference base.}{\samrule}
```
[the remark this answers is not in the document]
jkbonfield: Cigars mapping entirely within insertion are valid and sometimes useful too.

SAMstrict.tex
jkbonfield: This gets into the same murky waters as NM tag, specifically is N vs N a match?

SAMstrict.tex
jkbonfield: I think I've actually hit that one before! Hah.

jmarshall:
The issues for SAM and VCF strictness proposals are separate. Would it not be more practical to open a VCF strict proposal as a new separate PR?

tskir:
I agree with a comment from @jmarshall above about splitting this into two. Since _in its current form_ it concerns both VCF and SAM, I've added the appropriate labels for the time being

d-cameron:
Happy to progress these specs if someone has funding/budget/time to actually write a validator

tfenne:
I spent a little time with Claude this afternoon scoping out what a validator would look like for the SAM/BAM side of this PR's spec.  I think, broadly speaking, I have a good idea of the shape and what would be required, and I don't think it's a ton of work so long as we accept the the following:

- input must be either coordinate sorted *or* query grouped (i.e. GO:query, or SO:queryname, or SO:template-coordinate, etc.)
- it's ok for the validator to have a "fairly high" max memory usage (high single digit to low tens of GB)

I think it's either that, or accept that the input may have to be re-sorted in order to validate.  I.e. we either need to be able to see all read for a template/queryname together _or_ predict approximately where they'll be in the file, or we'll have to cache too many reads to do it in a single pass.

So I'm game to take take a shot at the SAM/BAM validator, once the spec is in good shape.  Claude and I also reviewed the Strict SAM/BAM spec and put together a detailed review of where the spec needs work prior to tooling being built, which follows below:

# Review notes on the draft SAM strict specification

These are intended to be constructive review notes on the work-in-progress *SAM strict* draft ([samtools/hts-specs#283](https://github.com/samtools/hts-specs/pull/283)), written while starting to build a validator against it. The enumeration of rules is already a strong first inventory. The notes below are aimed at one thing: **what would need to be tightened so that independent implementations agree on what "strict compliant" means and can check it reliably at whole-genome scale.**

We focus on substance and deliberately skip editorial/typo issues. The items here are the ones that change what an implementation does, or that leave "compliant" genuinely ambiguous.

## 1. The biggest gap: each rule needs a formal severity and a declared validation requirement

Right now a rule is a sentence of prose plus, sometimes, a single "category" tag (`SAM`, `SAMv1.5 best practice`, `VCFv4.3`, `BAM`). Many rules carry no tag at all. That single tag is being asked to encode three unrelated things at once, and it's missing the one piece an implementer most needs. We'd suggest breaking a rule's metadata into explicit, orthogonal fields.

### 1.1 Severity / conformance level

Separate *how bad a violation is* from *where the rule comes from*. At minimum, adopt RFC 2119 language and a two- (or three-) level severity: **MUST → error**, **SHOULD → warning**, and optionally **MAY/INFO**. Today this is entangled with provenance — `SAMv1.5 best practice` implies "warning" while `SAM` implies "error" — but that breaks down immediately: a rule can be a hard MUST *and* only be checkable under certain conditions, and provenance (`VCFv4.3`) tells you nothing about severity on its own.

### 1.2 Validation requirement (scope and inputs)

This is the field that's currently missing entirely, and it's the single most important one for implementability: **what does a validator need to have in hand to check this rule?** It determines memory footprint, whether the reference genome is required, whether a second pass or on-disk spill is needed, and — critically — whether the rule is even checkable given the file's sort order. 

A workable taxonomy:

| Scope tier | What the validator must hold | Example rules | Notes |
| --- | --- | --- | --- |
| Field-local | one field of one record | QNAME charset, MAPQ range, tag type | free, any order |
| Record-local | one whole record | CIGAR/SEQ length, hard/soft-clip placement, single-record FLAG bits | free, any order |
| Header-dependent | record + header | RNAME ∈ `@SQ` SN, POS ≤ LN | header is tiny |
| Reference-dependent | record + reference FASTA | `=`/`X` vs reference, CIGAR overhang | needs the reference; see §2 |
| Mate-scoped | record + its mate's primary | RNEXT/PNEXT agreement, `=` when same ref | needs both mates co-resident |
| Segment-scoped | all alignments of one read-end | SEQ/QUAL consistency, SA reciprocity, per-segment primary count | needs the scattered supplementaries/secondaries |
| Template-scoped | all records sharing QNAME | FLAG consistency across QNAME, first/last-segment counts | needs the whole template |
| File-global | aggregate over all records | sort-order, `@SQ`/`@RG`/`@PG` ID uniqueness | streaming aggregate |

Each rule should additionally declare its concrete input dependencies as booleans: **needs reference?**, **needs mate?**, **needs all supplementaries?**, **needs whole template?** An implementer reading the spec should be able to bucket every rule without guessing.

### 1.3 Input-ordering assumptions

Some rules are only checkable in one sort mode, and the spec is silent on this. Verifying "records are in `SO:coordinate` order" is impossible from a queryname-sorted file, and vice-versa. The cross-record rules (mate, segment, template) are cheap on query-grouped input (all of a template's records are contiguous — gather, check, discard) but require a bounded reconciliation buffer on coordinate-sorted input. The spec should state which input modes it supports (we'd suggest requiring `SO:coordinate`, `SO:queryname`, or `GO:query` and refusing to validate otherwise), declare which rules apply in which mode, and — importantly — **explicitly permit a conformant validator to report "not verified under resource limits"** for the unbounded cross-record rules on coordinate-sorted input. Without that escape valve, the spec is only satisfiable with unbounded RAM.

### 1.4 Present rules as a table, and give them stable IDs

The prose-per-rule format hides all of the above. A table with columns `ID | severity | scope | needs-ref | needs-mate | needs-supps | ordering-modes | profile` would make the spec both human- and machine-readable, and would make gaps obvious. Relatedly, **the rules have no identifiers.** A validator needs stable IDs to report, suppress, document, and test against (we've had to assign provisional ones like `CIGAR-04` ourselves). Assigning canonical IDs in the spec — and a machine-readable manifest of the table — would let every implementation speak the same language.

## 2. Major problems (beyond typos)

**The `category` tag conflates provenance, severity, and conditional applicability.** As above: `SAM`/`best practice`/`VCFv4.3`/`BAM` mix "which document this comes from," "how serious it is," and "whether it only matters when targeting VCF / when encoded as BAM." These are three axes and should be three fields. `VCFv4.3` in particular is really a *downstream-consumer profile* masquerading as a category (see §3).

**Cross-record scope is hand-waved, and it's the hard part.** The rules that require gathering scattered records — FLAG consistency across a QNAME, SEQ/QUAL consistency across a segment's alignments, mate RNEXT/PNEXT/RNAME agreement, the SA "record set" reciprocity — are stated as flatly as the field-local ones, with no acknowledgement that they need state proportional to the input's mate/split-read distribution. This is the central implementability blocker, and it's invisible in the current presentation.

**Sort-order rules don't define the collation.** "Order must be consistent with the `SO` tag" is undefined for `SO:queryname`, where `samtools` (natural order) and htsjdk/Picard (lexical) famously disagree. The spec must pin the queryname collation precisely. Likewise, "the `SO` and `GO` orderings must be consistent" is never defined (consistent how?), and `SO:unknown` interactions are unaddressed.

**`@SQ` `M5` is marked as a hard requirement.** The draft requires `M5` on every `@SQ` at `SAM`/error severity. The overwhelming majority of existing mapped BAMs (anything from a default `bwa`/`samtools` pipeline that isn't CRAM) have no `M5`, so as written this rejects most real-world files. MD5 presence is a reasonable requirement for a *reference-integrity / CRAM / variant-calling* profile, but not for a universal core. This is a good concrete example of provenance/severity/profile entanglement.

**Enumerated value lists will drift from the SAM spec.** The `@RG` `PL` list (`CAPILLARY, LS454, ILLUMINA, SOLID, HELICOS, IONTORRENT, ONT, PACBIO`) already omits platforms present in current SAMtags (`DNBSEQ`, `ELEMENT`, `ULTIMA`, `SINGULAR`, `GENAPSYS`, …), so at error severity this rejects valid modern files. Same risk for the `SO`/`GO` enumerations and for "`VN` must match a published SAM version" (requires maintaining a version list). The strict spec should **reference the canonical SAMtags lists rather than re-enumerate them**, or it guarantees itself into staleness.

**Reference-comparison rules are underspecified and mis-scoped as baseline.** The `=`/`X`-vs-reference rules require the reference FASTA and a precise comparison semantics. The draft uses *asymmetric* semantics — `=` "matches" if the IUPAC ambiguity sets overlap (so a read `N` matches anything), while `X` "must not match" under case-insensitive exact `ACGT` comparison. That asymmetry is defensible but surprising and needs to be spelled out with examples (what about `N` vs `X`? lowercase soft-masked reference?). More importantly, since almost all aligners emit `M` rather than `=`/`X`, these rules rarely fire yet impose a hard reference dependency on the whole validator. They belong in an opt-in reference profile, not the core.

**"CIGAR must consume a reference base" is too strong for a universal rule.** The draft flags this itself (the Seven Bridges `100I` case). Pure-insertion alignments against known insertions, and unplaced local-assembly contigs, are legitimate. This should be a warning or a profile rule, not a `SAM`/error.

**Multi-segment (>2) templates are barely handled.** The FLAG/RNEXT/PNEXT rules are written in paired (two-segment) language ("the next segment," "the mate"). The general ring semantics for N>2 segments (the draft's own aside that "the first segment is the 'next' of the final segment") is mentioned once but not carried through. Linked-read, concatemer, and other multi-segment templates need explicit treatment or an explicit statement that only 1- and 2-segment templates are covered.

**TLEN says nothing about the value.** The draft checks TLEN's range and "0 if `0x1` unset" but is silent on the signed leftmost-to-rightmost computation — almost certainly deliberate, because tools disagree, but the spec should *say* it is intentionally silent and why, rather than leave readers to assume the value is unconstrained by accident, or require that all insert sizes are consistent with one of the handful of prevailing definitions.

**Load-bearing terms are used before they're defined.** "read", "segment", "template", "chimeric alignment", and especially the "representative"/primary member of a chimeric set are all doing heavy lifting across half the rules, but there's no tight definitions section. Strict validation needs these pinned even more precisely than SAMv1 does.

**Behavior on degenerate-but-valid files is unspecified.** Empty files, header-only files, and fully-unaligned files (uBAM) are all valid SAM. The rules imply handling (e.g. "`@SQ` required if any reads mapped") but the spec should state plainly that unaligned data is first-class and enumerate which rules simply don't apply.

## 3. What "strict" is this? Scope and framing

Reading the full rule set, its center of gravity is clear: **this is a spec for reference-based variant and structural-variant calling.** The evidence is throughout — heavy formalization of SA tags and chimeric-alignment record sets (SV calling), RNAME restricted to the VCFv4.3-safe character set and away from the reserved symbolic alleles `DEL/INS/DUP/INV/CNV/BND` (variant calling), `=`/`X`-vs-reference comparison, required reference `M5` checksums, and POS-within-reference bounds. Those are exactly the invariants a variant/SV caller wants to be able to trust. That's a worthy target — but it isn't obviously *SAM* strict, and the framing question matters.

Whole classes of legitimate SAM usage get little or no attention, for example:

- **Spliced RNA-seq alignments** (STAR, HISAT2, minimap2 `-x splice`). The `N` (skipped-region/intron) operator is syntactically allowed but never sanity-checked: nothing about `N` needing `M`/`=`/`X` flanks (not `I`/`D`), minimum/maximum intron length, or `N` at read ends being nonsensical. RNA-seq also leans heavily on multi-mappers (many secondary alignments), which stresses the segment-gathering assumptions and interacts with the "secondary SEQ should be `*`" rule. Strand tags (`XS`, `ts`) are unaddressed.
- **Long reads** (ONT, PacBio). The BAM long-CIGAR overflow encoding — a placeholder CIGAR in the main field plus the real one in the `CG:B,I` tag when there are >65535 operations — is unmentioned; a validator naive to `CG` will mis-check CIGAR/SEQ length on exactly these files. Modified-base tags (`MM`/`ML`) and their consistency with SEQ are also out of view. The "no adjacent identical ops" and "indels at read ends" questions bite hardest here.
- **Unaligned BAM (uBAM)** from basecallers/demux. No `@SQ`, everything unmapped, but rich `RG`/barcode/`MM`/`ML` tags. Is a uBAM "strict compliant"? It should be first-class, and the reference/variant-oriented rules should cleanly not apply.
- **Single-cell, linked-read, and Hi-C** data, whose barcode conventions and (for Hi-C) deliberately "discordant" mate placements stress the mate-consistency rules and the memory model.

The deeper point is that the draft is mixing two different documents:

1. **A universal SAM-strict core** — internal-consistency invariants that must hold for *any* valid SAM regardless of downstream use: FLAG consistency, CIGAR/SEQ length agreement, tag format/type correctness, header well-formedness and ID uniqueness/referential integrity, mate reciprocity, SA reciprocity. These are consumer-agnostic and are the strongest part of the draft.
2. **Consumer profiles** layered on top: a *variant-calling* profile (reference-base match, `M5`, POS-within-reference), an *SV* profile (SA completeness), a *VCF-export* profile (RNAME character set), an *RNA-seq* profile (splice-op sanity), and so on. These are the SHOULD/conditional rules that only matter to a particular downstream.

There is no clean answer to the simple question "is this a well-formed SAM file?" independent of "is this ready for my variant caller?" The existing `category` tags (`SAM` / `best practice` / `VCFv4.3`) are actually a nascent, informal version of this profile idea — we'd encourage formalizing it: **declare the document's scope explicitly, factor out a small consumer-agnostic core, and express the rest as named, opt-in profiles.** As written, an honest title would be closer to *"SAM for reference-based variant/SV calling, strict."* If the ambition is a universal baseline, it needs the RNA-seq / long-read / uBAM representation, etc. and the core-plus-profiles split to earn the name *SAM strict*.

## 4. Implementation-readiness checklist

Pulling the above into the concrete set of changes that would let someone build a conformant, agreeing, WGS-scale validator from the spec alone:

- Add a **definitions section** pinning read / segment / template / chimeric alignment / representative-primary.
- Give every rule a **stable ID**, an explicit **severity** (MUST/SHOULD), and a **validation-requirement** record (scope tier + needs-ref/mate/supps/template + applicable ordering modes). Present it all as a **table** with a machine-readable manifest.
- **Declare supported input modes** (coordinate / queryname / query-grouped) and per-rule applicability; **permit resource-bounded "unverified" outcomes** for unbounded cross-record rules.
- **Pin the queryname collation**, and define what `SO`/`GO` "consistency" means.
- **Reference canonical SAMtags value lists** instead of re-enumerating `PL`/`SO`/`GO`/versions.
- **Specify the `=`/`X`-vs-reference comparison semantics precisely** (with examples) and make it, `M5`, and POS-within-reference an **opt-in reference/variant profile** rather than core.
- **Factor the document into a universal core plus named consumer profiles**, and name the document to match its actual scope.
- **Address the gaps**: `N`/splice sanity, the BAM `CG` long-CIGAR encoding, uBAM as first-class, multi-segment (>2) templates.
- **State behavior on empty / header-only / fully-unaligned files.**
- Consider a companion note on the **machine-readable output contract** (rule IDs, severities, record locations, exit codes) — arguably outside a format spec, but it's what makes independent validators comparable.

None of this detracts from the direction. The single highest-leverage change is §1 — turning each rule from a sentence into a row with a severity and an explicit statement of what a validator must hold to check it. Do that, and most of the ambiguities above resolve themselves, the SV/variant-calling lean becomes an explicit profile rather than a hidden assumption, and the spec becomes something two independent teams can implement and get the same answer from.

jkbonfield:
NB: this is purely my thoughts on the SAM world for now.

The TeX doesn't compile still for me.  I had to do a lot of \tt and \verb hackery amongst other things, but the output is still broken and `samstrictrule` doesn't render correctly.  That said, the intent is still obvious and markup is a different issue.

I think there are clear categories which we can all agree on.

- **Syntactically wrong**.  That's a different spec, but it's a prerequisite.
- **Inconsistent data**.  It's present, but plain wrong.  Eg read 1 flags claim the mate is reverse, but the mate's flags claim otherwise.  (This is the sort of thing that `samtools fixmates` is designed to fix, among other tools.)  There's a whole host of such things and they are clearly errors.
- **Correct but ill advised**.  E.g. the specification states "Bit 0x4 is the only reliable place to tell whether the read is unmapped. If 0x4 is set, no assumptions can be made about RNAME, POS, CIGAR, MAPQ, and bits 0x2, 0x100, and 0x800".  I do  regularly see outputs from bwa that had CIGAR fields for unmapped data.  That's a valid file, but it's a bit annoying if every tool has to run through the same set of checks for something which is arguably just a bug.  So these sorts of breakages are really warnings:  tools may be misled.  We can improve on the expectations and arguably fix decisions that weren't sensible in the specification.
    - **Meaningless**. Likewise a set of official auxiliary tags that have no meaning when unaligned.  They're not explicitly forbidden by the specification, and the spec doesn't even list them as being undefined.  However "NM" and "MD" on unaligned records has been seen in the wild!  (I think it was from bwa again sadly.  IIRC it occurs when an alignment spans the end of one chromosome to another in the concatenated sequences, which gets filtered out and turned into unmapped at a later stage, albeit incompletely.)
- **Provenance and completeness**.  Not incorrect, but a nicety.  We should have a version number and sort order explicitly listed.  SQ lines should refer to an assembly reference ID.  Readgroups should list the platform and library types. Etc.   This is very much at the warning level.  It sort of fits in with the "recommendations" approach.
- **Correct but incomplete data**.  This is simply a specific tool requirement.  Eg we've done a range query and there are some records missing.  Not every paired read has two reads in the file.  This could be correct SAM and 100% consistent, but it's a tool requirement.  I'm not particularly comfortable with this level of checking.  If a tool needs it, it should check.  (I don't think it's the job of a standards specification to work around buggy tooling.)
    -  **Absent data**.  This comes under the "correct but incomplete data" clause perhaps, but it's a different kind.  Incompleteness above is simply down to a sub-selection, so we cannot validate between records.  There is another class which is completeness within a record.  Eg a tool may require that every record has an RG tag, or all RNEXT/PNEXT/TLEN are defined for paired data and not left as `* 0 0`.  These are specific checks which could be opted in.  I'd argue again that it's not the job of this specification to state such things as it's pandering to each and every tool out there, of which there are 100s, but it may be a useful tool to have.  Some tools don't output QUAL fields for secondary alignments, or even in some scenarios no sequence either (just a CIGAR field).  There is good reason to do such things, but it's a fair thing to validate on as it can be useful to know upfront if your data has such cases.  More bizarre things here could be QNAME has to be present (it can legally be "*"), although maybe that comes under "correct but illadvised" or "Completeness".
    
The first two categories are not really up for debate.  We know the semantic rules that apply on top of the syntax as they're in the specification and they're hopefully clearly written into the specification.  The third already forms the basis of many sanity checks in various tools, but it would be sensible to codify them.

Everything beyond that makes me uneasy for a specification unless they are very strict classes.

Eg rNA seq data should have these attributes.

Edit: just to clarify, I also think we're mixing up two things here.

1. The strict SAM specification which formally adds a set of additional diagnostics that are sensible for all tool chains, tightening up arguably poor choices in the original specifications.
2. A set of additional tool-chain specific checks of prerequisites.  It'd be nice to have such a tool, but IMO it doesn't belong in a standards body as it's tool-implementation specific.

PR #285, opened 2018-02-21
Add POST method to htsget

daviesrob:
Query parameters are passed as JSON.  It is possible to request
more than one location in the returned data.

Very much a work-in-progress at the moment.  Comments welcome.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -4,7 +4,7 @@ title: htsget protocol
 suppress_footer: true
 ---
 
-# Htsget retrieval API spec v1.2.1
+# Htsget retrieval API spec v1.3.0
 
 # Design principles
 
@@ -684,7 +684,7 @@ Example service-info response:
    "type":  {
       "group":        "org.ga4gh",
       "artifact":     "htsget",
-      "version":      "1.2.1"
+      "version":      "1.3.0"
    },
    "htsget": {
       "datatype": "reads",
@@ -722,7 +722,7 @@ Example listing of htsget reads API and variants API registrations from a servic
       "type": {
          "group": "org.ga4gh",
          "artifact": "htsget",
-         "version": "1.2.1"
+         "version": "1.3.0"
       }
    },
    {
@@ -738,7 +738,7 @@ Example listing of htsget reads API and variants API registrations from a servic
       "type": {
          "group": "org.ga4gh",
          "artifact": "htsget",
-         "version": "1.2.1"
+         "version": "1.3.0"
       }
    }
 ]
--- a/htsget.md
+++ b/htsget.md
@@ -84,6 +84,7 @@ Error type	| HTTP status code | Description
 InvalidAuthentication	| 401	| Authorization provided is invalid
 PermissionDenied	| 403	| Authorization is required to access the resource
 NotFound		| 404	| The resource requested was not found
+PayloadTooLarge         | 413   | POST request size is too large
 UnsupportedFormat	| 400	| The requested file format is not supported by the server
 InvalidInput		| 400	| The request parameters do not adhere to the specification
 InvalidRange		| 400	| The requested range cannot be satisfied
@@ -119,7 +120,7 @@ If a request to the URL of an API method includes the `Origin` header, its conte
 The values of `Origin` and `Access-Control-Request-Headers` (if any) of the request will be propagated to `Access-Control-Allow-Origin` and `Access-Control-Allow-Headers` respectively in the preflight response.
 The `Access-Control-Max-Age` of the preflight response is set to the equivalent of 30 days.
 
-# Request
+# GET request
 
 ## Methods
 
@@ -279,6 +280,185 @@ SEQ	| Read bases
 QUAL	| Base quality scores
 
 Example: `fields=QNAME,FLAG,POS`.
+
+# POST request
+
+In addition to the GET method, servers may optionally also accept POST requests.
+The main differences to the GET method are that query parameters are encoded in JSON format and it is possible to request data for more than one genomic range.
+The data returned is a JSON "ticket", as described for the GET method.
+
+Htsget POST requests must be "safe" as defined in section 4.2.1 of [RFC 7231], that is they must be essentially read-only.
+
+It should be expected that htsget servers will have a limit on the size of POST request that they can accept.
+If an incoming POST request is too large, the server SHOULD reply with a `PayloadTooLarge` error.
+
+## URL parameters
+
+POST and GET requests should use the same URLs.
+
+POST requests MUST NOT use and of the query parameters defined for GET requests on the URL.
+If query parameters are present, the server SHOULD reply with an `InvalidInput` error.
+
+## Request body
+
+The POST request body is a JSON object containing the request parameters,
+which have similar names and meanings to those used in the GET query.
+A notable difference is that the  `referenceName`, `start` and `end` parameters are not used at the top level of the JSON object.
+Instead as they are put into an array under the `regions` key, which allows more than one range to be specified.
+
+<table><tbody><tr markdown="block"><td>
+
+`format`  _optional string_
+</td><td>
+
+Request data in this format.
+The allowed values for each type of record are:
+
+* Reads: BAM (default), CRAM.
+* Variants: VCF (default), BCF.
+
+The server SHOULD reply with an `UnsupportedFormat` error if the requested format is not supported.
+</td></tr>
+<tr markdown="block"><td>
+
+`class`
+_optional string_
+</td><td>
+
+Request different classes of data.
+As for GET requests, if `class` is not specified the response will comprise the full data stream including file headers, body data records and EOF marker.
+If `class` is specified, its value MUST be one of the following:
+
+<table>
+<tr markdown="block"><td>
+
+`header`
+</td><td>
+
+Request the SAM/CRAM/VCF headers only.
+
+The server SHOULD respond with an `InvalidInput` error if any parameters other than `format` are specified at the same time as `"class" : "header"`.
+</td></tr>
+</table>
+</td></tr>
+
+`fields`  _optional array of strings_
+</td><td>
+
+List of fields to include.  See Field filtering.
+  <table><tbody><tr markdown="block"><td>
+
+Field name _string_
+  </td></tr></tbody></table>
+</td></tr>
+<tr markdown="block"><td>
+
+`tags`  _optional array of strings_
+</td><td>
+
+List of tags to include, default: all.  If an empty array is specified, no tags are included.
+
+The server SHOULD respond with an `InvalidInput` error if `tags` and `notags` intersect.
+  <table><tbody><tr markdown="block"><td>
+
+Tag name _string_
+  </td></tr></tbody></table>
+</td></tr>
+<tr markdown="block"><td>
+
+`notags`  _optional array of strings_
+</td><td>
+
+List of tags to exclude, default: none.
+
+The server SHOULD respond with an `InvalidInput` error if `tags` and `notags` intersect.
+  <table><tbody><tr markdown="block"><td>
+
+Tag name _string_
+  </td></tr></tbody></table>
+</td></tr>
+<tr markdown="block"><td>
+
+`regions` _optional array of objects_
+</td><td>
+
+Regions to return.
+
+If not present, the entire file will be returned.
+When present, the array must contain at least one region specification.
+Note that regions will be returned in the order that they appear in the file, which may not match the order in the list.
+Any overlapping regions will be merged.
+
+The server SHOULD respond with an `InvalidInput` error if the region list is empty or not well-formed.
+  <table><tbody><tr markdown="block"><td>
+
+`referenceName` _string_
+  </td><td>
+
+The reference sequence name, for example "chr1", "1", or "chrX".
+
+The server SHOULD respond with an `InvalidInput` error if `referenceName` is not specified.
+
+The server SHOULD reply with a `NotFound` error if the requested reference does not exist.
+  </td></tr>
+  <tr markdown="block"><td>
+
+`start` _optional unsigned integer_
+  </td><td>
+
+The start position of the range on the reference, 0-based, inclusive.
+
+If not present, data will be returned starting from the first base in `referenceName`.
+
+The server SHOULD respond with an `InvalidRange` error if `start` and `end`
+are specified and `start` is greater than or equal to `end`.
+  </td></tr>
+  <tr markdown="block"><td>
+
+`end` _optional unsigned integer_
+  </td><td>
+
+The end position of the range on the reference, 0-based exclusive.
+
+If not present, data will be returned up to the end of the reference.
+
+The server SHOULD respond with an `InvalidRange` error if `start` and `end`
+are specified and `start` is greater than or equal to `end`.
+  </td></tr></tbody></table>
+</td></tr></tbody></table>
+
+### Example
+
+```json
+{
+   "format" : "bam",
+   "fields" : ["QNAME", "FLAG", "RNAME", "POS", "CIGAR", "SEQ"],
+   "tags" : ["RG"],
+   "notags" : ["OQ"],
+   "regions" : [
+      { "referenceName" : "chr1" },
+      { "referenceName" : "chr2", "start" : 999, "end" : 1000 },
+      { "referenceName" : "chr2", "start" : 2000, "end" : 2100 }
+   ]
+}
+```
+
+## Region list
+
+The `regions` parameter is an array of objects which describe the locations to be returned.
+Each location object contains a `referenceName`, which must always be present, and optional `start` and `end` tags.
+
+The first region in the example above shows how to return all reads for reference "chr1".
+
+To return reads covering a single base, the client should use an end position one greater than the start, as shown in the second region in the example which requests the 1000th base of reference "chr2".
+
+The regions list acts as a filter on the requested file.
+Records that overlap the requested regions will be returned in the order that they occurred in the original file.
+This may not be the same as the order in the list.
+A record will only be returned once, even if it matches more than one location in the list.
+
+As with the GET request, the server response may contain a super-set of the desired results.
+Clients will need to filter out any extraneous records if necessary.
 
 # Response
 
@@ -568,7 +748,6 @@ Example listing of htsget reads API and variants API registrations from a servic
 
 * add a mechanism to request reads from more than one ID at a time (e.g. for a trio)
 * allow clients to provide a suggested data block size to the server
-* add POST support (if and when request sizes get large)
 * [dglazer] add a way to request reads in GA4GH binary format [^d] (e.g. fmt=proto)
 
 ## Existing clarification suggestions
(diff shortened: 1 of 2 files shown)


htsget.md
```diff
@@ -357,15 +357,119 @@ Initial guidelines, which we expect to revise in light of future experience:
...
+<tr markdown="block"><td>`notags`  _optional array of strings_</td><td>
+List of tags to exclude, default: none.  If an empty array is specified, no tags are included.
+
+The server SHOULD respond with an `InvalidInput` error if `tags` and `notags` intersect.
+  <table>
+     <tr markdown="block"><td>Tag name _string_</td></tr>
+  </table>
+</td></tr>
+<tr markdown="block"><td>`regions` _object_</td><td>
+Regions to return.  If not present, the entire file will be returned.  If an empty object or null, just the header will be returned.
```
jeromekelleher: This is a good idea for allowing people to get the header.

htsget.md
```diff
@@ -357,15 +357,119 @@ Initial guidelines, which we expect to revise in light of future experience:
...
+Both elements must always be present.
+`start` must always be a number and gives the 0-based, inclusive start position.
+`end` must either be a number giving the 0-based end position, or the exact string `"end"` to indicate that the region finished at the last base of the given reference.
+The end position must be strictly greater than the start.
+To return all bases for `chr1`, the client should use the request
+`"regions" : { "chr1" : [[0, "end"]]}`.
+To return a single base, the client should use `[POS, POS+1]`, for example:
+`"regions" : { "chr1" : [[1000, 1001]]}`.
+
+If elements in the location list overlap, the returned data will be the union of the overlapping elements.
```
jeromekelleher: Does this make life hard for the server unnecessarily? You could imagine that if a client asks for the same region twice, then it's perfectly fine to just return the same data twice. I don't really know, just asking the question!

jeromekelleher:
Looks good to me @daviesrob, all very sensible. There are few things I'd like to follow up on:

## List of region objects, vs object of lists.

I think it would be quite a lot neater to use a list of region objects, rather than the object of lists notation currently used:
```json
{
   "format" : "bam",
￼   "fields" : ["QNAME", "FLAG", "RNAME", "POS", "CIGAR", "SEQ"],
￼   "tags" : ["RG"],
￼   "notags" : ["OQ"],
￼   "regions" : [
         {"referenceName": "chr1", "start": 0},
         {"referenceName": "chr2", "start": 1000, "end": 1001},
         {"referenceName": "chr2", "start": 2000, "end": 2100},
￼   ]
￼}
```
The nice thing here is that this maps exactly to the semantics of a the query parameters for a GET request, so there's no need for any special values to denote null start or end coordinates. Presumably an empty reqion also corresponds to everything, following the GET semantics.

## Ordering of returned blocks

From the meeting, it seems like people are keen on allowing the server some freedom about the order in which the requests are serviced. However, we do need to require that all blocks for a **single region** are adjacent in the response (otherwise, the client will have to demultiplex them, which seems like an unnecessary mess). Semantically, it would be simplest to require that regions are returned in the order that they are requested and I would imagine that this would be the expected outcome for a client.

htsget.md
daviesrob: I thought it would be best to define what happens in this case so we don't end up with a mixture of servers that repeat the data and ones that don't.  The server would have to do similar work anyway so that it can handle non-overlapping locations that have lots of reads in common because they are close enough together.

jeromekelleher: Definitely agree that we should define the behaviour, and I agree with your argument. The reason I bring up the alternative, is that the obvious and easy way to implement this from a server perspective (in terms of code, not necessarily efficiency) is to just treat this in this as a list of independent GET requests, and just concatenate the lists of chunks. The "front-end" server can't do this if it has to reason about overlaps within the requests. 

I agree that if minimising the traffic to the back-end server is the goal, then returning the union is the way to go.

daviesrob:
The idea behind using an object of lists was to keep the request as small as reasonably possible.  If you want to do something like look at SNPs across the Human genome it could get quite big (100s of thousands to millions of locations).  Keeping the overhead for each region low seems like a good idea so that the server can handle as many as possible on each round trip.

Would it be a good idea to add a recommended maximum number of regions?  If so, we might want to add [413](https://tools.ietf.org/html/rfc7231#section-6.5.11) to our list of error codes so the server can complain if it's asked to do too much.

On the ordering of the returned data, it might be best to think of the region list acting as a filter on the entire file.  The resulting file contains all of the reads that overlap the desired regions in the same order as they were in the original.  It's likely to be much more efficient than trying to return the data in exactly the requested order and you won't get problems like duplicated reads in the output (in the worst case where you request adjacent 1 base regions, each read could be duplicated as many times as its mapped length).  Anyone who does want a single region in a single group of reads could always fall back to making the requests one at a time (as now).

jmtcsngr:
I like the idea of having a maximum number of regions. Practically speaking most http server implementations will impose a max size on the post anyway. Security wise is also good idea to limit the payload to mitigate DoS attacks. I understand max number of regions may can not be expressed as max number of bytes in payload. But a compromise which considers both will be good, I think.

Probably I missed it, but I think is essential to require a content-type header in the post request so we can include a version of the "schema" to be future proof.

What we do with edge scenarios? Things like regions we can no serve (they don't exist, they are wrongly specified). Do we 500 on first problem? Some cases can be a challenge, e.g. unauthorized for specific regions is that a 403 or a 500? Any opinions?

htsget.md
jmarshall: That is nice, but I think we should still come up with a way to get the header with a GET request.

daviesrob:
I've added a commit to make it clearer that the regions list should act as a filter on the requested file.

As for the sort of query that we might expect, [this](https://github.com/samtools/samtools/files/1908053/regions.bed.txt) is an example of a fairly big regions list we were sent as part of a bug report.  It contains 258823 entries, although quite a few overlap (combining the overlaps results in 80868 unique regions).  Obviously, asking for these using 80868 GET requests would be impractical.

Trying a few options for how to format the original file (without merging overlaps) gave the following sizes:

```
Object of lists version:
perl -nae 'BEGIN {print qq["regions" : {\n]; } if ($l[0] ne $F[0]) { if (@l) { print "]";} print "${cc}$F[0] : ["; $c = ""; $cc = ",\n"; } print "${c}[$F[1],$F[2]]"; $c = ",\n"; @l = @F; }{ print "]\n}\n";' /tmp/regions.bed.txt | wc
 258825  258875 5540981

List of objects, as suggested by @jeromekelleher:
perl -nae 'BEGIN {print qq{"regions" : [\n}; } print qq[${c}{"referenceName":"$F[0]","start":$F[1],"end":$F[2]}]; $c = ",\n"; END { print "]\n"; }' /tmp/regions.bed.txt | wc
 258824  258826 15221948

List of objects, minified version:
perl -nae 'BEGIN {print qq{"regions" : [\n}; } print qq[${c}{"r":"$F[0]","s":$F[1],"e":$F[2]}]; $c = ",\n"; END { print "]\n"; }' /tmp/regions.bed.txt | wc
 258824  258826 10563134
```
In comparison, the `regions.bed.txt` file was 6163129 bytes (it's only got three columns, so is a region list and not a real .bed file).

jmarshall:
Re _List of region objects vs object of lists_, I very much like the fact that the object form is unordered, so there is (for those who understand JSON!) no question of the ordering of records in the results being influenced by the ordering of the query regions. So we're left with the ordering of records according to the server (i.e. mostly just the ordering in the BAM/VCF file backing the server) being the only ordering in play.

One disadvantage of the _object of lists_ version is that clients have to group together intervals on the same chromosome. So if a client has an arbitrary list of regions of interest, constructing a syntactically-valid region object is somewhat tedious.

daviesrob:
Rebased and variant format support added.

punchmk:
For the long responses, was there any prior discussion around either paging the response or doing asynchronous responses for large requests?

In the paging scenario the first page of data could be returned with a link to the next page and some other metadata. The plus side is you can manage the message size fairly easily by constraining page size, but the downside is that you have to make multiple requests to get the whole response. Additionally providing a consistent view by page of the data set could result in further complexity.

There is a different pattern where for large or long-running requests a 202 could be returned with a link to a resource that would provide access to the final result when it was ready. This would avoid long running requests blocking new requests, and could allow isolation of compute resources for streaming large responses back to clients asynchronously using chunked encoding or other means.

I'm new to the group so hopefully I'm not going too far off on tangents that you may have explored and discarded previously.

jmarshall:
This is :+1: LGTM with some minor concerns:

> `regions` When present, the array must contain at least one region specification.

I've spend too much time with WHAT-WG folks :smile: and think you should spell out what the server should do if it doesn't. e.g. _The server SHOULD respond with an InvalidInput error if the region list is empty or otherwise not well-formed._

> The server SHOULD respond with an InvalidRange error if start and end are specified and start is greater than end. […] The end position must be strictly greater than the start.

The text in the table allows `start==end` while the “Region list” text does not. I reckon this is not the place for 0-length requests so suggest the table text should be changed to agree: _InvalidRange if … start is greater than or equal to end_.

* The first few “Region list” paragraphs duplicate the information in the table. Could they be omitted or shortened?

    The helpful paragraphs about “To return all bases / a single base” could be integrated with the example, e.g. decrement the middle line of the example to

    ```
    … { "referenceName" : "chr2", "start" : 999, "end" : 1000 }, …
    ```
    and say _To return **records covering** [maybe add] a single base, the client should use **an** [fix typo] end position one greater than the start, as shown in the second region which requests the 1000th base of chr2_.

* Not part of your PR but as it mentions POST — the introductory “Design principles” section says

    > The structuring of POST inputs, redirects and other non-reads data will be protobuf3 compatible JSON

    which needs to be updated to say _non-reads/variants_ or something equivalent.

* Good work on fixing the table formatting!

htsget.md
```diff
@@ -393,7 +393,12 @@ The `referenceName`, `start` and `end` parameters are not used for POST requests
 
 `format`  _optional string_
 </td><td>
-Request read data in this format. Default: BAM. Allowed values: BAM,CRAM.
+
+Request read data in this format.
```
jmarshall: s/read data/data/

jeromekelleher:
LGTM also, and I agree with the points @jmarshall makes above.

However, I would vote **against** merging until we have some server and client implementations. One of the great strengths of htsget so far is that it has been implementation-driven, if we lose this principle then I think there's a very real risk of the protocol getting bloated, with partial implementations and poor interoperability.

jmarshall:
Thanks for those updates. Maybe the `end` paragraph needs “or equal to” too?

And when this gets merged, we probably want to think about bumping the spec version number.

htsget.md
```diff
@@ -372,15 +372,167 @@ Initial guidelines, which we expect to revise in light of future experience:
...
+Htsget POST requests must be "safe" as defined in section 4.2.1 of [RFC 7231], that is they must be essentially read-only.
+
+## URL parameters
+
+POST and GET requests should use the same URLs.
+
+## Query parameters
+
+The query parameters are similar to those used in the GET request, formatted into a JSON object.
+The `referenceName`, `start` and `end` parameters are not used for POST requests as they are replaced by `regions` which allows more that one range to be specified.
```
jmarshall: Typo, “than”.

jmarshall:
> ## Query parameters
>
> The query parameters are similar to those used in the GET request, formatted into a JSON object.

When you use POST, there is a JSON object as the body of the request. I think this JSON object contains equivalents of everything that would be written as `?foo=x&bar=y` on the existing GET requests — so is it true that there will never be anything significant to htsget in the URL's`?query` part for POST requests?

If so it would be good to mention that (and perhaps to say that clients SHOULD/MUST not have a
`?query` part in their URLs when making POST requests).

Also I found this section confusing as I thought the HTTP nomenclature was that “query parameter” specifically refers to the `?foo=x&bar=y` parameters in the query part of an URL — if this is true, it would be good to describe these POST body parameters as something else.

jmarshall:
Noted in meetings that “How to specify class in the POST request body needs to be added”.

This seems easy: add a `class` parameter to the table alongside the others, referring to the description of GET `class`.

However: POST requests are for the purpose of requesting more than one genomic range. At the moment, `class` is used only for requesting headers-only, i.e., requesting no genomic ranges at all.

So as long as `header` is the only defined value for the `class` parameter, there is little need to add it to POST — as anything you could do with POST-with-class you can do with GET-with-class.

Alternatively, it may be worth adding it to POST anyway for symmetry and generality reasons.

jmtcsngr:
Hi, #322 looks merged. Anyone is trying to implement these changes in servers/clients? We (sanger) have development time approved to try to build this.

GrishaH:
When implementing the POST method, how strict should the query parameters validations be - are we wanting to accept additional properties?
Shall we reject requests that contain unlisted properties? 
Should we mention this in the spec?

jmarshall:
@GrishaH: That's the question I was getting at in https://github.com/samtools/hts-specs/pull/285#issuecomment-426453605, particularly the square-bracketed part of that comment.

I can imagine scenarios in which particular servers use additional query parameters on these URLs for their own purposes, e.g., as part of authentication or in selecting between different databases, or just as a byproduct of [UTM-style gobbledegook](https://en.wikipedia.org/wiki/UTM_parameters). IMHO that is entirely appropriate and htsget implementations should happily ignore query parameters that have no meaning for them.

At present the spec is agnostic on this, which to my mind means it allows them. Perhaps it should be mentioned explicitly one way or the other.

The question applies to GET requests as well as POST.

daviesrob:
I agree, implementations should ignore parameters that they don't understand.  Although I'd be up for banning the UTM ones on principle :wink:

jmtcsngr:
Is it worth trying to document expected behaviour when limits in body length or number or regions are hit? We will very likely set a limit in the implementation. So will be interesting to see if we agree about error codes or error messages. This is something the client can easily overcome if the original query is split into multiple queries.

jmtcsngr:
We are playing with JSON Schema for validation of the post body. We have found it helpful as it is a less ambiguous way to describe and process the properties of the query. Do you think is worth adding a JSON Schema to this PR?

Now that the OpenAPI PR has been merged we also need to bring it up to date with the POST version I guess. Completely ignorant about it, but I think it should be possible to transpile a section of the open api yaml into a JSON schema.

jmarshall:
Now that there is an extant (server) implementation so this PR is no longer blocked by that, here is a summary of the still outstanding minor editorial issues previously identified in conversation on this PR:

* Questions about maximum size of POST requests (https://github.com/samtools/hts-specs/pull/285#issuecomment-532658543): HTTP doesn't specify this but does have status code 413 for the purpose, which we ought to add to our table of errors (https://github.com/samtools/hts-specs/pull/285#issuecomment-368042187)

* Is this JSON text protobuf3-compatible? (cf “Design principles” section and https://github.com/samtools/hts-specs/pull/285#issuecomment-404214218)

* `format` table entry: s/read data/data/ (https://github.com/samtools/hts-specs/pull/285#discussion_r201745971)

* Mention that POST requests should/must not have htsget-defined query parameters in the `?query` part of their URLs; avoid confusion by renaming section to “Payload parameters” to distinguish from what HTTP and htsget call “query parameters”. (https://github.com/samtools/hts-specs/pull/285#issuecomment-426453605 and following)

* Decide whether to add `class` to the POST request (https://github.com/samtools/hts-specs/pull/285#issuecomment-465621119)

* Since this PR was started, an OpenAPI description of the protocol has been added. We should probably add `post:` operation sections to _pub/htsget-openapi.yaml_.

Also additional items not previously noted:

* `notags`: Remove copy/pasted _If an empty array is specified, no tags are included_ sentence — an empty array here should mean the same thing as omitting `notags`. (Or change it to _…no tags are excluded_ but the sentence is pretty redundant IMHO.)

* Organisation: the PR adds the POST section to the bottom of the document. In particular, it's below the _Reliability & performance considerations_ and _Security considerations_ but these things apply to POST requests too.

   One suggestion would be to insert it just before the _Response_ section instead (or probably even better, just before _Field filtering_). Add to the _Request :: Methods_ intro section to list `POST /reads/<id> // POST /variants/<id>` too, clarify that _GET Query parameters_ is for GET and the new _POST Payload parameters_ section is for POST, and then the _Response_ section naturally is the same response for both types of request.

daviesrob:
Rebased with extra commits to attempt to address issues raised by @jmarshall.

I haven't tried to update the openapi specification yet.

mlin:
Re the protobuf3 JSON constraint -- this is somewhat historical, and today I'd reinterpret it around OpenAPI given how widely that's used throughout the GA4GH spec groups. Basically we don't want our request/response schemas to be problematic to express in OpenAPI. Does anyone have a strong reaction? I can prepare a small separate PR on htsget to that end. cc @jb-adams

mlin:
@jb-adams any comments/concerns on this PR based on your recent implementation efforts? Thanks!

jb-adams:
thanks @mlin just had another read through. I see my questions about `regions` are addressed, that is, a requested duplicate/overlapping region will be returned only once, and in the order they appear on the original file. Our reference server will likely have to do some pre-processing of the regions (sorting, collapsing overlaps) before returning the ticket.

We should add the `POST` request to the OpenAPI yaml as well, as @jmarshall mentioned. I'm happy to do this, I could work on it on my own fork and merge it here. I'll wait for confirmation before starting.

Lastly, should we rebase this branch onto `samtools:master` given the recent merge? We could either do this now or wait for the `service-info` PR to merge into `master` as well.

daviesrob:
Rebased again.  @jb-adams Feel free to push the openapi specification changes to this branch.

jb-adams:
thanks @daviesrob will check out a branch now

jb-adams:
[PR #2](https://github.com/daviesrob/hts-specs/pull/2) in @daviesrob 's fork adds the OpenAPI specification updates

daviesrob:
@jb-adams Thanks, I've included your openapi updates here.

mlin:
I just made a pass through the [most recent checklist](https://github.com/samtools/hts-specs/pull/285#issuecomment-603988382) from @jmarshall (thanks as always) and everything appeared complete to me but for bumping the version number to 1.3.0, if not 2.0.0!

mlin:
https://github.com/samtools/hts-specs/pull/493/files which is also ~ready for merge does bump to 1.3.0 so we just need a bit of coordination there

jmarshall:
Bumping to 2.0.0 could be interpreted by implementers as a signal that the existing GET request has changed in incompatible ways — which it hasn't.

The current version is 1.2.0. IMHO the version that adds both the service-info endpoint and POST requests should be 1.3.0. (There would be various ways to organise that: e.g. merge service-info without bumping the version or bumping it to 1.2.1; or merging the two commits with a merge commit that itself bumps the versions numbers from 1.2.0 to 1.3.0.)

daviesrob:
@mlin Did you intend to merge master into here?  I find it's usually cleaner to rebase as the resulting history is less tangled.

mlin:
@daviesrob My apologies, I'd just done the service-info merge so I thought I'd also try to clean up the trivial conflicts which GitHub then raised here, without bothering you. If you'd like to rollback my merge commit, rebase, bump the version to v1.3.0, and perform the squash-merge into master (!), that'd be most welcome!

daviesrob:
I'll take a look.  Probably best if I squash and rebase to a smaller number of commits, and leave it to you to merge.

daviesrob:
Hmm, that didn't quite work, I was supposed to end up with two commits.  Just a sec...

daviesrob:
That's better.  There's now a commit from me that adds POST to `htsget.md` and one from @jb-adams that adds it to the OpenAPI YAML file.  The rebased files exactly match e2575ea.

I assume any version number changes will go in another commit?

Issue #286, opened 2018-02-21
SAM Read Group (@RG) is SM required?


blankenberg:
Looking at the original spec ```SAM1.pages``` spec file (e.g: https://github.com/samtools/hts-specs/commit/dbdec8b0405fa2c4a4708a91722168c7e654bbab / https://github.com/samtools/hts-specs/raw/dbdec8b0405fa2c4a4708a91722168c7e654bbab/SAM1.pages) both ```ID``` and ```SM``` are listed as being required (pages formatted file removed mid 2013: https://github.com/samtools/hts-specs/commit/78aa8051579806832138692c6ec8f5d1abc054a6):
![image](https://user-images.githubusercontent.com/1231190/36502411-dfdfbdce-1717-11e8-8b3a-4ed35c3aa183.png)
however, when the LaTeX version appeared (https://github.com/samtools/hts-specs/commit/07dc1c67a717a6c5cc9d65eeb8b3c99612744cde) this requirement did not appear.

Certain tools, e.g. Picard Tools, require ```SM``` tag in read groups, which matches with the original spec.

Should ```SM``` be required or not?

bioinformed:
I'm trying hard not to be snarky, but this is what happens when we treat standards as a moving target.  

These documents affect clinical care (among other things that matter).  So can we be rational and version our specification documents and the resulting files properly?

blankenberg:
@bioinformed good point about versions. The specs are versioned and SAM/BAM files themselves can be 'optionally' versioned with the ```@HD``` record (```VN``` is required when ```@HD``` is specified). 

Looking closer at the pages file, its version is listed as```Version 0.1.2-draft (20090702)``` whereas the LaTex spec appeared as ```v1.3 draft```. 

However, this would be a rather significant specification change that is not mentioned in the version change history appendix, e.g. from version 1.5 (current, https://github.com/samtools/hts-specs/blob/7326e54db05075707fe133d003297681a5313c52/SAMv1.tex ):
```
\subsection*{1.3: July 2010 to April 2011}

\begin{itemize}
\item Re-add {\tt CC} and {\tt CP} auxiliary tags. (Mar 2011)
\item Add CIGAR N intron/skip operator. (Dec 2010)
\item Add {\tt BQ} BAQ tag. (Nov 2010)
\item Add {\tt RG PG} header field. (Nov 2010)
\item Add BAM description and index sections. (Nov 2010)
\item \textbf{Removal of FLAG letters.} (July 2010)
[...]

PR #287, opened 2018-02-22
Add changed requirement of @RG SM to version history

blankenberg:
xref https://github.com/samtools/hts-specs/issues/286

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -1176,6 +1176,7 @@ specification with its own version numbering.\footnote{
 \item Add {\tt RG PG} header field. (Nov 2010)
 \item Add BAM description and index sections. (Nov 2010)
 \item \textbf{Removal of FLAG letters.} (July 2010)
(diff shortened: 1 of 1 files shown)


jkbonfield:
Looks like a harmless enough change, although I wonder if technically this was a 1.3 update or somewhere in the pre-1.3 era.  Heng's date in the email though appears to match the 1.3 release.

jmarshall:
Nothing was publicly released between the last Pages document and these first LaTeX drafts in July 2010. This optional-SM text first appears in the second draft that Heng sent to the mailing list on 15 July 2010; both drafts that day identify themselves as “v1.3 draft”.

So the timing is right, but something like “The `SM` header field, previously mandatory for `@RG`, is now optional” would be clearer.

jmarshall:
Thanks — merged as 7df95c9c1b2d75a62fca3eceffb27d1cb7bd54f1.

PR #288, opened 2018-03-01
Changed gVCF format broken link

### LINKED ISSUE #225: VCF 4.3: broken link for gvcf in section 5.5
this link is broken https://support.basespace.illumina.com/knowledgebase/articles/147078-gvcf-file. I can't know what the original document was, but maybe this is the updated version https://help.basespace.illumina.com/articles/descriptive/gvcf-files/

cyenyxe:
Replaces broken link https://support.basespace.illumina.com/knowledgebase/articles/147078-gvcf-file with https://help.basespace.illumina.com/articles/descriptive/gvcf-files/

Fixes #225

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -1253,7 +1253,7 @@ Haplotype quality & 30 & 40 & 40 \\
 \subsection{Representing unspecified alleles and REF-only blocks (gVCF)}
 In order to report sequencing data evidence for both variant and non-variant
 positions in the genome, the VCF specification allows to represent blocks of reference-only calls in a single record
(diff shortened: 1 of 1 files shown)


ctsa:
BTW, if you're looking for a more reliable URL for this purpose, the first stable reference to gVCF as a clinical endpoint format was in the Isaac bioinformatics paper https://doi.org/10.1093/bioinformatics/btt314. Several posters/talks were presented earlier in 2012 but I don't think any of these were captured to a permanent link.

cyenyxe:
@ctsa My only concern about the DOI link is that the reference to gVCF can't be found easily, unless searching specifically for it. If the structure of the knowledge base changes again, then I agree we may have to use the DOI instead of a direct link.

ctsa:
Thanks, that sounds reasonable, I agree there isn't really an optimal choice here.

PR #289, opened 2018-03-01
INFO key 1000G is allowed as a special legacy value

### LINKED ISSUE #164: 1000G reserve key appears to break with new regular expression
I noticed that in the current version of the VCFv4.3 spec in section 1.6.1 (8) it has a new regular expression for INFO keys that enforces that the first character is alphabetical but on the same page it lists 1000G key as being one of the reserved keys. This appears to be invalid according to the new character restrictions.

cyenyxe:
1000G does not match the regular expression for INFO keys but needs to be supported for legacy reasons.

Fixes #164

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -331,7 +331,7 @@ There are 8 fixed fields per record.  Fixed fields are:
   \item INFO - additional information: (String, no semi-colons or equals-signs permitted; commas are permitted only as delimiters for lists of
   values; characters with special meaning can be encoded using the percent encoding, see Section~\ref{character-encoding}; space characters are allowed)
   INFO fields are encoded as a semicolon-separated series of short keys with optional values in the format: $<$key$>$=$<$data$>$[,data].
-  INFO keys must match the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, although ``1000G'' is allowed as a special legacy value. Duplicate fields are not allowed. Arbitrary keys are permitted, although the sub-fields listed in Table~\ref{table:reserved-info} are reserved (albeit optional).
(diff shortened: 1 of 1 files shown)


VCFv4.3.tex
```diff
@@ -331,7 +331,7 @@ \subsubsection{Fixed fields}
   \item INFO - additional information: (String, no semi-colons or equals-signs permitted; commas are permitted only as delimiters for lists of
   values; characters with special meaning can be encoded using the percent encoding, see Section~\ref{character-encoding}; space characters are allowed)
   INFO fields are encoded as a semicolon-separated series of short keys with optional values in the format: $<$key$>$=$<$data$>$[,data].
-  INFO keys must match the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, duplicate fields are not allowed. Arbitrary keys are permitted, although the sub-fields listed in Table~\ref{table:reserved-info} are reserved (albeit optional).
+  INFO keys must match the regular expression \texttt{\^{}[A-Za-z\_][0-9A-Za-z\_.]*\$}, although ``1000G'' is allowed as a special legacy value. Duplicate fields are not allowed. Arbitrary keys are permitted, although the sub-fields listed in Table~\ref{table:reserved-info} are reserved (albeit optional).
```
lbergelson: It might be best to add 1000G as an or clause in the regex.
Something like: `^([A-Za-z ][0-9A-Za-z .]*|1000G)$`
Then I would still include the note that 1000G is included as a legacy value.  It would let people use the regex to validate without any caveats which seems nice.

If it's kept this way, I would probably change the wording slightly.   Maybe "..., except for the key "1000G" which is allowed as a special legacy value."

cyenyxe:
I have made the change suggested by @lbergelson

Issue #276, opened 2018-01-05
A new SAM tag for transcript strand


lh3:
Tophat and many RNA-seq mappers output an "XS:A" in SAM. It is either "+" or "-", indicating the putative strand of the transcript the read is sequenced from. It is so widely used and effectively becomes a standard tag in the RNA-seq world. Do we want to add a "TS" (or "ST"; both are available) tag for the same purpose? In practice, I am not sure how many will use the new tag, but better late than never. If you are ok with this, I can send a pull request later.

dpryan79:
To my knowledge, only cufflinks and stringTie have ever made use of the "XS:A" tag. I'm not sure it's worth occupying a tag just for them.

lh3:
I expect for long RNA-seq reads, this tag will be used much more often, for example to generate BED12 and GFF. Some users are requesting it (lh3/minimap2#88). In the old days, when we map ESTs, transcript strand is critical information when is unknown.

yfarjoun:
can this information not be determined form the read ordinality and the mapping orientation?

lh3:
What's read ordinality? Either way, for non-stranded libraries, transcript strand can't be inferred.

jkbonfield:
Even if it doesn't get a large amount of use, the fact that more than one program wishes to generate this tag and that users have requested it makes me think it should be accepted.

I'm happy with either TS or ST as a tag type.

lh3:
Closed via PR #292.

PR #292, opened 2018-03-05
added the TS tag for "transcript strand"

lh3:
TS means: "Strand of the transcript the query sequence represents. Valid values: '+', '-' or '.' (for unknown)." See also #276.

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -114,6 +114,7 @@ or
   {\tt SM} & i & Template-independent mapping quality \\
   {\tt SQ} & ? & Reserved for backwards compatibility reasons \\
   {\tt TC} & i & The number of segments in the template \\
+  {\tt TS} & A & Transcript strand \\
   {\tt U2} & Z & Phred probability of the 2nd call being wrong conditional on the best being wrong \\
   {\tt UQ} & i & Phred likelihood of the segment, conditional on the mapping being correct \\
   {\tt X?} & ? & Reserved for end users \\
@@ -227,6 +228,9 @@ Template-independent mapping quality, i.e., the mapping quality if the read were
 
 \item[TC:i:\tagvalue{}]
 The number of segments in the template.
+
+\item[TS:A:\tagvalue{strand}]
+Strand (`{\tt +}' or `{\tt -}') of the transcript to which the read has been mapped.
 
 \item[U2:Z:\tagvalue{}]
 Phred probability of the 2nd call being wrong conditional on the best being wrong.
@@ -473,6 +477,11 @@ This appendix lists when standard tags were initially defined or significantly c
 
 \setlength{\parindent}{0pt}
 \newcommand*{\gap}{\vspace*{2ex}}
+
+\subsubsection*{March 2020}
+
+Transcript strand tag TS added, equivalent to the locally-defined XS tag
+produced by several RNA aligners.
 
 \subsubsection*{January 2019}
 Added the OA tag for recording original/previous alignment information.


jkbonfield:
Thanks for the PR Heng.

Your comment here is '?' for unknown (used in GFF and also the PT:Z: tag) but your PR is '.' (used in VCF world).  Which do you want?

Arguably neither as lack of tag means it is unknown, but do you want the ability to *explicitly* state that we do not know this information?

Also, personally I think "Strand of the transcript THAT the query sequence represents" reads better.  Any others have comments on this?

lh3:
GFF uses '.' for missing. I will stick with that, so use '.' for missing. I prefer an explicit missing value as missing tags is sometimes annoying.

> Also, personally I think "Strand of the transcript THAT the query sequence represents" reads better. Any others have comments on this?

Yes. Added "that".

lh3:
Hmm... Having an explicit missing value leads to two ways to represent "missing": 1) by not specifying the tag at all; 2) by using ".". This may be an issue. I will just follow the consensus/majority. Don't have a strong opinion either way.

jkbonfield:
GFF3 uses "." for data that is unstranded, not missing:

> Column 7: "strand"
> The strand of the feature. + for positive strand (relative to the landmark), - for minus strand, and . for features that are not stranded. In addition, ? can be used for features whose strandedness is relevant, but unknown."

My inclination though is just to have +/- and avoid the unknown scenario unless there is a good reason why we need to explicitly distinguish between no-data-unknown and have-data-but-still-unknown.

lh3:
That is fair. I have removed an explicit missing value.

jmarshall:
The point of the SAMtags document is that we have plenty of space for explanations outside the table of tags.

So just put `{\tt TS} & A & Transcript strand \\` in the table and add the rest of the text and more to a `\item[TS:A:\tagregex{[+-]}]` paragraph in the _Additional Template and Mapping data_ section.

SAMtags.tex
```diff
@@ -213,6 +214,9 @@ \subsection{Additional Template and Mapping data}
 \item[TC:i:\tagvalue{}]
 The number of segments in the template.
 
+\item[TS:A:\tagvalue{strand}]
+Strand (`{\tt +}' or `{\tt -}') of the transcript that the query sequence represents.
```
yfarjoun: Does a query sequence "represent" a transcript? If so, perhaps add another sentence to explain what this means to you.

jkbonfield:
This has been rotting for over a year.   My recollection was that we deemed it a bit minimal in wording, given the SAMtags document now gives much more room for a fuller description.

However we all agreed in principle it was fine.

If there are no objections, I'll merge this as is and we can open a new PR to revise the wording if someone feels inclined to phrase it more carefully.  We shouldn't just sit on things indefinitely and block progress just because it's not perfect.

jmarshall:
I have been working on this this week in anticipation of tomorrow's meeting.

jmarshall:
> However we all agreed in principle it was fine.

As noted in the meetings when it has been discussed, in principle it is fine but at present it does not give enough guidance as to what TS would be supposed to _mean_ for tools to implement it equivalently.

jkbonfield:
Yes, but even so having TS bookmarked as meaning transcript strand in the specification, even if a bit vague, is better than omitting it given it's likely in use already. (I haven't checked, but there is a trend for such behaviour that we all know!)

So the principle is still clear.  If the PR is OK in principle but needs some polishing that we don't realistically have time, we should merge and raise "X needs some polish" as an issue for us to address at a later stage.  Incremental improvements are better than PRs languishing for a full year.

jmarshall:
Thank you for the reminders. As you know, after the [2018-12-13 meeting](https://docs.google.com/document/d/1VZspg-j6-9ewf6uRWwBoXqyFpjbBbr22qXpD8_zbIOA/edit#heading=h.d9w45ptmulw7) dealing with this PR is assigned to me. Let's discuss it further at the meeting tomorrow, whether that's approving the rebased updated version or changing who this PR is assigned to.

yfarjoun:
👍

cmdcolin:
The merged PR says that TS is equivalent to XS however it is sometimes (or always) not ? See https://github.com/lh3/minimap2/issues/88#issuecomment-355101901

With TS from minimap for example, I have to flip the value of TS depending on whether the read is positive or negative strand, but with XS my experience was that I did not have to

Not sure if this level of detail is needed for the specs for TS but it was a gotcha for me when coding for both TS and XS and I remain to flip the TS value based on whether the read is positive or negative strand due to the behavior of minimap, and not sure how the nonstandardized status of XS affects this

jmarshall:
As noted in the comment you linked to, minimap outputs its own `ts` tag, which is different from `XS` and `TS`.

What has been standardised is intended to be the same behaviour as `XS`. The text here is essentially the same description as the [HISAT2 manual](https://ccb.jhu.edu/software/hisat2/manual.shtml#sam-output) gives for its XS.

cmdcolin:
Oh I see. Didn't realize it was a lower case ts tag in this case.

jmarshall:
Yeah, we'd need @lh3 to remind us why minimap2 outputs a subtly differently defined value for a seemingly similar transcript concept. And what his current thoughts on the most useful transcript-related value to have in your SAM file are.

Also I would love to do some more testing of HISAT2 or another RNA aligner and observe exactly what it does with reads and transcripts of various orientations, with a view to further clarifying the description in the specification. As a non-specialist, constructing some appropriate test data is sadly low down my to-do list.

jkbonfield:
From @lh3

> XS is a custom tag. I will make a proposal to the SAM spec before implementing anything. Deriving XS from ts is straightforward. This should not stop users now.

There is some ambiguity in the above statement as to whether the SAM spec proposal was intended to be a copy of XS or ts.   However given Heng was aware of differences between ts and XS at the time, and this PR was infact made by him it would appear matching XS is desired and deliberate.

Or putting it another way, it doesn't appear this PR was made and then a new ts was added later.  Phew. :-)

Issue #206, opened 2017-05-11
SAM SA tag: 0 or 1 based coordinates?


dpryan79:
This question originally comes [from biostars](https://www.biostars.org/p/252281/#252300), but it would be nice to have this clarified in the spec, lest different aligners start outputting off-by-one values from one another.

PR #294, opened 2018-03-05
SAM SA tag: make rname and strand requirements explicit.

jkbonfield:
Clarified SA tag following discussions on the OA tag PR https://github.com/samtools/hts-specs/pull/193

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -207,6 +207,7 @@ for any associated quality values.
 Other canonical alignments in a chimeric alignment, formatted as a semicolon-delimited list.
 Each element in the list represents a part of the chimeric alignment. Conventionally, at a supplementary line, the first element points to the primary line.
 \emph{Strand} is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to FLAG bit 0x10.
+\emph{Pos} is a 1-based coordinate.
 
 \item[SM:i:\tagvalue{}]
 Template-independent mapping quality.
@@ -301,7 +302,7 @@ Deprecated alternative to {\tt BC} tag originally used at Sanger.
 Original CIGAR, usually before realignment.
 
 \item[OP:i:\tagvalue{pos}]
-Original mapping position, usually before realignment.
+Original 1-based mapping position, usually before realignment.
 
 \item[OQ:Z:\tagvalue{qualities}]
 Original base quality, usually before recalibration.
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -206,6 +206,7 @@ for any associated quality values.
 \item[SA:Z:\tagregex{{\tt (}\emph{rname}{\tt ,}\emph{pos}{\tt ,}\emph{strand}{\tt ,}\emph{CIGAR}{\tt ,}\emph{mapQ}{\tt ,}\emph{NM}{\tt ;)}+}]
 Other canonical alignments in a chimeric alignment, formatted as a semicolon-delimited list.
 Each element in the list represents a part of the chimeric alignment. Conventionally, at a supplementary line, the first element points to the primary line.
+\emph{Strand} is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to FLAG bit 0x10.
 
 \item[SM:i:\tagvalue{}]
 Template-independent mapping quality.
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -487,7 +487,7 @@ Each bit is explained in the following table:
   next read.  If {\sf
     RNEXT} is `*', no assumptions can be made on {\sf PNEXT} and bit
   0x20.
-\item {\sf PNEXT}: Position of the primary alignment of the NEXT read in the template. Set as
+\item {\sf PNEXT}: 1-based Position of the primary alignment of the NEXT read in the template. Set as
   0 when the information is unavailable. This field equals {\sf POS} at the primary line of
   the next read. If {\sf PNEXT} is 0, no assumptions can be made on
   {\sf RNEXT} and bit 0x20.


jkbonfield:
There were some comments raised on this at the GA4GH conference call, but I can't see all of them in the minutes.

From what I recall we had issues with = being disallowed here.  Technically it's perfectly valid in SA as the alignment may be the same reference as the primary, however pragmatically it means if we remap a read then we also need to modify the SA tags so perhaps it's safer to not use =.  Does this make it a recommendation rather than a rule?

There were also concerns about using the word strand instead of explicit hex flag numbers, although this is a separate issue as that wording was in there already.  For reference, the flags part of the spec talks only about "reverse complemented".  Maybe we should keep strand as a short hand for the one-line description with explicit explanation in the text?  We still need to keep '+' and '-' though as that's how it is currently implemented, not by numeric flag values.

SAMtags.tex
```diff
@@ -206,6 +206,8 @@ \subsection{Additional Template and Mapping data}
 \item[SA:Z:\tagregex{{\tt (}\emph{rname}{\tt ,}\emph{pos}{\tt ,}\emph{strand}{\tt ,}\emph{CIGAR}{\tt ,}\emph{mapQ}{\tt ,}\emph{NM}{\tt ;)}+}]
 Other canonical alignments in a chimeric alignment, formatted as a semicolon-delimited list.
 Each element in the list represents a part of the chimeric alignment. Conventionally, at a supplementary line, the first element points to the primary line.
+\emph{Rname} should be stated explicitly and is not permitted to be `{\tt=}'.
```
yfarjoun: should->must (I think this is what we said in the past about language...right?)

jkbonfield: I think "must" makes more sense in this sentence as "should" is simply a recommendation and is at odds with the "not permitted" bit later on (in RFC2119, although we don't follow this for the SAM spec).

However there was discussion at the meeting which made me wonder if it's valid anyway.  It is completely valid, and the only concerns are about remapping some but not all reads, but arguably that is an issue of the remapper not updating fields correctly.  If we want to make it so that is easy to do, then we should also be banning filling out TLEN, PNEXT, RNEXT, etc as they all have similar problems with remapping of partial data.

Comments?

yfarjoun: I think that we do not need to make reverting to the original data __easy__, we just need to make it __possible__ (without re-running the mapper). which is why I do not find it necessary to record the original TLEN, PNEXT, RNEXT as they are recoverable.

jkbonfield: Note this is SA tag and not OA tag.  The comment about = only applies if we edit the primary (or "previous") read.  In editing such data though we're already in a world of hurt when it comes to updating lots of fields, so I don't think = vs explicit is going to cut it.

I'll drop that sentence.

jmarshall:
The strand-related concern was not so much about explicit hex flag numbers, but about tying the field back to the relevant flag field — as the term _strand_ is not particularly defined as something represented in the file in the SAM specs. Rather, the thing that's represented in the file is the reverse-complemented flag.

It is a bit of a shame the spec doesn't give any short names for the various flags, and §1.4 and §3 have long been reduced to talking about “Bit 0xWHATEVER”.

> _Strand_ should be '`+`' to indicate positive strand and '`-`' to indicate negative strand, corresponding to FLAG 0x10 being clear or set respectively.

Wordsmithy suggestions: this is discussing what the value **is** rather than how implementations might behave, so [just use _is_ rather than _should_/_must_](https://github.com/samtools/hts-specs/pull/230#discussion_r131352646); I think you mean _or_ rather than _and_.

Otherwise this is fine though in context it's a bit heavy: we're just telling people what the expected values are, we don't need to explain what the signs mean :smile:. If it was me, I'd just put something like

> _Strand_ is either '`+`' or '`-`', indicating positive/negative strand and corresponding to flag bit 0x10.

Also I just noticed #206 which asks us to make it explicit that POS here is 1-based.

jkbonfield:
I'm happy with the revised wording and will update.

Regarding POS, yes we could make it explicit here, but it feels out of place.  POS is 1 based throughout SAM so why single out this one specific location for changing?  Looking at the others though it's not often referred to though so fixing everything isn't hard.  If we wanted to make POS explicitly 1 based here then I think we should also add it to the PNEXT description in SAM.tex too and to the OP tag in this doc.

jkbonfield:
Rebased, squashed and pushed updates with the 1-based change as a separate commit so it's easy to drop if we don't want it.

In the end I decided against updating PNEXT as the text explicitly mentions it should match POS of the other read, which should be sufficient to clarify 1-based vs 0-based.  So I only updated SA and OP tags here. (PT already had it.)

jmarshall:
Great :smile:

Re POS, I agree I would like to think it's obvious, particularly as the glossary almost says SAM is 1-based throughout. However 1) this subfield appears as text in BAM too which is otherwise mostly 0-based, and 2) people have asked the question because they find it unclear, so it behooves us to clarify it…

[New commits arrive…]

:+1: though while I take your point about the circumstantial evidence in the PNEXT description, IMHO we might as well be explicit in the PNEXT `\item` too.

jkbonfield:
OK, in for a penny in for a pound.  SAMv1.tex updated now too. :-)

Ping @yfarjoun.

jmarshall:
The table is an **overview** so doesn't really need this detail there, but either way.

I suppose “Position” should stay capitalised so along with “NEXT read” the capitals still spell out PNEXT :vomit:

jkbonfield:
Updated.   Agreed the overview doesn't need the 1-based bit (I was unsure and added it only because POS has it, but it's extra verbage that is best dealt with in the explanatory section instead).

yfarjoun:
👍

PR #306, opened 2018-04-24
Fix types in documentation of array tags

klmr:
Optional fields of array type are specified via `S:‹type›,`. This was done correctly in one place (for the `CG` tag), but there was a typo for the `FZ` tag (resulting in not `S,‹type›:`). In addition, the overview table listed the types in a confusing way (separated by comma rather than colon).

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -60,7 +60,7 @@ or
   {\tt BQ} & Z & Offset to base alignment quality (BAQ) \\
   {\tt BZ} & Z & Phred quality of the unique molecular barcode bases in the {\tt OX} tag \\
   {\tt CC} & Z & Reference name of the next hit \\
-  {\tt CG} & B,I & BAM only: {\sf CIGAR} in BAM's binary encoding if (and only if) it consists of $>$65535 operators \\
+  {\tt CG} & B:I & BAM only: {\sf CIGAR} in BAM's binary encoding if (and only if) it consists of $>$65535 operators \\
   {\tt CM} & i & Edit distance between the color sequence and the color reference (see also {\tt NM}) \\
   {\tt CO} & Z & Free-text comments \\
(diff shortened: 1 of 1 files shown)


jmarshall:
Thanks for the typo fix.

Sadly the array types have been referred to as B,I and so on since they were introduced, even though in a table that looks a lot like it's saying the item is either of type B or of type I.

Colon is probably an improvement (thoughts, other maintainers?), in which case there are three instances of it in _SAMv1.tex_ too.

jkbonfield:
It's confusing regardless. Eg with `FZ:B:i,105,204,0,1,412,...` technically given the stated syntax of TAG:TYPE:VALUE the TYPE is `B`, not `B:i` nor `B,i` with the `i,` being added to the start of the VALUE.   The `B,i` notation has really been a shorthand for this, but perhaps it needs spelling out explicitly.

Furthermore if we state B:i is the TYPE then it sort of implies we don't precede VALUE by the code, which is also misleading.  Again, it could be clarified by expicitly explaining this to be short-hand.

I think therefore we've painted ourselves in a corner somewhat and whatever we say is misleading and must be clarified by appropriate example.  We have these already in the SAMaux document in the expansions of each aux type.  I'm rather loathe to change things now given the alternative has a different set of problems and doesn't technically fit our spec either.

klmr:
Fair enough, especially since it concerns only the table (this could be clarified by a footnote on the table anyway). That said, [the patch contains another correction that’s similar but unrelated](https://github.com/samtools/hts-specs/pull/306/files#diff-b768745dd64e78b732dc83c17eea56b4R357). Should I create a new PR for that or is that overkill?

jmarshall:
I've applied that part as 5565ac3ee3c82b0a463be1511b6303b4ba28fbd1.

PR #307, opened 2018-05-02
Bump version to VN:1.6 due to long CIGAR strings

jmarshall:
PR #227 was merged as dab57f44a4d5358e53d9f75a7d6c6ab2924451b3 on 28 November 2017, but the [discussion about bumping the version number](https://github.com/samtools/hts-specs/pull/227#issuecomment-342091422) to indicate the presence of this feature was never returned to:

> [@jkbonfield]
> I'm tempted however to suggest that we ought to bump the SAM version number, even though this is technically a BAM only change. Why? Because there is no BAM version number other than the sledgehammer complete format change and there is the risk that this BAM only change will leak into SAM anyway, hence having a fixed version where we know it can occur (or vice versa, where we know it cannot) acts in a similar way to CG header tag, but more useful IMO. Plus, like it or not, the BAM and SAM formats have been inextricably linked in the same document since day 1.

The SAM version number was bumped to 1.4 to signal the existence of B arrays; to 1.5 to signal the existence of the SUPPLEMENTARY flag bit. IMHO this new CIGAR representation warrants the same treatment: the new tag field will be unnoticed by older code, which will see the placeholder
CIGAR string — so it needs to be possible to signal the presence of CG tags via the `@HD-VN` version number header field.

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -424,7 +424,11 @@ Cellular barcode tags CB, CR, and CY added.
 
 \subsubsection*{November 2017}
 
-CG tag added.
+SAM version number {\tt VN:1.6} introduced, indicating the addition of the CG tag representation of very long CIGAR strings.
+Files that contain records with more than 65,535 CIGAR operators should not declare a version number lower than~1.6 in their {\tt @HD} headers.
+% Technically only BAM files containing records with CG tags need to avoid
+% declaring VN<1.6, but recommending that SAM and CRAM files with long CIGAR
+% strings also declare VN:1.6+ aids file format conversion.
 
 \subsubsection*{August 2017}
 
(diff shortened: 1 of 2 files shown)


jkbonfield:
The SAMtags.tex document also mentions CG:

```
B,I & BAM only: {\sf CIGAR} in BAM's binary encoding if (and only if) it consists of $>$65535 operators
```

That document also has a version number, although it's hidden away right at the end instead of the start (we ought to fix that at the same time to be honest).  So if we're bumping SAM spec then we should bump SAMtags in sync given it's the same addition appearing in both.

It's still a tricky one though and I can see both sides of the argument, but the *concept* is new even if it doesn't directly change SAM itself - just BAM - so I agree with bumping the version and it's unfortunate we forgot about this at the time.

jmarshall:
Rebased on top of the now-merged PR #312, and a note added to the SAMtags history.

jmarshall:
@jkbonfield — does this address your previous comments sufficiently?

PR #312, opened 2018-05-17
Add full SAMtags chronological history

jmarshall:
List tag additions and changes chronologically, as IMHO the main question readers want to answer is 
_When was the XY tag introduced?_

Remove SAMv1 history items pertaining to tags, as they now just duplicate SAMtags history items.

@jkbonfield @yfarjoun It would be good if we can get this in fairly rapidly, as I have some followups in mind, notably addressing James's comments on PR #307 to note the VN:1.6 bump in the SAMtags history too.

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -411,17 +411,111 @@ to \mailtourl{samtools-devel@lists.sourceforge.net}.
 
 \begin{appendices}
 \appendix
-\section{SAM Tags History}\label{sec:history}
+\section{Tag History}
 
-This lists the date of each tagged SAM version along with changes that
-have been made while that version was current.  
+This appendix lists when standard tags were initially defined or significantly changed, and other historical events that affect how tags are interpreted or what files they may appear in.
 
-\subsection*{1.5: 23 May 2013 to current}
-\begin{itemize}
-\item Cellular barcode tags CB, CR, and CY added. (May 2018)
-\item Add UMI-related tags (RX, QX, OX, BZ, MI) and clarified usage of sample barcode tag BC. (August 2017)
-\item SAMtags.txt (this file) created with tags from SAMv1 
-\end{itemize}
+\setlength{\parindent}{0pt}
+\newcommand*{\gap}{\vspace*{2ex}}
+
+\subsubsection*{May 2018}
+
+Cellular barcode tags CB, CR, and CY added.
+
+\subsubsection*{November 2017}
+
+CG tag added.
+
+\subsubsection*{August 2017}
+
+Unique molecular identifier tags BZ, MI, OX, QX, and RX added.
+
+Usage of sample barcode tag BC clarified.
+
+\subsubsection*{June 2017}
+
+Corrected the description of the E2 tag, which previously wrongly claimed E2 to be encoded as per {\sf QUAL}.
+
+\subsubsection*{September 2016}
+
+Predefined tags, previously listed as a brief table within the main SAM specification, have been split out into this new document.
+There is now space for clearer and more complete tag descriptions.
+
+\subsubsection*{February 2014}
+
+MC tag added.
+
+\subsubsection*{May 2013}
+
+SAM version number {\tt VN:1.5} introduced, indicating the addition of the {\sc supplementary} flag (bit 0x800).
+
+SA tag added.
+
+\subsubsection*{March 2012}
+
+Descriptions of CT and PT annotation tags significantly clarified.
+
+\subsubsection*{October 2011}
+
+Sample barcode tags QT and RT added, with RT being identified as a deprecated alternative to BC.
+
+% These were actually added in late September as RT/PT, but RT was changed to
+% CT (see samtools-devel, "Potential clash of RT tags (annotation vs barcode)",
+% October 2011) before read-annotation-RT appeared in the wild.
+Read annotation tags CT and PT added.
+
+\subsubsection*{September 2011}
+
+% This was actually August 29th, but let's call it September.
+FZ tag's type changed from {\tt H} to {\tt B,S}-array.
+
+BC and CO tags added.
+
+\subsubsection*{April 2011}
+
+SAM version number {\tt VN:1.4} introduced, indicating the addition of the {\tt B}-array tag type.
+Files that contain records with {\tt B}-array fields should not declare a version number lower than~1.4 in their {\tt @HD} headers.
+
+\gap
+FZ tag added, with type {\tt H}.
+
+MD tag description changed to allow IUPAC ambiguity codes in addition to {\tt ACGTN}.
+
+\subsubsection*{March 2011}
+
+CC and CP tags reinstated with their original meanings.
+
+\subsubsection*{November 2010}
+
+BQ tag added.
+
+\subsubsection*{July 2010}
+
+The specification was rewritten as a \LaTeX\ document specifying SAM version number {\tt VN:1.3}.
+
+\gap
+Tags FI, FS, OC, OP, OQ, and TC added.
+
+Tags GC, GQ, and GS added and reserved for backwards compatibility.
+
+Existing CC, CP, and MF tags noted as reserved for backwards compatibility.
+
+\subsubsection*{July 2009}
+
+The original SAM ``0.1.2-draft'' specification specified version number {\tt VN:1.0} and defined a total of thirty standard tags:
+
+\begin{center}
+\begin{tabular}{l@{\qquad}l@{\qquad}l@{\qquad}l@{\qquad}l}
+AM & CS & IH & NM & RG \\
+AS & E2 & LB & PG & S2 \\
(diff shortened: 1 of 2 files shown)


SAMtags.tex
```diff
@@ -411,17 +411,111 @@ \section{Locally-defined tags}
...
+Predefined tags, previously listed as a brief table within the main SAM specification, have been split out into this new document.
+There is now space for clearer and more complete tag descriptions.
+
+\subsubsection*{February 2014}
+
+MC tag added.
+
+\subsubsection*{May 2013}
+
+SAM version number {\tt VN:1.5} introduced, indicating the addition of the {\sc supplementary} flag (bit 0x800).
```
peterjc: I consider the SAM specification v1.5 milestone to also cover the de novo extensions in http://dx.doi.org/10.1101/020024 - can we add that here (with the URL)?

jmarshall: That version number bump entry is there to mark (theoretical) compatibility concerns (see the VN:1.4 entry for a real example). Rather to my surprise, the CIGAR P operator was already in VN:1.0; are there other incompatibilities?

However I see your point and will reread that paper and fit something in. The URL would probably be even better in the main section (e.g. SAMv1.pdf's Guide for Describing Assembly Sequences in SAM) rather than this appendix.

jkbonfield: Our version control is a bit untidy.  CO and RT tags weren't there at the start of 1.4 but where there quite early into the 1.4 cycle and so were there at the end.  So while you could say 1.5 is the first release they were in from the start, we deem them to be a 1.4 feature.  Maybe it needs clarification for the CT/PT section though with a link in the "Describe padded alignments" comment (2011).

PS. Generally any behaviour that you *must* understand warrants a version update, rather than simply having the ability to ignore the new fields (tags mostly fall into this feature).  A new flag did (SUPPLEMENTARY), as did a complete new type of tag as the BAM parsing breaks ("B"), and belatedly as we'll update version for >= 65536 CIGAR entries.

peterjc: Thanks.

And yes, from memory the CIGAR P operator was already there, but this was may have been the first attempt to document how it could/should be used - does that sound about right @jkbonfield ?

peterjc: That description of how and when the version was bumped makes sense to me now, it seemed a lot more fluid / unclear to me back at the time.

jkbonfield:
I haven't gone through it carefully yet (will do), but I concur with the basic idea of moving tag history out of SAMv1.tex and in to SAMtags.tex.  They fit much better there as I doubt we've ever be wanting to actually validate SAM version numbers against tag types - unless we've ever actually redefined a tag.  (I hope not.)

SAMtags.tex
jkbonfield: Yes, CIGAR P was there from very early on and there is even an example of it in the original samtools paper, complete with the expected output from tview.  Imagine my surprise when I discovered that samtools doesn't actually render the alignment in the paper as advertised!

jmarshall:
Thanks for reviewing.

I'm making some tweaks to the CT/PT tags section and Padded SAM section (whose cross-references to each other haven't been updated for the SAMv1/SAMtags split, oops) so will add add a link to the paper etc along with that.

SAMtags.tex
```diff
@@ -411,17 +411,111 @@ \section{Locally-defined tags}
...
+
+\subsubsection*{August 2017}
+
+Unique molecular identifier tags BZ, MI, OX, QX, and RX added.
+
+Usage of sample barcode tag BC clarified.
+
+\subsubsection*{June 2017}
+
+Corrected the description of the E2 tag, which previously wrongly claimed E2 to be encoded as per {\sf QUAL}.
```
yfarjoun: why do we want to record the incorrect claim here?

SAMtags.tex
yfarjoun: hmmm. not sure why this line should be in the tags document...seems like the main text is more appropriate.

SAMtags.tex
```diff
@@ -411,17 +411,111 @@ \section{Locally-defined tags}
...
+
+\subsubsection*{March 2012}
+
+Descriptions of CT and PT annotation tags significantly clarified.
+
+\subsubsection*{October 2011}
+
+Sample barcode tags QT and RT added, with RT being identified as a deprecated alternative to BC.
+
+% These were actually added in late September as RT/PT, but RT was changed to
```
yfarjoun: what's the purpose of the comment?

SAMtags.tex
```diff
@@ -411,17 +411,111 @@ \section{Locally-defined tags}
...
+\subsubsection*{September 2011}
+
+% This was actually August 29th, but let's call it September.
+FZ tag's type changed from {\tt H} to {\tt B,S}-array.
+
+BC and CO tags added.
+
+\subsubsection*{April 2011}
+
+SAM version number {\tt VN:1.4} introduced, indicating the addition of the {\tt B}-array tag type.
```
yfarjoun: again, I'm not sure why the SAM version should be mentioned in this document.

SAMtags.tex
```diff
@@ -411,17 +411,111 @@ \section{Locally-defined tags}
...
+
+CC and CP tags reinstated with their original meanings.
+
+\subsubsection*{November 2010}
+
+BQ tag added.
+
+\subsubsection*{July 2010}
+
+The specification was rewritten as a \LaTeX\ document specifying SAM version number {\tt VN:1.3}.
```
yfarjoun: not relevant to tags

SAMtags.tex
jkbonfield: I think it's simply there as a way to anchor the dates back to SAM specifications.

yfarjoun: in that case, perhaps simply write 
```
\subsubsection*{May 2013 (SAM v1.5)}
```
or similar, and not duplicate the specifics of the version?

SAMtags.tex
jmarshall: The first half of the sentence records a substantive change to the spec, and the second half answers the question that everyone reading the first half will have: _What was it that changed?_

I was going to reply that the previous text was not self-consistent so obviously wrong to interested readers… therefore we should just delete this entry.

However (see 434cda9238e9099418362016380460d35b0dc380) as the `\tagvalue` was also just plain wrong, I think this was sufficiently consistently wrong that it's worth listing the correction. Will change it to _which was previously unclear whether E2 contains bases or base qualities_.

SAMtags.tex
jmarshall: Satisfying pedants, who want to see the existence of the RT/PT history acknowledged. But I didn't want it in the PDF muddying the waters for the reader of a more normal interest level.

SAMtags.tex
jmarshall: In the case of VN:1.4 and array tags, I would have thought the relevance was obvious — and made obvious by this paragraph. Please (re)read it and suggest how it could be made clearer.

SAMtags.tex
jmarshall: Actually the fact that the specification was **almost entirely rewritten** is of interest to tags. See for example the work in progress in https://github.com/samtools/hts-specs/compare/master...jmarshall:ancient-text of recovering the text from the SAM1.pages document for those tags where the original text was clearer.

(Quick quiz before you look at that branch: do you have any idea what _SM:i Template-independent mapping quality_ is supposed to mean?)

jmarshall:
Gee, I wish PR #218 could have enjoyed this level of review too :smile:

I was actually fishing for comments on the chronologically-orientated (rather than the previous version-first-orientated) history, but I assume from the lack of comments on that that you all are happy with it.

@yfarjoun's main concern is about whether the VN version number bumping items are of interest in this document. While I am the first to agree that over-emphasising the version number in SAM is wrong and/or pointless (see e.g. [here](https://github.com/samtools/hts-specs/pull/218#issuecomment-311947914)), I think these entries provide context that many readers are interested in. So my reply is three-fold:

1. The VN:1.4 item and the upcoming VN:1.6 are vital to tags;

2. If those two are going to be in this tags history, then the other VN bumps should be too;

3. The version number bumps are context that I think lots of people are rightly or wrongly interested in, and they “anchor the dates back to SAM specifications” as @jkbonfield noted. A brief hint of why the version number was bumped turns the entry into a sentence, and I think is useful to have at hand for tag spec readers even if it partly duplicates the main spec's full bump reasoning.

…and happily we're not working to a word limit here!

jmarshall:
Thanks for the detailed review. There was method behind most of the madness :smile:

I've pushed a commit with the post-review adjustments I plan to squash and merge later today. Further changes to the relevant main sections (rather than history appendix) to follow to address adding the paper link etc mentioned by @peterjc, probably as another PR.

jmarshall:
Merged as f521a579650f3ae6e91b6ed2c632d35f6ef848ff.

PR #315, opened 2018-05-24
Remove deprecated RT tag

jmarshall:
The barcode RT tag was added in 2011 (in ff06ea66c9d78be64b2a86ee1fe11dc10f0fe045) as an already-deprecated equivalent of the BC tag that had been locally in use at Sanger.

We've been adding a lot of barcode-related stuff lately, so it would be good to simplify the barcodes section a little by removing this.

Sanger folks — @dkj @daviesrob @jkbonfield @jenniferliddle et al — any objections to removing the barcode definition of RT from the specification? I imagine you've long since moved over to using BC…

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -100,11 +100,11 @@ or
   {\tt PT} & Z & Read annotations for parts of the padded read sequence \\
   {\tt PU} & Z & Platform unit \\
   {\tt Q2} & Z & Phred quality of the mate/next segment sequence in the {\tt R2} tag \\
-  {\tt QT} & Z & Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag \\
+  {\tt QT} & Z & Phred quality of the sample barcode sequence in the {\tt BC} tag \\
   {\tt QX} & Z & Quality score of the unique molecular identifier in the {\tt RX} tag \\
   {\tt R2} & Z & Sequence of the mate/next segment in the template \\
   {\tt RG} & Z & Read group \\
-  {\tt RT} & Z & Barcode sequence (deprecated; use {\tt BC} instead) \\
+  {\tt RT} & ? & Reserved for backwards compatibility reasons \\
   {\tt RX} & Z & Sequence bases of the (possibly corrected) unique molecular identifier \\
   {\tt S2} & ? & Reserved for backwards compatibility reasons \\
   {\tt SA} & Z & Other canonical alignments in a chimeric alignment \\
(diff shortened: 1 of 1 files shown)


jmarshall:
Any comments on removing RT from the spec? @dkj @daviesrob @jkbonfield @jenniferliddle or other Sangerites? Bueller?

jkbonfield:
I'm happy for it to go as it's long overdue.

I wonder if it should even be listed at all given every other uppercase pairing not listed is also reserved for future use (ie just comment it out so we can see it in the TeX but not in the pdf, for our own presence of mind and as a reminder that it was once used).

dkj:
The listing of `RT` perhaps weakens the "every other uppercase pairing not listed is also reserved": can we better emphasize all other uppercase tags not starting [XYZ] are "reserved" for HTS-specs?

Given there are ongoing and more recent breaches of the reserved space and the data with `RT` is
at least 6 years old agree it is probably time to purge its explicit declaration. However, I'd advocate that you leave it in as a comment.

jmarshall:
I think that's confirmation from the appropriate Sanger people that they are happy for the RT tag, used locally at Sanger long ago, to be removed from the spec.

I'd propose to merge this as is, adding a “reserved” entry for RT to the §1 table alongside the others already there. And I've filed issue #319 to track removing all these “reserved for backwards compatibility” entries and saying more about uppercase tags being reserved — to be addressed separately from this RT PR.

SAMtags.tex
```diff
@@ -100,11 +100,11 @@ \section{Standard tags}
   {\tt PT} & Z & Read annotations for parts of the padded read sequence \\
   {\tt PU} & Z & Platform unit \\
   {\tt Q2} & Z & Phred quality of the mate/next segment sequence in the {\tt R2} tag \\
-  {\tt QT} & Z & Phred quality of the sample-barcode sequence in the {\tt BC} (or {\tt RT}) tag \\
+  {\tt QT} & Z & Phred quality of the sample barcode sequence in the {\tt BC} tag \\
   {\tt QX} & Z & Quality score of the unique molecular identifier in the {\tt RX} tag \\
   {\tt R2} & Z & Sequence of the mate/next segment in the template \\
   {\tt RG} & Z & Read group \\
-  {\tt RT} & Z & Barcode sequence (deprecated; use {\tt BC} instead) \\
+  {\tt RT} & ? & Reserved for backwards compatibility reasons \\
```
yfarjoun: I think that the type ("Z") should be kept for backwards compatibility reasons. 

Imagine that we had a validator that checked the types found in a file against the spec. Even though a RT is no longer in use, an old file that had RT tag could be either valid or invalid, and you could know that from the type. 

What's the value in "forgetting" the type of the reserved RT tag?

SAMtags.tex
```diff
@@ -312,9 +312,6 @@ \subsection{Barcodes}
 The qualities here may have been corrected (Raw bases and qualities can be stored in {\tt OX} and {\tt BZ} respectively.)
 The lengths of the {\tt QX} and the {\tt RX} tags must match.
 In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the quality strings with a space (`{\tt \textvisiblespace}') between the different strings.
-
-\item[RT:Z:\tagvalue{sequence}]
-Deprecated alternative to {\tt BC} tag originally used at Sanger.
```
yfarjoun: I'd keep a sentence here explaining that while we no longer condone the use of RT for any purpose, it cannot be used for any new purpose due to the fact that it had been used in the past.

SAMtags.tex
jmarshall: Putting ? here is consistent with “it cannot be used for any new purpose”. But I take your point about the previously-expected type being useful historical information and have added it.

jmarshall:
I've added a note about RT previously being type Z. The rest of your comments are about what we should say about these six (soon seven) removed tags in general, not RT specifically. So I'd like to squash/merge this as is, and address the reset of your comments as #319. Okay, @yfarjoun?

yfarjoun:
👍

jmarshall:
Thanks — merged as 791825d3f93af18bcf68314b554fc4a2c355ddfd.

PR #324, opened 2018-06-26
Clarification to CRAM tag encodings.

jkbonfield:
This is now explicit that the type codes follow BAM conventions
(cCsSiI) and not SAM (i), along with examples to demonstrate this.
Also rearranged this so the text describing the dictionary is in the
preservation map and the text describing the "encodings" is in the Tag
Encoding Map.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -597,7 +597,22 @@ section\tabularnewline
 \hline
 \end{tabular}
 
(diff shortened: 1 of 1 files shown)


meganshand:
This is helpful, thank you!

jmthibault79:
OK - will wait for rebase

Issue #300, opened 2018-03-09
BGZF block conventions


betteridiot:
One of the recent moves made was ensuring that if a read doesn't fit in the remainder of the current block it will start a new one. It is also well known that the last concatenated block contains the EOF marker. However, in some really low-level work, I found that the entirety of my BAM header was contained to the first BGZF block (e.g. BSIZE/ISIZE for first block was 1750 and 5271, respectively). The next block's BSIZE/ISIZE was 12682 and 65245, respectively as well.

Does this mean that there is a special convention regarding the header which attempts to enforce that the header is isolated from other data within its respective blocks?

nh13:
I may be wrong, but I don't see anything in the BAM spec that requires this, but some tools will produce such data, for example Picard's `ReplaceSamHeader` tool, which uses the [`reheaderBamFile` method in htsjdk](https://github.com/samtools/htsjdk/blob/c9cab31191e786b6e1b6e83c4f47a39df726c1c6/src/main/java/htsjdk/samtools/BamFileIoUtils.java#L42).  Interestingly in the latter, the end of the header could be in a block with the first alignment/read/record, and so `reheaderBamFile` calls [this method](https://github.com/samtools/htsjdk/blob/c9cab31191e786b6e1b6e83c4f47a39df726c1c6/src/main/java/htsjdk/samtools/BamFileIoUtils.java#L86-L93) to writes the rest of the header into its own block. 

You can also see in htsjdk where it (writes the header to a BAM file and then flushes the block](https://github.com/samtools/htsjdk/blob/c9cab31191e786b6e1b6e83c4f47a39df726c1c6/src/main/java/htsjdk/samtools/BAMFileWriter.java#L208).

jmarshall:
Several implementations began starting a new BGZF block after the headers and before reads that would otherwise be split across BGZF blocks as a result of [this June 2010 samtools-devel thread](https://sourceforge.net/p/samtools/mailman/samtools-devel/thread/AANLkTilCcBXyXt6zVgviEbKWuEhIxgzUV1noqGaZBcS0%40mail.gmail.com/#msg25496753) (starting with [this message](https://sourceforge.net/p/samtools/mailman/message/25496753/)).

The motivation for this is discussed in the thread — having the headers in a separate block was to facilitate tools like `ReplaceSamHeader`.

Nobody got around to adding these conventions to the spec, but I don't think there was any particular reason for not doing so.

jkbonfield:
[...]

PR #325, opened 2018-07-12
[DRAFT] Cache control for re-using previously-downloaded headers

jmarshall:
This proposal is a follow-up to #322. It will require rebasing etc as #322 develops, so I don't anticipate updating or polishing this until the `class` proposal has landed in **master**.

However, if that facility is to be used by clients to enable re-using previously-downloaded headers and this is to be done safely, then I think using HTTP cache control will be the natural way to do it and extrapolating ETag/etc to the htsget ticket is a natural extension. But this is somewhat moot in the absence of implementations, hence this separate PR.


jdidion:
Thanks for the additions here and in #322. Practical question: I am trying to implement a POC htsget server that splits up BAM files into header and body pieces. I'm at a loss for how to actually create a header BAM file and a body BAM file, such that the header BAM is valid by itself but can be concatenated with the body BAM to create the final BAM. In pseudo-python, the way I want my client to interact with the server is:

```
urlobjs = get_urls_from_server()
# The final BAM, containing both headers and body
outbam = open('out.bam', 'wb')
for urlobj in urls:
    url_content = fetch(urlobj['url'])
    if urlobj['class'] == 'header':
        headers = decode_bam(url_content).headers
        # ...do something with the headers...
    outbam.write(url_content)
```

I realize that I can use samtools to split my BAM file into header-only and body-only by first converting to SAM, splitting into header-only SAM and body-only SAM, and then converting both of those back into BAM. But concatenating those two files does not create a valid BAM. I guess I could just write out the body BAM and then use samtools reheader to add the header, but that's quite slow for large BAM files. Any other suggestions?

jmarshall:
You need to find the boundary file offset between the header and the body, which requires understanding the format in a way that a general-purpose read-the-records API won't provide. So for BAM, you need to

* figure out the length in uncompressed bytes of the BAM header, essentially by adding up
`l_text + sum(l_name[1…n_ref])` plus the constant-size fields;

* figure out how many BGZF blocks at the start of the file are used for those headers, by adding up blocks' `isize` fields until the total equals the uncompressed header size.

At that point, you'll have the header-body boundary (in “compressed space”) that you're looking for.

Note that this assumes that a new BGZF block is started for the first body data record (i.e. the header-body boundary is also at a BGZF block boundary) — this has never been stated in the SAM specification, but is something that the main implementations have done for BAM since 2010 (see #300). It seems to me that implementing htsget requires BAM files to have this property. (And similarly for BCF files, but IFAIK the main implementations don't do this for them!)

In practice, you'd more likely find this boundary by looking in a BAI/etc index for the virtual file offset of the first body data record — e.g. (presumably) by finding the smallest `ioffset` in any of the linear indices.

jdidion:
@jmarshall that makes sense, thanks for the explanation. I have a python library for parsing index files I've been meaning to release for a while - seems like that will be useful here. I'll work on putting together a library and command line tool that can be used to split up BAM/BCF/etc files for serving by htsget.

Issue #319, opened 2018-06-01
SAMtags only weakly reserves unassigned uppercase tags


jmarshall:
The table in §1 of the SAMtags document lists seven entries (GC, GQ, GS, MF, RT, S2, and SQ) as

> Reserved for backwards compatibility reasons

and lists X?, Y?, and Z? as

> Reserved for end users

while not mentioning the reserved status of uppercase tags not listed in the table.

It has been pointed out by @dkj on PR #315 that explicitly listing these perhaps weakens the status of every other uppercase tag not listed as also reserved.

Furthermore §2 “Locally-defined tags” tells users that lowercase and X?/Y?/Z? are reserved for their local use, but does not explicitly state that non-lowercase tags are actually reserved for their not use. This vagueness was semi-intentional at the time, but perhaps we should be tighter now.

Suggested improvements:

* Be more prescriptive in the “Locally-defined tags” section;

* Remove “Reserved for backwards compatibility reasons” entries from the §1 table, leaving them listed only in the history appendix (and leaving comments in the TeX table source to remind maintainers that these tags have previous meanings).

jkbonfield:
Agreed.  Infact I assumed it already did say this.

The same logic applies the SAMv1.tex which has this to say on header tags:

> As with alignment optional fields (see Section 1.5), you can freely add new tags for further data fields.  Tags containing lowercase letters are reserved for local use and will not be formally defined in any future version of this specification

Again maybe it needs rewording to explicitly state that uppercase tags are reserved, although it does have a reasonable footnote describing best practice.  Similarly the aforementioned section 1.5 in SAMv1.tex has the same wording again.  We may as well be belt and braces with this.

jkbonfield:
Following on from @yfarjoun's comment in #315 I wonder now whether it may be useful to have a one line summary like:

> Deprecated tags previously defined in earlier versions: GC, GQ, GS, MF, RT, S2, SQ.

(Perhaps with :Z, :i etc if we can figure out what their type codes were.)  The purpose of this is simply to document what *could* legitimately be in a SAM file for use by a tag validator, albeit with a warning about deprecation.

I have no strong views on it really, just musings.

jmarshall:
I was planning to retroactively add entries to the chronology listing the removal dates of GC, GQ, etc and their types and half a sentence of description, like the entry #315 adds for RT.

No strong views on whether having them sprinkled through the chronology suffices or whether a sentence like yours in §1 substantially improves it.

PR #326, opened 2018-07-23
Miscellaneous SAM fixes

### LINKED ISSUE #135: SAM and zero length Z, H, and B fields
We occasionally find a tag like `MC:Z:` with no string after it, ie the zero length string. 
This tag in BAM can be read by both samtools and picard, but in SAM both tools emit an error.

The SAM specification states Z has to take the form `[ !-~]+`, ie 1 or more characters, so the above MC tag is illegal.

However I can't really see any justification for it except accidental over specification. (It's always been this way though.)

Is this something we care about improving, or just accept that the tag was bugged and indeed it shouldn't be possible to specify empty strings.

jmarshall:
Minor SAMv1 and SAMtags improvements, notably:

* Permit no numeric entries in SAM B array regexp. Fixes #135 (for B fields; already fixed for Z and H).

* Clarify several tag descriptions using ancient SAM1.pages text where the pre-LaTeX document was clearer.

* Add types and brief descriptions of removed historic tags. (Addresses a minor aspect of #319.)

* Add links between CT/PT and the Assembly Sequences Guide and a link to @peterjc et al's _SAM/BAM format v1.5 extensions for de novo assemblies_ paper.

* Improve SAMtags's `\tagregex` macro so as to better format complicated patterns like those for CT and PT. (It's not great that both the verbatim characters and the *?+ metacharacters are in the same \tt, and we can do better. This is an improvement for now though.)

(Others truly minor, see commit messages.)


jkbonfield:
Thumbs up from me - it's a lot of crossing *t*s and dotting *i*s, but well worth it.

SAMtags.tex
```diff
@@ -55,7 +56,7 @@ \section{Standard tags}
   \hline
   {\bf Tag} & {\bf Type} & {\bf Description} \\
   \hline
-  {\tt AM} & i & The smallest template-independent mapping quality of segments in the rest \\
+  {\tt AM} & i & The smallest template-independent mapping quality in the template \\
```
yfarjoun: this look like a change in the definition. I've never used AM, but the reading of the original definition seems to imply that it does not take the mapping quality of this record into account, while the new definition does. Am I wrong about my interpretation? is this correction justified?

SAMtags.tex
```diff
@@ -170,7 +172,7 @@ \subsection{Additional Template and Mapping data}
 in SAM.
 
 \item[IH:i:\tagvalue{count}]
-Number of stored alignments in SAM that contains the query in the current
+Number of alignments stored in the file that contain the query in the current
```
yfarjoun: can you expand here on the definition of query. given the term `queryname` I am not 100% clear about whether this pertains to only one "read", or to all the reads from the template.

SAMtags.tex
```diff
@@ -187,17 +189,17 @@ \subsection{Additional Template and Mapping data}
...
-\item[MQ:i:\tagvalue{}]
+\item[MQ:i:\tagvalue{score}]
 Mapping quality of the mate/next segment.
 
-\item[NH:i:\tagvalue{}]
-Number of reported alignments that contains the query in the current record.
+\item[NH:i:\tagvalue{count}]
+Number of reported alignments that contain the query in the current record.
 
 \item[NM:i:\tagvalue{}]
```
yfarjoun: distance?

SAMtags.tex
```diff
@@ -213,8 +215,8 @@ \subsection{Additional Template and Mapping data}
 \emph{Strand} is either `{\tt +}' or `{\tt -}', indicating positive/negative strand, corresponding to FLAG bit 0x10.
 \emph{Pos} is a 1-based coordinate.
 
-\item[SM:i:\tagvalue{}]
-Template-independent mapping quality.
+\item[SM:i:\tagvalue{score}]
+Template-independent mapping quality, i.e., the mapping quality if the read were mapped as a single read rather than as part of a read pair or template.
 
 \item[TC:i:\tagvalue{}]
```
yfarjoun: `\tagvalue{number}`

SAMtags.tex
```diff
@@ -353,21 +359,23 @@ \subsection{Annotation and Padding}
...
 tags separated by `{\tt |}', each annotating a sub-region of the read.
 Each tag consists of \emph{start}, \emph{end}, \emph{strand},
 \emph{type} and zero or more \emph{key}{\tt =}\emph{value} pairs, each
 separated with semicolons. \emph{Start} and \emph{end} are 1-based
 positions between one and the sum of the {\tt M/I/D/P/S/=/X}
 {\sf CIGAR} operators, i.e. {\sf SEQ} length plus any pads.  Note
-any editing of the CIGAR string may require updating the `{\tt PT}'
+any editing of the CIGAR string may require updating the {\tt PT}
 tag coordinates, or even invalidate them.
 As in GFF3, \emph{strand} is one of `{\tt +}' for forward strand tags,
```
yfarjoun: link to GFF3 format?

SAMtags.tex
```diff
@@ -500,13 +508,14 @@ \subsubsection*{July 2010}
 \gap
 Tags FI, FS, OC, OP, OQ, and TC added.
 
-Tags GC, GQ, and GS listed as reserved for backwards compatibility.
+Tags GC:Z, GQ:Z, and GS:Z, briefly proposed for representing repeatedly-sequenced reads, noted as reserved for backwards compatibility.
+Existing tags MF:i (MAQ pair flag), SQ:H (suboptimal bases), and S2:H (mate's suboptimal bases) removed and noted as reserved for backwards compatibility.
 
-Existing CC, CP, and MF tags removed and noted as reserved for backwards compatibility.
+CC and CP tags temporarily removed.
```
yfarjoun: what does "temporarily" removed mean?

SAMtags.tex
jmarshall: This PR doesn't touch NM. I'd leave that to #308…

SAMtags.tex
jmarshall: Would you like to suggest a suitable link?

SAMtags.tex
jmarshall: It's clear in the context of the March 2011 entry.

SAMtags.tex
jmarshall: I would guess that “query in the current record” implies it's just this read's sequence, and there's samtools-devel emails from 2011 that can be interpreted as agreeing with that. BWA doesn't implement NH/IH/HI, and HISAT2 seems to calculate NH in both interpretations in different circumstances. So I'd rather not expand on this without doing further research!

SAMtags.tex
jmarshall: AS is already described as a _score_. This PR adds `\tagvalue{score}` to AM, MQ, PQ, and SM. I suppose AM/MQ/SM are properly mapping qualities, which (in bwa at least) are alignment scores scaled in some way, but therefore are not exactly scores themselves. And PQ is a Phred likelihood, with whatever relationship to alignment scores or mapping qualities that might have.

So perhaps `\tagvalue{mapq}` for AM/MQ/SM? Would rather not `number` as leaving it blank would convey the same amount of information…

SAMtags.tex
jkbonfield: It confused me initially too, but maybe just add "(until March 2011)" as a hint.   It's all ancient history though and I doubt most people will be reading back that far.

SAMtags.tex
jkbonfield: I haven't any objections to the change, but in review I did wonder how this tag is ever meant to work...

"Stored" is a dangerous word.  If we do a sub-range query then we likely invalidate this tag, and not doing so turns sub-range queries from fast operations into complex nightmares.  It's simply not a sensible tag as it's written and doesn't match common practice.

I'd guess than "Number of generated alignment records" may match the expected usage more where the aligner can inform us of how many records it wrote, but as you say its usage right now is patchy and probably ill defined.  The language clarification is fine and I don't mind punting more substantive changes to another PR.

SAMtags.tex
jkbonfield: I avoided "distance" because frankly it gets into the whole world of preconceived ideas as to what an edit distance is.  I figured "count" was sufficient for NM as it's used a lot elsewhere and the actual description for what is being counted follows.

SAMtags.tex
yfarjoun: the official, yet sparse:
https://useast.ensembl.org/info/website/upload/gff3.html

unofficial, but less sparse:
https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md

peterjc: I though the official version was at http://www.sequenceontology.org/ but they moved it to GitHub recently making https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md the official source now?

PR #330, opened 2018-08-15
Add 206, 303, 501 return codes for /sequence/{id} in refget-openapi.yaml..

raskoleinonen:
Suggestion to add 206, 303, 501 return codes for /sequence/{id} defined in the specs.

### DIFF
--- a/pub/refget-openapi.yaml
+++ b/pub/refget-openapi.yaml
@@ -98,6 +98,14 @@ paths:
                 type: string
                 example: >-
                   MSSPTPPGGQRTLQKRKQGSSQKVAASAPKKNTNSNNSILKIYSDEATGLRVDPLVVLFLAVGFIFSVVALHVISKVAGKLF
+        '206':
+          description: Successful retrieval of subsequence as a single string with no line breaks
+        '302':
+          description: Redirecting the client where the sequence can be retrieved
+        '303':
+          description: Redirecting the client where the sequence can be retrieved
+        '307':
+          description: Redirecting the client where the sequence can be retrieved
         '400':
           description: Invalid input; normally due to range parameter usage
         '404':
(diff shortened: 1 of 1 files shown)


andrewyatz:
Same again :). Looks great just need a more descriptive commit message and it'll go in

jmarshall:
It would be nice to keep the list of codes in numerical order. Also probably wants to be 302/303/307 rather than just 303.

raskoleinonen:
It is not straightforward to modify the commit message (https://help.github.com/articles/changing-a-commit-message/). Might be easier if you reject this pull request and I'll just create a new one. 

I can fix the numerical order at the same time.

About 302/303/307.

Specs says in 'Protocol essentials': 302, 303 and 307 are all valid response codes to use.
And later in 'Method: get sequence by ID':  '... or 303 redirecting the client to where it can retrieve the sequence' 
without mentioning 302 or 307.

jmarshall:
> And later in 'Method: get sequence by ID': '... or 303 redirecting the client to where it can retrieve the sequence' without mentioning 302 or 307.

303 was relaxed in _Protocol essentials_ to 3XX(302/303/307) in a late commit to PR #327; the text in the _Method_ paragraph should say 3XX rather than 303.

(It is pretty straightforward to update a PR consisting of just one commit: `git commit --amend` which you'd be doing anyway; then `git push -f rasko patch-1` (or whatever your remote is called). Just adds a `-f` compared to adding another commit to the PR. Or just push a second commit to the PR and I'm sure @andrewyatz can tidy them up when merging it :smile:)

andrewyatz:
Aye can you add the other missing 30X codes (302 and 307).

raskoleinonen:
Done and ordered return codes in numerical order.

While checking 30X return codes for my own interest I stumbled also on:
301 Moved Permanently

andrewyatz:
Yes. 301 would be a pretty bad thing to send through on this API unless we were sure that URL location would never ever move. We specifically think 301 is possibly harmful to the API though someone may provide a 301 to indicate the server location has permanently moved. But not for this trick of redirecting a client to sequence.

andrewyatz:
Merged. Thanks!

PR #331, opened 2018-08-15
Change Range to be exclusive at the end in refget specification.

raskoleinonen:
Suggestion to change Range to be exclusive at end to have the same semantics as for start and end parameters.

Excerpt from https://tools.ietf.org/html/draft-ietf-http-range-retrieval-00
Examples of the Content-Range: HTTP Response Header
The first 500 bytes:
Content-Range: bytes 0-499

Also, we should check the current wording in:
https://github.com/samtools/hts-specs/blob/master/pub/refget-openapi.yaml
          name: Range
          required: false
          description: >-
            Specify a substring as a range not using query parameters. Note
            start is in Ensembl-style start coordinates i.e. the first base of a
            sequence is 1 not 0 as in the start parameter. Example is 'bytes:
            1-10'


andrewyatz:
Passing this onto Matt as I think he's better to comment. However same again the commit heading should be changed to reflect what the actual change has been

raskoleinonen:
It is not straightforward to modify the commit message (https://help.github.com/articles/changing-a-commit-message/). Might be easier if you reject this pull request and I'll just create a new one,

raskoleinonen:
Closing this pull request and will create a new one with proper commit message.

PR #332, opened 2018-08-15
Nest service-info under "sequence" and metadata endpoint parameters under "metadata" in refget-openapi.yaml.

raskoleinonen:
service-info endpoint
------------------------
Nested under /sequence in the 0.2 specs.

https://github.com/samtools/hts-specs/blob/master/refget.md
Method: Fetch information on the service
GET /sequence/service-info

metadata endpoint
------------------------
Wrapped within "metadata" object in the 0.2 specs.

### DIFF
--- a/pub/refget-openapi.yaml
+++ b/pub/refget-openapi.yaml
@@ -20,7 +20,7 @@ tags:
   - name: Services
     description: Housekeeping services
 paths:
-  /service-info:
+  '/sequence/service-info/:
     get:
       summary: Retrieve a summary of features this API deployment supports
       operationId: serviceInfo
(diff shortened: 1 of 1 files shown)


jmarshall:
Please use specific subject lines in your pull requests. For example, this one could be something like “Fix /sequence/service-info/ in refget-openapi.yaml”. There's no point having a [list of PRs](
https://github.com/samtools/hts-specs/pulls) that all say the same thing:

image

raskoleinonen:
My apologies. Done.

andrewyatz:
Can you change the first line of your commit message to indicate the change you've made (similar to the name you've given this issue). Sorry otherwise it's just going to appear as `Update refget_openapi.yaml`

andrewyatz:
Otherwise the change looks good

raskoleinonen:
It is not straightforward to modify the commit message (https://help.github.com/articles/changing-a-commit-message/). Might be easier if you reject this pull request and I'll just create a new one,

andrewyatz:
As the GitHub docs say if you can clone your fork of hts-specs then changing the commit is a case of switching to the branch in question and doing the change. In this case I assume you have to do:

```bash
git clone https://github.com/raskoleinonen/hts-specs
git checkout patch-3
git commit --amend
# edit the message
git checkout patch-2
# edit the message on the other branches you have
git push --force
```

This is off the top of my head and also after doing this earlier on today :)

raskoleinonen:
Ok, tried this but managed to change only 1 of 3 commit messages (https://github.com/samtools/hts-specs/pull/330). Clone + checkout + amend for each branch and then git push -force. Tried also checkout  + amend + git push -force separately for the two remaining branches. All commit changes have been effected locally but the push was successful only for one (https://github.com/samtools/hts-specs/pull/330). I might be still doing something wrong...

andrewyatz:
Did you do the push from each branch? That might have been my mistake

jmarshall:
Actually all three have been updated. (It would have been the `git push --force` separately for each branch that did the trick.) Sometimes you need to refresh the GitHub webpage to see the effects; all three look good via git.

raskoleinonen:
Right, the commit messages updates are now visible in GitHub commit view but the pull request view text remains stubbornly unchanged for this ticket even if I use a different browser etc.

Ok, I'll leave this pull request open in the hope that it might be good enough to be merged.

Issue #291, opened 2018-03-01
Resolving ambiguities introduced when supporting colons in hg38 names


cyenyxe:
I have re-read the whole hg38 contig name pull request thread #258 and wanted to suggest some related, longer term ideas without polluting it.

In that thread, most people are in favor of supporting colons in contig names, and making the parsers resolve breakend ambiguities in favor of contig:position or contig:position-position.

This is a legitimate use case, but I am concerned because we have been trying to reduce the ambiguities in the spec, and this change will introduce a new one. Would people be open to slightly modify the breakend notation in future versions of the spec, for instance by replacing the colon with a dot or comma? I think these characters are not allowed in contig names according to the SAM spec, but could someone please confirm it?

The representation would change from something like:

```
G]HLA-A*01:01:01:01:123456]
```

To something like the following:

```
G]HLA-A*01:01:01:01.123456]
G]HLA-A*01:01:01:01,123456]
```

daviesrob:
Unfortunately the [SAM specification](http://samtools.github.io/hts-specs/SAMv1.pdf) allows pretty much anything in a reference name.  The first character is not allowed to be `*` or `=`.  The regex for the rest is `/[!-~]*/` which includes all printable non-whitespace characters in US-ASCII.

yfarjoun:
perhaps we could limit the name-space for contig names in a way that
doesn't interfere with current usage and would allow for this suggestion.

jkbonfield:
As added to https://github.com/samtools/hts-specs/pull/193 also, comma is already a problem for us in SAM world due to SA tags and the new OA tag.  Given it's currently unused in all the legal fai files we found, I'd be in favour of simply banning comma so we can get some sanity back in the world.

daviesrob:
I should mention my previous [table of punctuation characters found in reference files](https://github.com/samtools/hts-specs/pull/220#issuecomment-305537180):
```
 # 203
 % 203
 * 525
 + 1
 , 496
 - 154226
 . 1826561
 : 1577
 = 26
 _ 4961932
 | 1098333
```

So comma isn't _too_ bad (all the commas came from a single badly-formatted file), but dot is not a good choice.  Semicolon looks fairly safe at the moment.

It would be a very good idea if everyone with access to large collections of reference sequences did a similar analysis.  We can compare results and see if these are typical or not.

yfarjoun:
Here are our results:
```
     12 #
    527 *
    357 ,
1451132 -
1492749 .
  86114 /
[...]

Issue #258, opened 2017-10-25
Existing VCF spec is incompatible with certain hg38 contig names


droazen:
In hg38 sequence dictionaries we frequently encounter contigs such as this one:

```
@SQ	SN:HLA-A*01:01:01:01	LN:3503	M5:01cd0df602495b044b2c214d69a60aa2	AS:38	UR:/seq/references/Homo_sapiens_assembly38/v0/Homo_sapiens_assembly38.fasta	SP:Homo sapiens
```

These contigs include the `:` character in their names, which is problematic for VCF since section `1.6.1` of the current spec specifically forbids colons in contig names:

```
The colon symbol (:) must be absent from all
chromosome names to avoid parsing errors when dealing with breakends.
```

This is inconsistent with the SAM spec, which allows colons in contig names (they are required to match the regular expression `\*|[!-()+-<>-~][!-~]*`). 

With increasing adoption of hg38 (particularly in the last year or so), this is becoming a bit of a crisis. The maintainers of the VCF spec met recently and agreed that this issue needs to be resolved in the next version of the spec (in addition to agreeing on an officially-sanctioned practical workaround that can be used with the current version of the spec).

I see a couple of possible options here:

1. Remove the ban on colons in contig names in the next version of the spec (VCF 4.4). This would resolve the current incompatibility between the VCF and SAM specs, and since it constitutes a relaxation of an existing restriction rather than imposing a new restriction, it should not cause backwards-compatibility issues.

2. Keep the restriction, but require that colons in contig names be escaped within VCF records, and automatically encoded/decoded by parsers (rather than escaped manually by users, which would be impractical). This option is consistent with section 1.2 of the existing spec:

```
Characters with special meaning (such as field delimiters ’;’ in INFO or ’:’ FORMAT fields) 
must be represented using the capitalized percent encoding:
%3A : (colon)
```

I think whatever approach we choose, tools will have to be updated to accept intervals such as "HLA-A*01:01:01:01:1-5", and use the sequence dictionary to resolve the parsing ambiguity introduced by the colons. Requiring users to escape these contig names manually seems far too onerous.

Thoughts on this issue? I'm hoping to trigger a discussion that will lead to a consensus + resolution in time for VCF 4.4, plus agreement on a workaround that can be used within the existing spec.

yfarjoun:
So, the issue with the colon (AFAIK) is that people like to be able to write

>contig:start-end

[...]

PR #333, opened 2018-08-23
Disallow commas and other punctuation in RNAME etc

jmarshall:
The suggestion here is to disallow `\ , "'`` () <> [] {}`, i.e., backslashes, commas, various quote marks, and various brackets. Hopefully this is consistent with the statistics listed in #291. Additional counts for some of these characters would be useful  — see https://github.com/samtools/hts-specs/pull/333#issuecomment-422778596 below.

The existing spec uses regular expressions containing ASCII ranges, which is fairly opaque. This PR lists the punctuation characters explicitly and then invents some extra character class notation to refer to the set of allowed characters succinctly and hopefully clearly:

rname

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -32,6 +32,9 @@
 \newcommand*{\firstbytebox}[2]{\byteboxAux{#1}{#2}{\put(0,0){\line(0,1){\bytetotalheight}}}}
 \newcommand*{\bytebox}[2]{\byteboxAux{#1}{#2}{}}
 
+\newcommand*{\cclass}[1]{{\rm\sf :#1:}}
+\newcommand*{\caret}{\textsuperscript{$\wedge$}}
+
 \begin{document}
 
 \input{SAMv1.ver}
@@ -176,6 +179,31 @@ almost entirely overlapping with each other; except the single-best optimal
 mapping, all the other mappings get mapping quality $<$Q3
 and are ignored by most SNP/INDEL callers.}
 
+\subsubsection{Character set restrictions}\label{sec:charset}
+
+Reference sequence names, CIGAR strings, and several other field types are used as values or parts of values of other fields in SAM and related formats such as VCF.
+To ensure that these other fields' representations are unambiguous, these field types disallow particular delimiter characters.
+
+Query or read names may contain any printable ASCII characters in the range \verb"[!-~]" apart from `\verb"@"', so that SAM alignment lines can be easily distinguished from header lines.
+(They are also limited in length.)
+
+Reference sequence names may contain any printable ASCII characters in the range {\tt [!-\verb:~:]} apart from backslashes, commas, quotation marks, and brackets---i.e., apart from `{\tt \verb:\:\,,\,"`'\,()\,[]\,\verb:{}:\,<>}'---and may not start with `{\tt *}' or `{\tt =}'.%
+\footnote{Characters that are \emph{not} disallowed include `{\tt |}', which historically appeared in reference names derived from NCBI FASTA files, and `{\tt :}', which appears in HLA allele names.
+Appendix~\ref{sec:parse-region} describes approaches for parsing \emph{name}{\tt [:}\emph{begin}{\tt -}\emph{end}{\tt ]} region notation unambiguously even though \emph{name} may itself contain colons.}
+
+Thus they match the following regular expression:
+\begin{center}
+{\tt [\verb"0-9A-Za-z!#$%&+./:;?@^_|~-"][\verb"0-9A-Za-z!#$%&*+./:;=?@^_|~-"]*}
+\end{center}
+
+% Pedantically this should be [[:rname:]^*=][[:rname:]]*, but we take advantage
+% of POSIX (Issue 7) section 9.3.5/8 to elide the excess brackets for clarity.
+\newcommand*{\rnameRegexp}{[\cclass{rname}\caret*=][\cclass{rname}]*}
+
+\noindent
+For clarity, elsewhere in this specification we write this set of allowed characters as a character class~{\tt [\cclass{rname}]} and extend the POSIX regular expression notation to use {\tt\caret *=} to indicate the omission of `{\tt *}' and `{\tt =}' from the character class.
+Thus this regular expression can be written more clearly as {\tt\rnameRegexp}.
+
 \subsection{The header section}
 Each header line begins with the character `{\tt @}' followed by
 one of the two-letter header record type codes defined in this section.
@@ -235,7 +263,7 @@ For example, {\tt @HD SO:queryname SS:coordinate:TLEN} would indicate that the d
 The {\tt SN} tags and all individual {\tt AN} names in all {\tt @SQ} lines
 must be distinct.
   The value of this field is used in the
-  alignment records in {\sf RNAME} and {\sf RNEXT} fields. Regular expression: {\tt [!-)+-\char60\char62-\char126][!-\char126]*}\\\cline{2-3}
+  alignment records in {\sf RNAME} and {\sf RNEXT} fields. Regular expression: {\tt\rnameRegexp}\\\cline{2-3}
   & {\tt LN}* & Reference sequence length. \emph{Range}: $[1,\,2^{31}-1]$\\\cline{2-3}
   & {\tt AH} & Indicates that this sequence is an alternate locus.%
 \footnote{See \url{https://www.ncbi.nlm.nih.gov/grc/help/definitions} for descriptions of \emph{alternate locus} and \emph{primary assembly}.}
@@ -246,13 +274,12 @@ A comma-separated list of alternative names that tools may use when referring
 to this reference sequence.%
 \footnote{For example, given `{\tt @SQ SN:MT AN:chrMT,M,chrM LN:16569}',
 tools can ensure that a user's request for any of `MT', `chrMT', `M',
-or~`chrM' succeeds and refers to the same sequence.
-Note the restricted set of characters allowed in an alternative name.}
+or~`chrM' succeeds and refers to the same sequence.}
 These alternative names are not used elsewhere within the SAM file;
 in particular, they must not appear in alignment records' {\sf RNAME}
 or~{\sf RNEXT} fields.
 \emph{Regular expression}: \emph{name}{\tt (,}\emph{name}{\tt )*}
-where \emph{name} is {\tt [0-9A-Za-z][0-9A-Za-z*+.@\_|-]*}\\\cline{2-3}
+where \emph{name} is {\tt\rnameRegexp}\\\cline{2-3}
   & {\tt AS} & Genome assembly identifier. \\\cline{2-3}
   & {\tt DS} & Description.  UTF-8 encoding may be used.\\\cline{2-3}
   & {\tt M5} & MD5 checksum of the sequence.  See Section~\ref{sec:ref-md5}\\\cline{2-3}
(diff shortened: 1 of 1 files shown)


yfarjoun:
I'm assuming that `,` is in the list because otherwise SA and OA are broken, despite being present at some references. 

What will the folks that have `,` in their ref-names do? Should the implementations enable (with warnings) these characters or not?

jmarshall:
I mean to re-collect the statistics on this PR, but IIRC all the commas observed in the various repositories are from one malformatted file, as noted in https://github.com/samtools/hts-specs/issues/291#issuecomment-369656079. So such folks have not been observed to exist.

jmarshall:
Previous discussion of punctuation characters in RNAMEs can be found in #124 (HLA and colons), #167 (VCF and commas/equals), #220 (SQ-AN; statistics), #258 (VCF and HLA/colons), #291 (VCF breakends; statistics).

This table collects the statistics of punctuation character counts from various archives of reference FASTA files:

Char | Sanger (Rob)	| Broad (Yossi)	| ENA (Cristina)	| Notes
---	| ---	| ---	| --- | ---
`!`	| .	| .	| .
`"`	| .	| .	| .
`#`	| 203	| 12	| 16927
`$`	| .	| .	| .
`%`	| 203	| .	| .
`&`	| .	| .	| .
`'`	| .	| .	| .
`(`	| .	| .	| .
`)`	| .	| .	| .
`*`	| 525	| 527	| 1
`+`	| 1	| .	| .
`,`	| 496	| 357	| 231	| Want to disallow
`-`	| 154226	| 1451132	| 122563947
`.`	| 1826561	| 1492749	| 521540419
`/`	| .	| 86114	| 236951
`:`	| 1577	| 233731	| 30181	| Can't disallow, used in HLA alleles
`;`	| .	| .	| 72892	| Might like to disallow, but many in ENA data
`<`	| .	| .	| .
`=`	| 26	| 2034	| 186611
`>`	| .	| .	| .
`?`	| .	| .	| .
`@`	| .	| 17	| 3713
`[`	| .	| .	| .
`\`	| .	| .	| 0
`]`	| .	| .	| .
`^`	| .	| .	| .
`_`	| 4961932	| .	| .
\`	| .	| .	| .
`{`	| .	| .	| .
\|	| 1098333	| 1735713	| 949	| Can't disallow, historically used in NCBI FASTA
`}`	| .	| .	| .
`~`	| .	| .	| .

`.` indicates counts not reported. It would be very good to fill in the gaps for the characters proposed to be disallowed: in particular quotes and the various bracket characters.

Re commas, @daviesrob reported that all the commas he observed originate from one malformatted file. @yfarjoun, @cyenyxe: Is something similar the case for the other two repositories?

jmarshall:
> What will the folks that have `,` [etc] in their ref-names do? Should the implementations [allow] (with warnings) these characters or not?

I was previously thinking it was important that implementations shouldn't need to enforce several different subsets of characters, and therefore that we shouldn't mark this change via HD-VN. Now I don't think that's nuanced enough. Based on conversation at last month's meeting, the considerations are:

1. Implementations that would like to validate the subset of RNAME characters shouldn't need to enforce several different subsets;

2. Implementations shouldn't gratuitously reject existing data files containing acceptable at the time but now disallowed RNAME characters;

3. Implementations need to be able to signal their displeasure with disallowed RNAME characters if they are seen in newly-created data files. (Otherwise people will just carry on (mis)using them.)

So implementations should accept commas etc for HD-VN≤1.5 but diagnose commas etc for HD-VN≥1.6. (Assuming that they essentially copy VN when copying records around but emit the current spec VN when creating new data files.) It would be up to the implementation whether this means silence for ≤ 1.5 and a warning for ≥ 1.6, or warning≤1.5 and error≥1.6, or etc. (The SQ-AN subset was more restricted prior to this PR, but I would suggest that there is no point enforcing this for ≤1.5.)

If a ≤1.5 file has a `,` in an RNAME then this should IMHO be accepted (perhaps with a warning) as nothing will go wrong immediately and it is useful to be able to process the data. Problems would ensue if that RNAME was propagated to an SA or OA tag, and IMHO it should be up to the implementation whether to accept, warn, or reject it at that point. (The statistics collected show that this scenario would be vanishingly rare.)

Probably VN:1.6 is rare enough in the wild that using it to signal the intention to comply with this restricted RNAME subset is fine. Or we might perhaps want to bump the spec to VN:1.7 to signal this.

If others agree with this ≤1.5 / ≥1.6 approach, I'll add text to this effect to the history appendix entry.

Consideration (1) means that we have only this one chance to get the subset details right!

yfarjoun:
I'm OK with outlawing `#`, `[`, `]`, and the comma `,` 

In the end it's only 3 references that have them, and i think that we can deal.
the other operators we do not have in our references...so unreported == 0 (from Broad)

I'd like to note that * is in hg38...so I think the ship has sailed FAR with that one.

jkbonfield:
Do we need to extend the regexp notation?

My interpretation of POSIX is we'd normally use things like `[[:alpha;]]` rather than `[:alpha:]` and can extend them.  Thus `[[:alpha:][:digit:]]` is equivalent to `[[:alnum:]]`.

This in turn means if we chose the left hand side of the rname as `[:rname:]` class, rather than the right (note we don't say which it is - it's just "this" in the spec now), then we can define the RNAME field as matching `[[:rname:]][*=[:rname:]]*`.  Ugly, but I think it's valid without inventing new syntax.

jkbonfield:
I like the inclusion of appendix A - it's not specification, but guidance on tool chain usage which feels appropriate.  Do we also want to expand on potential disambiguation techniques?  Eg given we're outlawing curly braces, we could also permit {*rname*}:1-2 to be equivalent to *rname*:1-2, thus permitting `{foo:100-200}` vs `{foo}:100-200` as disambiguation syntax.  Again this isn't a specification requirement, but a hint for a new mechanism

yfarjoun:
I like the proposed hint for disambiguation. it should probably be put into the spec so that different tools can all use it.

jkbonfield:
For completeness here, if I recall in the last File Formats meeting @jmarshall explained the logic for a new regex syntax.  My understanding is it is a preference to define the characters that may be used, with an exception for the first character, rather than swapping these around purely for the convenience of sticking with standard regexps.

Really this is just a case of regexp being a poor choice compared to a grammar.  I'm not entirely convinced with the argument, but I can see it has some benefit too so I'll accept it if others are like minded.

Apart from that foible, I'm happy with this PR.

jmarshall:
I think in the previous meeting we agreed with https://github.com/samtools/hts-specs/pull/333#issuecomment-422835741 so I'll add a bit talking about VN foibles. And that I'll expand the appendix to suggest a convention of `{rname}:1-2` syntax.

The thing about regexp syntax comes down to **clarity of exposition**. For me, the downside of inventing our own tiny piece of ad hoc regexp-related syntax is outweighed by upside of being able to express the set of rname characters directly:

1. in text: _any printable ASCII characters in the range `[!-~]` apart from backslashes, commas, quotation marks, and brackets—i.e., apart from [the exact characters]_
2. in the exact voluminous regexp
3. in a legible pseudo-regexp that shows the foible of “must not start with * or =” directly.

I think it's clear that “we write **this set** of characters” refers to the set described in (1), and even if that's not clear to a particular reader the meaning is made clear by having (2) there too.

I tried it both ways around (and several other ways too) and I believe that this way leads to the clearest exposition, which is why it's in the proposal. James's understanding described in the previous comment is correct; people who remain displeased by the idea of inventing ad hoc regexp syntax should workshop this themselves and see if they can come up with something clearer. :smile:

SAMv1.tex
```diff
@@ -178,6 +181,31 @@ \subsection{Terminologies and Concepts}
...
+Query or read names may contain any printable ASCII characters in the range \verb"[!-~]" apart from `\verb"@"', so that SAM alignment lines can be easily distinguished from header lines.
+(They are also limited in length.)
+
+Reference sequence names may contain any printable ASCII characters in the range {\tt [!-\verb:~:]} apart from backslashes, commas, quotation marks, and brackets---i.e., apart from `{\tt \verb:\:\,,\,"`'\,()\,[]\,\verb:{}:\,<>}'---and may not start with `{\tt *}' or `{\tt =}'.%
+\footnote{Characters that are \emph{not} disallowed include `{\tt |}', which historically appeared in reference names derived from NCBI FASTA files, and `{\tt :}', which appear in HLA allele names.
+Appendix~\ref{sec:parse-region} describes an approach for parsing \emph{name}{\tt [:}\emph{begin}{\tt -}\emph{end}{\tt ]} region notation unambiguously even though \emph{name} may itself contain colons.}
+
+Thus they match the following regular expression:
+\begin{center}
+{\tt [\verb"0-9A-Za-z!#$%&+./:;?@^_|~-"][\verb"0-9A-Za-z!#$%&*+./:;=?@^_|~-"]*}
```
yfarjoun: this doesn't seem to be a legal regexp.in particular the range `~-"` is invalid.

yfarjoun: sorry. bad parsing of tex.

yfarjoun:
While we are here...should we protect the sequence name `unmapped` from ever being used? in GATK we use this to point to the unmapped & unplaced reads...it would break if someone were to use `unmapped` as a name of a sequence...

yfarjoun:
given that htsget is currently grappling with how to access unplaced reads, I don't think that it is implementation specific.... #320

jmarshall:
> should we protect the sequence name `unmapped` from ever being used

This is a question about in-band vs. out-of-band signalling (i.e., do we allow for special values that are syntactically the same as normal values). In programming languages, this called reserved words vs. identifiers, and the usual approaches are grandfathering in the list of reserved words right from the start or adding new reserved words by using distinguishing syntax (e.g. a leading `_` that's not allowed for normal identifiers, or having some namespace syntax like `::`).

IMHO it's very late in the day to be thinking about adding some reserved words to our RNAME rules. Particularly as SAM has always reserved all RNAMEs starting with `*` or `=`, so we already have leading character distinguishing syntax. So I'm afraid IMHO GATK should be using `*unmapped` or `=unmapped` to do this in a reasonable way.

As for htsget:— there has been a discussion about whether to use SAM's already distinguished-syntax `*` for unplaced unmapped reads queries, or whether to use URL query parameter facilities to be truly distinctive and out-of-band. What you can be sure of is that htsget won't be considering using anything in-band like an otherwise undistinguished `referenceName=unmapped`. :smile:

jkbonfield:
Technically SAM doesn't reserve words starting with "\*" or "=", but words that are *entirely* "\*" or "=".
My suggestion though is to relax that slightly, as John describes above.

I think "implementation defined" is a good way of dealing with them too.  So we can say "\*" and "\=" have reserved meanings, anything starting "\*" or "\=" is implementation dependent, and everyone else you have to honour as valid user data.

I've given a use case I hope to put into practice already here - fake references for data clustering purposes.

jmarshall:
The current spec has long said that `@SQ-SN` values must match `[!-)+-<>-~][!-~]*`.

You are correct that the RNAME and RNEXT fields assign particular meanings to words that are _entirely_ "*" or "=". However due to that regexp the spec **also** reserves all RNAMEs starting with `*` or `=` — though it doesn't yet define any meanings for such RNAMEs.

jkbonfield:
Ah yes, you're correct.  I hadn't spotted \* was missing from the RNAME regex.  This is inherently the problem of having negated regexps done this way; it's hard to remember my ASCII collating order to know what is between ) and +.    The new regexp is a huge improvement in this regard too. :-)

jmarshall:
> I think in the previous meeting we agreed with https://github.com/samtools/hts-specs/pull/333#issuecomment-422835741 so I'll add a bit talking about VN foibles. And that I'll expand the appendix to suggest a convention of {rname}:1-2 syntax.

I've just pushed a draft of those two things. I think we're all in agreement on the basic thrust of this PR, so once that's reviewed and wordsmithed, I think this is ready to merge!

It wasn't that easy to compose something useful to say about the VN foibles and note that contrary to the suggested text in this latest commit, samtools/htsjdk#1238 as merged strictly enforces the RNAME regexp regardless of the input file's \@HD-VN value. Hmmm…

jmarshall:
@jkbonfield @yfarjoun — please review this last commit about {rname}:1-2 delimiter syntax and VN.

If we agree with adding this to VN:1.6 (rather than bumping to a new VN:1.7), I think this is ready for merging.

jkbonfield:
I'm fine with it and if I recall the bracing style matches the in the languishing https://github.com/samtools/htslib/pull/708 PR.

I don't think we need a version number bump (but I wouldn't feel overly concerned if we decided it warranted it).

jmarshall:
Rebased on to current **master**.

It appears that we've never really particularly discussed the suggested candidates for punctuation to disallow!

After removing the characters disallowed by this pull request, we are left with the following accepted RNAME punctuation characters:

`. _ - + = @ # * / : | ;    ` Characters seen in the wild that are of (mostly) obvious utility in naming

`! $ % & ? ^ ~    ` Characters not seen in the wild and of dubious utility in naming

Of the dubious ones only `%` (203 at Sanger) has a non-zero count in the table in https://github.com/samtools/hts-specs/pull/333#issuecomment-422778596.

There would be something to be gained from disallowing some of these dubious characters (`^$` as regexp metacharacters or `^chr5` not-chr-5 notation; `?&%` in URLs; `!` in C-Shell or not- notation). However it also behooves us to have understandable rules about what characters are not allowed and not to arbitrarily restrict what has been allowed for years (`chr5??!` in an incomplete assembly? :smile:).

So I mention this for completeness, and propose to merge this as is, with the somewhat clear and memorable rule being: backslash-comma-quotes-brackets forbidden…

jkbonfield:
It's a great step forward, and if we later rethink it further it can be a separate PR.  However realistically I'm not so worried about shell and regexp metacharacters as they can be escaped.  The main concern we had was actual functionality of the specification being broken - eg commas being used elsewhere in strings that contain names, angle brackets in VCF contig names, etc.

Besides, if we forbid ? then how am I going to get my ANSI trigraphs through now? ;-)

jmarshall:
It is indeed a great step forward. I **really** don't want to restrict it further in future though, as that wouldn't be fair to users or to implementations (https://github.com/samtools/hts-specs/pull/333#issuecomment-422835741 consideration (1)).

Issue #329, opened 2018-08-15
Distinctness of different kinds of refget digest IDs


jmarshall:
Sequences are accessed via

    GET /sequence/<id>

where `<id>` is described as

> A string specifying the sequence to be returned. The identifier shall be a checksum derived from the sequence using one of the supported checksum algorithms, or an alias for the sequence supported by the server.

However there doesn't appear to be a way for the client to say which checksum algorithm or what kind of alias they are intending to be using.

The currently-defined checksum algorithms are `md5` (ids are 32 hex digits) and `trunc512` (ids are 48 hex digits). So fortunately it is trivial for the server to determine which of `md5` or `trunc512` is being used by looking at the `<id>`'s length. The format of “aliases” is not described however, and it appears a maliciously-constructed alias could collide with an extant checksum `<id>`.

1. Is the intention that any more checksum algorithms defined in the future will also have easily distinguished `<id>` values?

2. Would it be a good idea to prevent aliases from colliding with valid checksum `<id>` values?
For example, servers might only consider an `<id>` to potentially be an alias if it has 30 or fewer characters or if it contains non-hexdigit characters.

andrewyatz:
Good questions. 

> Is the intention that any more checksum algorithms defined in the future will also have easily distinguished <id> values?

That was the idea. I would only want to expand the supported checksum range as/when we've found there to be an issue with an existing supported checksum i.e. if trunc512 and md5 resulted in a collision we would expand the digest size used by trunc512.

> Would it be a good idea to prevent aliases from colliding with valid checksum <id> values?
For example, servers might only consider an <id> to potentially be an alias if it has 30 or fewer characters or if it contains non-hexdigit characters.

I'm not totally sure. I viewed the aliases section as a bit where an API can say "I believe this is a known alias for this ID". Nothing more. Those known aliases could be other checksums e.g. if UniParc implemented this they could provide their crc64 checksums as an alias. Part of me feels that this is a buyer beware situation. 

What might be a good solution is to extend the API to, as you said, specify the type of checksum being used. That way aliases can collide but they would never resolve.

yfarjoun:
I think it would be bet to have different access points for the different
retrieval types. While we are at it, I would add a version...so that you'll
have

GET /sequence/v1/md5/<md5-id>

GET /sequence/v1/trunc512/<trunc512-id>

GET /sequence/v1/alias/<alias-id>


This will make it easy to change a version if we so choose, and add or
remove capabilities.

[...]

PR #334, opened 2018-08-29
Clarification of checksum ID identification

andrewyatz:
See issue #329.

The specification currently states there are two checksum identifiers available of different length. Length should be used to detect which one has been given. If we need to adopt other checksums then the spec will evolve to include other methods to identify the checksum used.


andrewyatz:
I'm just going to merge this considering there's been no further comments on this

PR #335, opened 2018-09-04
Make "future enhancements" an unordered list

jmarshall:
At present, a change implementing e.g. item 3 is tempted to sully its diffs by renumbering items 4 to 6 after removing item 3. e.g. both PRs #285 and #320 do this.

Also remove the item about adding variants, which was implemented by PR #301.

@mlin: this is a trivial editorial change, so as editor you might want to just apply it. Then when those two significant PRs are rebased past this commit, their one-line changes to the _Possible future enhancements_ section will become clearer.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -375,12 +375,11 @@ The data block URL and headers might contain embedded authentication tokens; the
 
 # Possible future enhancements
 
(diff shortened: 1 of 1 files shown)


jeromekelleher:
+1

PR #336, opened 2018-09-12
Product review committee changes v1

andrewyatz:
Modifications related to the first PRC

Changes have been made in response to the first PRC for refget. This includes

- Adding a better description to the problem  the specification is solving
- Clarification of the OpenAPI schema
- Improving text in protocol essentials (Range and query parameters are allowed)
- Adding an example to `/sequence/<id>`
- What happens when start and end are the same value (a zero length string is returned)
- Adding Reece Hart as a contributor
- Typos


andrewyatz:
First set of changes now in. We have a couple more changes to discuss before we okay this

andrewyatz:
Going to merge this ahead of the announcement today

PR #339, opened 2018-09-15
Fix refget-openapi.yaml's Range header description

jmarshall:
PR #331 noted that the text for `Range` in the YAML file was inconsistent with that in the specification.

While it was a Google Docs document, the specification was corrected to reflect that HTTP Range byte ranges are zero-based fully-inclusive. Update the OpenAPI description correspondingly.

Also fix typos.

### DIFF
--- a/pub/refget-openapi.yaml
+++ b/pub/refget-openapi.yaml
@@ -14,9 +14,9 @@ info:
     url: https://www.ga4gh.org/
 tags:
   - name: Metadata
-    description: Services linked to retriving metadata from an idetifier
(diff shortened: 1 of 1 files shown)


jmarshall:
Also: the YAML file mentions a `/ping` endpoint. This is not described in the spec document. Should it be added to the spec or removed from the YAML (perhaps because it has been superseded by `/sequence/service-info`)?

andrewyatz:
Quite right about `/ping`. It's a custom endpoint I had on my implementation (the original YAML was based on my server version). It should be removed and I will get that going now

Issue #341, opened 2018-09-18
Add a table of contents to the refget specification


andrewyatz:
The GA4GH product review committee suggested adding a table of contents to the refget specification. If done in JavaScript this could be reused on any specification, which renders in HTML such as refget and htsget.

andrewyatz:
I'll experiment with both options. The liquid package might be better as that's compatible with any HTML versus the kramdown option, which seems to only work for markdown files. Still though both are excellent finds and 👍 for a non-JavaScript solution

andrewyatz:
Closing this as even though @mlin referred to this in #346 this should still be closed

PR #343, opened 2018-09-20
Support table of contents in refget and htsget. See issue #341.

andrewyatz:
During refget's PRC, the reviewers suggested adding a table of contents to the specification. Kramdown can support this via the `{:toc}` tag. Anything that should not be included in the TOC is tagged with `{:.no_toc}`.

The config is setup to build the TOC for all `h1`/`h2`/`h3` tags, which on first inspection looks sufficient.

Because both refget and htsget specifications are written in Markdown it made sense to apply this to both documents.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -13,3 +13,4 @@ exclude: ["*.sh", "*.tex", img, new, Makefile, MAINTAINERS.md, README.md]
 markdown: kramdown
 kramdown:
   parse_block_html: true
+  toc_levels: 1..3
--- a/refget.md
+++ b/refget.md
@@ -5,6 +5,10 @@ suppress_footer: true
 ---
 
 # Refget API Specification v0.2
+{:.no_toc}
+
+* Do not remove this line (it will not be displayed)
+{:toc}
 
 ## Design principles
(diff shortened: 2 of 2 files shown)


andrewyatz:
@jmarshall had a look at this last Thursday and thought it was fine. @mlin if you’re unhappy with the change let me know by tomorrow else I’ll merge this in

htsget.md
```diff
@@ -57,13 +61,13 @@ _object_
 Container for response object.
 <table>
 <tr markdown="block"><td>
-`error`  
+`error`
```
jmarshall: Don't remove the trailing whitespace on these lines. They are [kramdown line breaks](https://kramdown.gettalong.org/quickref.html#paragraphs). (I think the two-backslashes notation didn't exist at the time and might be worth converting to, but that would be a separate change.)

andrewyatz: Sorry you're right. I shouldn't have removed them. I think it's my text editor setting that strips trailing whitespace out. On reflection I'll revert the htsget changes and let those guys introduce in TOC if they want as you've suggested.

andrewyatz:
Changes to `htsget.md` have been reverted.

PR #359, opened 2018-11-13
CRAM spec updates; large!

### LINKED ISSUE #353: Clarification and expansion of CRAM Section 13.3: Codings
I am working on updating the HTSJDK code base for better compliance with the CRAM spec, and I'm having some difficulty understanding how some of the Codings are intended to work.  While I have been able to work through some of my confusion through external sources, it would be helpful to me (and future implementors) if the spec contained everything necessary to implement these encodings.

### Beta coding
* is this exclusive to CRAM, or is this a widely-known encoding?  Neither https://en.wikipedia.org/wiki/Beta_Code nor https://en.wikipedia.org/wiki/Beta_encoder appear relevant here.
* update the examples table to include the offset and clarify that all values in a series are stored with the same bit length
* clarify that the length parameter is the (constant) bit length per value
* why would one choose Beta encoding?  (example given for Golomb is for a geometric distribution)

### Golomb
* how would one choose the M parameter?
* describe the decoding algorithm
* add an example for the Generate Codeword c-ii-B case

### Subexponential
* is there a widely-known source reference for this encoding?  the best I could find was a reference in [this paper](https://link.springer.com/content/pdf/10.1007/s10677-004-4254-0.pdf) with the same encoding/decoding algorithm described here.
* the decoding section should be rewritten.  The instructions to "append" and "form a run" are unclear in this context.  If my understanding is correct, this is equivalent:
  * count 1 bits (as **i**) until the 0 separator bit is seen
  * calculate **tail-len** from **i** as described
  * after the separator 0 bit, read **tail-len** bits and add **2^(tail-len)**

### Huffman
* link to https://en.wikipedia.org/wiki/Canonical_Huffman_code
* provide a brief overview of the process:
  * generate frequency table from input values
  * use frequency table to generate the Huffman codebook
  * canonicalize the codebook
  * decode using the canonical codebook

### Byte Array Length
* be more explicit that the data is stored as a block of array lengths, followed by a block of array values
* how are the length and values sub-encodings specified?  Do we set encoding ID (from the table in section 3) as parameter values?
* how do we pass encoding parameters to the sub-encodings?
* a table like that for Encoding in Section 2.3 would be helpful here

### Byte Array Stop
* why does this need a reference to an external block?
* is the external block encoded using one of the CRAM encodings, or are the byte arrays stored as-is?
* is this a copy/paste error?  "For example this could be a golomb encoding"

jkbonfield:
This is a good thrash at the CRAM spec, fixes #353 and also takes a good stab at making the data series decode order explicit rather than guesswork.  (The tables were almost correct already, but it was never explicitly mentioned.)

It's now using algorithmicx.sty.  I could use use algorithm.sty for slightly better formatting, but it's GPL and for want of a single extra line of formatting I couldn't be bothered to figure out what that means to a text document! (TeX of course is actually a language, but what is the "program" here?)

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -17,6 +17,9 @@
 \usepackage{fixltx2e}
 \usepackage{amssymb}
 \usepackage{fancyhdr}
+\usepackage{amsmath}
+\usepackage{algpseudocode}
+\usepackage{threeparttable}
 \renewcommand{\headrulewidth}{0pt}
 \renewcommand{\footrulewidth}{0pt}
 
@@ -170,8 +173,9 @@ bits. Alternatively, two codes can be combined, where the first contains the num
 of bits to read. 
 
 \subsection{\textbf{Writing bytes to a byte stream}}
+\label{subsec:writing-bytes}
 
-The interpretation of byte stream is straightforward. CRAM uses little \emph{endiannes} 
+The interpretation of byte stream is straightforward. CRAM uses little \emph{endianness} 
 for bytes when applicable and defines the following storage data types:
 
 \begin{description}
@@ -286,9 +290,9 @@ NULL & 0 & none & series not preserved\tabularnewline
 EXTERNAL & 1 & int block content id & the block content identifier used to associate 
 external data blocks with data series\tabularnewline
 \hline
-GOLOMB & 2 & int offset, int M & Golomb coding\tabularnewline
+Deprecated (GOLOMB) & 2 & int offset, int M & Golomb coding\tabularnewline
 \hline
-HUFFMAN\_INT & 3 & int array, int array & coding with int values\tabularnewline
+HUFFMAN & 3 & int array, int array & coding with int/byte values\tabularnewline
 \hline
 BYTE\_ARRAY\_LEN & 4 & encoding\texttt{<}int\texttt{>} array length, encoding\texttt{<}byte\texttt{>} 
 bytes & coding of byte arrays with array length\tabularnewline
@@ -300,7 +304,7 @@ BETA & 6 & int offset, int number of bits & binary coding\tabularnewline
 \hline
 SUBEXP & 7 & int offset, int K & subexponential coding\tabularnewline
 \hline
-GOLOMB\_RICE & 8 & int offset, int log2m & Golomb-Rice coding\tabularnewline
+Deprecated (GOLOMB\_RICE) & 8 & int offset, int log$_{\text{2}}$m & Golomb-Rice coding\tabularnewline
 \hline
 GAMMA & 9 & int offset & Elias gamma coding\tabularnewline
 \hline
@@ -570,6 +574,7 @@ into the file.
 \end{itemize}
 
 \subsection{\textbf{Compression header block}}
+\label{subsec:compression-header}
 
 The compression header block consists of 3 parts: preservation map, data series 
 encoding map and tag encoding map.
@@ -600,55 +605,31 @@ section\tabularnewline
 \subsubsection*{Data series encodings}
 
 Each data series has an encoding. These encoding are stored in a map with byte[2] 
-keys:
+keys and are decoded in approximately this order\footnote{The precise order is defined in section~\ref{sec:record}.}:
 
+\begin{threeparttable}[t]
 \begin{tabular}{|l|l|>{\raggedright}p{100pt}|>{\raggedright}p{220pt}|}
 \hline
 \textbf{Key} & \textbf{Value data type} & \textbf{Name} & \textbf{Value}\tabularnewline
 \hline
-BF & encoding\texttt{<}int\texttt{>} & bit flags & see separate section\tabularnewline
+BF & encoding\texttt{<}int\texttt{>} & BAM bit flags & see separate section\tabularnewline
 \hline
-AP & encoding\texttt{<}int\texttt{>} & in-seq positions & 0-based alignment start 
-delta from previous record *\tabularnewline
+CF & encoding\texttt{<}int\texttt{>} & CRAM bit flags & see specific section\tabularnewline
 \hline
-FP & encoding\texttt{<}int\texttt{>} & in-read positions & positions of the read 
-features\tabularnewline
+RI & encoding\texttt{<}int\texttt{>} & reference id & record reference id from
+the BAM file header\tabularnewline
 \hline
 RL & encoding\texttt{<}int\texttt{>} & read lengths & read lengths\tabularnewline
 \hline
-DL & encoding\texttt{<}int\texttt{>} & deletion lengths & base-pair deletion lengths\tabularnewline
-\hline
-NF & encoding\texttt{<}int\texttt{>} & distance to next fragment & number of records 
-to the next fragment*\tabularnewline
-\hline
-BA & encoding\texttt{<}byte\texttt{>} & bases & bases\tabularnewline
-\hline
-QS & encoding\texttt{<}byte\texttt{>} & quality scores & quality scores\tabularnewline
-\hline
-BB & encoding\texttt{<}byte[ ]\texttt{>} & stretches of bases & bases\tabularnewline
-\hline
-QQ & encoding\texttt{<}byte[ ]\texttt{>} & stretches of quality scores & quality scores\tabularnewline
-\hline
-FC & encoding\texttt{<}byte\texttt{>} & read features codes & see separate section\tabularnewline
-\hline
-FN & encoding\texttt{<}int\texttt{>} & number of read features & number of read 
-features in each record\tabularnewline
-\hline
-BS & encoding\texttt{<}byte\texttt{>} & base substitution codes & base substitution 
-codes\tabularnewline
-\hline
-IN & encoding\texttt{<}byte[ ]\texttt{>} & insertion & inserted bases\tabularnewline
+AP & encoding\texttt{<}int\texttt{>} & in-seq positions & 0-based alignment start
+delta from previous record *\tabularnewline
 \hline
 RG & encoding\texttt{<}int\texttt{>} & read groups & read groups. Special value 
 `-1' stands for no group.\tabularnewline
 \hline
-MQ & encoding\texttt{<}int\texttt{>} & mapping qualities & mapping quality scores 
- \tabularnewline
+RN\tnote{a} & encoding\texttt{<}byte[ ]\texttt{>} & read names & read names\tabularnewline
 \hline
-TL & encoding\texttt{<}int\texttt{>} & tag ids  & list of tag ids, see tag encoding 
-section\tabularnewline
-\hline
-RN & encoding\texttt{<}byte[ ]\texttt{>} & read names & read names\tabularnewline
+MF & encoding\texttt{<}int\texttt{>} & next mate bit flags & see specific section\tabularnewline
 \hline
 NS & encoding\texttt{<}int\texttt{>} & next fragment reference sequence id & reference 
 sequence ids for the next fragment \tabularnewline
@@ -658,15 +639,30 @@ for the next fragment\tabularnewline
 \hline
 TS & encoding\texttt{<}int\texttt{>} & template size & template sizes\tabularnewline
 \hline
-MF & encoding\texttt{<}int\texttt{>} & next mate bit flags & see specific section\tabularnewline
+NF & encoding\texttt{<}int\texttt{>} & distance to next fragment & number of records
+to the next fragment\tnote{b}\tabularnewline
 \hline
-CF & encoding\texttt{<}int\texttt{>} & compression bit flags & see specific section\tabularnewline
+TL\tnote{c} & encoding\texttt{<}int\texttt{>} & tag ids  & list of tag ids, see tag encoding
+section\tabularnewline
 \hline
-TM & encoding\texttt{<}int\texttt{>} & test mark & a prefix expected before every 
-record, for debugging purposes.\tabularnewline
+FN & encoding\texttt{<}int\texttt{>} & number of read features & number of read
+features in each record\tabularnewline
 \hline
-RI & encoding\texttt{<}int\texttt{>} & reference id & record reference id from 
-the BAM file header\tabularnewline
+FC & encoding\texttt{<}byte\texttt{>} & read features codes & see separate section\tabularnewline
+\hline
+FP & encoding\texttt{<}int\texttt{>} & in-read positions & positions of the read
+features\tabularnewline
+\hline
+DL & encoding\texttt{<}int\texttt{>} & deletion lengths & base-pair deletion lengths\tabularnewline
+\hline
+BB & encoding\texttt{<}byte[ ]\texttt{>} & stretches of bases & bases\tabularnewline
+\hline
+QQ & encoding\texttt{<}byte[ ]\texttt{>} & stretches of quality scores & quality scores\tabularnewline
+\hline
+BS & encoding\texttt{<}byte\texttt{>} & base substitution codes & base substitution
+codes\tabularnewline
+\hline
+IN & encoding\texttt{<}byte[ ]\texttt{>} & insertion & inserted bases\tabularnewline
 \hline
 RS & encoding\texttt{<}int\texttt{>} & reference skip length & number of skipped 
 bases for the `N' read feature\tabularnewline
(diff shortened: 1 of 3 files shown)


lbergelson:
@jmthibault79 Relevant to your work.

jmthibault79:
Thank you!  I will review and comment inline.

CRAMv3.tex
```diff
@@ -920,138 +921,325 @@ \section{\textbf{Record structure}}
...
 \hline
-1 & int & BF & CRAM bit flags & see CRAM record bit flags\tabularnewline
+int & CF  & compression bit flags & see compression bit flags\tabularnewline
 \hline
-2 & int & CF  & compression bit flags & see compression bit flags\tabularnewline
+- & - & Positional data & See section \ref{subsec:positions}\tabularnewline
 \hline
-3 & int & RI & ref id & reference sequence id, not used for single reference slices, 
-reserved for future multiref slices. \tabularnewline
+- & - & Read names & See section \ref{subsec:readnames}\tabularnewline
```
jmthibault79: Reference is `names` below (line 1076) so this does not render

CRAMv3.tex
```diff
@@ -920,138 +921,325 @@ \section{\textbf{Record structure}}
...
 \hline
-4 & int & RL & read length & the length of the read\tabularnewline
+- & - & Mate records & See section \ref{subsec:mate}\tabularnewline
 \hline
-5 & int & AP & alignment start & the alignment start position *1\tabularnewline
+- & - & Auxiliary tags & See section \ref{subsec:tags}\tabularnewline
 \hline
-6 & int & RG & read group & the read group identifier\tabularnewline
+- & - & Sequences & See sections \ref{subsec:mapped}and \ref{subsec:unmapped}\tabularnewline
```
jmthibault79: Needs a space after `{subsec:mapped}`

CRAMv3.tex
```diff
@@ -920,138 +921,325 @@ \section{\textbf{Record structure}}
...
+\subsubsection*{\textbf{BAM bit flags (BF data series)}}
+
+The following flags are duplicated from the SAM and BAM specification, with identical meaning.
+Note however some of these flags can be derived during decode, so may be omitted in the CRAM file and the bits computed based on both reads of a pair-end library residing within the same slice.
+
+\begin{tabular}{|>{\raggedright}p{144pt}|>{\raggedright}p{144pt}|>{\raggedright}p{144pt}|}
+\hline
+\textbf{Bit flag} & \textbf{Comment} & \textbf{Description}\tabularnewline
+\hline
+0x1 & ! 0x40 \&\& ! 0x80 & template having multiple segments in sequencing\tabularnewline
```
jmthibault79: Could you add more to this comment?  I have a few possible interpretations for the correct behavior so I'd like that to be made explicit here.

* Reading
1. bit `0x1` is strictly ignored, but stored as-is for round-tripping equality.  Always use the `0x40` and `0x80` values as the source of truth.
1. store bit `0x1` as-is.  If it conflicts with `0x40` and `0x80`, use the `0x1` value.  
1. bit `0x1` should be validated against the calculation from the `0x40` and `0x80` values, and the record rejected if these do not match
1. bit `0x1` should be corrected (overwritten) by the calculation from the `0x40` and `0x80` values

* Writing
1. write out the `0x1` bit from the `0x40` and `0x80` values, regardless of what it was previously
1. write out the `0x1` bit as the original value, even if it conflicts with the `0x40` and `0x80` values

CRAMv3.tex
```diff
@@ -920,138 +921,325 @@ \section{\textbf{Record structure}}
...
 \hline
-8 & byte[ ] & RN & read name & the read names (if preserved)\tabularnewline
+0x2 &  & each segment properly aligned according to the aligner\tabularnewline
 \hline
-9 & *2 & *2 & mate record & *2 (if not the last record)\tabularnewline
+0x4 &  & segment unmapped\tabularnewline
 \hline
-10 & int & TL & tag ids & tag ids *3\tabularnewline
+0x8 & calculated* or stored in the mate's info & next segment in temlate unmapped\tabularnewline
```
jmthibault79: spelling: template

CRAMv3.tex
```diff
@@ -920,138 +921,325 @@ \section{\textbf{Record structure}}
...
+is unknown and that any encoded reference differences are present only to
+recreate the CIGAR string.\tabularnewline
 \hline
 \end{tabular}
 
-*1 The AP data series is delta encoded for reads mapped to a single reference slice 
-and normal integer value in all other cases. 
 
-*2 See \emph{mate record} section.
+The following pseudocode describes the general process of decoding an entire CRAM record.
```
jmthibault79: This is excellent, thank you.

CRAMv3.tex
```diff
@@ -920,138 +921,325 @@ \section{\textbf{Record structure}}
...
 
-\begin{tabular}{|>{\raggedright}p{36pt}|>{\raggedright}p{70pt}|>{\raggedright}p{74pt}|>{\raggedright}p{85pt}|>{\raggedright}p{177pt}|}
+The AP data series is delta encoded for reads mapped to a position-sorted slice containing data from a single reference, and as a normal integer value in all other cases.
+
+\begin{tabular}{|>{\raggedright}p{70pt}|>{\raggedright}p{75pt}|>{\raggedright}p{90pt}|>{\raggedright}p{171pt}|}
+\hline
+\textbf{Data series type} & \textbf{Data series name} & \textbf{Field} & \textbf{Description}\tabularnewline
 \hline
- & \textbf{Data series type} & \textbf{Data series name} & \textbf{Field} & \textbf{Description}\tabularnewline
+int & RI & ref id & reference sequence id (only used in multiref slices)\tabularnewline
```
jmthibault79: Should this be set to a particular constant when unused - or ignored?

CRAMv3.tex
```diff
@@ -920,138 +921,325 @@ \section{\textbf{Record structure}}
...
+\hline
+int & RG & read group & the read group identifier expressed as the N\textsuperscript{th} record in the header, starting from 0 with -1 for no group\tabularnewline
 \hline
 \end{tabular}
 
-*1 See read feature record specification below.
+\vskip 20pt
+\begin{algorithmic}[1]
+\Procedure{DecodePositions}{}
+\If{$slice.ref\_id = -2$}
```
jmthibault79: Please indicate here that `-2` means multi-ref

CRAMv3.tex
jkbonfield: These came from the SAM specification.  I expect Vadim added them because he wanted the CRAM specification to be self contained, which is reasonable.  It's overly terse and I agree confusing.

I'll review them and see what the SAM spec has to say too.  Do we want to just lift the SAM text verbatim, along with all the bullet points elaborating on them?  I think probably we can find a way to add them to the description part of the text, perhaps with footnotes for further explanatory bits.

CRAMv3.tex
jkbonfield: Resolving things I've done locally just for ease of tracking.  I'll push an update once I've completed my round of revisions.

Thanks for taking the time to review.

CRAMv3.tex
jkbonfield: Thanks. :)

I am aware the codec section could be rewritten using algorithmicx too for consistency, but I added it after having gone through them.  I think this is maybe a second round of updates.  There's a lot more to do and incremental improvements is the easiest way given how much of the spec still needs a refresh.

CRAMv3.tex
jkbonfield: Looking at my C code, I see I store BF verbatim (this is inefficient perhaps).  During read I start with the stored value and then modify it in some manners.

- Add 0x1 if a mate pair is present in this slice.
- Add 0x8 if mate is in this slice and is unmapped
- Add 0x20 if mate is in this slice and reverse complemented.

In other words it copes with these bits being missing and will repopulate (or even correct them if wrong), but it does store them.  I'm unsure why I did it this way.  Maybe the Java code doesn't (or didn't) store all these flags, or perhaps it's related to earlier CRAM versions.  I recall one of them having some bits missing - CRAMv1 maybe - and always generating them.  This doesn't work well though when the mate-pair is outside this slice.

I'd need to check the Java code to see what that does.

CRAMv3.tex
```diff
@@ -1139,96 +1327,94 @@ \subsubsection*{Base substitution codes (BS data series)}
...
-4 & int & NP & mate alignment start position \tabularnewline
-\hline
-5 & int & TS & the size of the template (insert size)\tabularnewline
-\hline
-\end{tabular}
-
-\subsubsection*{}
-
-\subsubsection*{Next mate bit flags (MF data series)}
+\subsubsection*{Decoe mapped read pseudocode}
```
jmthibault79: spelling: Decode

CRAMv3.tex
```diff
@@ -1708,36 +1894,42 @@ \subsubsection*{Introduction}
...
 It can also be used for monotonically decreasing distributions with the maximum 
 not equal to zero. For example, given offset is 10 and the value to be encoded 
 is 1, the actually encoded value would be offset+value=11. Then when decoding, 
 the offset would be subtracted from the decoded value. 
 
-\subsubsection*{Beta coding}
+\subsubsection{External: codec ID 1}
 
-\subsubsection*{Definition}
+Can encode types \textit{Byte}, \textit{Integer}.
```
jmthibault79: + Long

CRAMv3.tex
```diff
@@ -1708,36 +1894,42 @@ \subsubsection*{Introduction}
...
-2 & 10\tabularnewline
+\textbf{Data type} & \textbf{Name} & \textbf{Comment}
+\tabularnewline
 \hline
-4 & 100\tabularnewline
+itf8 & external id & id of an external block containing the byte stream\tabularnewline
 \hline
 \end{tabular}
 
+\subsubsection{Beta coding: codec ID 6}
```
jmthibault79: Curious about the ordering: why not in ID order?

CRAMv3.tex
```diff
@@ -1746,13 +1938,38 @@ \subsubsection*{Parameters}
...
+itf8 & offset & offset is subtracted from each value during decode\tabularnewline
 \hline
 itf8 & length & the number of bits used\tabularnewline
 \hline
 \end{tabular}
 
-\subsubsection*{Gamma coding}
+\subsubsection*{Examples}
+
+If we have integer values in the range 10 to 15 inclusive, the largest value would traditionally need 4 bits, but with an offset of -10 we can hold values 0 to 5, using a fixed size of 3 bits.
```
jmthibault79: Great example.

CRAMv3.tex
```diff
@@ -1996,37 +2226,74 @@ \subsubsection*{Parameters}
...
+\item Obtain symbol code lengths.
+\begin{itemize}
+\item If encoding:\\
+- Compute symbol frequencies.\\
+- Compute code lengths from frequencies.
+\item If decoding:\\
+- Read code lengths from codec parameters.
+\end{itemize}
+
+\item Compute canconical huffman codewords from code lengths\footnote{\url{https://en.wikipedia.org/wiki/Canonical_Huffman_code}}.
```
jmthibault79: Spelling: canonical

CRAMv3.tex
```diff
@@ -2035,22 +2302,31 @@ \subsubsection*{Parameters}
...
+\subsubsection*{BYTE\_ARRAY\_LEN: codec ID 4}
+
+Can encode types \textit{Byte[]}.
+
+Byte arrays are captured length-first, meaning that the length of every array element is written using an additional encoding.
+For example this could be a HUFFMAN encoding or another EXTERNAL block.
+The length is decoded first followed by the data, followed by the next length and data, and so on.
+
+This encoding can therefore be considered as a nested encoding, with each pair of nested encodings containing their own set of parameters.
+The byte stream for parameters of the BYTE\_ARRAY\_LEN encoding is therefore the concatenation of the length and value encoding parameters as described in section~\ref{subsec:writing-bytes}.
```
jmthibault79: Could you expand on this a little further with an example?  I'll make an attempt here:

As an example, if we used BETA encoding for length and EXTERNAL encoding for values, the parameters would be [ 6 (BETA encoding ID), len(beta params), beta params, 1 (EXTERNAL encoding ID), len(external params), external params ]

CRAMv3.tex
jmthibault79: Setting 0x1 if it has a mate pair fits the description, but I'm not understanding what that has to do with the 0x40 and 0x80 flags.

CRAMv3.tex
jmthibault79: Yes, I like that

CRAMv3.tex
jkbonfield: I've no idea to be honest!  I didn't reorder anything, so this is as per the original author.  I'll take a look, but I can see a benefit to having golomb-rice immediately next to golomb.

Maybe we can list the recommended ones in order and then the deprecated ones that none of us use (in their own order).   It's arguable whether deprecating them is a change of specification that goes above simple clarification, but the practical side of it is they don't exist in the wild anyway.

CRAMv3.tex
jkbonfield: I'm struggling to understand that myself too actually.  I'll look at the Java code - maybe there is a clue there.  Agreed it isn't clear.

jkbonfield: I just deleted it. :-)

It makes no sense and seems to have no apparent relationship to how either of our implementations work.  I guess it's something lost in the mists of time.

CRAMv3.tex
jkbonfield: I've added Long to io_lib for my CRAMv4 prototype and I see it's in htsjdk too, but htslib has no support for external long.

This section is ambiguous though as it describes Int and Byte being acceptable and then mentions Long in the next sentence.  In practice CRAM v3 (and before) have no longs in any data series, with long restricted purely to ltf8 encoding within container and slice headers.

We should be consistent though.  For v3 I think just drop long from external as it's not a possibility.

CRAMv3.tex
jkbonfield: Fixed reordering.  Also shuffled some other things around in this appendix.

CRAMv3.tex
jkbonfield: Good idea.

I went with HUFFMAN + external.  This has the benefit of showing how constant sized values can still utilise the byte array len mechanism, but in an efficient manner that doesn't actually use any storage for length field.  (I use this for integer aux tags.)

jkbonfield:
Another round of updates.

Hopefully this fixes most of the concerns.  Thanks for your comments @jmthibault79

CRAMv3.tex
jmthibault79: OK.  The HTSJDK implementation of External Long is broken anyway.

CRAMv3.tex
```diff
@@ -920,143 +922,363 @@ \section{\textbf{Record structure}}
...
 \hline
-11 & byte[ ] & - & tag values & tag values *3\tabularnewline
+0x800 &  & Supplementary alignment\tabularnewline
 \hline
 \end{tabular}
+\begin{tablenotes}
+\item[a] Bit 0x4 is the only reliable place to tell whether the read is unmapped.  If 0x4 is set, no assumptions may be made about bits 0x2, 0x100 and 0x800.
+\item[b] For segments within the same slice.
+\item[c] Bits 0x40 and 0x80 reflect the read ordering within each template inherent in the sequencing technology used, which may be independent from the actual mapping orientation.
+If 0x40 and 0x80 are both set, the read is part of a linear template (one where the template sequence is expected to be in a linear order), but it is neither the first nor the last read.
```
jmthibault79: This is counterintuitive so I want to confirm: "first" = true and "last" = true means it is neither first nor last?

Maybe this would be better represented by a two-bit flag 0xC0 with 4 meaningful values?

CRAMv3.tex
```diff
@@ -920,143 +922,363 @@ \section{\textbf{Record structure}}
 other hand, data series that do not require binary coding can be stored separately 
 in external blocks with some other compression applied to them independently.
 
+As CRAM data series may be interleaved within the same blocks\footnote{Interleaving can sometimes provide better compression, however it also adds dependency between types of data meaning it is not possible to selectively decode one data series if it co-locates with another data series in the same block.} understanding the order in which CRAM data series must be decoded is vital.
```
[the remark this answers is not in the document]
jkbonfield: No, we can use the same external block ID and interleave to external blocks too.  This can improve compression ratios too if we are putting data with similar frequency distributions in the same block or data with strong correlation between values.

Eg imagine we had one tag X0 which is number of mismatches in this read and X1 which is number of mismatches in this plus mate.  There will be a strong correlation between the value of X1 and the value of X0 (at the very least X1 >= X0), so putting both together would mean an order-1 entropy encoder will probably do a decent job.

In practice we don't do this any more, so it remains untapped potential.

CRAMv3.tex
jkbonfield: This comes directly from the SAM specification, which basically has identical text bar a bit of clarification on "linear template".

I don't want to confuse things too much for people who are familiar with SAM and while I agree there is potential for a better way to describe it, I think that ship has sailed.

CRAMv3.tex
```diff
@@ -1319,440 +1529,347 @@ \subsubsection*{BAM index}
...
-a & a &  646 \\
-  & b & 1725 \\
-  & c &  862 \\
-  & d &  862 \\
-\hline
-b & r & 4095 \\
-\hline
-c & a & 4095 \\
+
+For example, the bytes specifying a BYTE\_ARRAY\_LEN encoding, including the codec and parameters, for a 16-bit X0 auxiliary tag (``X0C'') may use HUFFMAN encoding to specify the length (always 2 bytes) and an EXTERNAL encoding to store the value to an external block with ID 200.
```
jmthibault79: Great!

CRAMv3.tex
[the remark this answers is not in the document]
jmthibault79: Interesting!  I didn't realize that, thanks.

CRAMv3.tex
jmthibault79: ok

CRAMv3.tex
```diff
@@ -600,55 +605,30 @@ \subsubsection*{Preservation map}
...
 \hline
 RG & encoding\texttt{<}int\texttt{>} & read groups & read groups. Special value 
 `-1' stands for no group.\tabularnewline
 \hline
-MQ & encoding\texttt{<}int\texttt{>} & mapping qualities & mapping quality scores 
- \tabularnewline
-\hline
-TL & encoding\texttt{<}int\texttt{>} & tag ids  & list of tag ids, see tag encoding 
-section\tabularnewline
+RN\footnote{Note this is decoded after MF if ...} & encoding\texttt{<}byte[ ]\texttt{>} & read names & read names\tabularnewline
```
jmthibault79: I don't see this footnote in the rendered PDF.  Same problem for TL.

jmthibault79:
Done with this round of review, with one additional question: should we deprecate Gamma and Subexponential encodings as well?  My (limited) understanding is that they're similar to Golomb and Golomb-Rice in lack of use.

jkbonfield:
I think I have seen them in use in earlier versions of CRAM, unlike the golomb ones which got ditched prior to 1.0.

For CRAM v4 I'd like to either remove all of these and just use external codecs, or perhaps better to elevate them to be ways of writing day into external codecs (ie give them all a block ID instead of hard-coded to CORE) and permit bit-packing plus the possibility of a subsequent entropy encoding step.

If we had a bit-wise arithmetic coder coupled to a bit-packing method it *may* be better.  I haven't explored it though.

CRAMv3.tex
```diff
@@ -1319,440 +1529,347 @@ \subsubsection*{BAM index}
...
-cfreq_{i-1} + freq_{i-1} & \quad \textrm{if $i \geq 1$}
-\end{array}
-\right. $
-% \\*[8pt]
-% \indent $cfreq_{i} = \displaystyle \sum_{j=0}^{i-1} freq_j$
-}
+\subsubsection*{Decoding}
 
+\begin{enumerate}
+\item Read $i$ in unary form, counting the number of leading 1s (prefix) in the codeword (discard the trailing 0 bit).
```
jmthibault79: Can we make this `u` as well, since it's the same as above?

CRAMv3.tex
```diff
@@ -920,143 +922,363 @@ \section{\textbf{Record structure}}
...
+  \State $reference\_id\gets slice\_header.reference\_sequence\_id$
+\EndIf
+\State $read\_length \gets$ \Call{ReadItem}{RL, Integer}
+\If{$container\_pmap.AP\_delta \ne 0$}
+    \State $alignment\_position \gets$ \Call{ReadItem}{AP, Integer} + $last\_position$
+    \State $last\_position \gets alignment\_position$
+\Else
+    \State $alignment\_position \gets$ \Call{ReadItem}{AP, Integer}
+\EndIf
+\State $read\_group \gets$ \Call{ReadItem}{AG, Integer}
```
jmthibault79: typo: should be RG

jmthibault79:
I'm trying to map HTSJDK DataSeries to these updates, and there are a few I'm not finding here.  Could you add them?  (Or explain why they don't belong here)

* MQ - Mapping Quality
* TC - Tag Count
* TN - Tag Name and Type
* TM and TV - Test Marks (maybe these are intentionally out of spec)

CRAMv3.tex
jkbonfield: Fixed, thanks.

CRAMv3.tex
jkbonfield: Thanks.  I've reformatted these using LaTeX threeparttable.

CRAMv3.tex
jkbonfield: Done.

jkbonfield:
>     * MQ - Mapping Quality

MQ is an oversight, and now added (along with QS too for mapped reads).  It was in the pseudocode and the main table, but not the record description.

>     * TC - Tag Count
>     * TN - Tag Name and Type

TC and TN are the old CRAM 1.0 format way of doing aux tags.  They have been replaced by TL and the TD preservation map.

>     * TM and TV - Test Marks (maybe these are intentionally out of spec)

I think these are tests internal to the Java code.  They don't appear to exist in the files I've looked at.

Anyway I pushed a new copy of the PR now with the latest round of changes.

I also went for broke and added pseudo-code for the rANS decoder (I gave up at the encoder, but it's less important).  I have a noddy implementation to test it works (it didn't originally!) to check my description is correct.  It's hard to get right, but surprisingly simple when written down that way.

I think this is enough for now though.

jkbonfield:
@rbuels: would you please take a look at these CRAM spec modifications.  They are a lot of clarifications in codecs, record structure and the rANS decoder.  There is still much more to do for clarity - all the early sections haven't been touched.

Apart from labelling some codecs as deprecated because none of our code generates data that way, none of this is actually changing the file format in any way.  The intent is simply to clarify and in some cases to fix mistakes in the specification.

[CRAMv3.pdf](https://github.com/samtools/hts-specs/files/2600934/CRAMv3.pdf)

rbuels:
Some very nice improvements here! The pseudocode sections in particular are going to be very useful for future implementors of the spec.

rbuels:
I also am very fond of the explicit deprecations, those could save somebody a lot of time!

jkbonfield:
Thanks for the comments all.  Sorry @rbuels these all came too late for your implementation!  Was there anything specific you found incomprehensible that you had to refer to the source code to work out? (Probably lots, sadly.)

The work above is incomplete as there is much to of the spec to refresh, but I propose merging it after our next conference call as a step-wise improvement.  Part 1 of several, hopefully.

PR #361, opened 2018-11-21
Clarify SAM queryname sort order.

jkbonfield:
The queryname sort order doesn't have anything to say regarding the lexicographical order.  This has caused confusion before, especially given htsjdk and htslib differ in their sort orders.  This ship has sailed, but we can at least explicitly acknowledge this in the specification.

Addresses an issue raised here https://github.com/samtools/samtools/issues/979

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -248,16 +248,19 @@ meanings can be avoided.}
   key is the {\sf RNAME} field, with order defined by the order of {\tt @SQ} lines in the header.  The
   minor sort key is the {\sf POS} field.  For alignments with equal {\sf RNAME} and {\sf POS}, order is
   arbitrary.  All alignments with `{\tt *}' in {\sf RNAME} field follow alignments with some other
-  value but otherwise are in arbitrary order.\\\cline{2-3}
+  value but otherwise are in arbitrary order.
+  For queryname sort, no explicit requirement is made regarding the ordering other than that it be applied consistently throughout the entire file.%
+  \footnote{It is known that widely used software libraries have differing definitions of the queryname sort order, meaning care should be taken when operating on multiple files of varying provenance.
+  Tools may wish to use the sub-sort field to explicitly distinguish between natural and lexicographical ordering.  See Section~\ref{sec:sub-sort}.}\\\cline{2-3}
   & {\tt GO} & Grouping of alignments, indicating that similar alignment records
     are grouped together but the file is not necessarily sorted overall.
     \emph{Valid values}: {\tt none} (default), {\tt query} (alignments are
     grouped by {\sf QNAME}), and {\tt reference} (alignments are grouped by
     {\sf RNAME}/{\sf POS}).\\\cline{1-3}
-  & {\tt SS} & Sub-sorting order of alignments. Valid values are of the form \emph{sort-order}{\tt :}\emph{sub-sort}, where \emph{sort-order} is the same value stored in the {\tt SO} tag and \emph{sub-sort} is an implementation-dependent colon-separated string further describing the sort order.%
+  & {\tt SS} & Sub-sorting order of alignments. Valid values are of the form \emph{sort-order}{\tt :}\emph{sub-sort}, where \emph{sort-order} is the same value stored in the {\tt SO} tag and \emph{sub-sort} is an implementation-dependent colon-separated string further describing the sort order, but with some predefined terms defined in Section~\ref{sec:sub-sort}.
+    For example, if an algorithm relies on a {\tt coordinate} sort that, at each coordinate, is further sorted by query name then the header could contain {\tt @HD SO:coordinate SS:coordinate:queryname}.%
 \footnote{The repetition of \emph{sort-order} enables a limited form of validation.
 For example, {\tt @HD SO:queryname SS:coordinate:TLEN} would indicate that the data has been re-sorted (by query name) by a non-{\tt SS}-aware tool and the {\tt SS} field should be ignored.}
-    For example, if an algorithm relies on a {\tt coordinate} sort that, at each coordinate, is further sorted by query name the header could contain {\tt @HD SO:coordinate SS:coordinate:queryname}.
     If the primary sort is not one of the predefined primary sort orders, then {\tt unsorted} should be used and the subsort is effectively the major sort.
     For example, if sorted by an auxillary tag {\tt MI} then by coordinate then the header could contain {\tt @HD SO:unsorted SS:unsorted:MI:coordinate}.
     \newline
@@ -337,6 +340,24 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   \cline{1-3}
 \end{longtable}
 \end{center}
+
+\subsubsection{Defined sub-sort terms}\label{sec:sub-sort}
+
+While the {\tt SS} sub-sort field allows implementation-defined keywords, some terms are predefined with specific meanings.
+
+\begin{description}
+\item[lexicographical] sort order is defined as a character-based dictionary sort with the character order as defined by the POSIX C locale.
+  For example ``abc'', ``abc17'', ``abc5'', ``abc59'' and ``abcd'' are in lexicographical order.
+
+\item[natural] sort order is similar to lexicographical order except that runs of adjacent digits are considered to be numbers embedded within the text string, ordered numerically when compared to each other and ordered as single digits when compared to the surrounding non-digit characters.
+Runs that differ only in the number of leading zeros (thus are numerically tied) are ordered by more-zeros coming before fewer-zeros.
+The characters `\texttt{-}' and `\texttt{.}' are considered as ordinary characters, so apparently negative or fractional values are not treated as part of an embedded number.
+  For example, ``abc'', ``abc+5'', ``abc-5'', ``abc.d'', ``abc03'', ``abc5'', ``abc008'', ``abc08'', ``abc8'', ``abc17'', ``abc17.+'', ``abc17.2'', ``abc17.d'', ``abc59'' and ``abcd'' are in natural order.
+
+\item[umi] is a lexicographical sort by the UMI tag.
+  The {\tt MI} tag should be used for comparing UMIs. The RX tag may be used in its absence but is not guaranteed to be unique across multiple libraries.
+
+\end{description}
 
 \subsubsection{Reference MD5 calculation}\label{sec:ref-md5}
 The {\tt M5} tag on {\tt @SQ} lines allows reference sequences to be uniquely identified through the MD5 digest of the sequence itself.


jmarshall:
There is lots of background to this in #5.

Needs a `%` at the end of the preceding line (or lines joined as `…the entire file.\footnote{…`) to avoid the excess space before the footnote mark.
[Lexicographic order](https://en.wikipedia.org/wiki/Lexicographical_order) means “what `strcmp()` does” and the point is that `queryname` doesn't require use of a plain lexicographic sort. So just using “ordering” or similar in this text.

jkbonfield:
There can be multiple lexicographical orderings depending on locale, which is why I said it that way.  However on re-read I think it's not sufficient anyway as the sort order doesn't currently match any locale for samtools.  So yes I think maybe "ordering" is more appropriate.

Regarding sub-sort - that's an orthogonal issue and hence a separate PR IMO.  I was looking to just clarify existing behaviour rather than to promote a new one.

lbergelson:
This feels like we're essentially just abandoning queryname sort and just defining it as query grouped.

jkbonfield:
Not really.  You still need to have a sort order rather than collation.  Now possibly that sort ordering is just "hash(name)" which looks more like grouping again, but that's really clutching at straws and not in the intended spirit.  I don't know how you define "sorted" to forbid utterly whacky orderings while still retaining the abillity to do alphabet vs alphanumeric style sorting.

Maybe to be explicit if we wanted to add lexicographical to subsort then we could add alphanumeric too?  Basically our aim is to distinguish between

a1
a2
a7
a9
a10
a12
a88

and

a1
a10
a12
a2
a7
a88
a9

Both of these are in active use and both are defined as queryname order.  The horse has bolted so we cannot change that now, but we can perhaps provide ways of helping (not requiring) tools to be more explicit.

jmarshall:
> Maybe to be explicit if we wanted to add lexicographical to subsort then we could add alphanumeric too?

There's any number of orderings that could be described as “alphanumeric”, and it's non-trivial to describe the ordering that samtools uses (the mailing list thread pointed to near the start of #5 tried to describe it in pseudocode).

OTOH “lexicographical” is well-known, easily described (and already described elsewhere), not tied to particular implementations, and unique (provided you tie down locale vagaries)…

jkbonfield:
Maybe another way of looking at it is keeping order but redefining the token size.  Htsjdk strictly uses bytes.  Samtools uses strings of bytes or digits.

Eg

a12 is "a" "1" "2" or "a" "12".

"a" "12" can be greater than "a" "2" for some definition of greater, while "a" "1" "2" is not greater than "a" "2" using the same greater definition.

Imagine names like

abc_xyz_12
abc_xyz_2
abcd_xyz_7
pnm_xyz_12

Byte wise, "d" > "\_" so abcd_xyz_7 comes after pnm_xyz_12.  Token wise, using an arbitrary definition of "\_" as a delimiting character, "abcd" < "pnm" so the order is reverse.  My point is there are many ways we can sort by name that are equally valid, but it mostly boils down to byte by byte vs token by token coupled with a locale collation order.

yfarjoun:
if the purpose is to permit the two implementations (htsjdk and samtools) but not open the possibilities up entirely, I think that we should specify "lexicographic" and "alphanumerical" (or similar) as specific sub-sorts and define their meaning in the spec. That way we can validate the sort order internally. Theoretically, we would also be able to index a queryname-sorted bam if we did that....

jkbonfield:
I originally used "alphanumeric" here, but on googling it found that "natural" is the common term used.  Wikipedia (https://en.wikipedia.org/wiki/Natural_sort_order) defines it quite succintly as:

> Natural sort order is an ordering of strings in alphabetical order, except that multi-digit numbers are ordered as a single character.

I think that's reasonable provided we give an example, "e.g. abc5, abc17, abc59 are in natural sort order".

yfarjoun:
The example isn't complete without specifying whether 'abc123' come before or after 'abcd' and the logic for that decision. 

Note that the wikipedia page doesn't explain that either.

jkbonfield:
Hmm maybe, but we don't explain that for lexicographical either - it's just part of the locale.  Maybe we should state natural sort is a lexicographical sort with the extra tweak that multi-digit numbers are ordered as if they are a single character.

If we really want to be explicit regarding abc123 vs abcd then we'd have to also define in the collation order where multi-digit values sit in regard to single digit numbers and letters / punctuation, but it's basically "the obvious place" as there really is only one point they can be if you think about it; just after all the other single digit values and before the next non-digit.

How would we phrase this?  Maybe this:

> Natural sort order is an ordering of strings in the lexicographical order, except that multi-digit numbers are ordered as a single character (collating immediately after the single-digit numbers).  For example, "abc", "abc5", "abc17", "abc59" and "abcd" are in natural order.

While we're at it for lexicographical order (defined before Natural so we can refer to it):

> Lexicographical sort order is defined to be a character based dictionary sort with the character order as defined by the POSIX C locale.  For example "abc", "abc17", "abc5", "abc59" and "abcd" are in lexicographical order.

"Dictionary" is redundant as it's where Lexicographical comes from, but it's there for clarity as it is a term more people understand.

While we're at it, what was the conclusion regarding name spaces?  I don't think we came to a sensible conclusion.  We could take the pragmatic approach of the above are defined keywords and if used must have this meaning.  If you have additional sub-sorts that you wish to reserve for specific meaning then please raise an issue at GitHub?

yfarjoun:
I'm glad you brought up 05...and we should resolve it. I suggest that in case of ambiguity, lexicographic takes over, so that 05 would precede 5.

Also, make sure you specify multi-digit **natural** numbers, unless you want to deal with negatives, and floating-point numbers. 

I agree that we should reserve at-least "natural" and "lexicographic" as subsort. 

I would also like to reserve "duplicate" to imply orders that put reads from the same duplicate-set together, and "umi" to imply sorting (lexicographically) by the UMI tag RX.

jkbonfield:
Agreed.  It's the easy solution and, HURRAH, what samtools does which is what we're trying to document (the two likely scenarios) afterall.  For reference, samtools thinks this is the name sort order:

```abc
abc-
abc.5
abc0
abc001
abc01
abc1
abc1-
abc1.2
abc1.5
abc1x
abc2
abc5
abc10
abc17
abc59
abcd
```

1.5 is before 1x and 17 simply because "." is before x and 7 in the collation order.  An update to the natural order definition:

> Natural sort order is an ordering of strings in the lexicographical order, except that multi-digit natural numbers (excluding leading zeros) are ordered as a single character (collating immediately adjacent to the single-digit numbers). For example, "abc", "abc05", "abc5", "abc17", "abc59" and "abcd" are in natural order.

The samtools code is very similar in operation to this definition.  It looks for digits.  When it finds it it skips zeros, then it looks for all subsequent digits to make a number and sorts based on which of those numbers are larger.

Note we haven't mentioned stable sort anywhere, but it'd be opening up an entirely new can of worms and that is an "encoder" property rather than "decoder" one so we shouldn't pin down implementations there.  Ie if we swap two records named "foo" around then the name sort order is still correct.

Do you wish to have a stab at defining duplicate and umi text?

yfarjoun:
I'd suggest putting in negative numbers and floating points into the example to show that they do not collate with the naturals.

Regarding stability, I think that it's worth mentioning that encoder order will be different form validation order due to the need to break ties between records with the same name.


> Do you wish to have a stab at defining duplicate and umi text?

Sure. Here's a straw-man to shoot down :-)

>duplicate - an unspecified order that ensures that reads that would be considered part of a "duplicate set" (i.e. that at most one of them will _not_ have bit 0x400 set if duplicate-marked) are collated together. This will depend on the algorithm by which duplication-marking is performed, and for example, whether UMI and Library are taken into consideration.

>umi - Sorted lexicographically by the UMI. The "MI" tag should be used for comparing UMIs.  In its absence, the RX tag, which is not guaranteed to be unique across multiple libraries, may be used.

jkbonfield:
I updated the PR with a new section, but I haven't added anything about private vs public name spaces.  (We need to decide whether we care about this or not.)

SAMv1.tex
```diff
@@ -307,6 +309,25 @@ \subsection{The header section}
...
+\subsubsection{Defined sub-sort terms}\label{sec:sub-sort}
+
+While sub-sort permits implementation dependent sorting, some terms are defined to have specific meanings.
+
+\begin{description}
+\item[lexicographical] sort order is defined as a character based dictionary sort with the character order as defined by the POSIX C locale.
+  For example ``abc'', ``abc17'', ``abc5'', ``abc59'' and ``abcd'' are in lexicographical order.
+
+\item[natural] sort order is an ordering of strings in the lexicographical order, except that multi-digit natural numbers (positive integers excluding leading zeros) are ordered as a single character (collating immediately after the single-digit numbers).
+  For example, ``abc-5'', ``abc'', ``abc05'', ``abc5'', ``abc17'', ``abc59'' and ``abcd'' are in natural order.
```
yfarjoun: add "abb" and "abd" and "abc+5" "abc-6" "abc/" and "abc:" as well.

yfarjoun: Need to specify that leading zeros are Tiebroken by lexicographic order so that "abc05" "abc5" "abc05" is not properly sorted. (unlike SamTools now.)

jmarshall: Decimal fractions won't be very natural! Consider 6.1, 6.05, 6.12 which would be in "natural" order!

SAMv1.tex
```diff
@@ -216,20 +216,25 @@ \subsection{The header section}
...
+    For example, if an algorithm relies on a {\tt coordinate} sort that, at each coordinate, is further sorted by query name the header could contain {\tt @HD SO:coordinate SS:coordinate:queryname}.%
 \footnote{The repetition of \emph{sort-order} enables a limited form of validation.
 For example, {\tt @HD SO:queryname SS:coordinate:TLEN} would indicate that the data has been re-sorted (by query name) by a non-{\tt SS}-aware tool and the {\tt SS} field should be ignored.}
-    For example, if an algorithm relies on a {\tt coordinate} sort that, at each coordinate, is further sorted by query name the header could contain {\tt @HD SO:coordinate SS:coordinate:queryname}.
     If the primary sort is not one of the predefined primary sort orders, then {\tt unsorted} should be used and the subsort is effectively the major sort.
     For example, if sorted by an auxillary tag {\tt MI} then by coordinate then the header could contain {\tt @HD SO:unsorted SS:unsorted:MI:coordinate}.
     \newline
     \emph{Regular expression}: {\tt (coordinate|queryname|unsorted)(:[\verb"A-Za-z0-9_-"]+)+}\\\cline{1-3}
+%xyzzy start
```
yfarjoun: delete this please.

SAMv1.tex
jkbonfield: I'm wary of creating an example list which is so vast it goes on for ages!  However agreed we can add more and fractions are definitely worth putting in the examples.

For what it's worth, the text DOES already cover minuses, fractions and leading zeros by stating "_multi-digit natural numbers (positive integers excluding leading zeros)_".  If people read the spec then it's clear -050.5 is - 0 50 . 5 with two natural numbers in there for purpose of sorting (50 and 5).

SAMv1.tex
jkbonfield: Oops, yeah sorry that was me getting fed up finding it and putting temporary string in so I could search.  Will do. :-)

SAMv1.tex
jkbonfield: Ok maybe not that clear - samtools actually works differently to how I expected.  It does use leading zeros in the calculation, with ties for the same equality of value being those with the most leading zeroes come first.

Thus this is ordered:  abc03 abc5 abc008 abc08 abc8 abc17

jkbonfield:
More exhaustive analysis of samtools sort and testing to ensure the natural sort order description matches the implementation.

SAMv1.tex
```diff
@@ -304,6 +307,25 @@ \subsection{The header section}
...
 
+\subsubsection{Defined sub-sort terms}\label{sec:sub-sort}
+
+While sub-sort permits implementation dependent sorting, some terms are defined to have specific meanings.
+
+\begin{description}
+\item[lexicographical] sort order is defined as a character based dictionary sort with the character order as defined by the POSIX C locale.
+  For example ``abc'', ``abc17'', ``abc5'', ``abc59'' and ``abcd'' are in lexicographical order.
+
+\item[natural] sort order is an ordering of strings in the lexicographical order, except that multi-digit natural numbers (non-negative integers including leading zeros) are ordered as a single character with value equal to the numerical value, slotting into the collation order in the same location as single-digit numbers.  Numeric ties caused through leading zeros are broken with more-zeroes coming before fewer-zeros.  Apparent negative and fractional values are not treated as part of the number, with `\texttt{-}' and `\texttt{.}' being interpreted as all other single characters for collation purposes.
```
jmarshall: Thanks for adding the stuff about `-` and `.` just being ordinary characters. Alas it's still pretty murky. Here's another attempt that might provide some more inspiration…

\item[natural] sort order is similar to lexicographical order except that runs of adjacent digits are considered to be numbers embedded within the text string, ordered numerically when compared to each other and ordered as single digits when compared to the surrounding non-digit characters. Runs that differ only in the number of leading zeros (thus are numerically tied) are ordered by more-zeros coming before fewer-zeros. The characters `-` and `.` are considered as ordinary characters, so apparently negative or fractional values are not treated as part of an embedded number.

SAMv1.tex
```diff
@@ -304,6 +307,25 @@ \subsection{The header section}
...
+  For example, ``abc'', ``abc+5'', ``abc-5'', ``abc.d'', ``abc03'', ``abc5'', ``abc008'', ``abc08'', ``abc8'', ``abc17'', ``abc17.+'', ``abc17.2'', ``abc17.d'', ``abc59'' and ``abcd'' are in natural order.
+
+\item[duplicate] is an unspecified order that ensures that reads that would be considered part of a ``duplicate set'' (i.e. that at most one of them will \emph{not} have bit 0x400 set if duplicate-marked) are collated together.
+  This will depend on the algorithm by which duplication-marking is performed, and for example, whether UMI and Library are taken into consideration.
```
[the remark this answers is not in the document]
yfarjoun: you are correct...but technically, so is queryname sort-order...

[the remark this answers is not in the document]
jmarshall: Not really. `GO:query` implies much less than `SO:queryname`.

For `SO:queryname` there's an obvious lexicographic sort ordering based on QNAME that is being selected, even if implementations disagree on details.

This description of `duplicate` explains the semantic intention for determining equivalence of reads, but gives no hints about what ordering between the collated equivalence groups might look like.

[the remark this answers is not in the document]
yfarjoun: You are right. I'll suggest a tightening of the duplicate order.

yfarjoun:
```
\item[duplicate] ensures that reads that would be considered part of a ``duplicate set'' (i.e. that at most one of them will \emph{not} have bit 0x400 set if duplicate-marked) are collated together. 
The relative position of duplicate "sets" is determined by the unclipped 5' end of the non-duplicate record in that set.

File: SAMv1.tex [hunk 5]
```diff
@@ -216,16 +216,19 @@ \subsection{The header section}
   key is the {\sf RNAME} field, with order defined by the order of {\tt @SQ} lines in the header.  The
   minor sort key is the {\sf POS} field.  For alignments with equal {\sf RNAME} and {\sf POS}, order is
   arbitrary.  All alignments with `{\tt *}' in {\sf RNAME} field follow alignments with some other
-  value but otherwise are in arbitrary order.\\\cline{2-3}
+  value but otherwise are in arbitrary order.
+  For queryname sort, no explicit requirement is made regarding the ordering other than that it be applied consistently throughout the entire file.%
+  \footnote{It is known that widely used software have differing definitions of the queryname sort order, meaning care should be taken when operating on multiple files of varying provenance.
```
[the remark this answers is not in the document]
jkbonfield: "libraries" I accept, but widely-used?  Seems wrong to me.

https://en.wikipedia.org/wiki/Glossary_of_British_terms_not_widely_used_in_the_United_States

SAMv1.tex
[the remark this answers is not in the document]
jkbonfield: Agreed it's more GO and SO, but we have sub-sort tag not sub-group so it's natural it goes there unless we wish to expand even more on our tags.

I think it just needs clarifying this is a partial sort, so more of a group, but placed in sub-sort for simplicity.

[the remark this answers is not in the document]
jmarshall: What value of _sort-order_ would this partner with? i.e., what is `XX` in `SO:XX SS:XX:duplicate`?

I'm not suggesting inventing a sub-group field. I'm saying this (as described) should just be `GO:duplicate`.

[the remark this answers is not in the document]
jkbonfield: I propose defining SO as _sort-order:sub-sort_ or _group-order:sub-group_.  The two should be unambiguous, given we don't have both defined at the same time (can we?) and the keywords don't clash either.

SAMv1.tex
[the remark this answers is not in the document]
jmarshall: This is a case of an attributive compound adjective as described in your second link. However as I don't care I'll defer to that page's unsourced claim that there is an exception to this hyphen when the compound is joined around `-ly`. Interesting; and whatevs.

SAMv1.tex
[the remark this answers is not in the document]
jmarshall: Okay, what value of _group-order_ would this (i.e., `duplicate`) partner with?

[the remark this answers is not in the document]
jkbonfield: PS.  Argh!  Why did none of us spot GO:query vs SO:queryname at the time?  Bah humbug...  (I include myself in the list of failings for that one.)

SAMv1.tex
jkbonfield: Happy to go with that.

SAMv1.tex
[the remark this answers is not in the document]
jmarshall: > Why did none of us spot GO:query vs SO:queryname at the time

Because that was back in the 1000 genomes Pages word processor document days, and most of the current “us” were not involved at the time, nor had any input or visibility into what G1K were doing.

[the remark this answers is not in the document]
jkbonfield: Actually I take back the stuff above.  I misunderstood the argument I think, as maybe have you John.

The duplicate sub-sort is still *sorted* on the primary key, so it has nothing to do with group order.  It is strictly a sub-field of sort, not group.  Why does GO enter in this at all?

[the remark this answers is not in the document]
jmarshall: I considered that possibility, but that's not what the proposed text says. Or, at least…

It might be useful to have an ordering that is e.g. “coordinate sorted, except that records that are duplicates are grouped together”. However such a file is no longer sorted by coordinate!

So if we wanted this to mean that, then we have to add a bunch of text explaining that we don't mean that files have to be **entirely** sorted. I don't think we want to go down that road…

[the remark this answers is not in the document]
jkbonfield: I think we need @yfarjoun to explain the intended purpose.  I'm now confused.

We obviously interpreted "unspecified order" differently.  @yfarjoun, does that mean unsorted primary order, or that duplicate sub-sort doesn't have any requirements as to what the primary sort is?  The way I read it was more that in eg SS:queryname:duplicate we're explicitly stating only a maximum of one field will be non-duplicate.

Can you give an explanation of what problem you're solving with this that isn't, for example, solved by GO:query or SO:queryname already?

jkbonfield:
> ```
> \item[duplicate] ensures that reads that would be considered part of a ``duplicate set'' (i.e. that at most one of them will \emph{not} have bit 0x400 set if duplicate-marked) are collated together. 
> The relative position of duplicate "sets" is determined by the unclipped 5' end of the non-duplicate record in that set.
> ```

Does "relative position" mean that the primary field is a coordinate sort, but coordinate sort only on one read per set?  So the data is actually sorted (not simply grouped), but sorted per cluster of reads instead of sorted for each and every read?  We need an explicit example of what the primary sort order would be for this sub-sort.

yfarjoun:
let's just drop the duplicate sort....

jkbonfield:
Updated with @jmarshall's comments and removed the duplicate definition.

jkbonfield:
Note: it predated the .circleci config, so I rebased on master and force pushed.  Sorry for the spammage.

jmarshall:
Dude, you pushed to **origin** instead of your own fork again! 🤣

jkbonfield:
Ah crap - I'm having a bad day!  Sorry I didn't know I did that the first time either.

SAMv1.tex
```diff
@@ -247,16 +247,19 @@ \subsection{The header section}
...
     are grouped together but the file is not necessarily sorted overall.
     \emph{Valid values}: {\tt none} (default), {\tt query} (alignments are
     grouped by {\sf QNAME}), and {\tt reference} (alignments are grouped by
     {\sf RNAME}/{\sf POS}).\\\cline{1-3}
-  & {\tt SS} & Sub-sorting order of alignments. Valid values are of the form \emph{sort-order}{\tt :}\emph{sub-sort}, where \emph{sort-order} is the same value stored in the {\tt SO} tag and \emph{sub-sort} is an implementation-dependent colon-separated string further describing the sort order.%
+  & {\tt SS} & Sub-sorting order of alignments. Valid values are of the form \emph{sort-order}{\tt :}\emph{sub-sort}, where \emph{sort-order} is the same value stored in the {\tt SO} tag and \emph{sub-sort} is an implementation-dependent colon-separated string further describing the sort order, but with some controlled terms defined in Section~\ref{sec:sub-sort}.
+    For example, if an algorithm relies on a {\tt coordinate} sort that, at each coordinate, is further sorted by query name the header could contain {\tt @HD SO:coordinate SS:coordinate:queryname}.%
```
nh13: either "query name _then_ the header could" or "query name, the header could"

SAMv1.tex
```diff
@@ -247,16 +247,19 @@ \subsection{The header section}
...
+  Tools may wish to use the sub-sort field to explicitly distinguish between natural and lexicographical ordering.  See Section~\ref{sec:sub-sort}.}\\\cline{2-3}
   & {\tt GO} & Grouping of alignments, indicating that similar alignment records
     are grouped together but the file is not necessarily sorted overall.
     \emph{Valid values}: {\tt none} (default), {\tt query} (alignments are
     grouped by {\sf QNAME}), and {\tt reference} (alignments are grouped by
     {\sf RNAME}/{\sf POS}).\\\cline{1-3}
-  & {\tt SS} & Sub-sorting order of alignments. Valid values are of the form \emph{sort-order}{\tt :}\emph{sub-sort}, where \emph{sort-order} is the same value stored in the {\tt SO} tag and \emph{sub-sort} is an implementation-dependent colon-separated string further describing the sort order.%
+  & {\tt SS} & Sub-sorting order of alignments. Valid values are of the form \emph{sort-order}{\tt :}\emph{sub-sort}, where \emph{sort-order} is the same value stored in the {\tt SO} tag and \emph{sub-sort} is an implementation-dependent colon-separated string further describing the sort order, but with some controlled terms defined in Section~\ref{sec:sub-sort}.
```
nh13: here we use "some controlled terms" but in the sub-sort section it is "implementation-defined keywords" and "predefined terms".  How about using the latter?

SAMv1.tex
```diff
@@ -337,6 +340,24 @@ \subsection{The header section}
...
+  For example ``abc'', ``abc17'', ``abc5'', ``abc59'' and ``abcd'' are in lexicographical order.
+
+\item[natural] sort order is similar to lexicographical order except that runs of adjacent digits are considered to be numbers embedded within the text string, ordered numerically when compared to each other and ordered as single digits when compared to the surrounding non-digit characters.
+Runs that differ only in the number of leading zeros (thus are numerically tied) are ordered by more-zeros coming before fewer-zeros.
+The characters `\texttt{-}' and `\texttt{.}' are considered as ordinary characters, so apparently negative or fractional values are not treated as part of an embedded number.
+  For example, ``abc'', ``abc+5'', ``abc-5'', ``abc.d'', ``abc03'', ``abc5'', ``abc008'', ``abc08'', ``abc8'', ``abc17'', ``abc17.+'', ``abc17.2'', ``abc17.d'', ``abc59'' and ``abcd'' are in natural order.
+
+\item[umi] is a lexicographical sort by the UMI tag.
+  The {\tt MI} tag should be used for comparing UMIs. In its absence, the RX tag, which is not guaranteed to be unique across multiple libraries, may be used.
```
nh13: Kirk, that's a lot, of, commas.
```suggestion
  The {\tt MI} tag should be used for comparing UMIs. The RX tag may be used in its absence but is not guaranteed to be unique across multiple libraries.
```

SAMv1.tex
```diff
@@ -337,6 +340,24 @@ \subsection{The header section}
...
+\begin{description}
+\item[lexicographical] sort order is defined as a character-based dictionary sort with the character order as defined by the POSIX C locale.
+  For example ``abc'', ``abc17'', ``abc5'', ``abc59'' and ``abcd'' are in lexicographical order.
+
+\item[natural] sort order is similar to lexicographical order except that runs of adjacent digits are considered to be numbers embedded within the text string, ordered numerically when compared to each other and ordered as single digits when compared to the surrounding non-digit characters.
+Runs that differ only in the number of leading zeros (thus are numerically tied) are ordered by more-zeros coming before fewer-zeros.
+The characters `\texttt{-}' and `\texttt{.}' are considered as ordinary characters, so apparently negative or fractional values are not treated as part of an embedded number.
+  For example, ``abc'', ``abc+5'', ``abc-5'', ``abc.d'', ``abc03'', ``abc5'', ``abc008'', ``abc08'', ``abc8'', ``abc17'', ``abc17.+'', ``abc17.2'', ``abc17.d'', ``abc59'' and ``abcd'' are in natural order.
+
+\item[umi] is a lexicographical sort by the UMI tag.
```
nh13: How late is it to argue that this should be called `molecule`, so we don't have acronyms as keywords?  A UMI is a unique _molecular_ identifier.

SAMv1.tex
jkbonfield: The above text also uses implementation-dependent, so I'll change the controlled terms to predefined terms and the implementation-defined bit in the later section to implementation-dependent for consistency.  Thanks.

SAMv1.tex
jkbonfield: This was a suggested addition by @yfarjoun I believe.  It's not something I've seen or used before.  Yossi care to comment?

I'm happy with either umi or molecule for what it's worth, although would note perhaps umi is more accurate as we'll typically be sorted by the *identifier* which may contain non sequence bases.

jkbonfield:
Thanks @nh13.  Pushed another update.

SAMv1.tex
nh13: Makes sense; for my comment, take it or leave it.

jmarshall:
:+1: please merge it, with one minor editorial comment:

> While SS sub-sort field is implementation-dependent, some terms are predefined with specific meanings.

Needs **the**: _While the SS sub-sort field…_

Personally I preferred the phrasing from a commit or two ago because it emphasised that there are implementation-dependent terms within a spec-defined colon syntax:

> While the SS sub-sort field allows implementation-defined terms, some are predefined with specific meanings.

But whichever you prefer is fine, just add the “the” :smile:

jkbonfield:
Thanks.  Rebased and squashed ready to merge.  Will do tomorrow.

PR #363, opened 2018-11-30
Update string encoding example with Mark's current location

ThomasColthurst:


VCFv4.3.tex
```diff
@@ -1654,18 +1654,18 @@ \subsubsection{Type encoding}
 \end{tabular}
 \vspace{0.3cm}
 
-Suppose you want to encode the string MarkDePristoWorksAtTheBroad, a string of size 27.
+Suppose you want to encode the string MarkDePristoWorksAtGoogle, a string of size 25.
```
yfarjoun: This was an annoying idea when first put in, but given his status was strange to comment on....at this point, I think that it should either be left alone, or, if changed, find something that will always be true. possibly NOT pointing to a particular person or company....

yfarjoun:
👎

PR #366, opened 2018-12-13
Acknowledge the TLEN dichotomy.

jkbonfield:
Quick first draft.  WIP as I need to head off and wanted something in before the file formats meeting. :-)

Replaces #23 and #27 

The observation is that these issues have been around pretty much forever.  They are still hanging around because this is a **hard problem**.  The basic problem is that the specification changed from 5' to 5' to leftmost to rightmost, but most software implementations did not.  New software however adhered to the new spec, meaning we ended up with both definitions in use.  This PR is simply an acceptance that the horse has bolted and to embrace the diversity instead of pretending we can fix it.

Right now it's definitely the case that the dominant use is still 5' to 5'.  We could bless this as the preferred solution, but I haven't yet done this.  Right now I just list the two known definitions, but don't even restrict it to these.  I haven't explained the historical issue of out by one errors either.  I'm not expecting the spec footnote to be a complete list of our cockups through the years.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -9,6 +9,7 @@
 \usepackage[pdfborder={0 0 0},hyperfootnotes=false]{hyperref}
 \usepackage[title]{appendix}
 \usepackage[calc]{picture}
+\usepackage{tikz}
 
 \newlength{\bytewidth}
 \newlength{\bytetotalheight}
@@ -560,12 +561,68 @@ Each bit is explained in the following table:
   the next read. If {\sf PNEXT} is 0, no assumptions can be made on
   {\sf RNEXT} and bit 0x20.
 \item {\sf TLEN}: signed observed Template LENgth. If all segments are
-  mapped to the same reference, the unsigned observed template length
-  equals the number of bases from the leftmost mapped base to the
-  rightmost mapped base. The leftmost segment has a plus sign and the
-  rightmost has a minus sign. The sign of segments in the middle is
-  undefined. It is set as 0 for single-segment template or when the
-  information is unavailable.
+  mapped to the same reference sequence, the absolute value of TLEN
+  equals the distance between the mapped end of the template and the
+  mapped start of the template, inclusively (i.e., $\mbox{end}-\mbox{start}+1$).%
+  \footnote{Thus a segment
+    aligning in the forward direction at base 100 for length 50 and a
+    segment aligning in the reverse direction at base 200 for length
+    50 indicate the template covers bases 100 to 249 and has length
+    150.}
+  Note that \textit{mapped base} is defined to be one that aligns to the
+  reference as described by CIGAR, hence excludes soft-clipped bases.
+  The TLEN field is positive for the leftmost segment of the template,
+  negative for the rightmost, and the sign for any middle segment is
+  undefined.  If segments cover the same coordinates then the choice
+  of which is leftmost and rightmost is arbitrary, but the two ends
+  must still have differing signs.
+  It is set as 0 for a single-segment template or when the
+  information is unavailable (e.g., when the first or last segment
+  of a multi-segment template is unmapped or when the two are mapped
+  to different reference sequences).
+
+% If future updates cause this footnote to spill over the page boundary
+% try e.g. \enlargethispage{-10em} to adjust where this page ends.
+  The intention of this field is to indicate where the other end of
+  the template has been aligned without needing to read the remainder
+  of the SAM file.  Unfortunately there has been no
+  clear consensus on the definitions of the template mapped start
+  and end.  Thus the exact definitions are implementation-defined.%
+\footnote{The earliest versions of this specification used $5'$ to
+  $5'$ (in original orientation, TLEN\#1; dashed parts of the reads
+  indicate soft-clipped bases) while later ones used leftmost to
+  rightmost mapped based (TLEN\#2).  Note: these two definitions agree in most
+  alignments, but differ in the case of overlaps where the first
+  segment aligns beyond the start of the last segment.
+
+\begin{center}
+\begin{tikzpicture}[
+    thick, xscale=2, yscale=1.8,
+    read/.style={line width=0.4mm},
+]
+\draw[|-, read] (0,0)   -- (0.8,0)  node[midway, above] {Read 1};
+\draw[->, read, dashed] (0.85,0) -- (1.2,0);
+\draw[|-, read] (3,0) -- (2.2,0) node[midway, above] {Read 2};
+\draw[->, read, dashed] (2.25,0) -- (1.8,0);
+\draw[|<->|,gray!120]     (0,-.2) -- +(3,0) node[midway, below] {TLEN \#1/\#2};
+\node at (1.5,-0.7) {\textit{Unambiguous scenario}};
+\end{tikzpicture}
+\hspace{3cm}% No space means side by side
+\begin{tikzpicture}[
+    thick, xscale=1,yscale=0.9,
+    read/.style={line width=0.4mm},
+]
+\draw[|-, read]         (2.2,0)   -- (5.2,0)    node[very near end, above] {Read 1};
+\draw[->, read, dashed] (5.3,0)   -- (6,0);
+\draw[-|, read]         (1,.3)    -- (4,.3)     node[very near start , above] {Read 2};
+\draw[->, read, dashed] (0.9,.3)  -- (0,.3);
+\draw[|<->|,gray!120]   (1,-0.3)  -- (5.2,-0.3) node[midway, below] {TLEN\#2};
+\draw[|<->|,gray!120]   (2.2,0.65)-- (4,0.65)    node[midway, above] {TLEN\#1};
+\node at (2.7,-1) {\textit{Ambiguous scenario}};
+\end{tikzpicture}
+\end{center}
+}
+
 \item {\sf SEQ}: segment SEQuence. This field can be a `*' when the
   sequence is not stored. If not a `*', the length of the sequence must
   equal the sum of lengths of {\tt M/I/S/=/X} operations in {\sf CIGAR}.


jkbonfield:
Bump

SAMv1.tex
```diff
@@ -500,11 +500,33 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
-  equals the number of bases from the leftmost mapped base to the
-  rightmost mapped base. The leftmost segment has a plus sign and the
-  rightmost has a minus sign. The sign of segments in the middle is
-  undefined. It is set as 0 for single-segment template or when the
-  information is unavailable.
+  equals the distance between the mapped end of the template and the
+  mapped start of the template, inclusively.\footnote{Thus a segment
+    aligning in the forward direction at base 100 for length 50 and a
+    segment aligning in the reverse direction at base 200 for length
+    50 indicate the template covers bases 100 to 250 and has length
```
yfarjoun:
```suggestion
    50 indicate the template covers bases 100 to 249 and has length
```

SAMv1.tex
```diff
@@ -500,11 +500,33 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
+    151.}
```
yfarjoun:
```suggestion
    150.}
```

SAMv1.tex
```diff
@@ -500,11 +500,33 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
+  Note that mapped base is defined to be one that aligns to the
+  reference as described by CIGAR, hence exclude soft-clipped bases.
+  The TLEN field is positive for leftmost segment of the template,
+  negative for the rightmost and the sign for any middle segment is
```
yfarjoun: what about two reads with the same exact alignment? which is negative and which positive?

SAMv1.tex
jkbonfield: Facepalm!  Thanks :)

SAMv1.tex
jkbonfield: I thought the spec already stated this case gives an arbitrary choice, but I can't see it.  Maybe I'm remembering something else.

We should add something that states leftmost and rightmost can be arbitrarily chosen when they are both equal.

jkbonfield:
I'll have another full read through, but it looks like you've addressed my comments and it looks good.  Thanks.

SAMv1.tex
```diff
@@ -500,11 +500,36 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
   {\sf RNEXT} and bit 0x20.
 \item {\sf TLEN}: signed observed Template LENgth. If all segments are
   mapped to the same reference, the unsigned observed template length
```
jmarshall: “…to the same reference sequence“ would be clearer.

SAMv1.tex
```diff
@@ -500,11 +500,36 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
+  of the SAM file.  Unfortunately there has been no
+  clear consensus on the definition the template mapped start
+  and end.  Thus the exact definitions are implementation defined.%
+\footnote{The earliest versions of this specification used $5'$ to
+  $5'$ (in original orientation) while later ones used leftmost to
+  rightmost.  Note: these two definitions agree in many alignments for
+  many paired library strategies, but differ in the case of overlaps
+  where the first segment aligns beyond the start of the last segment.
+  When implementing a new alignment algorithm it is recommended to
+  choose one of the above two definitions.}
```
jmarshall: Maybe we can be more definitive and clearer than “choose one of the above” :smile:

jkbonfield: Maybe, but the whole point of the PR was to admit defeat that we've had years to make a choice and dismally failed to do so.  If we start to have the discussion on which to recommend I suspect we'll be here for another few years!  Unless everyone has since made their mind up in the same direction. :-)

jkbonfield:
Fixed minor language suggestions, rebased and squashed.

There is still not recommendation on which method to use, but that's sort of the whole point of this PR.  We've *never* been able to recommend, or we'd have closed this issues years ago.

SAMv1.tex
yfarjoun: Similarly unclear is which is leftmost and which is rightmost...are we talking about the 5' ends? or the alignment POS?

jkbonfield: "leftmost mapped base".  How can that be unclear except in the case of ties, which I think we covered.

Fundamentally, we can't fix this.  It's been hanging around like a bad smell for *years*.  If it was easy to nail down to a single definition it'd have been done by now.  This PR is simply a pragmatic acceptance that we failed and there are (at least) two definitions out there already.  I'm open to better wording, but not open to ways of nailing it down to one meaning as that ship sadly sailed ages ago.

yfarjoun:
sorry, James. I'm not trying to be difficult. Nor am I trying to make up a new definition...

What I mean is that the statement

>The TLEN field is positive for leftmost segment of the template,
  negative for the rightmost

is not well defined for short inserts:

                    |---------------------------------------------->
     <----------------------------------------------|

Since the POS for these two reads will be the same (due to softclipping) but we probably think of the top one as "left most segment" and the bottom one as "rightmost"..though, I'm not sure!

jkbonfield:
Hmm, it's mostly covered by

> If segments cover the same coordinates then the choice of which is leftmost and rightmost is arbitrary, but the two ends must still have differing sign.

Although that's talking about coordinates (plural) meaning both ends.  If you soft-clip both ends in your example and get both left and right mapped coordinates being the same, then the statement above covers it.

Either way, "arbitrary" is about as good as we're going to get I fear.

jkbonfield:
I'm open to better wording if you can suggest some.

jmarshall:
Unfortunately this is still an incomprehensible mess. If we're going to try to improve the spec w.r.t. TLEN, I think we should try to make it a comprehensible mess.

IMHO this would involve something closer to the original “9. TLEN” short paragraph and a new §1.4.1 discussing TLEN calculations _with pictures_ (or a _Recommended practice_ subsection, but I'm not sure that fits) and noting when the two definitions are the same and when they are different. And the history appendix should note when the spec changed from 5'-to-5' to leftmost-to-rightmost — namely the 2010 transition from the Pages document to LaTeX, in particular e6ca319516752ecdbc25b5880c41b1a4970dd0ec.

It's all still a mess. I've been doing some testing and I see that for the very-overlapping situation (https://github.com/samtools/hts-specs/pull/366#issuecomment-471516869) both bwa mem (lh3/bwa#137) and minimap2 (no issue yet filed) observably output TLEN as 5'-to-5'. Sigh…

It would be good to check what current Picard alters this to (if anything) and to check some other authors' aligners.

jkbonfield:
I just reread it and don't find it hard to understand.  It's long, but that's necessarily so.

Maybe we should just cull the footnote explaining alternative implementations and leave it as "implementation defined".

Certainly I don't really see the point in analysing what aligners do and what processing software expects.  The whole point of this PR was specifically so we didn't have to do that and we'd simply admit to ourselves that different people have a different interpretation.  What benefit is there to gathering a corpus of who does what?

yfarjoun:
here's a proposed illustration, happy to modify it...
[tlen_options.pdf](https://github.com/samtools/hts-specs/files/3364334/tlen_options.pdf)

jkbonfield:
Thanks @yfarjoun.  I'll look at doing it as a LaTeX drawing, but it'll have to be a week or two due to other committements at the moment.

jkbonfield:
Updated with a tikz pic.  It only uses the most basic absolute positioning and I'm sure I can be clever with nodes, but it's sufficient.

SAMv1.tex
```diff
@@ -539,12 +540,66 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
   the next read. If {\sf PNEXT} is 0, no assumptions can be made on
   {\sf RNEXT} and bit 0x20.
 \item {\sf TLEN}: signed observed Template LENgth. If all segments are
-  mapped to the same reference, the unsigned observed template length
-  equals the number of bases from the leftmost mapped base to the
-  rightmost mapped base. The leftmost segment has a plus sign and the
-  rightmost has a minus sign. The sign of segments in the middle is
-  undefined. It is set as 0 for single-segment template or when the
-  information is unavailable.
+  mapped to the same reference sequence, the unsigned observed template length
```
yfarjoun: I would prefer  " the absolute value of TLEN equals...." otherwise is sounds like there's a "signed observed template length" and also a "signed observed template length"...

SAMv1.tex
```diff
@@ -539,12 +540,66 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
+  reference as described by CIGAR, hence exclude soft-clipped bases.
+  The TLEN field is positive for leftmost segment of the template,
+  negative for the rightmost and the sign for any middle segment is
+  undefined.  If segments cover the same coordinates then the choice
+  of which is leftmost and rightmost is arbitrary, but the two ends
+  must still have differing sign.
+  It is set as 0 for single-segment template or when the
+  information is unavailable (e.g. for a single-segment template or
+  when the first or last segment of a multi-segment template is
+  unmapped or are mapped to different references).
```
yfarjoun: reference _sequences_

SAMv1.tex
```diff
@@ -539,12 +540,66 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
 \item {\sf TLEN}: signed observed Template LENgth. If all segments are
-  mapped to the same reference, the unsigned observed template length
-  equals the number of bases from the leftmost mapped base to the
-  rightmost mapped base. The leftmost segment has a plus sign and the
-  rightmost has a minus sign. The sign of segments in the middle is
-  undefined. It is set as 0 for single-segment template or when the
-  information is unavailable.
+  mapped to the same reference sequence, the unsigned observed template length
+  equals the distance between the mapped end of the template and the
+  mapped start of the template, inclusively.\footnote{Thus a segment
```
yfarjoun: Should we get rid of a possible +1 ambiguity here by stating that a template with left-most base aligned at START and rightmost base aligned at END will have a TLEN of END - START + 1?

SAMv1.tex
```diff
@@ -539,12 +540,66 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
+  It is set as 0 for single-segment template or when the
+  information is unavailable (e.g. for a single-segment template or
+  when the first or last segment of a multi-segment template is
+  unmapped or are mapped to different references).
+
+  The intention of this field is to indicate where the other end of
+  the template has been aligned without needing to read the remainder
+  of the SAM file.  Unfortunately there has been no
+  clear consensus on the definition the template mapped start
+  and end.  Thus the exact definitions are implementation defined.%
```
yfarjoun:
```suggestion
  and end.  Thus the exact definitions are implementation-defined.%
```

yfarjoun:
thanks for converting my drawing to tikZ!

SAMv1.tex
jkbonfield: Is that not what the maths already indicates?  Do you mean explicitly stating it again symbolically in addition to the numbers?  I'll give that a go.

jkbonfield: Note however your phrasing has locked it down to one of the two specifically implementations of TLEN in the wild. ;-)

jkbonfield:
Thanks @yfarjoun.  I've dealt with your comments I hope.

SAMv1.tex
yfarjoun: not really, since I was talking about templates, not reads....

SAMv1.tex
```diff
@@ -539,12 +540,67 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
+  reference as described by CIGAR, hence exclude soft-clipped bases.
+  The TLEN field is positive for leftmost segment of the template,
+  negative for the rightmost and the sign for any middle segment is
+  undefined.  If segments cover the same coordinates then the choice
+  of which is leftmost and rightmost is arbitrary, but the two ends
+  must still have differing sign.
+  It is set as 0 for single-segment template or when the
+  information is unavailable (e.g. for a single-segment template or
+  when the first or last segment of a multi-segment template is
+  unmapped or are mapped to different reference sequences).
```
yfarjoun:
```suggestion
  unmapped or when the two are mapped to a different reference sequences).
```

jkbonfield:
Done.  Needs squashing, but for clarity of review process I haven't done this yet.

yfarjoun:
:+1:

yfarjoun:
@jmarshall what do you think? can we merge this?

jmarshall:
Apologies for the excruciating delay… I've pushed my wording suggestions (formatting and a few IMHO missing words). Also I found the “Ambiguous scenario” diagram confusing as IMHO it wasn't obvious that the dashed parts of the lines weren't part of the reads. Suggested diagram update to just have the direction arrowheads on the reads themselves, no dashed extra bit.

It's still terrible having a diagram in a footnote — and I think when you rebase this you may find the footnote gets split across pages. But it's an improvement and we can lift the information out of being a footnote sometime later when we can all face looking at this again. :smile:

 So :+1: from me for merging.

jkbonfield:
The dashed bits were meant to be part of the reads, but unmapped data - eg softclips.  However it didn't state this in the footnote so that's where the ambiguity comes from.  The point was to demonstrate that the leftmost and rightmost is *mapped* coordinate and you can't simply add/subtract read length from pos.  The more common (and not ambiguous) instance of the right hand plot is where the left/right mapped pos for both overlapping reads are identical because soft-clips have been applied.   Here it is clear the soft clips are further out.

However I'll concede it's probably an over elaborate diagram for pointing out this specific issue and simplifying it works better.  Thanks.

jmarshall:
\<penny drops> I see! Okay good idea, seems like it's useful to illustrate the soft clipping, but it does need to be described in the footnote text.

jmarshall:
So you could revert my diagram changes and add something like e.g. “(in original orientation, TLEN#1; dashed parts of the reads indicate soft-clipped bases)” or even briefer to explain this enough.

Re the unfortunate page break when this is rebased: I experimented with dropping the “Un/ambiguous scenario” legends and squeezing the four lines on the RHS a bit closer together, and this sufficed to get the footnote text and diagram on the same page. YMMV as to whether this is worth doing (perhaps temporarily)…

jkbonfield:
Adding `\enlargethispage{-10em}` before the paragraph containing the footnote ends the page slightly early, meaning the footnote is referenced at the start of the next page and it then doesn't spill over.  It could be an appropriate temporary fudge.

I'll also explore whether the image can just be a bit smaller.  Should be doable I think.

jkbonfield:
I abridged the text somewhat. We really don't need to be explicitly requesting aligners stick to the vagaries of different interpretations.   IMO this is sufficient.  I also shrunk the pic slightly.

Rebased and it's now on the same page.  Added a comment to explain how to hack this in future if other changes trigger the splitting behaviour again.

jmarshall:
Dunno why the bot is saying “Artifact not found” (maybe something about the timing of the  force-pushing? 🤷‍♂) but I built it locally and :+1: from me to that last abridging and everything else.

One small comment for clarity is maybe add “mapped base” to the footnote: “while later ones used leftmost to rightmost mapped base (TLEN#2)”. Otherwise please please squash and merge and celebrate!

SAMv1.tex
```diff
@@ -560,12 +561,68 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
...
+% try e.g. \enlargethispage{-10em} to adjust where this page ends.
+  The intention of this field is to indicate where the other end of
+  the template has been aligned without needing to read the remainder
+  of the SAM file.  Unfortunately there has been no
+  clear consensus on the definitions of the template mapped start
+  and end.  Thus the exact definitions are implementation-defined.%
+\footnote{The earliest versions of this specification used $5'$ to
+  $5'$ (in original orientation, TLEN\#1; dashed parts of the reads
+  indicate soft-clipped bases) while later ones used leftmost to
+  rightmost mapped based (TLEN\#2).  Note: these two definitions agree in most
```
jmarshall:
```suggestion
  rightmost mapped base (TLEN\#2).  Note: these two definitions agree in most
```

jmarshall:
Typo, and don't forget to add `(PR #366)` to the commit message when merging to **master** so we can get to this glorious conversation thread easily :smile:

jkbonfield:
Argh, now I see your message.  Wonder why not before!  (I added PR  366 after reminding myself about it from checking MAINTAINERS.md, not from your comment).

I'll do a quick typo push. Grumble!

jmarshall:
I already pushed a typo fix :smile:

jkbonfield:
So I see.  Thanks :)

PR #367, opened 2018-12-14
Clarification to CRAM tag encodings.

jkbonfield:
This is now explicit that the type codes follow BAM conventions
(cCsSiI) and not SAM (i), along with examples to demonstrate this.
Also rearranged this so the text describing the dictionary is in the
preservation map and the text describing the "encodings" is in the Tag
Encoding Map.

Replaces #324

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -691,11 +691,12 @@ QS & encoding\texttt{<}byte\texttt{>} & quality scores & quality scores\tabularn
 \subsubsection*{Tag encodings}
 \label{subsubsec:tags}
 
-The tag dictionary describes the combinations of tag id / type that occur on each alignment record.
+The tag dictionary (TD) describes the unique combinations of tag id / type that occur on each alignment record.
 For example if we search the id / types present in each record and find only two combinations -- X1:i BC:Z SA:Z: and X1:i: BC:Z -- then we have two dictionary entries in the TD map.
 
 Let $L_{i}=\{T_{i0}, T_{i1}, \ldots, T_{ix}\}$ be a list of all tag ids for a record $R_{i}$, where $i$ is the sequential record index and $T_{ij}$ denotes $j$-th tag id in the record.
 The list of unique $L_{i}$ is stored as the TD value in the preservation map.
+Maintaining the order is not a requirement for encoders (hence ``combinations''), but it is permissible and thus different permutations, each encoded with their own elements in TD, should be supported by the decoder.
 Each $L_{i}$ element in TD is assigned a sequential integer number starting with 0.
 These integer numbers are referred to by the TL data series.
 Using TD, an integer from the TL data series can be mapped back into a list of tag ids.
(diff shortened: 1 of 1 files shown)


CRAMv3.tex
```diff
@@ -691,43 +691,38 @@ \subsubsection*{Data series encodings}
...
-be sorted list of all $x$ tag ids for a record $R_{i}$, where $i$ is the sequential
-record index and $T_{ij}$ denotes $j$-th tag id in the record.
-The list of unique $L_{i}$ is assigned sequential
-integer numbers starting with 0. These integer numbers represent the TL data series. 
-The list of unique $L_{i}$ is stored as the TD value in the preservation map, see section \ref{subsec:compression-header}.
+The tag dictionary describes the combinations of tag id / type that occur on each alignment record.
+For example if we search the id / types present in each record and find only two combinations -- X1:i BC:Z SA:Z: and X1:i: BC:Z -- then we have two dictionary entries in the TD map.
+
+Let $L_{i}=\{T_{i0}, T_{i1}, \ldots, T_{ix}\}$ be a list of all tag ids for a record $R_{i}$, where $i$ is the sequential record index and $T_{ij}$ denotes $j$-th tag id in the record.
+The list of unique $L_{i}$ is stored as the TD value in the preservation map.
```
cmnbroad: Should this be explicit about whether tag order counts (are different permutations of the same set of tags permitted to appear in the dictionary) ?

CRAMv3.tex
```diff
@@ -691,43 +691,38 @@ \subsubsection*{Data series encodings}
...
-The TL (tag list) data series represents combined information about the number 
-of tags in a record and their ids. 
-
-Let $L_{i}=\{T_{i0}, T_{i1}, \ldots, T_{ix-1}\}$
-be sorted list of all $x$ tag ids for a record $R_{i}$, where $i$ is the sequential
-record index and $T_{ij}$ denotes $j$-th tag id in the record.
-The list of unique $L_{i}$ is assigned sequential
-integer numbers starting with 0. These integer numbers represent the TL data series. 
-The list of unique $L_{i}$ is stored as the TD value in the preservation map, see section \ref{subsec:compression-header}.
+The tag dictionary describes the combinations of tag id / type that occur on each alignment record.
```
cmnbroad:
```suggestion
The tag dictionary (TD) describes the unique combinations of tag id / type that occur on each alignment record.
```

CRAMv3.tex
```diff
@@ -691,43 +691,38 @@ \subsubsection*{Data series encodings}
...
 \tabularnewline
 \hline
-TAG NAME 1:TAG TYPE 1 & encoding\texttt{<}byte[ ]\texttt{>} & read tag 1 & tag values 
+TAG ID 1:TAG TYPE 1 & encoding\texttt{<}byte[ ]\texttt{>} & read tag 1 & tag values
 (names and types are available in the data series code)\tabularnewline
 \hline
 ... &  & ... & ...\tabularnewline
 \hline
-TAG NAME N:TAG TYPE N & encoding\texttt{<}byte[ ]\texttt{>} & read tag N & ...\tabularnewline
+TAG ID N:TAG TYPE N & encoding\texttt{<}byte[ ]\texttt{>} & read tag N & ...\tabularnewline
```
cmnbroad: How should floating point tag values to be written ?

jkbonfield: This is already covered in "the same as BAM", but granted we should probably be explicit and lift over the text describing C, c, S, s, I, i, f, B, H and Z (assuming I remembered all).

I think that can be the topic of another PR though - the general issue of describing CRAM in its own right rather than building on top of SAM.

CRAMv3.tex
cmnbroad: Thanks for the background - I like the additional text you proposed.

cmnbroad:
:+1: Thx for the updates - looks good to me.

Issue #52, opened 2014-11-14
sam spec change: ambiguous reverse complement bit for unmapped reads


amcpherson:
In the sam spec http://samtools.github.io/hts-specs/SAMv1.pdf, it says the following about the flag:

	0x10 SEQ being reverse complemented

and further down in the notes:

	Bit 0x4 is the only reliable place to tell whether the read is unmapped. If 0x4 is set, no
	assumptions can be made about RNAME, POS, CIGAR, MAPQ, bits 0x2, 0x10, 0x100
	and 0x800, and the bit 0x20 of the previous read in the template.

Thus for an unmapped read, we dont know if the reverse complement flag is valid. This has important implications for extracting the original read data from a bam file, since we dont know whether unmapped reads are represented as-is in the bam, or have been reverse complemented. It would be fairly simple to add the requirement that 0x10 is always valid, regardless of 0x4.

If I understand the bwa code correctly, the reverse complement flag is always set for a reverse complemented read (though confusingly, that flag takes on the same value of its mate if its mate is mapped, reverse complementing the read unnecessarily). If we updated the spec, bwa would already be within spec. I am not sure about other aligners.

In the communities I work with bam files are ubiquitous for storing sequencing data, and with good reason since they are smaller than compressed fastqs. For people relying on bams as a storage format, and wishing to use unmapped reads to predict, for instance, structural variants, it is important the bams allows for consistent extraction of the read sequences as they came off the sequencer.

PR #371, opened 2019-01-08
Add note that SAM alignments are always represented on the forward strand

### LINKED ISSUE #368: SAM alignments are always represented on the forward strand
The original pre-LaTeX Pages SAM specification had a note after the §1.4 QNAME/FLAG/RNAME/POS/etc table saying

> 7. All mapped reads are represented on the forward genomic strand.  The bases are reverse complemented from the unmapped read sequence and the quality scores and cigar strings are recorded consistently with the bases.  This applies to information in the mate tags (R2, Q2, S2, etc.) and any other tags that are strand sensitive.  The strand bits in the flag simply indicates whether this reverse complement transform was applied from the original read sequence to obtain the bases listed in the SAM file.

and it described the 0x10/0x20 FLAG bits as

> 0x0010  strand of the query (0 for forward; 1 for reverse strand)
> 0x0020  strand of the mate

The second sentence appears to have accidentally omitted “_For reads that have been mapped to the reverse strand,_ the bases are…”.

I'm happy with the modern spec's definitions of the 0x10/0x20 bits as “SEQ being reversed complemented” with the corresponding paragraph talking about “whether SEQ has been reverse complemented and QUAL reversed” (though it probably needs to talk about R2/Q2/S2/etc or their modern equivalents also being recorded consistent with that).

However I don't think we have anywhere a clear statement that **in SAM, mapped reads are represented on the forward genomic strand**?!

jmarshall:
Adds a clear statement that in SAM, mapped reads are represented on the forward genomic strand. Fixes #368 by resurrecting and updating the old text from the original Pages SAM spec.

Spell out the implications of this for the Bit 0x10 text too, i.e., unset=forward, set=reverse.

And minor fixups: strand nomenclature is forward/reverse rather than positive/negative; <strike>formatting and wordsmithing in the first half of the paragraph (i.e., before the table)</strike>.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -397,6 +397,10 @@ of the mandatory fields in the SAM format:
 \footnotetext{Reference sequence names may contain any printable ASCII characters with the exception of certain punctuation characters, and may not start with `{\tt *}' or `{\tt =}'.
 See Section~\ref{sec:charset} for details and an explanation of the {\tt [\cclass{rname}]} notation.}
 
+\noindent
+All mapped segments in alignment lines are represented on the forward genomic strand.
+For segments that have been mapped to the reverse strand, the recorded {\sf SEQ} is reverse complemented from the original unmapped sequence and {\sf CIGAR}, {\sf QUAL}, and strand-sensitive optional fields are reversed and thus recorded consistently with the sequence bases as represented.
+
 \begin{enumerate}
 \item {\sf QNAME}: Query template NAME. Reads/segments having identical {\sf QNAME}
 	are regarded to come from the same template. A {\sf QNAME} `{\tt *}'
@@ -443,7 +447,8 @@ Each bit is explained in the following table:
   \item Bit 0x10 indicates whether {\sf SEQ} has been reverse complemented
     and {\sf QUAL} reversed.
     When bit~0x4 is unset, this corresponds to the strand to which the
-    segment has been mapped.
+    segment has been mapped: bit~0x10 unset indicates the forward strand,
+    while set indicates the reverse strand.
     When 0x4 is set, this indicates whether the unmapped read is stored
     in its original orientation as it came off the sequencing machine.
   \item Bits 0x40 and 0x80 reflect the read ordering within each template
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -217,7 +217,7 @@ for any associated quality values.
 \item[SA:Z:\tagregex{{\tt (}\emph{rname}{\tt ,}\emph{pos}{\tt ,}\emph{strand}{\tt ,}\emph{CIGAR}{\tt ,}\emph{mapQ}{\tt ,}\emph{NM}{\tt ;)}+}]
 Other canonical alignments in a chimeric alignment, formatted as a semicolon-delimited list.
 Each element in the list represents a part of the chimeric alignment. Conventionally, at a supplementary line, the first element points to the primary line.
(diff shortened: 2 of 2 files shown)


SAMv1.tex
```diff
@@ -413,7 +415,8 @@ \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
   \item Bit 0x10 indicates whether {\sf SEQ} has been reverse complemented
     and {\sf QUAL} reversed.
     When bit~0x4 is unset, this corresponds to the strand to which the
-    segment has been mapped.
+    segment has been mapped: bit~0x10 unset indicates the forward strand,
+    while set indicates the reverse strand.
```
jkbonfield: This is not consistent with the paragraph above: "If 0x4 is set, no assumptions can be made about {\sf
      RNAME}, {\sf POS}, {\sf CIGAR}, {\sf MAPQ}, and bits 0x2, 0x100,  and 0x800".

That implies when 0x4 is set, 0x10 still has meaning.  My recollection is that this was previously also listed as an undefined case but it was changed in order to permit SAM to FASTQ conversion to work with placed-but-unmapped reads without making assumptions on undefined fields.  If we have a requirement that 0x4 being set is that the sequence is in the original orientation as it came off the instrument, then it should be stated as such - eg if 0x4 is set then the 0x10 flag must be unset.  Instead it's probably easier just to remove any statements about 0x10 only having meaning when 0x4 is unset, in the knowledge that practically speaking it'll just be 0 anyway for the unmapped data so has no impact.

Apart from this rather old problem, which isn't a fault of this PR itself, I'm happy with it.

jmarshall: Yep, the following sentence is closely related :smile:.

We would love to be able to just say 0x4 set ⇒ 0x10 unset, but all this verbiage is to allow for the BWA bug whereby some reads that end up being output as unmapped have already been revcomped internally, and BWA just outputs them that way around. This was #52.

jkbonfield: Ah yes that was it.  I was trying to remember why the orientation was still important knowledge for the unmapped data.  Far too much of SAM has been specified in order to work around bwa!

jmarshall:
Rebased. @yfarjoun please review — I'd like to merge this by the end of this week, and continued silence will be taken as assent :smile:

SAMv1.tex
```diff
@@ -367,12 +367,10 @@ \subsubsection{Reference MD5 calculation}\label{sec:ref-md5}
...
 \subsection{The alignment section: mandatory fields}\label{sec:alnrecord}
-In the SAM format, each alignment line typically represents the linear
-alignment of a segment. Each line has 11
-mandatory fields. These fields always appear in the same order and must be
-present, but their values can be `0' or `*' (depending on the field) if the
-corresponding information is unavailable. The following table gives an overview
-of the mandatory fields in the SAM format:
+In the SAM format, each alignment line typically represents the linear alignment of a segment.
+Each line has eleven mandatory fields.
+These fields always appear in the same order and must be present, though their values may be `{\tt 0}' or `{\tt *}' (depending on the field type) when the corresponding information is unavailable.
```
yfarjoun: This makes it sounds like the values could also be something else when the information is unavailable....perhaps reword as 
> These fields always appear in the same order and must be present when available.
When unavailable, their values must be `{\tt 0}' or `{\tt *}' (depending on the field type).

jmarshall: That's the original Heng paragraph, basically unchanged. What it actually wants is a larger rewrite! I'll merge the main part of this PR and the OA tag and then come back to it…

Issue #370, opened 2019-01-07
CRAM: clarification of slice and container headers


jmthibault79:
@cmnbroad and I were talking last week about our continuing HTSJDK refactoring work, and we have some questions regarding the Slice Header and Container Header in the CRAM 3 spec, particularly regarding which combinations of read types are valid to combine in a slice or container, and how these impact the header values. 

## Read Types

First, we should clarify how different read types are treated in **8.5 Slice header block**.  Mapped is clear, but unmapped reads can be placed at a location or not.  It can make sense to include placed unmapped reads in slices with mapped reads on the associated reference, but I would think unplaced unmapped reads should be strictly assigned to their own slice.  Also, what is the definition of an "unsorted read" here?  Is it a read from a CRAM file which isn't coordinate-sorted?  It's not clear to me that this implies anything in particular about its mapping status.  Or is "unsorted" used here to mean "unmapped unplaced"?  We should clarify, either way.

## Slice Types

Certain combinations of reads and references make sense as logical slices, but others do not.  We should make this explicit, and document conventions for how to encode these types in the slice headers.  We've come up with the following set, which may require some iteration:

1. All reads are either mapped to or unmapped but placed on the same external reference
1. All reads are either mapped to or unmapped but placed on the same embedded reference
1. All reads are unmapped and unplaced
1. Multi-Reference.  Anything can go here **except embedded-reference reads** because we have no way of specifying multiple embedded references.

The Slice Header fields for each Slice Type are:

Columns = Slice Type
Rows = Header Fields

types -> | Single External Ref | Single Embedded Ref | Unmapped Unplaced | Multiple External Ref
-- | -- | -- | -- | --
Ref Seq ID | [A real Ref ID] | **Q: Do embedded refs have an ID?** | -1 UNMAPPED | -2 MULTI
Align Start | min(read starts) | min(read starts) | -1 UNMAPPED | -1 UNMAPPED
Align Span | max(read ends) -min(read starts) | max(read ends) -min(read starts) | 0 UNMAPPED | 0 UNMAPPED
Embed Ref Block Content ID | -1 NONE | [A real Block ID] | -1 NONE | -1 NONE
Ref MD5 | [MD5 of the subset of the Ref Sequence which covers the span] | [MD5 of the subset of the Ref Sequence which covers the span] | 16 0-bytes | 16 0-bytes

## Container Header vs. Slice Header

**7 Container structure** conflicts with **8.5 Slice header block**.

[...]

PR #372, opened 2019-01-11
Slice Header updates for 370

jmthibault79:
Renamed `MAPPED_SLICE_HEADER` Block Content Type to `SLICE_HEADER`

Updated slice header text and table:
* remove "unsorted"
* clarify unmapped-placed and unmapped-unplaced distinction
* add -2 multi-ref flag
* multi-ref may have unmapped-placed but must only be external-ref

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -521,6 +521,7 @@ byte[4] & CRC32 & CRC32 hash value for all preceding bytes in the block\tabularn
 
 CRAM has the following block content types:
 
+\begin{threeparttable}[t]
 \begin{tabular}{|>{\raggedright}p{143pt}|>{\raggedright}p{45pt}|>{\raggedright}p{116pt}|>{\raggedright}p{114pt}|}
 \hline
 \textbf{Block content type} & \textbf{Block content type id} & \textbf{Name} & \textbf{Contents}\tabularnewline
@@ -529,7 +530,7 @@ FILE\_HEADER & 0 & BAM header block & BAM header\tabularnewline
 \hline
 COMPRESSION\_HEADER & 1 & Compression header block & See specific section\tabularnewline
 \hline
-MAPPED\_SLICE\_HEADER & 2 & Slice header block & See specific section\tabularnewline
+SLICE\_HEADER\tnote{a} & 2 & Slice header block & See specific section\tabularnewline
 \hline
  & 3 &  & reserved\tabularnewline
 \hline
@@ -538,7 +539,10 @@ EXTERNAL\_DATA & 4 & external data block & data produced by external encodings\t
 CORE\_DATA & 5 & core data block & bit stream of all encodings except for external\tabularnewline
 \hline
 \end{tabular}
-
+\begin{tablenotes}
+\item[a] Formerly MAPPED\_SLICE\_HEADER.  Now used by all slice headers regardless of mapping status.
+\end{tablenotes}
+\end{threeparttable}
 
 \subsection{\textbf{Block content id}}
 
@@ -737,15 +741,27 @@ Hence consuming 1 byte for types `C' and `c', 2 bytes for types `S' and `s', 4 b
 
 The slice header block is never compressed (block method=raw). For reference mapped 
 reads the slice header also defines the reference sequence context of the data 
-blocks associated with the slice. Mapped and unmapped reads can be stored within 
-the same slice similarly to BAM file. Slices with unsorted reads must not contain 
-any other types of reads.
+blocks associated with the slice. Mapped reads can be stored along with
+\textbf{placed unmapped}\footnote{Unmapped reads can be \textit{placed} or \textit{unplaced}.
+By placed unmapped read we mean a read that is unmapped according to bit 0x4 of the 
+BF (BAM bit flags) data series, but has position fields filled in, thus "placing"  it on a reference sequence. In contrast, 
+unplaced unmapped reads have have a reference sequence ID of -1 and alignment position of 0.}
+reads on the same reference within the same slice.  
+
+Slices with the Multiple Reference flag (-2) set as the sequence ID in the header may contain reads
+mapped to multiple external references, including unmapped\footnotemark[\value{footnote}] reads (placed on these references or unplaced), 
+but multiple embedded references cannot be combined in this way.  When multiple references are 
+used, the RI data series will be used to determine the reference sequence ID for each record. This 
+data series is not present when only a single reference is used within a slice.
+
+The Unmapped (-1) sequence ID in the header is for slices containing only unplaced 
+unmapped\footnotemark[\value{footnote}] reads.
 
 A slice containing data that does not use the external reference in
 any sequence may set the reference MD5 sum to zero.  This can happen
 because the data is unmapped or the sequence has been stored verbatim
 instead of via reference-differencing.  This latter scenario is
-recommended for unsorted or non-coordinate sorted data.
+recommended for unsorted or non-coordinate-sorted data.
 
 The slice header block contains the following fields.
 
@@ -753,14 +769,17 @@ The slice header block contains the following fields.
 \hline
 \textbf{Data type} & \textbf{Name} & \textbf{Value}\tabularnewline
 \hline
-itf8 & reference sequence id & reference sequence identifier or -1 for unmapped 
-or unsorted reads\tabularnewline
+itf8 & reference sequence id & reference sequence identifier or\linebreak{}
+-1 for unmapped reads\linebreak{}
+-2 for multiple reference sequences\tabularnewline
 \hline
-itf8 & alignment start & the alignment start position or -1 for unmapped or unsorted 
-reads\tabularnewline
+itf8 & alignment start & the alignment start position.\linebreak{}
+Ignored on read and set to 0 on write if the slice is multiple-reference
+or contains unmapped unplaced reads\tabularnewline
 \hline
-itf8 & alignment span & the length of the alignment or 0 for unmapped or unsorted 
-reads\tabularnewline
+itf8 & alignment span & the length of the alignment.\linebreak{}
+Ignored on read and set to 0 on write if the slice is multiple-reference
+or contains unmapped unplaced reads\tabularnewline
 \hline
 itf8 & number of records & the number of records in the slice\tabularnewline
 \hline
@@ -774,7 +793,9 @@ itf8 & embedded reference bases block content id & block content id for the embe
 reference sequence bases or -1 for none\tabularnewline
 \hline
 byte[16] & reference md5 & MD5 checksum of the reference bases within the slice 
-boundaries or 16 \textbackslash{}0 bytes when unused\tabularnewline
+boundaries.  If this slice has reference sequence id of -1 (unmapped) or -2 (multi-ref) 
+the MD5 should be 16 bytes of \textbackslash{}0. For embedded references, the MD5 
+can either be all-zeros or the MD5 of the embedded sequence.\tabularnewline
 \hline
 byte[] & optional tags & a series of tag,type,value tuples encoded as
 per BAM auxiliary fields.\tabularnewline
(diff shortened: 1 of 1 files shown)


CRAMv3.tex
```diff
@@ -1519,9 +1530,9 @@ \subsubsection*{CRAM index}
 \begin{enumerate}
 \item Sequence id
 
-\item Alignment start
+\item Alignment start (ignored for unmapped slices)
```
jmthibault79: require that it be set to some sentinel value here?  maybe 0 or -1

jmthibault79:
Added some text to **12 Indexing** about multi-ref and unmapped slices.

CRAMv3.tex
jkbonfield: Probably 0 if we wanted to pin it down.  -1 is ridiculously huge in CRAM anyway.

However practically speaking the main requirement is that it has no meaning and is ignored by the decoder.  If you do tie it down then alignment span should have the same action applied because they belong together.

CRAMv3.tex
```diff
@@ -732,9 +734,14 @@ \subsection{\textbf{Slice header block}}
 
 The slice header block is never compressed (block method=raw). For reference mapped 
 reads the slice header also defines the reference sequence context of the data 
-blocks associated with the slice. Mapped and unmapped reads can be stored within 
-the same slice similarly to BAM file. Slices with unsorted reads must not contain 
-any other types of reads.
+blocks associated with the slice.
+Mapped reads can be stored along with unmapped-but-placed reads on the same reference
```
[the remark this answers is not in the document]
jmarshall: I'd suggest duplicating (with whatever terminology changes are appropriate for CRAM) the SAM spec's footnote on the matter:

> By *placed unmapped* read we mean a read that is unmapped according to its FLAG but whose RNAME and POS fields are filled in, thus “placing” it on a reference sequence. In contrast, *unplaced* unmapped reads have ‘*’ and 0 for RNAME and POS.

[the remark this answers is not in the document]
jkbonfield: Agreed, a footnote is better and using the SAM one makes sense.  We'd want to amend it to say "unplaced unmapped reads have a reference sequence ID of -1 and alignment position of zero".

jmthibault79:
Thank you for the comments @jkbonfield and @jmarshall .  Back to you.

CRAMv3.tex
```diff
@@ -732,9 +736,19 @@ \subsection{\textbf{Slice header block}}
...
+\textbf{placed unmapped}\footnote{By placed unmapped read we mean a read that is unmapped according to 
+its BAM bit flags but has position fields filled in, thus "placing"  it on a reference sequence. In contrast, 
+unplaced unmapped reads have have a reference sequence ID of -1 and alignment position of 0.}
+reads on the same reference within the same slice, as is possible in BAM.  
+Slices may also contain reads mapped to
+different external references, by setting the Multiple Reference flag (-2) as the sequence ID in the header.
+These can also include placed unmapped reads on these references, but multiple embedded references 
+cannot be combined in this way.  When multiple references are used, the RI data series will be used to 
+determine the reference sequence ID for each record. This data series is not required when only a single 
+reference is used within a slice.   Slices with unplaced unmapped\footnotemark[\value{footnote}] reads
```
jmthibault79: I think this is wrong and I'm over-constraining here.

This was based on an interpretation of the old text "Slices with unsorted reads must not contain any other types of reads."  What is an "unsorted read" independently of the context of other reads?  No, it's the overall file header which determines sortedness.  For any sorting besides Coordinate, Slices and Containers will often have reads with a variety of references, so Multi-Ref (-2) by necessity.  What would it gain us to exclude Unmapped-Unsorted reads from these slices?  It could be quite bad in the pathological case: (mapped, unmapped, mapped, unmapped, mapped) would require a new slice at every record boundary!

So I'd like to remove this.  What do you think?

jmthibault79: Added this change as commit bf44654

CRAMv3.tex
```diff
@@ -732,9 +736,21 @@ \subsection{\textbf{Slice header block}}
 
 The slice header block is never compressed (block method=raw). For reference mapped 
 reads the slice header also defines the reference sequence context of the data 
-blocks associated with the slice. Mapped and unmapped reads can be stored within 
-the same slice similarly to BAM file. Slices with unsorted reads must not contain 
-any other types of reads.
+blocks associated with the slice. Mapped reads can be stored along with
```
cmnbroad: I'd propose including a table, similar to the one in https://github.com/samtools/hts-specs/issues/370, right in the spec. With the columns for the three Slice "reference context" types, and rows for every other header field affected by the reference context value (maybe with an additional row for `RI`), it would nicely summarize the whole reference context state space for Slice header.

CRAMv3.tex
```diff
@@ -732,9 +736,21 @@ \subsection{\textbf{Slice header block}}
...
+\textbf{placed unmapped}\footnote{Unmapped reads can be \textit{placed} or \textit{unplaced}.
+By placed unmapped read we mean a read that is unmapped according to 
+its BAM bit flags but has position fields filled in, thus "placing"  it on a reference sequence. In contrast, 
+unplaced unmapped reads have have a reference sequence ID of -1 and alignment position of 0.}
+reads on the same reference within the same slice, as is possible in BAM.
```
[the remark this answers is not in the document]
jmthibault79: Happy to remove it if we prefer.  This was my attempt to preserve the existing "similarly to BAM file" wording.

CRAMv3.tex
```diff
@@ -732,30 +736,67 @@ \subsection{\textbf{Slice header block}}
...
+By placed unmapped read we mean a read that is unmapped according to 
+its BAM bit flags but has position fields filled in, thus "placing"  it on a reference sequence. In contrast, 
+unplaced unmapped reads have have a reference sequence ID of -1 and alignment position of 0.}
+reads on the same reference within the same slice, as is possible in BAM.  
+
+Slices with the Multiple Reference flag (-2) set as the sequence ID in the header may contain reads
+mapped to multiple external references, including unmapped\footnotemark[\value{footnote}] reads (placed on these references or unplaced), 
+but multiple embedded references cannot be combined in this way.  When multiple references are 
+used, the RI data series will be used to determine the reference sequence ID for each record. This 
+data series is not required when only a single reference is used within a slice.
```
jkbonfield: Pedant mode: Maybe "not present" to be explicit that it is banned.

This doesn't actually matter in modern implementations as we put each data series into its own block, but the earlier crams used HUFFMAN, BETA, GAMMA etc and interleaved everything into the CORE block.  In that situation it is important to *not* write out the RI field when it's not needed otherwise we get out of sync when decoding.

CRAMv3.tex
jkbonfield: I agree in name sorted data it could be pathologically bad.

I'm not sure what we do in this scenario, but it's probably quite likely we just mix them in the same slice.  I don't see any issue with stating multi-ref (-2) and having ref -1 (unmapped) for some of the reads.  It feels natural.  This text is probably just a hangover from when we had multiple slice types.

CRAMv3.tex
jmthibault79: done

CRAMv3.tex
cmnbroad: Thanks for adding the table. I'm not sure if this is a spec comment or a file format comment, but it seems like "Embedded Reference" is another Slice "type" that is mutually exclusive with the existing 3 types in the table. What should the refID value be for such a slice (or is that defined elsewhere?) Should it have it's own column in the table ?

cmnbroad: Also, while were at it, should this table (the one I suggested -;)) overlaps some with the table that immediately follows. Can/should they be merged ?

jmthibault79: Responded to the Embedded Reference question at #370.

I don't think the tables should be merged.  The main slice header fields table is conceptually similar to other header tables in the spec, and follows the common (type, name, description) pattern.  The new slice types summary table describes the logical interactions of the fields instead, so I see that as having a different purpose.

cmnbroad: Ok, well, I'm all in favor of continuing to tighten up this part of the spec, but I certainly don't have any objections to merging what we have so far, which looks good.

jkbonfield: Sorry I've been absent recently - illness.  I'll take a look at the start of next week.  Thanks for the updates.

CRAMv3.tex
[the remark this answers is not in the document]
jkbonfield: The specification elsewhere now refers to BF and BAM bit flags and CF as CRAM flags.  This is perhaps a bit sloppy, although it does list their meanings.

I am happy to refer to BF as BAM bit flags, but perhaps we should be explicit and state e.g. "unmapped according to bit 0x4 of the BF (BAM bit flags) data series, but has position fields filled in thus ...".

CRAMv3.tex
```diff
@@ -732,30 +736,67 @@ \subsection{\textbf{Slice header block}}
...
+\tabularnewline
+\hline
+ \textbf{Aln Span} & Span including all reads & 0 / ignored & 0 / ignored
+\tabularnewline
+\hline
+ \textbf{Ref MD5} & Required for external reference;\linebreak{}
+optional for embedded & 16 bytes of \textbackslash{}0 & 16 bytes of \textbackslash{}0
+\tabularnewline
+\hline
+\end{tabular}
```
jkbonfield: I am unsure how much extra value this table brings, given it is immediately followed by the slice description anyway which basically has the same information in it.  The exception is the MD5 description in the table below which could be made more like this table.  Right now it states:

> MD5 checksum of the reference bases within the slice boundaries or 16 \0 bytes when unused

"When unused" is as clear as mud and I prefer your explanation in the table above.  Can we update the table below to be:

> MD5 checksum of the reference bases within the slice boundaries.  If this slice has reference sequence id of -1 (unmapped) or -2 (multi-ref) the MD5 should be 16 bytes of \textbackslash{}0.   For embedded references, the MD5 can either be all-zeros or the MD5 of the embedded sequence.

Note this last sentence is subtly different to how htsjdk currently works I believe.  If I recall, putting an embedded reference sequence which didn't exactly match the external reference (eg a consensus instead of the genuine reference) demonstrated that the code validated the MD5sum against the external reference and not the embedded reference.  I'm not convinced this is correct behaviour although it does raise the question of whether it is even permitted to have the embedded reference differing to the external reference.

jmthibault79: OK, updated the text.

I don't have a strong opinion on the new table.

cmnbroad: I'm a big fan of tables over text, but I agree that the new table doesn't add much as it stands (the original suggestion was to merge it with the table below - but as @jmthibault79 pointed out, that would diverge from the pattern used by other tables in the spec). So I'm fine leaving it out.

CRAMv3.tex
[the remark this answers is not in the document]
jmthibault79: updated for these comments

jkbonfield:
I approve what we have here.  There are additional things listed in #370 so this is a big step forward but not a complete fix of that.

We can either squash and merge this and produce a new PR for the container updates, or continue with adding things to this.  I'm easy either way.

jmthibault79:
Thanks!  I'll remove the table and squash.

jmthibault79:
I'll need someone with write access to merge.

jkbonfield:
Merged as 5d9fc522de812f40f8e5c58159804ebed510e425

I removed some trailing white space from some of the lines, to be tidy.  I think @jmarshall initially gave me the hints to make these appear in diffs, using .gitconfig:

```
[color]
        ui = auto
```

We'll need to get one of the samtools maintainers to grant you and @cmnbroad commit access here.

jmarshall:
Done — I've added @cmnbroad and @jmthibault79 to **Specification authors**, so you should now be able to push to **hts-specs**. Use this power wisely :smile:

jmthibault79:
Thanks @jmarshall

cmnbroad:
Great - thx @jmarshall !

PR #373, opened 2019-01-22
Avoid \usepackage{ulem} [trivial]

jmarshall:
By adding `\renewcommand{\ULthickness}{4pt}` and eyeballing the resulting thick bars, verified that the `ulem` package was acting on the `\emph{…}` text (making it underlined rather than italics) and no other text in the two CRAM specs.

Removed `\usepackage{ulem}`. Changed the Hsi-Yang Fritz, et al. reference to a full citation footnote. Changed section references to bold or cross-references, and left other underlined text (which are all introductions/definitions of terms) to revert to italics. (Details in commit message.)

Also a second commit for your consideration, tidying up the formatting of the arithmetic in the _Choosing the container size_ appendix.

@jkbonfield et al: hopefully this is trivial enough to just merge :smile:

### DIFF
--- a/CRAMv2.1.tex
+++ b/CRAMv2.1.tex
@@ -13,7 +13,6 @@
 
 \usepackage{graphicx}
 \usepackage{array}
-\usepackage{ulem}
 \usepackage{fixltx2e}
 \usepackage{amssymb}
 \usepackage{fancyhdr}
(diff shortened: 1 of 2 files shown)


jmarshall:
Thanks. Does that mean (1) JM please merge; (2) JKB will merge; or (3) @cmnbroad should have a look too…

cmnbroad:
LGTM, it its (3).

jmarshall:
Thanks, I have made it so. One of these days we should update the CRAM section in MAINTAINERS.md btw…

PR #374, opened 2019-01-24
Improve htsget table contents formatting when displayed on GitHub [trivial]

jmarshall:
I finally figured out how to get markdown nested inside an HTML \<table\> into a common subset of kramdown and GitHub-flavoured markdown, so that the _italics_ and `code` etc appear as such (rather than as raw markup) in the rest of the tables when viewing _htsget.md_ within the GitHub repository web view.

@mlin: This has no text changes (and I've verified that it makes no difference to how [htsget.html](http://samtools.github.io/hts-specs/htsget.html) is displayed), so IMHO as editor you should feel free to check it over and merge this without conversation or delay.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -51,18 +51,21 @@ For errors that are specific to the `htsget` protocol, the response body SHOULD 
 
 <table>
 <tr markdown="block"><td>
+
 `htsget`
 _object_
 </td><td>
 Container for response object.
 <table>
 <tr markdown="block"><td>
+
 `error`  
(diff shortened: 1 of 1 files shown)


jmarshall:
Thanks. Out of interest, which merge button did you use (“Squash and merge”?) and did you edit the first line of the commit message yourself or was that partially modified by the GitHub UI? I'm curious how much of the change in the message from the original commit may be related to Isaacs/github#1368.

Issue #352, opened 2018-10-15
<*> allele vs <NON_REF> 


lbergelson:
In the VCF 4.3 spec the `<*>` is specified as the *unspecified alt allele*.  

GATK uses the `<NON_REF>` allele instead for this purpose and treats `<*>` as a deprecated version of the `*` allele.  This is obviously not matching the spec.  Are there other software suites that generate GVCFs that use `<*>`?

@pd3 I'm wondering why `<*>` was selected as the canonical unspecified alt.  Do other common callers produce gVCFs with `<*>`?

pd3:
Historically, samtools was the first caller to start using the unspecified allele and originally it was `*`. After this was codified in the specification as `<*>`, samtools/bcftools started producing `<*>` instead.  I don't know why GATK uses `<NON_REF>` instead of `<*>`.

lbergelson:
Interesting, the history seems complicated.  I did bit of digging to try and understand it.

* It looks like samtools used `X` and then `<X>` as well before moving to `<*>`.  
 
* According to git history samtools adopted `<*>` in 2015 (samtools/samtools@78c23438066eb38fd6e04924559ae6db32cef9b0) 
*  bcftools supported it in 2014 (samtools/bcftools@802ff30)

* The spec included both `<X>` and `<NON_REF>` up until 2014, when it changed to standardize on `<*>` (samtools/hts-specs@4a91745)

* GATK started using`<NON_REF>` since 2013 so it's pretty old.  (broadgsa/gatk-protected@7c3cd48fdfa571fc4735188b9548f8e7db976b06)

Unfortunately GATK never got the message to change to `<*>`, and it doesn't seem likely to happen due to how many `<NON_REF>` gvcfs there are out there now.  

It might be worth considering adding back `<NON_REF>` to the spec since practically it's used in petabytes of gvcf files and it's unlikely to go away.

yfarjoun:
Given that we have consistently said that gVCFs are NOT deliverable, I
think that we have an opportunity to change the <NON_REF> allele at EVERY
GATK version...

nh13:
@yfarjoun my best google-fu could not find a list of supported or unsupported deliverables for GATK, but searching for ‘<NON_REF>’ did yield a non-trivial number of tools that rely or leverage the allele.  So likely there are gVCFs that are stored that have this allele, waiting to be combined in joint calling mid-project, or otherwise. I just don’t buy your argument. 

[...]

PR #380, opened 2019-02-06
Suggesting <NON_REF> as the unspecified alt allele

lbergelson:
* Adding <NON_REF> to the spec and recommending its use over the <*> allele which is easily confused with *.
* closes https://github.com/samtools/hts-specs/issues/352

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -1362,7 +1362,7 @@ Haplotype quality & 30 & 40 & 40 \\
 In order to report sequencing data evidence for both variant and non-variant positions in the genome, the VCF specification allows to represent blocks of reference-only calls in a single record using the END INFO tag, an idea originally introduced by the gVCF file format\footnote{\url{https://help.basespace.illumina.com/articles/descriptive/gvcf-files/}}.
 The convention adopted here is to represent reference evidence as likelihoods against an unknown alternate allele.
 Think of this as the likelihood for reference as compared to any other possible alternate allele (both SNP, indel, or otherwise).
-Either of the two symbolic alleles $<$*$>$ or $<$NON\_REF$>$ may be used to represent this unspecified alternate allele.
+Either of the two symbolic alleles $<$NON\_REF$>$ or $<$*$>$ may be used to represent this unspecified alternate allele.
 It is recommended to use $<$NON\_REF$>$ because $<$*$>$ is easily confused with the unrelated `*' allele.
 
 Example records are given below:
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -1362,19 +1362,20 @@ Haplotype quality & 30 & 40 & 40 \\
 In order to report sequencing data evidence for both variant and non-variant positions in the genome, the VCF specification allows to represent blocks of reference-only calls in a single record using the END INFO tag, an idea originally introduced by the gVCF file format\footnote{\url{https://help.basespace.illumina.com/articles/descriptive/gvcf-files/}}.
 The convention adopted here is to represent reference evidence as likelihoods against an unknown alternate allele.
 Think of this as the likelihood for reference as compared to any other possible alternate allele (both SNP, indel, or otherwise).
-A symbolic alternate allele $<$*$>$ is used to represent this unspecified alternate allele.
+A symbolic alternate allele is used to represent this unspecified alternate allele.
+Either $<$*$>$ or $<$NON\_REF$>$ may be used to represent the unspecified alternate allele, but $<$NON\_REF$>$ is preferred due to the ease of confusion between $<$*$>$ and the unrelated `*' allele.
 
 Example records are given below:
 \scriptsize
 \begin{flushleft}
 \begin{tabular}{ l l l l l l l l l l }
 \#CHROM & POS & ID & REF & ALT & QUAL & FILTER & INFO & FORMAT & Sample \\
-1 & 4370 & . & G & $<$*$>$ & . & . & END=4383 & GT:DP:GQ:MIN\_DP:PL & 0/0:25:60:23:0,60,900 \\
+1 & 4370 & . & G & $<$NON\_REF$>$ & . & . & END=4383 & GT:DP:GQ:MIN\_DP:PL & 0/0:25:60:23:0,60,900 \\
 1 & 4384 & . & C & $<$*$>$ & . & . & END=4388 & GT:DP:GQ:MIN\_DP:PL & 0/0:25:45:25:0,42,630 \\
-1 & 4389 & . & T & TC,$<$*$>$ & 213.73 & . & . & GT:DP:GQ:PL & 0/1:23:99:51,0,36,93,92,86 \\
-1 & 4390 & . & C & $<$*$>$ & . & . & END=4390 & GT:DP:GQ:MIN\_DP:PL & 0/0:26:0:26:0,0,315 \\
-1 & 4391 & . & C & $<$*$>$ & . & . & END=4395 & GT:DP:GQ:MIN\_DP:PL & 0/0:27:63:27:0,63,945 \\
-1 & 4396 & . & G & C,$<$*$>$ & 0 & . & . & GT:DP:GQ:P & 0/0:24:52:0,52,95,66,95,97 \\
+1 & 4389 & . & T & TC,$<$NON\_REF$>$ & 213.73 & . & . & GT:DP:GQ:PL & 0/1:23:99:51,0,36,93,92,86 \\
+1 & 4390 & . & C & $<$NON\_REF$>$ & . & . & END=4390 & GT:DP:GQ:MIN\_DP:PL & 0/0:26:0:26:0,0,315 \\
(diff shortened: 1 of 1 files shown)


lbergelson:
I don't know why circle-ci ran on this and failed...  Strange.

VCFv4.3.tex
```diff
@@ -1347,19 +1347,20 @@ \subsection{Representing unspecified alleles and REF-only blocks (gVCF)}
 In order to report sequencing data evidence for both variant and non-variant positions in the genome, the VCF specification allows to represent blocks of reference-only calls in a single record using the END INFO tag, an idea originally introduced by the gVCF file format\footnote{\url{https://help.basespace.illumina.com/articles/descriptive/gvcf-files/}}.
 The convention adopted here is to represent reference evidence as likelihoods against an unknown alternate allele.
 Think of this as the likelihood for reference as compared to any other possible alternate allele (both SNP, indel, or otherwise).
-A symbolic alternate allele $<$*$>$ is used to represent this unspecified alternate allele.
+Either of the two symbolic alleles $<$*$>$ or $<$NON\_REF$>$ may be used to represent this unspecified alternate allele.
```
cyenyxe: Given that we are recommending the usage of NON_REF, maybe it should be listed first?

lbergelson:
@cyenyxe That's a good question about 4.3 vs 4.2.  Gatk doesn't output 4.3 due to technical problems in htsjdk.   We're planning on fixing it but it hasn't happened yet.  This PR isn't a deciding factor either way.  

I think we could probably backport port the gvcf blurb to the 4.2 spec since it's really more of a clarification about a specific use of VCF than a change in the spec and clearly gvcfs have existed before 4.2.  

If you think we should include this note in the 4.2 spec I can move that back as well.

VCFv4.3.tex
lbergelson: Good point, I've switched them.

cyenyxe:
I could be fine with backporting the gVCF part, as it doesn't break compatibility and would make GATK VCFs (not a small number) fully compliant against the spec. What do you think @pd3?

pd3:
@lh3 came up with the idea of the space-holder for an unseen allele in samtools/bcftools and when GATK started using it as well, we agreed on the symbolic star allele `<*>`. Samtools/bcftools swiftly changed from its original `X` and `<X>`. Atlhough GATK developers were active in drafting the specification and agreed on `<*>` years ago, somehow it still remains an issue. I don't understand why and in my opinion allowing `<NON_REF>` now would set a very bad precedent.

cyenyxe:
The problem with <*> is that it is an extremely confusing notation (see comments from non-GATK developers in #352). We can always create a brief poll to understand how people are using them, and which one would they prefer in the long term.

But the truth is that there are many VCF files out there already using NON_REF.

pd3:
I agree the notation is not well described in the specification, but I disagree it is extremely confusing.

The correct way forward for GATK is to switch to producing `<*>` and silently continue supporting also `<NON_REF>` for backward compatibility, rather than enforce all other tools to start supporting both. It is programmatically very easy to do and if both alleles should become a standard as in this proposal, it would have to support both anyway!

tfenne:
A few thoughts:

1. @pd3 I have personally interacted with enough folks confused by `*` vs `<*>` that I find it hard to agree that it's not confusing.
2. I'm not sure the historical argument holds water.  As @lbergelson points out in #352 at least as far as he could see GATK first started using `<NON_REF>` in 2013, _before_ any of the other tools mentioned used an allele for this purpose.
3. To be fair though, there was a lot of turnover on the GATK team IIRC in the 2013-2015 timeframe, so it's possible that someone agreed to the change to `<*>` and that agreement got lost in the churn.  I certainly don't think anyone was acting in bad faith by not changing to `<*>` at the time.  Is there any chance that discussion happened in email/git issues vs. phone calls?
4. As a user of VCF I feel like VCF could adopt the PERL slogan of "there's more than one way to do it" - so many things in VCF are encoded differently by different callers, having two ways of doing it seems perfectly normal.  

Here's an idea: how about instead of extending the spec to officially recognize `<*>` and `<NON_REF>`, having a way to use the `##ALT` header line to explicitly declare in a VCF what if any allele is used for this purpose by adding a `NONREF` type?  And then perhaps stating that if not present `<*>` and `<NON_REF>` are used by existing tools and may be expected?

pd3:
@tfenne, on my hard drive lives the file samtools/bam2bcf.c with a timestamp from Dec 20 2011 with the following line 
````
kputc(bc->unseen == i? 'X' : "ACGT"[bc->a[i]], &s);
````
and very likely it is not the first version of it.

My points are following:
1. The current GATK's `<NON_REF>` allele does not break the VCF specification, therefore you can continue using it.
2. We tried to standardize this years ago and the consensus was that the symbolic star allele`<*>` would become reserved in the specification, not `<X>` and not `<NON_REF>`.
3. If people find `<*>` vs `*` confusing, this is still only a problem of documenting it better.

Although I personally would have preferred to continue using the `<X>` notation, we switched to `<*>` for the sake of supporting the standard. I find it disconcerting that now, years after, the issue is still alive. 

As for your suggestion, the use of the symbolic non-ref allele is very specific and I am not convinced that it is worthwhile to complicate further the `##ALT` meta-information lines.

**Comment by jkbonfield** [comment]
My take on this, as an independent non-VCF person, is that if things don't break the specification but are in common usage then it doesn't harm to document them.

However we need to be clear to distinguish normative specification text and explanatory / clarificatory text.

If the scale of the problem is simply too vast (eg the TLEN problem of 5' to 3' vs left-most to right-most is *dominated* by code doing the wrong thing, because the spec changed from one thing to another and hardly anyone updated their code), then documenting both is a pragmatic alternative but not an ideal scenario.  I'd be wary of blessing one over the other though - if you go down this road you are acknowledging both will exist forever.

If you want to get only method and feel it is achievable, then don't label both as permitted - just document the unsatisfactory situation and emphasise the recommended practice is the current behaviour.

**Comment by lh3** [comment]
> Although I personally would have preferred to continue using the `<X>` notation

I prefer `<X>`, too, but I am fine with `<*>`, too.

> on my hard drive lives the file samtools/bam2bcf.c with a timestamp from Dec 20 2011 with the following line

Bcftools was separated from samtools in samtools v0.1.9, released on 2010-10-27. Bcftools had been using `X` as the unspecified allele from the very beginning. As I vaguely remember, the VCF spec didn't have the concept of symbolic allele at the time. I could be wrong, though.


Personally, I am ok to add `<NON_REF>` as an alternative to `<*>` in principle, but I think `<*>` should be the recommended encoding in the spec. That was the decision. Also, if we add `<NON_REF>`, htsjdk/GATK should also support both. @pd3: would supporting `<NON_REF>` cause much trouble to htslib?

> Here's an idea: how about instead of extending the spec to officially recognize `<*>` and `<NON_REF>`, having a way to use the `##ALT` header line to explicitly declare in a VCF what if any allele is used for this purpose by adding a `NONREF` type? And then perhaps stating that if not present `<*>` and `<NON_REF>` are used by existing tools and may be expected?

If we add `<NON_REF>`, I much prefer to make them always equivalent. The VCF spec is complex enough. Having a new header line seems too complicated.

**Comment by cyenyxe** [comment]
After several offline discussions, I have updated this to acknowledge the existence of <NON_REF>, but keep <*> as the preferred allele.

File: VCFv4.3.tex [hunk 2]
```diff
@@ -1362,7 +1362,7 @@ \subsection{Representing unspecified alleles and REF-only blocks (gVCF)}
 In order to report sequencing data evidence for both variant and non-variant positions in the genome, the VCF specification allows to represent blocks of reference-only calls in a single record using the END INFO tag, an idea originally introduced by the gVCF file format\footnote{\url{https://help.basespace.illumina.com/articles/descriptive/gvcf-files/}}.
 The convention adopted here is to represent reference evidence as likelihoods against an unknown alternate allele.
 Think of this as the likelihood for reference as compared to any other possible alternate allele (both SNP, indel, or otherwise).
-A symbolic alternate allele $<$*$>$ is used to represent this unspecified alternate allele.
+A symbolic alternate allele $<$*$>$ is used to represent this unspecified alternate allele. This is preferred over the symbolic allele $<$NON\_REF$>$, used by the Genome Analysis Toolkit (GATK).
```
lbergelson: @cyenyxe This seems like the right solution, the community has definitely overruled my initial suggestion to promote `<NON_REF>` over `<*>`
```suggestion
A symbolic allele $<$*$>$ is used to represent this unspecified alternate allele. 
This is preferred over the equivalent allele $<$NON\_REF$>$.
```

I think we have a redundant `alternate` in there, and we should split multiple sentences on multiple lines to match the formatting.  

I think it's good to specify that they have the same meaning, and I don't know if it's necessary to include the GATK by name here since we don't usually refer to specific implementations.  

It might also be useful to highlight in some way the term `unspecified alternate allele` if that's how we intend it to be pronounced, since my major complaint with `<*>` is that it's pronunciation causes confusion.

PR #383, opened 2019-02-19
Fix rANS pseudocode.

jkbonfield:
- Consistency of s vs sym between ReadFrequences0 and ReadFrequencies1
- Added missing calls to ReadFrequencies0 and ReadFrequencies1
- ReadFrequencies1 now updates F_{sym} and not F_{i}.
- Added missing i++ in RansDecode1.

This pseudocode has been validated using a basic JavaScript
implementation at https://github.com/jkbonfield/hts-specs/tree/cram_reference/cram_reference

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -2535,29 +2535,28 @@ Here we also define some of the functions for manipulating the rANS state, which
 \Statex (Reads a table of Order-0 symbol frequencies $F_i$
 \Statex (and sets the cumulative frequency table $C_{i+1} = C_i+F_i$)
 \Procedure{ReadFrequencies0}{$F, C$}
-\State $s \gets$ \Call{ReadByte}{}\Comment{Next alphabet symbol}
-\State $last\_sym \gets s$
+\State $sym \gets$ \Call{ReadByte}{}\Comment{Next alphabet symbol}
+\State $last\_sym \gets sym$
 \State $rle \gets 0$
 \Repeat
   \State $f \gets$ \Call{ReadITF8}{}
-  \settowidth{\maxwidth}{$C_s$}
-  \State \algalign{F_s}{\gets} $f$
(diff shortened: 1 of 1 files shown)


jkbonfield:
The cram_reference branch is an ongoing effort to build a simple to read reference implementation.  I'm starting with the core algorithms that have pseudocode (which will also focus on CRAMv4), but long term this could include the entirety of CRAM.

I don't intend for us to bless any specific CRAM code, but a reference implementation purely for purposes of spec clarification rather than efficiency may serve a purpose.

CRAMv3.tex
```diff
@@ -2622,6 +2622,7 @@ \subsubsection*{rANS order-1}
 
 \Statex
 \Procedure{RansDecode1}{$output$, $nbytes$}
+  \State \Call{ReadFrequencies1}{$F,\ C$} \Comment{Creates F and C vectors}
```
cmnbroad: Might be worth updating the comment to reflect the fact that `F` and `C` are (2d) arrays in this context.

cmnbroad:
Interesting - I didn't know about the cram reference code. Will have to take a look.

jkbonfield:
The reference code is new and not in a PR yet.  It's also a work in progress.  Primarily I wrote it purely to double check that my pseudocode is actually correct, as the C rANS codec is very complex due to excessive (but useful) optimisations and it doesn't look particularly like the spec.

It's proving a useful exercise though for CRAM 4 and in documenting it this way I've thought about the problem harder and come up with improvements to the spec too, so it's actually a bit of a circular process of coding and spec writing.

jkbonfield:
Rebased off master.  Also added comment about 2D F and C arrays.

If I recall this was already basically agreed on, so will merge tomorrow unless I hear otherwise.

cmnbroad:
Yeah, looks good to me 👍.

PR #328, opened 2018-07-26
Add Continuous integration to automatically generate PDF and diff-pdfs using circleCI

yfarjoun:
This PR adds a continuous integration option that creates the PDFs and serves them up for perusal. 

This informs circle-ci to make all the pdfs and publish them to an accessible location.
See https://circleci.com/gh/samtools/hts-specs for build information.

PR #385, opened 2019-02-21
Add htsget 1.2.0, OpenAPI v3.0.2 spec

brainstorm:
As mentioned over twitter:

https://twitter.com/braincode/status/1097808553723670528

I've tried to comply with the official htsget spec, from the human-readable official spec:

https://samtools.github.io/hts-specs/htsget.html

But please let me know if there's something off. Things that concern me:

1. [OpenAPI v3 only seems to have **signed** int32, while on the spec an **unsigned** int32 is required by the htsget spec](https://twitter.com/braincode/status/1098415940222431232). Granted, this is just for code scaffolding/stubs, but I hope nobody implements it wrong by accident.
1. ~~I could write in the OAuth aspects in it, but the htsget spec explicitly keeps it out since it's "beyond 
 the scope of this [htsget] specification". I feel inclined to add it regardless to ease code generation, thoughts?~~. Just went ahead and added it.
1. I've added all fields as shown in the spec for BAM, but for VCF, I just picked up some (most used ones?) from [the VCF4.2 spec](https://samtools.github.io/hts-specs/VCFv4.2.pdf). Would it make sense to list them in the hts-spec as well?

### DIFF
--- a/pub/htsget-openapi.yaml
+++ b/pub/htsget-openapi.yaml
@@ -9,8 +9,6 @@ info:
   contact:
     name: GA4GH
     email: security-notification@ga4gh.org
-  license:
-    name: GPLv3
 paths:
   /reads/{id}:
     get:
@@ -34,7 +32,7 @@ paths:
             type: string
         - in: "query"
           name: referenceName
-          description: VReference sequence name
+          description: Reference sequence name
           example: "chr1"
           required: false
           schema:
@@ -137,7 +135,7 @@ paths:
             type: string
         - in: "query"
           name: referenceName
-          description: VReference sequence name
+          description: Reference sequence name
           example: "chr1"
           required: false
           schema:
@@ -160,19 +158,6 @@ paths:
             type: integer
             format: int64
             minimum: 0
-        - in: "query"
-          name: fields
-          description: A list of variant fields to include, such as INFO, SAMPLE, FORMAT, etc.
-          example: "fields=INFO,SAMPLE"
-          required: false
-          schema:
-            enum:
-              - "INFO"
-              - "SAMPLE"
-              - "FILTER"
-              - "FORMAT"
-              - "ALT"
-            type: string
         - in: "query"
           name: tags
           description: A comma separated list of tags to include, by default all.
--- a/pub/htsget-openapi.yaml
+++ b/pub/htsget-openapi.yaml
@@ -12,7 +12,7 @@ info:
   license:
     name: GPLv3
 paths:
-  /reads:
+  /reads/{id}:
     get:
       summary: Gets the reads from a pre-indexed id
       operationId: searchReadId
@@ -20,8 +20,8 @@ paths:
         Searches a pre-indexed object id.
       parameters:
         - in: "path"
-          name: identifier of the read object
-          description: id
+          name: id
+          description: identifier of the read object
           required: true
           schema:
             type: string
@@ -113,7 +113,7 @@ paths:
           - read:genomic_reads
 
 
-  /variants:
+  /variants/{id}:
     get:
       summary: Gets the variants from a pre-indexed id
       operationId: searchVariantId
--- a/pub/htsget-openapi.yaml
+++ b/pub/htsget-openapi.yaml
@@ -1,0 +1,263 @@
+openapi: 3.0.0
+servers:
+  - description: htsget genomics api
+    url: https://virtserver.swaggerhub.com/brainkod/htsget/1.1.1
+info:
+  description: "This data retrieval API bridges from existing genomics bulk data transfers to a client/server model"
+  version: "1.1.1"
+  title: htsget
+  contact:
+    name: Roman Valls Guimera
+    email: brainstorm+htsget@nopcode.org
+  license:
+    name: GPLv3
+paths:
+  /reads:
+    get:
+      summary: Gets the reads from a pre-indexed id
+      operationId: searchReadId
+      description: |
+        Searches a pre-indexed object id.
+      parameters:
+        - in: "path"
+          name: identifier of the read object
+          description: id
+          required: true
+          schema:
+            type: string
+        - in: "query"
+          name: format
+          description: File format, BAM (default), CRAM.
+          example: "BAM"
+          required: false
+          schema:
+            type: string
+        - in: "query"
+          name: referenceName
+          description: VReference sequence name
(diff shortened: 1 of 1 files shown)


brainstorm:
Adding the Oauth2 bit might not have been the best idea since AWS API Gateway chokes on it:

skarmavbild 2019-02-21 kl 15 08 41

delagoya:
@brainstorm thanks a couple of notes on quick review: 

1. unsigned int32 comes from trying to minimize the bytes necessary to store a genomic position in highly compress formats like BAM or CRAM. Since there are query parameters, I think it is OK to define it as `int64` with a `minimum` of `0`

2. Per your AWS API Gateway error, there are some OAS OAuth2 examples. But this seems specific to API Gateway and would not be a good thing to add to a official specification, but rather define a API Gateway specific schema in your own repo, import the schema components using  `$ref`, and add in the `custom-authorizer` there.

3. No comment on this yet, will need to dig into the schema details.

jmarshall:
Re (1), that OpenAPI documentation says `format` is free text so you would seem to be entirely justified in putting `uint32` there.

(IMHO the spec saying "32-bit unsigned int" here is over-constraining things. It's a text format so just "unsigned int" or "int, minimum 0" says the same thing as far as the protocol is concerned. However "32-bit" has been there since the original Google Docs document and never come up in discussion as far as I know.)

Re (3), please raise this as a separate issue. Subsetting VCF fields via the `fields` parameter appears to have been overlooked when variants were added to htsget in PR #233.

Also

* The contact information will need changing before merging. This could be the ga4gh-htsget mailing list; or the refget OpenAPI spec uses a GA4GH Security Notifications email address here.

daviesrob:
I think it would be good to avoid 32 bit limitations where not strictly needed.  SAM does support positions longer than that, and we're planning to add support for them in HTSlib (see samtools/htslib#709).  While not necessary for human sequencing, it is for quite a few other organisms.

brainstorm:
Thanks everyone for the prompt and valuable feedback, I just added those the proposed changes and opened #386 as suggested.

pub/htsget-openapi.yaml
```diff
@@ -0,0 +1,286 @@
...
+    url: https://virtserver.swaggerhub.com/brainkod/htsget/1.1.1
+info:
+  description: "This data retrieval API bridges from existing genomics bulk data transfers to a client/server model"
+  version: "1.1.1"
+  title: htsget
+  contact:
+    name: GA4GH
+    email: security-notification@ga4gh.org
+  license:
+    name: GPLv3
```
jmarshall: [“The license information for the exposed API.”](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.0.md#info-object) Probably best to omit `license` for the time being.

pub/htsget-openapi.yaml
```diff
@@ -0,0 +1,286 @@
...
+        - in: "query"
+          name: format
+          description: File format, BAM (default), CRAM.
+          example: "BAM"
+          required: false
+          schema:
+            type: string
+        - in: "query"
+          name: referenceName
+          description: VReference sequence name
```
jmarshall:
```suggestion
          description: Reference sequence name
```

pub/htsget-openapi.yaml
```diff
@@ -0,0 +1,286 @@
...
+          description: A list of variant fields to include, such as INFO, SAMPLE, FORMAT, etc.
+          example: "fields=INFO,SAMPLE"
+          required: false
+          schema:
+            enum:
+              - "INFO"
+              - "SAMPLE"
+              - "FILTER"
+              - "FORMAT"
+              - "ALT"
```
jmarshall: Probably best to leave this out until #386 is resolved.

pub/htsget-openapi.yaml
mlin: Yea, various parties may be allergic to GPL in particular -- omitting it is probably easier than selecting an alternative? :wink:

mlin:
Sorry for the terrible belatedness of me adding a couple of comments here:

1. This is a really terrific companion to the spec, so first thank you!!
2. The example responses should probably illustrate HTTP URLs in addition to data: URIs, since the redirect is a core mechanic of the protocol, which servers designed for scalable operations would usually rely on. We could make an example for publicly accessible 1000 Genomes data or such.
3. The possible error responses could be fleshed out in more detail following the prose spec.

I would really like to help with these, but have bandwidth problems so would also welcome if somebody else is able to chew on them.

Thanks again this is great!

brainstorm:
Thanks everyone for the feedback and code review, I'm refactoring and while I'm at it, I'll introduce 1.2.0 spec details as well (class), WIP here: https://app.swaggerhub.com/apis/brainkod/htsget/1.2.0#/default/searchReadId

brainstorm:
Spec bumped to 1.2.0, thanks @mlin for the feedback and the heads up @ohofmann.

I hope there's nothing else missing/broken or out of sync... looks valid/ok on SwaggerHub:

https://app.swaggerhub.com/apis/brainkod/htsget/1.2.0

OT: Who is in charge of the CircleCI config on this repo?

brainstorm:
👋

jmarshall:
The example of a `/variants/{id}` response is a BAM stream. It needs to be VCF or BCF instead.

> OT: Who is in charge of the CircleCI config on this repo?

You can figure that out by looking at the relevant PR (#328). However it's immaterial — if this PR is rebased to after PR #328 landed, Circle CI will be happy. As this PR does not involve PDFs, it doesn't matter either way.

brainstorm:
Thanks @jmarshall, good catch, fixed now!

jmarshall:
Agreed at the meeting that this is now ready for merging. Two remaining minor concerns:

1. The swaggerhub URL at the top is specific to `brainkod`:

```yaml
servers:
    url: https://virtserver.swaggerhub.com/brainkod/htsget/1.2.0
```

   Can this be replaced with a non-specific URL like the one in _refget-openapi.yaml_?

2. Last time I looked the treatment of error return codes was a bit haphazard, however we're happy for that to be tidied up at a later date.

@brainstorm: If you would like to address (1), I'll take care of rebasing and merging this. Thanks!

jmarshall:
Merged as 766424faf913b54ff789ac480f6b2bebb0633f7d. Thanks!

PR #388, opened 2019-02-22
CRAM: The first record in a slice should have alignment-delta = 0

jmthibault79:
Update the DecodePositions pseudocode to match htslib/htsjdk behavior: the state of Last Position is initialized with the Slice's Alignment Start, and therefore the Alignment Delta for the first record in a Slice is 0.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -623,8 +623,9 @@ the BAM file header\tabularnewline
 RL & encoding\texttt{<}int\texttt{>} & read lengths & read lengths\tabularnewline
 \hline
 AP & encoding\texttt{<}int\texttt{>} & in-seq positions & if \textbf{APDelta} = true: 0-based alignment start
-delta from previous record, resetting to 0 for the first record in the slice\linebreak{}
-alignment start position if \textbf{APDelta} = false\tabularnewline
+delta from the previous record.  When the record is the first in the slice,
+its alignment start will be equal to that of the slice, so its alignment delta is 0.\linebreak{}
+if \textbf{APDelta} = false: encodes the alignment start position directly\tabularnewline
 \hline
 RG & encoding\texttt{<}int\texttt{>} & read groups & read groups. Special value 
 `-1' stands for no group.\tabularnewline
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -1093,7 +1093,7 @@ int & RG & read group & the read group identifier expressed as the N\textsupersc
   \State $reference\_id\gets slice\_header.reference\_sequence\_id$
 \EndIf
 \State $read\_length \gets$ \Call{ReadItem}{RL, Integer}
-\If{$container\_pmap.AP\_delta \ne 0$}
+\If{$container\_pmap.AP\_delta \ne 0$ \textbf{and} $slice\_header.reference\_sequence\_id \geq 0$}
     \State Initialise $last\_position$ to $slice\_header.alignment\_start$ if first record in slice
     \State $alignment\_position \gets$ \Call{ReadItem}{AP, Integer} + $last\_position$
     \State $last\_position \gets alignment\_position$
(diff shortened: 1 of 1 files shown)


jmthibault79:
Requesting review from @jkbonfield and @cmnbroad

jkbonfield:
Good spot.  The C code does the same so this was just a mistake in my spec pseudocode.

jkbonfield:
Note this will be zero on a multi-ref slice, but that's fine - it's still the value held in the slice header alignment start field.  It does indicate why this field MUST be zero though and isn't just something to be ignored.

jmthibault79:
Oh no, now you've got me thinking again.  :)  Should we delta-encode multi-ref?  Recalling that we only delta-encode when we're coordinate-sorted, some questions that some to mind are:
*  what do we do with the first record of a new reference?  e.g. `2:10` will have a very large negative delta from `1:20,000,000`.  Maybe a single large outlier is OK, but it's going to reduce the efficiency for some of the encodings.
* what about unmapped-unplaced reads?  Those are fine in themselves since they don't have relevant starts, but anything following these won't be able to delta-encode.  We'll need to ensure that nothing in the Slice comes after these.

Some older thoughts, maybe now outdated: https://github.com/samtools/htsjdk/issues/1218

jkbonfield:
We unset AP delta for non-position sorted data, as delta a random bunch of positions is larger than the original positions.

I'd forgotten about multi-ref and non-delta encoding though.  I'd just assumed it did delta (starting at zero), but I see your point it sets AP delta to off.  It's not tragically inefficient if we're encoding say 10 references into a single 10k read slice, but it is if we're using it to encode unsorted data.  The choice should be up to the encoder though rather than forcing it.  It's an interesting question though - does htslib (and io_lib) trigger multi-ref too early?  The trade off between improved compression ratios on larger blocks of data and fewer container/slice overheads is maybe lost by the larger size of the AP data series.

Changing this is a spec change though, albeit backwards compatible.  Maybe something for the CRAM 4 wish list.  For now we just have to define how it works in the existing version.

jmthibault79:
Sorry, I just realized that I overlooked something: the spec requires non-delta-encoding for multiref.  My comments above were based on the assumption that that multiref delta-encoding was something that we needed to handle.  My apologies if you interpreted my question as a request to enable this.

I added a commit which makes this explicit.

CRAMv3.tex
```diff
@@ -1093,8 +1093,8 @@ \subsection{\textbf{CRAM positional data}}
   \State $reference\_id\gets slice\_header.reference\_sequence\_id$
 \EndIf
 \State $read\_length \gets$ \Call{ReadItem}{RL, Integer}
-\If{$container\_pmap.AP\_delta \ne 0$}
-    \State Initialise $last\_position$ to 0 if first record in slice
+\If{$container\_pmap.AP\_delta \ne 0$ \textbf{and} $slice\_header.reference\_sequence\_id \geq 0$}
+    \State Initialise $last\_position$ to $slice\_header.alignment\_start$ if first record in slice
```
cmnbroad: Might as well use pseudocode, otherwise LGTM:
```suggestion
    \If{$first\_record\_in\_slice$}
        \State $last\_position\gets$ $slice\_header.alignment\_start$
    \EndIf
```

jkbonfield:
I'm happy with the change to the pseudocode and good spot on the bug of 0 vs ref start.  However the AP description in the table above still needs correcting to use slice header start pos instead of 0.

jmthibault79:
The first record in a (coord-sorted single-ref) slice will have Alignment Start equal to the slice, so it's correct that its delta will be 0.

However: it would be useful to have a reminder here that this will be literal Alignment Start when APDelta = false.  So I'll add something like that.

jkbonfield:
Ah I see what you mean, I didn't read it carefully enough although maybe that is an issue with the wording.

When you say it's a delta to the previous read, resetting to zero for the first, I took it to mean delta vs zero, not that delta is zero (that's kind of the obvious case).  Maybe it should be explicit and say delta to the previous read, or the alignment start if there is no previous read?

jmthibault79:
ok, how about this text?

jkbonfield:
Looks good to me.

PR #389, opened 2019-02-28
Remove vanilla LaTeX *.sty files from the repository

jmarshall:
Instead add note to _MAINTAINERS.md_ describing which packages to add to incomplete TeX installations to provide the required styles (macOS users generally use a full TeX installation via MacTeX so no bespoke details are needed for them) and/or where to put ad-hoc style files under $HOME so LaTeX will find them.


jmarshall:
Thanks, merged. Re TeX paths, `kpse…` is a shortened version of kpathsea — neither of which I've ever been able to remember and have to google for every time I want to do something with the paths! From an [interview with the author](http://tug.org/interviews/berry.html):

> _Do I presume correctly that Kpathsea stands for “Karl's path searching?”_
>
> Yes, I'm afraid so. I didn't want to call it anything general like just “pathsearch”, and couldn't think of a better name. If I had it to do over again, I'd probably call it tpathsea (TeX Path Searching) or something else, but it's too late. Olaf Weber has done a fine job maintaining Kpathsea and Web2c for quite a few years now, so it should really be okpathsea. Which is a good description, too, come to think of it — it's not perfect, it certainly has its problems, but overall, I think it's ok.

PR #397, opened 2019-03-28
Expand CRAM index section.

### LINKED ISSUE #376: CRAM: indexing for multiple reference slices
We should clarify that the CRAI and BAI indices should be ordered by reference sequence and alignment, with the Unmapped Slice last.

But what should we do about multiple-reference slices?  I'm picturing two scenarios here:
1. coordinate-sorted CRAMs with small ordered sub-slices combined into multi-ref Slices/Containers
2. non-coordinate-sorted CRAMs, in the pathological case consisting entirely of multi-ref Slices/Containers, each containing all references

We can handle case 1, by splitting multi-ref slice index entries into their constituent references.  (HTSJDK attempts to do this, but my understanding is that bugs in this approach exist currently.)

Case 2 seems hopeless to me.  Would it appropriate to state in the spec that indexes are explicitly not supported when reads are not coordinate-sorted?

jkbonfield:
Fixes #376

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -19,6 +19,7 @@
 \usepackage{amsmath}
 \usepackage{algpseudocode}
 \usepackage{threeparttable}
+\usepackage{tikz}
 \renewcommand{\headrulewidth}{0pt}
 \renewcommand{\footrulewidth}{0pt}
 
@@ -1533,6 +1534,8 @@ slices.
 
 \subsubsection*{General notes}
 
+Indexing is only valid on coordinate (reference ID and then leftmost position) sorted files.
+
 Please note that CRAM indexing is external to the file format itself and may change 
 independently of the file format specification in the future. For example, a new 
 type of index file may appear.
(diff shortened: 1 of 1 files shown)


CRAMv3.tex
```diff
@@ -1539,31 +1542,80 @@ \subsubsection*{General notes}
...
-\item Slice bytes
+\item Slice size in bytes (including slice header and all blocks).
 \end{enumerate}
 
-Each line represents a slice in the CRAM file. Please note that all slices must 
-be listed in the index file.
+Each line represents a slice in the CRAM file.
+Please note that all slices must be listed in the index file.
+
+Multi-reference slices must may need to have multiple lines for the same slice; one for each reference contained within that slice.
```
jmthibault79: rm "must"

CRAMv3.tex
```diff
@@ -1539,31 +1542,80 @@ \subsubsection*{General notes}
...
+\node [below] at (4,-2.3) {Slice offset 3};
+
+\draw[<-] (0,-0.1) -- (0,-1) node[below, text width=1.5cm]{Container absolute offset};
+
+\draw[|<->] (5.5,1.8) -- (7.5,1.8);
+\draw[|<->] (7.5,1.8) -- (9.5,1.8);
+\draw[|<->|] (9.5,1.8) -- (11.5,1.8);
+\node [above] at (6.5,1.8) {Size 1};
+\node [above] at (8.5,1.8) {Size 2};
+\node [above] at (10.5,1.8){Size 3};
```
cmnbroad: Nice!

CRAMv3.tex
```diff
@@ -1529,6 +1530,8 @@ \section{\textbf{Indexing}}
 
 \subsubsection*{General notes}
 
+Indexing is only valid on coordinate (reference and position) sorted files.
```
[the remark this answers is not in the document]
jkbonfield: Agreed we should be consistent here with SAM wording.  Better to just copy the explanation instead of linking to an external reference.

CRAMv3.tex
```diff
@@ -1539,31 +1542,80 @@ \subsubsection*{General notes}
...
-Each line represents a slice in the CRAM file. Please note that all slices must 
-be listed in the index file.
+Each line represents a slice in the CRAM file.
+Please note that all slices must be listed in the index file.
+
+Multi-reference slices may need to have multiple lines for the same slice; one for each reference contained within that slice.
+In this case the index reference sequence ID will be the actual reference number (from the ``RI'' data series) and not -2.
+
+Slices containing solely unmapped unplaced data (reference ID -1) still require values for all columns, although the alignment start and span will be ignored.
+It is recommended that they are set to small constant values (0 or 1).
```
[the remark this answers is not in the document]
jkbonfield: Rather embarrassingly I put 0 or 1 because my own code uses both. ;-)

It doesn't actually matter as we define them to be ignored on read, but a constant single byte value keeps the index small and compressible.  I'll just change it to zero I think and at some point fix my code.

CRAMv3.tex
jkbonfield: It's my first time exploring pgf/tikz.  I already see I can write this in a friendlier fashion, but it's a good start and I like the ability to version control our diagrams.

jmarshall: Yes indeed :smile:

Trivially, perhaps wrap `\begin{center}` around it? That appears to be the way to centre a `tikzpicture`…

jkbonfield:
Updated following review, thanks.

I added an extra commit to adjust the one remaining use of "Container structure" as it seems pointless to need a separate PR for this, although when squashing we should probably retain that as its own commit.

CRAMv3.tex
jkbonfield: Now centred, and rewritten the pic layout.

Useful resources I found:

http://www-groups.mcs.st-andrews.ac.uk/~alanc/pub/c_tikzref/c_tikzref.pdf

https://stuff.mit.edu/afs/athena/contrib/tex-contrib/beamer/pgf-1.01/doc/generic/pgf/version-for-tex4ht/en/pgfmanualse8.html

PR #398, opened 2019-03-29
Clarify container header fields.  Resolves #396

### LINKED ISSUE #396: Clarify use of "blocks" in the Container Header fields
Containers consist of a Container Header and one or more Blocks.  There are three scenarios for the division of Blocks in a Container:
1. A BAM Header Block only, of Block Content Type **_FILE_HEADER_**.
2. A **_COMPRESSION_HEADER_** Block, plus one or more Slices, which each contain a **_SLICE_HEADER_** Block, a **_CORE_DATA_** Block, and any number of **_EXTERNAL_DATA_** Blocks.
3. A **_COMPRESSION_HEADER_** Block only, to indicate EOF.

Update the Container Header description to be explicit about how to derive the values for the following Container Header fields:

**_length_**: Does this refer strictly to Core and External Data Blocks, or does it include everything beyond the Container Header?  (i.e. does it include the File Header and/or the Compression Header lengths, when applicable?)

**_number of blocks_**: same question

**_landmarks_**: Add text indicating that this refers strictly to Slice start boundaries, and no other types of blocks.

jmthibault79:


jmarshall:
(As a reminder: please create PR branches in a fork of the hts-specs repository — see [MAINTAINERS.md](https://github.com/samtools/hts-specs/blob/master/MAINTAINERS.md). But this time it was a useful test to highlight #399 :smile:)

jmthibault79:
Oops, sorry about that.

CRAMv3.tex
```diff
@@ -458,10 +459,11 @@ \section{\textbf{Container structure}}
...
 ltf8 & bases & number of read bases\tabularnewline
 \hline
-itf8 & number of blocks & the number of blocks\tabularnewline
+itf8 & number of blocks & the total number of blocks in this container\tabularnewline
 \hline
-itf8[ ] & landmarks & Each integer value of this array is a byte offset into the 
-blocks byte array. Landmarks are used for random access indexing.\tabularnewline
+itf8[ ] & landmarks & the locations of slices in this container as byte offsets from the end of 
+this container header, used for random access indexing.
+The landmark count must equal the slice count.\tabularnewline
```
cmnbroad: Wouldn't this definition imply that the first landmark entry would always be 0, since the first slice header would always be immediately after the container header ? That doesn't seem to match the cram_dump output that @jkbonfield put in the comment in https://github.com/samtools/hts-specs/issues/396, nor what I see in htsjdk code:
```
final int containerHeaderSize = landmarks[0];
final int containerTotalByteSize = containerHeaderSize + containerByteSize;
```
Not sure which is incorrect, but the code seems to count the header size in the landmark offset, as if it were an offset from the beginning of the header/container, not the end of the header ?

jmthibault79: landmarks[0] = compression header length.  We could add a note here saying so.

I can make a quick PR to name this value more precisely in htsjdk.

jmthibault79: Also, these states are equivalent for containers:
* no landmarks = no slices = no compression header

cmnbroad: Ah I see - I wasn't properly distinguishing compression header from container header. I was thinking of them as one and a single header, but that doesn't match the spec treatment. I think your suggestion of noting that here is a good one, and would make that explicit.

jmthibault79: Thanks for the new picture.  Yes, it has helped to clarify this confusion.

I updated the text here in a new commit.

jkbonfield: I got muddled with which this PR was (too many related ones), sorry, but yes the picture elsewhere clarifies things there and your explicit comment regarding landmark[0] being size of the compression header helps a lot.

jkbonfield:
> Changed PDFs as of [bd2c42c](https://github.com/samtools/hts-specs/commit/bd2c42c525c2ef64e28686c933770585f45ee969): [CRAMv3](https://circleci.com/gh/samtools/hts-specs/211/artifacts/0//root/project/pdfs/CRAMv3.pdf) ([diff](https://circleci.com/gh/samtools/hts-specs/211/artifacts/0//root/project/diff/CRAMv3.pdf)).

The diffs artifact doesn't seem to contain the contents of this commit.  It's just one page long. @yfarjoun

CRAMv3.tex
```diff
@@ -440,7 +440,8 @@ \section{\textbf{Container structure}}
 \textbf{Data type} & \textbf{Name} & \textbf{Value}
 \tabularnewline
 \hline
-int32 & length & byte size of the container data (blocks)\tabularnewline
+int32 & length & the sum of the byte lengths of all blocks in this container;
```
jkbonfield: "byte lengths of all blocks" may be ambiguous.  Are we talking the size of all data[] parts of blocks, or also including the block structure itself?  We know it's the latter, but it may not be clear from this wording given "byte lengths" and the fact that blocks have a "byte[]" array in their structure.

To be totally unambiguous, maybe use "the sum of the lengths of all blocks (both structures and data) in this container".

jmarshall:
> The diffs artifact doesn't seem to contain the contents of this commit. It's just one page long.

Yep, it's the wrong one page. Yossi's been looking into that — cf #400 and the issue linked from there.

cmnbroad:
One additional issue, though perhaps for another PR. Otherwise this looks good to me.

CRAMv3.tex
```diff
@@ -440,7 +440,8 @@ \section{\textbf{Container structure}}
 \textbf{Data type} & \textbf{Name} & \textbf{Value}
 \tabularnewline
 \hline
-int32 & length & byte size of the container data (blocks)\tabularnewline
+int32 & length & the sum of the lengths of all blocks in this container (headers and data); 
+equal to the total byte length of the container minus this header structure\tabularnewline
```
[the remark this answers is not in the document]
jmthibault79: I agree that it's weird that `blocks[]` is in the header.  But I'd prefer a separate PR.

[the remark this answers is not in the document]
jkbonfield: Agreed.  I think container struct may have been my error in explanation here.  If I recall the spec uses container header normally.  Indeed we should go through it all and check we have the same term.  Good spot.

PR #400, opened 2019-03-30
Yf make diff work on long input

yfarjoun:
testing using --changes-only as per discussion in https://github.com/ftilmann/latexdiff/issues/169

(used commit from another PR to test...)


yfarjoun:
hmmm. this is better (works on the cram commit) but fails in an obscure way on the longtable example. 

@jmarshall can you see if you understand the log messages? I'm failing to reproduce the error locally....

jmarshall:
The operative parts of this have since been applied as PR #459.

PR #406, opened 2019-05-07
Adds crypt4gh (GA4GH encrypted file format proposal)

daviesrob:
crypt4gh.tex describes a format for encrypted files which allows for both streaming and random access.

It currently has 'preliminary' stamped across it.  This is because it may still be necessary to make breaking changes following security reviews etc.  Be warned!

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -41,6 +41,11 @@ Past VCF/BCF maintainers include Cristina Yenyxe Gonzalez Garcia, Ryan Poplin, a
 
 Past refget maintainers include Matt Laird.
 
+### crypt4gh
+
+* Alexander Senf (@AlexanderSenf)
+* Robert Davies (@daviesrob)
+
 [ga4gh-ff]:  https://www.ga4gh.org/howwework/workstreams/#lsg
 
 
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ PDFS =	BCFv1_qref.pdf \
 	BCFv2_qref.pdf \
 	CRAMv2.1.pdf \
 	CRAMv3.pdf \
+	crypt4gh.pdf \
 	CSIv1.pdf \
 	SAMv1.pdf \
 	SAMtags.pdf \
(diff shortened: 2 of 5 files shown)


yfarjoun:
There's an undefined reference: data:chacha20_encryption

jmarshall:
**References** appears in the TOC twice. [Looks like](https://en.wikibooks.org/wiki/LaTeX/Bibliography_Management#Other_methods) you want either `\addcontentsline` or `\renewcommand{\refname}{…}` but not both.

jmarshall:
Eventually, to appear on the website, this will want to be added to _README.md_ and _index.md_, and appropriate maintainers should be listed in _MAINTAINERS.md_. See those files in f16fedfa64bd57ad7317a815462b34dcf7763ad6 for what was needed for the last spec to be added.

daviesrob:
> There's an undefined reference: data:chacha20_encryption

The ref is defined [at this label](https://github.com/samtools/hts-specs/blob/eb6ce3c/crypt4gh.tex#L418).  It works in my checkout, and I can follow the link in the generated pdf.  Maybe latex needs to be run again to pick it up?

> **References** appears in the TOC twice.

Removing `\addcontentsline` seems to give the desired result.  I'll push a fix.

daviesrob:
I think the undefined reference is a red herring.  It looks like `latexdiff-vc` produces a zero-length pdf when the .tex file it's diffing doesn't exist in the old revision.  [It actually patches the file all the way back to /dev/null, and then decides to put the current version back](https://github.com/ftilmann/latexdiff/blob/e8da103/latexdiff-vc#L451).  The result is a diff of the file against itself, so no changes.  If you ask it to diff a revision where the file does exist then it works as expected (as long as something changed).

Possibly the Makefile needs something to deal with the not-in-old-revision case?

jmarshall:
I've added a commit (to be dropped before merging!) to get @hts-specs-bot to comment with PDFs regardless of the CircleCI diff problem. (If we're adding another new LaTeX document in future, we should commit a placeholder _foo.tex_ to **master** first and then base a PR off that, so this works better next time.)

daviesrob:
Thanks, @jmarshall  - a brute-force approach but it works :smile:

daviesrob:
Squashed and rebased ready for merge.  Unfortunately this will probably have broken circle-ci because the hack to make it work has been removed.  Final merge should be OK, though.

After merge the PDFs will need to be rebuilt, which I should be able to do as long as there's no major objection.

PR #409, opened 2019-05-13
Allow C and Java native text spellings of NaN and infinities

jmarshall:
The VCF v4.3 spec currently says of Float fields that they are:

> 32-bit, formatted to match the regular expression `^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$`, `NaN`, or `+/-Inf`.

It is not explicit whether `NaN` and `Inf` are intended to be case sensitive, but as they are adjacent to a regexp saying `…[eE]…` the reader's best guess is probably that they are indeed intended to specify that exact capitalisation.

C's `printf` outputs `+`/`-`/no sign as appropriate, followed by `nan` and `inf`/`infinity` (with `%f` etc) or `NAN` and `INF`/`INFINITY` (with `%F` etc); there is no programmer control over whether infinity is 3 or 8 letters, but current glibc and BSD libc output it as 3 letters. Its `scanf` and `strtod` accept all of those completely case-insensitively.

Java's [`Double.toString`](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#toString-double-) outputs `NaN`, `Infinity`, or `-Infinity` signed, capitalised, and spelt exactly thus. Its [`Double.valueOf`](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#valueOf-java.lang.String-) accepts an optional `+`/`-` sign followed by `NaN` or `Infinity` capitalised and spelt exactly thus.

Thus the VCF specification's mixed capitalisation is impossible to output using C's builtin functions, and its `+/-Inf` abbreviation is impossible to parse or output using Java's builtin functions.

In practice, htslib/bcftools outputs `nan` (samtools/bcftools#755) and htsjdk/picard outputs `Infinity` — regardless of what the spec says.

This PR fixes that by changing the VCF spec to allow the union of C and Java representations, i.e., allowing what `scanf` accepts, namely `{+,-,}{NAN,INF,INFINITY}` case-insensitively.

With this change, both input and output can be done with C's native functions and similarly Java can output Floats with `Double.toString`. Java input will still need special case code to parse otherwise-cased and abbreviated NaNs and infinities, but that's inevitable given `Double.valueOf`'s inflexibility.

See also previous discussions in https://github.com/samtools/bcftools/issues/755#issuecomment-455240246 and https://github.com/samtools/hts-specs/issues/89#issuecomment-455503658.

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -94,7 +94,7 @@ Characters with special meaning (such as field delimiters `\verb|;|' in INFO or 
 
 
 \subsection{Data types}
-Data types supported by VCF are: Integer (32-bit, signed), Float (32-bit, formatted to match the regular expression \verb|^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$|, \texttt{NaN}, or \texttt{+/-Inf}), Flag, Character, and String.
+Data types supported by VCF are: Integer (32-bit, signed), Float (32-bit IEEE-754, formatted to match one of the regular expressions \verb|^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$| or \verb"^[-+]?(INF|INFINITY|NAN)$" case insensitively), Flag, Character, and String.
 For the Integer type, the values from $-2^{31}$ to $-2^{31}+7$ cannot be stored in the binary version and therefore are disallowed in both VCF and BCF, see \ref{BcfTypeEncoding}.
 
 \subsection{Meta-information lines}
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -94,7 +94,9 @@ Characters with special meaning (such as field delimiters `\verb|;|' in INFO or 
 
 
 \subsection{Data types}
-Data types supported by VCF are: Integer (32-bit, signed), Float (32-bit IEEE-754, formatted to match one of the regular expressions \verb|^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$| or \verb"^[-+]?(INF|INFINITY|NAN)$" case insensitively), Flag, Character, and String.
+Data types supported by VCF are: Integer (32-bit, signed), Float (32-bit IEEE-754, formatted to match one of the regular expressions \verb|^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$| or \verb"^[-+]?(INF|INFINITY|NAN)$" case insensitively),%
+\footnote{Note Java's {\tt Double.valueOf} is particular about capitalisation, so additional code is needed to parse all VCF infinite/NaN values.}
+Flag, Character, and String.
 For the Integer type, the values from $-2^{31}$ to $-2^{31}+7$ cannot be stored in the binary version and therefore are disallowed in both VCF and BCF, see \ref{BcfTypeEncoding}.
 
 \subsection{Meta-information lines}


jmarshall:
Incidentally, C also accepts `NAN([0-9a-zA-Z_]…)` to encode different NaN values in some implementation-defined way. The current dominant library implementations do not output that, and this PR does not propose to add such notation to VCF.

pd3:
+1

cyenyxe:
> @jmarshall  This makes sense. Actually fixing the issue in the java implementation is a nasty can of worms though, since it makes use of libraries that unsurprisingly, use `Double.valueOf` and it's complicated to work around that.

Given this comment, it could be a good idea to briefly mention that native support for these values varies among programming languages.

cyenyxe:
@pd3 @lbergelson could you please review this again after the latest changes?

lbergelson:
The formatting in the pdf is a bit strange now.  The regex runs off into the margin and almost off the page.  I'm not sure how to fix it though, my latex is weak.  @yfarjoun you're a latex master aren't you?

image

lbergelson:
The text seems fine to me though.

VCFv4.3.tex
```diff
@@ -94,8 +94,17 @@ \subsection{Character encoding, non-printable characters and characters with spe
 
 
 \subsection{Data types}
-Data types supported by VCF are: Integer (32-bit, signed), Float (32-bit, formatted to match the regular expression \verb|^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$|, \texttt{NaN}, or \texttt{+/-Inf}), Flag, Character, and String.
-For the Integer type, the values from $-2^{31}$ to $-2^{31}+7$ cannot be stored in the binary version and therefore are disallowed in both VCF and BCF, see \ref{BcfTypeEncoding}.
+The data types supported by VCF are:
+
+\begin{itemize}
+  \item Integer (32-bit, signed): Values from $-2^{31}$ to $-2^{31}+7$ cannot be stored in the binary version and therefore are disallowed in both VCF and BCF, see \ref{BcfTypeEncoding}.
+  \item Float (32-bit IEEE-754): Formatted to match one of the regular expressions \verb|^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$| or \verb"^[-+]?(INF|INFINITY|NAN)$" case insensitively.
```
yfarjoun:
```suggestion
  \item Float (32-bit IEEE-754): Formatted to match one of the regular expressions  \verb"^[-+]?(INF|INFINITY|NAN)$" or \verb|^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$| case insensitively.
```
(so that the text doesn't fall out of the margin.)

jmarshall:
I'm not sure that programming language foibles need to be mentioned; but it can be useful and there is precedent in the SAM spec (which talks about Java GZIPInputStream's (historical?) problems).

I see you've now pushed to this branch, so I've pushed such a footnote to [nan-inf-footnote](https://github.com/jmarshall/hts-specs/commit/f6749fc61ad27fa1ca5b1f536fbd561a00608327).

VCFv4.3.tex
jmarshall: Numeric is the common case and ∞/NaN is the corner case, so swapping them around would not be ideal exposition. (And “case insensitively” pertains to the ∞/NaN regex in particular.)

lbergelson: We could join them into one regular expression and put that on it's own line.

jmarshall: I tried that, but considered that it was more legible as two separate regexps. YMMV.

The real problem here is that a monster regexp is far from an ideal way to say “floating point number text goes here” — see also https://github.com/samtools/hts-specs/issues/89#issuecomment-111417134.

jmarshall:
As no-one else has done anything about fixing up @cyenyxe's reformatting of this paragraph, I have rebased everything onto current **master** and replaced the PR branch with my commit adding the requested footnote (as mentioned in https://github.com/samtools/hts-specs/pull/409#issuecomment-514821980).

This does everything that has been requested (ie adds a footnote about the Java API function) while maintaining the existing single-paragraph formatting, and has no formatting issues. Can this be re-reviewed and merged, please?

Afterwards, if someone wants to revisit @cyenyxe's conversion of the paragraph to an itemised list of data types, it's available in this [jmarshall/nan-inf-cyenyxe branch](https://github.com/jmarshall/hts-specs/commit/57f2d826b2557dee614d579d67225c19974baa35).

PR #410, opened 2019-05-19
Fixes refget-openapi.yaml

reece:
Corrects three issues:

1. Missing quote in `service-info` path led to unparsable yaml
1. Metadata component `required` block was outdented one level, causing it to not apply to the metadata properties
1. The schema disagreed with the spec in a metadata attribute. It used `circular_locations`, whereas the spec uses `circular_supported`

Also, I changed the servers block to remove details for a particular implementation.


andrewyatz:
Thank you Reece. All of this looks perfect.

PR #411, opened 2019-05-21
Trivial VCF formatting improvements

jmarshall:
Avoid end-of-sentence spacing after i.e. / e.g. / cf. Where i.e. or e.g. delimits separate clauses, use a comma to set it off, e.g., as the previous comma does in this sentence.

Previously (PR #391) I did this only in VCFv4.3 (and neglected `cf.`). This extends this minor improvement to the other two documents.

No changes to the format or descriptions, so this should be straightforward to review and merge.


jmarshall:
@cyenyxe @lbergelson @pd3 — It would be great if one of you would glance at and merge this trivial formatting change. I have other PRs that I would like to rebase on top of this…

Issue #408, opened 2019-05-13
CRAM spec read feature clarifications


cmnbroad:
A few somewhat nitty questions on the CRAM spec (section 10.6, Read Features) based on some code cleanup I'm doing in htsjdk. I can make a spec PR based on the results of the discussion.

From the part under "Base substitution codes (BS data series)":

> There are 5 possible bases ACGTN, 4 possible substitutions for each base and 20 substitutions in total.

That seems to imply that only upper case differences are represented as substitutions, and that source case is otherwise preserved (?) However, htsjdk has code that handles `Substitution` features that are case transitions, but only on read; it only *writes* `Substitution` read features in the case where the reference and read base are different, but both upper case. Otherwise if they're different it generates a `ReadBase` feature. `ReadBase` features are referenced in the spec, but not spelled out anywhere. Is the intent that preservation of base case is prescribed by the spec, or is that left to implementation policy ? Is it required that `ReadBase` features are used to represent case differences, and/or are case `Substitution`s legal ?

> Same-frequency ties are broken using alphabetical order

The htsjdk implementation breaks ties using the order 'ACGTN', which isn't alphabetical order. Although I don't think it matters since it doesn't affect the file format itself.

> The whole substitution matrix is written as 5 bytes, one for each reference base in the alphabetical order: A, C, G, T and N

Not alphabetical...

> Note: the last two bits of each substitution code are redundant but still required to simplify the reading. 

Not sure how to interpret this ?

jkbonfield:
Regarding case, I doubt it was intended to consider case as important as this was originally aimed to replace BAM which is already ignoring case (it munges it all to uppercase).  The decision was before my time though.  My input here was simply to use a lookup table instead of the base itself.

If we wish to consider a case-preserving CRAM then we'll either need to expand on this further as punt it to CRAMv4 or spell out explicitly that all base case changes need to be encoded as a read base feature (B).   My gut feeling is to clarify that whether an implementation wishes to consider case changes as a difference that needs encoding as a feature is implementation dependent - we can state that explicitly in a footnote.  FWIW, htslib treats c and C as equal.

I don't think read base needs explicitly spelling out - it's a base.  We could clarify by saying it has to be upper case, or maybe it has to be one of the recognised IUPAC ambiguity codes, etc, but I think it's basically the byte that cases from the input source data and anything beyond that is implementation specific checking.

[...]

PR #412, opened 2019-05-28
Update CRAM spec section on substitution matrix and codes.

cmnbroad:
Based on the discussion in https://github.com/samtools/hts-specs/issues/408. @jkbonfield I left the section on substitution frequency sorting in, since it motivates the need for the substitution matrix (otherwise the codes could just be fixed). See what you think.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -1389,8 +1389,9 @@ in the substitution matrix.
 
 Each of the 4 possible substitutions for a given reference base is assigned a 2-bit integer
 code (see below) with a value ranging from 0 to 3 inclusive. The 4 2-bit codes are packed
-into a single byte,  in the fixed order ACGTN, and the entire substitution matrix is
-written as 5 such bytes, one for each reference base, also in the order ACGTN.
+into a single byte, high 2-bits first, for each base ACGTN (minus the reference base itself).
+The entire substitution matrix is written as 5 such bytes, one for each reference base, also
+in the order ACGTN.
 
 \subsubsection*{Substitution Code Assignment}
 
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -1395,11 +1395,11 @@ in the order ACGTN.
 
 \subsubsection*{Substitution Code Assignment}
 
-To assign the susbtitution code for a given reference base/read base, the subsitutions for
+To assign the susbtitution code for a given reference base/read base, the substitutions for
 each reference base may optionally be sorted by their frequencies, in descending order, with
 same-frequency ties broken using the fixed order ACGTN. Although sorting by substitution
 frequency is not required by the CRAM format, assigning substitution codes based on frequency
-maximizes compression by ensuring that the most frequent subsitutions use the shortest possible
+maximizes compression by ensuring that the most frequent substitutions use the shortest possible
 codes.
 
 For example, let us assume the following substitution frequencies for base A: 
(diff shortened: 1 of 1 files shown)


CRAMv3.tex
```diff
@@ -1378,19 +1378,30 @@ \subsubsection*{Read feature codes}
...
-A base substitution matrix assigns integer codes to all possible substitutions. 
+\subsubsection*{Substitution Matrix Format}
 
-Substitution matrix is written as follows. Substitutions for a given reference 
-base are sorted by their frequencies in descending order then assigned numbers 
-from 0 to 3. Same-frequency ties are broken using alphabetical order. For example, 
-let us assume the following substitution frequencies for base A: 
+Each of the 4 possible substitutions for a given reference base is assigned a 2-bit integer
+code (see below) with a value ranging from 0 to 3 inclusive. The 4 2-bit codes are packed
+into a single byte,  in the fixed order ACGTN, and the entire substitution matrix is
```
jkbonfield: This is a little confusing as it could be implied that our 2-bit codes are packed together in the order ACGTN (although that would be 5 2-bit codes and not 4).  it doesn't particularly spell out that this is ACGTN minus the specific reference base being encoded. 

"Each of the 4 possible substitutions for a given reference base is assigned a 2-bit integer code (see below) with a value ranging from 0 to 3 inclusive. The 4 2-bit codes are packed into a single byte, high 2-bits first, for each base ACGTN (minus the reference base itself).   The entire substitution matrix is written as 5 such bytes, one for each reference base, also in the order ACGTN."

CRAMv3.tex
```diff
@@ -1410,12 +1421,11 @@ \subsubsection*{Base substitution codes (BS data series)}
 
 AN: 3
 
-and they are written as a single byte, 10 01 00 11 = 147 decimal or 0x93 in this 
-case. The whole substitution matrix is written as 5 bytes, one for each reference 
-base in the alphabetical order: A, C, G, T and N.
+The substitution matrix entry for this reference base then is written as a single byte, with the
```
jkbonfield: Maybe "The first byte of the substitution matrix entry for reference base A is written as a single byte, ..." and "This will then be followed by 4 more bytes representing substitutions for reference base C, G, T and N."

CRAMv3.tex
```diff
@@ -1410,12 +1421,11 @@ \subsubsection*{Base substitution codes (BS data series)}
...
+codes in the order CGTN: 10 01 00 11 = 147 decimal, or 0x93 in this case.
 
-Note: the last two bits of each substitution code are redundant but still required 
-to simplify the reading. 
+Note that the last two bits of the substitution matrix entry are redundant, since the remaining
```
jkbonfield: Maybe we should just remove this sentence.  It really doesn't add anything and just raises questions. :)

If we keep it, then it should state that the last two bits of each substitution byte are redundant, rather than imply just the last 2 bits of the entire thing.

CRAMv3.tex
cmnbroad: Yeah, done.

jkbonfield:
Thanks for the updates.  There are a couple outstanding comments here that haven't been answered, but they're not show-stoppers and I can live with this as-is (it's a big improvement already) so I'll leave it to @cmnbroad to decide whether he agrees with my suggestions or not.

So merge if you're done (or ask me if you're not comfortable doing that).  Otherwise, prod me when you're ready for reviewing again.

CRAMv3.tex
cmnbroad: Yeah, done.

CRAMv3.tex
cmnbroad: Also done (removed this line).

cmnbroad:
@jkbonfield Not sure why I didn't respond to those comments before, but I think I got them all now, so please have another look when you can.

jkbonfield:
Looks good to merge.  Thanks.

Issue #381, opened 2019-02-07
clarifications on vcf pedigree header


matthdsm:
Hi, 

Could someone clarify the section on pedigree headers in vcf? What are the requirements on field names?

More specific, how do you recreate a ped file representation in vcf? The spec is very vague about this.
For example, according to other issues, a family can be annotated as such:
```
##PEDIGREE=<ID=ChildID,Father=FatherID,Mother=MotherID>
```
or
```
##PEDIGREE=<ID=SampleID,Name_1=Ancestor_1,...,Name_N=Ancestor_N>
```
Is it possible to add some more details to the spec? Are the `Mother` and `Father` fields set or just arbitrary strings? How do you annotate siblings or even more complex relations? How do you add family id's?
Other ped fields such as affected status and ethnicity can be handled by the `##SAMPLE` header, so no issues there.

A suggestion could be to update the spec with fixed fieldnames for parents, and to add family id to the sample header or something?

Thank you for the input

ping @chapmanb, as this might be of interest to you too.

Thanks again
M

PR #413, opened 2019-05-29
Fix PEDIGREE example errors; draft of familial PEDIGREE section

jmarshall:
Two separate commits, the first fixing obvious errors and the second an outline of a suggested new “Describing family relationships” section:

1. Add missing `##` and other words; replicate 2016's PR #176's fixes in the other copies of the spec.

2. Outline of a new “Describing family relationships” section that would address #381 by explaining how the `##PEDIGREE` trio metadata line shown in the VCF spec is supposed to work and how it corresponds to an external PED file.

    Previous discussion has suggested that this functionality is thought to be stillborn. If so, the examples should be removed from the specs. Otherwise if the functionality is used out there, it would be good for the spec to describe the (fairly obvious) way in which it would be intended to be used.

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -253,7 +253,7 @@ It is possible to record relationships between genomes using the following synta
 ##pedigreeDB=URL
 \end{verbatim}
 
-\noindent See \ref{PedigreeInDetail} for details.
+\noindent See \ref{ClonalPedigree} (clonal relationships) and \ref{FamilialPedigree} (trios and families) for details.
 
 
 \subsection{Header line syntax}
@@ -1238,7 +1238,7 @@ However, a more evolved algorithm could attempt actually deconvolving the two ge
 \normalsize
 
 \subsubsection{Clonal derivation relationships}
-\label{PedigreeInDetail}
+\label{ClonalPedigree}
 In cancer, each VCF file represents several genomes from a patient, but one genome is special in that it represents the germline genome of the patient.
 This genome is contrasted to a second genome, the cancer tumor genome.
 In the simplest case the VCF file for a single patient contains only these two genomes.
@@ -1258,7 +1258,7 @@ The general format of a PEDIGREE line describing asexual, clonal derivation is:
 \end{verbatim}
 
 This line asserts that the DNA in genome DerivedID is asexually or clonally derived with mutations from the DNA in genome OriginalID.
-This is the asexual analog of the VCF format that has been proposed for family relationships between genomes, i.e., there is one entry per trio of the form:
+This is the asexual analog of the VCF format that has been proposed for family relationships between genomes (see~\ref{FamilialPedigree}), in which there is one entry per trio of the form:
 
 \begin{verbatim}
 ##PEDIGREE=<ID=ChildID,Mother=MotherID,Father=FatherID>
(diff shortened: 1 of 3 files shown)


VCFv4.3.tex
```diff
@@ -1380,6 +1380,19 @@ \subsection{Representing unspecified alleles and REF-only blocks (gVCF)}
 \end{flushleft}
 \normalsize
 
+\subsection{Describing family relationships}
```
cyenyxe: I would prefer to make this part of the PEDIGREE meta entry description. I can also address the TODO.

VCFv4.3.tex
```diff
@@ -253,7 +253,7 @@ \subsubsection{Pedigree field format}
 ##pedigreeDB=URL
```
cyenyxe: This is not [the line I'm interested in](https://github.com/samtools/hts-specs/pull/413/files#diff-6fd0045c2c7902d4181b2e8d315cc56dR249) but I can't comment on the other one. PED does not support more than 2 ancestors, do we want to do it in VCF? I have never seen this used so I don't think a lot of people will miss it if we drop it. That will make trivial to add some examples for trios.

VCFv4.3.tex
jmarshall: Pedigree lines are putatively used in two separate scenarios: clonal relationships and trios/families. There is already a separate section covering the clonal scenario, so it makes some sense to cover the other one in its own separate section too.

Similarly the minimal §1.4.8 META and SAMPLE meta entry description is expanded upon by §5.4.10.

cyenyxe: The only issue I see is that the clonal relationships are described in the context of breakends and are subsections of that one. On the other hand, this would be a completely separate one which would perfectly fit in the PEDIGREE section.

VCFv4.3.tex
cyenyxe: @pd3 @lbergelson what should we do about this? Dropping support for more than 2 ancestors would render some files incorrect, but as I said in my previous comment I have never seen that syntax being used.

cyenyxe:
If people are happy with the slight changes to the document structure, I will add an example about how do PEDIGREE entries compare against PED files.

VCFv4.3.tex
lbergelson: I don't think there would be a lot of pushback against dropping multiple ancestor lines because they seem pretty ambiguously defined at the moment and I haven't ever seen one used... That said, we should maybe not be making breaking changes to an existing spec?

Are they intended only for asexual ancestry where each is the parent of the next?  Or are they just an unsorted bag of ancestors that could represent any tree of parentage?  

i.e does `<ID=SampleID,Name_1=Ancestor_1,...,Name_N=Ancestor_N>`
imply
``` 
SampleID -> Ancestor_1 -> Ancestor 2 
```
or could it also mean
```
 Ancestor_1 <- SampleID -> Ancestor_2
```
I lean towards removing or deprecating it if we don't know exactly what it means and no one seems to be using it.

I'm also not clear on which of these are controlled vocabulary.  Is there a specific ontology of relationships that are allowed?  Are we allowed to specify something like `Sibling` in the case where we don't have parent in the vcf or is that handled with dummy trios that point to unique but not present parents ID's

Are you allowed to include only 1 parent or are trios required?  

I assume the example would address some of these questions.

jmmut:
I added some commits to add a minimal PED to VCF example. If people think that a complete example is preferable than a minimal one, I can add some corner cases. 

Also, note that I referenced other sections instead of putting a complete VCF with the phenotypes and genotypes. I think it's better not to repeat details in different sections for the risk of getting outdated, but in this case I can be convinced otherwise.

Finally, the third commit adds syntax to express sibling/twin relationships.

The only thing I'm not sure about is the generic ancestor line `<ID=SampleID,Name_1=Ancestor_1,...,Name_N=Ancestor_N>`. To me it looks like it means `Ancestor_1 <- SampleID -> Ancestor_2` but I agree with deprecating (or at least explaining better) because it's not clear enough.

jmarshall:
Rebased now that the first commit (typographical errors in examples) has been split off as PR #583 and merged to **master**.

Issue #236, opened 2017-08-30
The maximum BCF header size


pd3:
The BCF specification does not explicitly state what is the maximum header size, but htslib implementation assumes a 4 byte integer https://github.com/samtools/htslib/blob/develop/vcf.c#L909-L913

This is a problem for assemblies, with millions of contigs the header size can easily exceed the limit
(https://github.com/samtools/bcftools/issues/672).

It would be good explicitly state this limit in the specification. Also in the next version of BCF it might be worthwhile to reserve 8 bytes for the header length.

jmarshall:
Mostly a duplicate of #138.

cyenyxe:
In order to resolve this ticket and #138, could any of you please confirm that the only change required in the spec is to state that the maximum header size is 4 bytes? Thanks!

jmarshall:
To document the existing format, the spec needs to say the things that the quick reference document says: that there is a BCF\2\2 magic string (BCF\2\1 in the 4.1 and 4.2 specs) followed by a 4-byte `uint32_t` length followed by the `NUL`-terminated VCF header text. In particular, it needs to mention the length field and the NUL termination, neither of which it currently mentions. The best way to do this would be to copy the Field/magic/l_text/text rows from the table in the quick reference, so that this is described via a similar table to BCFv2_qref or BAM in SAMv1 §4.2.

The quick reference document (which is currently the only documentation of this outwith the implementations) says that the VCF header text is always NUL-terminated and that _l_text_ includes the NUL padding character. (This differs from BAM, where the NUL is optional.) In practice, HTSlib has always written a NUL terminator here and it appears HTSlib has been able to accept non-NUL-terminated BCF headers since v1.4 (see samtools/htslib@37c85e4b9737abc88871daf724552959b612e7cb). I haven't researched what other implementations require w.r.t. NUL termination. I would suggest the compatible route is to stick with the existing documentation from the qref and require the NUL terminator character.

Whether and how to extend this (perhaps in a future VCF version) to allow more than 4Gb of headers is a separate question.

pd3:
I added the "enhancement" label. This issue was not intended as a duplicate of #138, which only comments on the missing l_text field in the specification, but asks if the new version of BCF shouldn't extend this field to 8 bytes.

PR #417, opened 2019-06-06
Add the missing VCF header length and header NULL termination in the …

### LINKED ISSUE #138: Clarify BCF header format
VCFv4.3 §6.2 says

> The BCF2 header begins with the “BCF2 magic” 5 bytes that encode `BCFXY` where X and Y are bytes indicating the major number (currently 2) and the minor number (currently 2). This magic can be used to quickly examine the file to determine that it’s a BCF2 file. Immediately following the BCF2 magic is the standard VCF header lines in text format, beginning with `##fileformat=VCFvX.Y`.

However BCFv2_qref and the htslib implementation have a 32-bit length before the header text:

| Field | Description | Type | Value |
| --- | --- | --- | --- |
| magic | BCF2 magic string | char[5] | BCF\2\1 |
| ￼￼￼l_text | Length of the header text, including any NULL padding | uint32_t | ￼￼￼ |
| text | NULL-terminated plain VCF header text | char[l_text] |  |

The main VCF/BCF spec should describe the BCF header length and text more completely.

pd3:
…BCF specification. Resolves #138

### DIFF
--- a/VCFv4.1.tex
+++ b/VCFv4.1.tex
@@ -1070,10 +1070,10 @@ l\_indiv	uint32\_t	Data length of FORMAT and individual genotype fields
 CHROM	int32\_t	Given as an offset into the mandatory contig dictionary
 POS	int32\_t	0-based leftmost coordinate
 rlen	int32\_t	Length of the record as projected onto the reference sequence.  May be the actual length of the REF allele but for symbolic alleles should be the declared length respecting the END attribute
-n\_allele\_info	int32\_t	n\_info, where n\_allele is the number of REF+ALT alleles in this record, and n\_info is the number of VCF INFO fields present in this record
-n\_fmt\_sample	uint32\_t	n\_sample, where n\_fmt is the number of format fields for genotypes in this record, and n\_samples is the number of samples present in this sample.  Note that the number of samples must be equal to the number of samples in the header
 QUAL	float	Variant quality; 0x7F800001 for a missing value
-ID	typed string
+n\_allele\_info	 int32\_t	 The low 16 n\_allele bits represent the number of REF+ALT alleles in this record and the high 16 n\_info bits represent the number of VCF INFO fields in this record \\ \hline
+n\_fmt\_sample	 uint32\_t	 The low 24 n\_samples bits represent the number of samples in this record and the hight 8 n\_fmt bits represent the number of format fields for genotypes. Note that the number of samples must be equal to the number of samples in the header \\ \hline
+ID	       typed string  Variant identifier; 0x07 for a missing value \\ \hline
 REF+ALT	list of n\_allele typed strings	the first allele is REF (mandatory) followed by n\_alleles - 1 ALT
 alleles, all encoded as typed strings
 FILTER	Typed vector of integers	a vector of integer offsets into dictionary, one for each FILTER field value.  ``.'' is encoded as MISSING
@@ -1144,6 +1144,7 @@ The final four bits encodes an unsigned integer that indicates the type of the u
 \vspace{0.3cm}
 \begin{tabular}{|l | l | l|} \hline
 Lowest 4 bits & Hexadecimal encoding & Corresponding atomic type \\ \hline
+0 & 0x00 & A missing value without an explicit type provided \\ \hline
 1 & 0x?1 & Integer [8 bit] \\ \hline
 2 & 0x?2 & Integer [16 bit] \\ \hline
 3 & 0x?3 & Integer [32 bit] \\ \hline
@@ -1152,7 +1153,7 @@ Lowest 4 bits & Hexadecimal encoding & Corresponding atomic type \\ \hline
 \end{tabular}
 \vspace{0.3cm}
 
-Note this is not used in BCF2, but its type is reserved in case this becomes necessary.  In BCF2 characters are simply represented by strings with a single element 0,4,6,8-15 reserved for future use.
+Note this is not used in BCF2, but its type is reserved in case this becomes necessary.  In BCF2 characters are simply represented by strings with a single element 4,6,8-15 reserved for future use.
 
 \vspace{0.3cm}
 
@@ -1274,9 +1275,6 @@ $0/1\mid2$ & is tetraploid with a single phased allele & 0x02 04 07 \\ \hline
 The final example is something seen on chrX when we have a haploid male and a diploid female. The spare male allele is just assigned the missing value.
 \vspace{0.3cm}
 
-\textbf{Misc. notes}
-
-A type byte value of 0x00 is an allowed special case meaning MISSING but without an explicit type provided.
 
 
 
@@ -1359,11 +1357,14 @@ Just for discussion, suppose instead that ALT was ALT=C,T.  The only thing that 
 HM3;AC=3;AN=6;AA=C
 Let's assume that the header dictionary elements for HM3, AC, AN, and AA are at 80, 81, 82, and 83 respectively.  All of these can be encoded by 1-element INT8 values (0x11), with associated hex values of 0x50, 0x51, 0x52, and 0x53 respectively.
 
-First is HM3.  The entry begins with the key: 0x11 0x50.  Next we have a Flag value to indicate the field is present, represented as a 1 element INT8 value of 1.  Altogether we have:
+First is HM3.  The entry begins with the key: 0x11 0x50.
+The HM3 key has no value and therefore is represented as the special case of a missing value without an explicit type provided.
+Altogether we have:
+
 
 \vspace{0.3cm}
 \begin{tabular}{|l| l|} \hline
-0x11 0x50 0x11 0x01 & HM3 flag is present \\ \hline
+0x11 0x50 0x00 & HM3 flag is present \\ \hline
 \end{tabular}
 \vspace{0.3cm}
 
@@ -1426,7 +1427,7 @@ So the genotype block contains:
 
 We need to determine a few values before writing out the final block:
 
-l\_shared = 52 (Data length from CHROM to the end of INFO)
+l\_shared = 51 (Data length from CHROM to the end of INFO)
 
 l\_indiv = 42 (Data length of FORMAT and individual genotype fields)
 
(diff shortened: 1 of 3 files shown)


jmarshall:
`NUL` or `'\0'`, not `NULL`.

VCFv4.1.tex
```diff
@@ -1358,11 +1357,14 @@ \subsubsection{Encoding the INFO fields}
 HM3;AC=3;AN=6;AA=C
 Let's assume that the header dictionary elements for HM3, AC, AN, and AA are at 80, 81, 82, and 83 respectively.  All of these can be encoded by 1-element INT8 values (0x11), with associated hex values of 0x50, 0x51, 0x52, and 0x53 respectively.
 
-First is HM3.  The entry begins with the key: 0x11 0x50.  Next we have a Flag value to indicate the field is present, represented as a 1 element INT8 value of 1.  Altogether we have:
+First is HM3.  The entry begins with the key: 0x11 0x50.
+The HM3 key has no value and therefore is represented as the special case of a missing value without an explicit type provided.
```
lbergelson: I think I would find this confusing if I came on it without the current context.  I might rework it to include the word `flag`.  
```suggestion
The HM3 key is a flag with no associated value and therefore is represented as the special case of a missing value without an explicit type provided.
```

pd3:
@lbergelson Thanks for the suggestion, updated.

Big headers occasionally do appear, see this case with 58,407,655 scaffolds  https://github.com/samtools/bcftools/issues/672

VCFv4.3.tex
```diff
@@ -1882,12 +1880,12 @@ \subsubsection{Encoding the INFO fields}
 
 First is HM3.
 The entry begins with the key: 0x11 0x50.
-Next we have a Flag value to indicate the field is present, represented as a 1 element INT8 value of 1.
+The HM3 key is a flag with no associated value and therefore is represented as the special case of a missing value without an explicit type provided.
```
jmarshall:
```suggestion
The HM3 field is a flag with no associated value and therefore is represented as the special case of a missing value without an explicit type provided.
```

and similarly in the other two documents.

VCFv4.3.tex
```diff
@@ -1593,6 +1594,7 @@ \subsubsection{Type encoding}
 \vspace{0.3cm}
 \begin{tabular}{|l | l | l|} \hline
 Lowest 4 bits & Hexadecimal encoding & Corresponding atomic type \\ \hline
+0 & 0x00 & A missing value without an explicit type provided \\ \hline
```
jmarshall: Use `MISSING` to show that this is talking about the particular VCF/BCF meaning of _missing_.

VCFv4.3.tex
```diff
@@ -1602,7 +1604,7 @@ \subsubsection{Type encoding}
 \vspace{0.3cm}
 
 Note this is not used in BCF2, but its type is reserved in case this becomes necessary.
-In BCF2 characters are simply represented by strings with a single element 0,4,6,8--15 reserved for future use.
+In BCF2 characters are simply represented by strings with a single element 4,6,8--15 reserved for future use.
```
jmarshall: What does this paragraph even mean? What is “this” that's not used in BCF2? It's been like this since the document was first added (8b0a59271f13eea9cff30c6e372b7fe675f815f0). 

 Since you're changing it please fix it. It looks like it's referring to something that doesn't exist any more. So perhaps this paragraph can just become something to the effect of:

> Something about strings are type=7 of the appropriate length, and characters are simply represented as length=1 strings. Type codes 4,6,8--15 are reserved for future use.

First sentence optional; note fullstop added before `4,6,…`.

And similarly in the other two documents.

VCFv4.3.tex
```diff
@@ -1415,7 +1415,8 @@ \subsection{Header}
 
 The BCF2 header begins with the ``BCF2 magic'' 5 bytes that encode {\tt BCF\em XY} where {\em X} and {\em Y} are bytes indicating the major number (currently 2) and the minor number (currently 2).
 This magic can be used to quickly examine the file to determine that it's a BCF2 file.
-Immediately following the BCF2 magic is the standard VCF header lines in text format, beginning with \verb|##fileformat=VCFvX.Y|.
+Immediately following the BCF2 magic is an unsigned 4-byte integer containing the length of the header text which includes the NUL terminator character.
+Follow the standard NUL-terminated VCF header lines in text format, beginning with \verb|##fileformat=VCFvX.Y|.
```
jmarshall: This is now technically correct, describing the length field and NUL-termination that were previously omitted as noted in https://github.com/samtools/hts-specs/issues/236#issuecomment-423999865.

However it's completely incomprehensibly expressed. Either the PR submitter or the VCF maintainers should rewrite this so that it is readable by people who don't already know what the format is.

And similarly in the other two documents.

VCFv4.3.tex
jmarshall: To be precise: The two separate ideas and sentences run together as “_In BCF2 characters are simply represented by strings with a single element 0,4,6,8--15 reserved for future use_” is a historical typographical error and **needs** to be fixed. The previous sentence (“_Note this is not used in BCF2, but its type is reserved in case this becomes necessary_”) is a mystery that it **would be nice** to fix while working in the area, perhaps by simply removing it.

daviesrob:
I thought I would have a go at reviving this, so I don't get annoyed again by the BCF part of the specification not matching reality.  Most of the issues raised by @jmarshall hav hopefully been dealt with.  I only changed the section 6.2 (Header) section in VCFv4.3 for now so it's easy to compare with the earlier text in the other files.  If the spec. maintainers like it, it can be copied to the others.

There's still a few issues that we might want to work on.  For example, I'm not sure what the `##dictionary` part in the header section refers to.  Also, should we try to break up `n_fmt_sample` in the site encoding section, as I've done with `a_allele_info`?

I haven't done anything to VCFv4.4, partly because it doen't exist in this branch yet, and partly because we may want to completely remove and rewrite the BCF part in that version.

VCFv4.3.tex
```diff
@@ -1509,8 +1509,8 @@ \subsubsection{Site encoding}
                         Must be the length of the REF allele or the declared length of a symbolic allele respecting
                         the END attribute \\ \hline
 QUAL      & float     & Variant quality; 0x7F800001 for a missing value \\ \hline
-n\_allele\_info & int32\_t & The low 16 n\_allele bits represent the number of REF+ALT alleles in this record and the
-                             high 16 n\_info bits represent the number of VCF INFO fields in this record \\ \hline
+n\_allele & uint16\_t & The number of REF+ALT alleles in this record \\ \hline
+n\_info   & uint16\_t & The number of INFO fields in this record \\ \hline
```
jmarshall, line 1528: The HTSlib implementation has `n_info` first then `n_allele`, both before and after samtools/htslib#1023's rewrite. The text about low and high bits is incorrect and was only added in the second commit on this PR (3a931e1c2a8fa53764f0d008fcf9907d7010b184) — **master** just has “`n_info`, where `n_allele` is the number of REF+ALT alleles in this record, and `n_info` is the number of VCF INFO fields present in this record”.

daviesrob: Oops.  Should be fixed now.

VCFv4.3.tex
```diff
@@ -1413,10 +1413,25 @@ \subsection{Overall file organization}
...
-The BCF2 header begins with the ``BCF2 magic'' 5 bytes that encode {\tt BCF\em XY} where {\em X} and {\em Y} are bytes indicating the major number (currently 2) and the minor number (currently 2).
-This magic can be used to quickly examine the file to determine that it's a BCF2 file.
-Immediately following the BCF2 magic is an unsigned 4-byte integer containing the length of the header text which includes the NUL terminator character.
-Follow the standard NUL-terminated VCF header lines in text format, beginning with \verb|##fileformat=VCFvX.Y|.
+The BCF2 header contains the following items:
+
+{\small
+\begin{tabular}{|l | l | p{30em} | } \hline
+\textbf{Field} & \textbf{Type} & \textbf{Notes} \\ \hline
+magic          & char[3]     & The characters 'B', 'C', 'F' \\ \hline
```
jmarshall, line 1421:
```suggestion
magic          & char[3]     & The characters `B', `C', `F' \\ \hline
```

And perhaps put the literal characters in `\tt` too.

VCFv4.3.tex
```diff
@@ -1413,10 +1413,25 @@ \subsection{Overall file organization}
...
+major\_version & uint8\_t    & Currently 2 \\ \hline
+minor\_version & uint8\_t    & Currently 2 \\ \hline
+l\_text        & uint32\_t   & Length of the ``text'' section, including the terminating NUL character \\ \hline
+text           & char string & VCF format header text, NUL-terminated. \\ \hline
```
jmarshall, line 1425:
```suggestion
text           & char[l\_text] & VCF format header text, NUL-terminated. \\ \hline
```

jmarshall:
Rob's additions do indeed address all the issues I previously raised.

`##dictionary` remains mysterious. It doesn't appear to have been implemented in HTSlib.

I think it would help to break up `n_fmt_sample` too, though the resulting `uint24_t` might deserve a footnote to the effect of _i.e., an 8-bit value and a 24-bit value overlaid within a single `uint32_t`_.

The “used for string values” sentence added to the **BCF2 types** table makes the table very wide. This might be better back in the following paragraph, and IMHO it is helpful to mention that string values are represented as Character vectors and character values are represented as atomic Characters or equivalently Character vectors of length 1.

The header table looks good (though it needs a bit of space around it). The “Immediately following the minor_version …” sentence just repeats information in the table. IMHO that whole paragraph should be replaced by something like

> The ``text'' field contains the standard VCF header lines in text format, from `##fileformat…` to `#CHROM…` inclusive, terminated by a NUL character.

and perhaps merged into the “The magic field …” paragraph.

This PR has been going on so long that it is now almost unreviewable in its entirety :smile: and there is also PR #456 which was @lbergelson's stab at addressing the issues I raised. Probably this PR should be rebased and squashed appropriately, anything salient from the other PR brought across, and then we can do a final review of the eventual change in its entirety.

VCFv4.3.tex
daviesrob: Hmmm.  What would be better?
```
The characters `{\tt B}', `{\tt C}', `{\tt F}'
```
or
```
The characters ``{\tt BCF}''
```

jkbonfield: It's `char[3]`. It's pretty clear "BCF" means those 3 chars in that order without being tediously explicit.

lbergelson:
Closing, this has been replaced with https://github.com/samtools/hts-specs/pull/456

PR #427, opened 2019-06-27
Fix minor typo in the predecessors of BCF2

r6eve:


lbergelson:
@r6eve Good catch.  Thank you.

PR #429, opened 2019-07-02
The QUAL field should go immediately after rlen in BCFv2.x

athos:
I'm not sure whether or not the table in the section 6.3.1 specifies the field order, not only the type and semantics of each field, but anyway it seems confusing since it's not consistent with the actual field order.


pd3:
This is already addressed by this commit https://github.com/samtools/hts-specs/pull/417 which also addresses a few other issues.

athos:
Ah, thanks. I couldn't find that. I'm closing this PR then.

PR #430, opened 2019-07-10
htsget: add `samples` query parameter, principally to select subset of VCF columns

mlin:
e.g.

```
GET /htsget/1000genomes/variants?format=VCF&samples=NA12878,NA12877
```

Previously I circulated a different version of this with a repeated `sample=x` query parameter. This single, comma separated list is more consistent with the existing `fields` & `tags` query parameters.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -254,6 +254,16 @@ A comma separated list of tags to exclude, default: none.
 
 The server SHOULD respond with an `InvalidInput` error if `tags` and `notags` intersect.
 </td></tr>
+<tr markdown="block"><td>
+
+`samples`  
+_optional_
+</td><td>
+A comma separated list of samples to include, default: all. For VCF responses, this selects a subset of columns from a multi-sample VCF dataset.
+
+The server SHOULD respond with an `InvalidInput` error if a nonexistent `sample` is listed.
+</td></tr>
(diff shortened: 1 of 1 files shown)


mlin:
Can a sample name have a comma in it? (@robdavies)

mlin:
Correct me if wrong, but it is not forbidden for a VCF sample name to contain a comma.

One possibility is to specify URI encoding of each element in the comma-separated list, so any comma within the sample name would be percent-encoded. A comma-separated list of individually URI-encoded elements seems like it would be a slightly tortured construct, though.

Another possibility is reverting to the first straw man idea of providing the list through repeated query parameters, where each individual parameter would then by query string encoded as usual, e.g.

```
GET /htsget/1000genomes/variants?format=VCF&samples=NA12878&samples=NA12877&samples=quick%2C%20brown%20fox
```

The wart is that it's dissimilar from the existing `fields` & `tags` parameters.

nh13:
Another idea: the default delimiter be comma and then have a parameter to control the delimiter?

mlin:
@daviesrob suggestion: percent-encoded tab delimiter

yfarjoun:
Also, can sampleID be empty?

PR #434, opened 2019-07-24
Define Local Alleles in VCF to allow for sparser format

yfarjoun:
This is the local-alleles part of #420

### DIFF
--- a/VCFv4.4.tex
+++ b/VCFv4.4.tex
@@ -482,9 +482,9 @@ See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for s
       GQ		& 1		& Integer	& Conditional genotype quality \\
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
-      LAA		& .		& Integer	& Local Alternate Alleles the 1-based index into the alternate alleles indicating which are relevant for the current sample \\
-      LAD		& .		& Integer	& Local Allele Depth for the reference and each of the local alternate alleles (see: LAA) \\
-      LPL		& .		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the Reference and the LAA alleles only (see LAA) \\
+      LAA		& .		& Integer	& Strictly increasing, 1-based indices into ALT, indicating which alternate alleles are relevant (local) for the current sample \\
+      LAD		& .		& Integer	& Local allele read depth for the reference and each of the local alternate alleles listed in LAA \\
+      LPL		& .		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the reference and the local alternative alleles listed in LAA \\
       MQ		& 1		& Integer	& RMS mapping quality \\
       PL		& G		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer \\
       PP		& G		& Integer	& Phred-scaled genotype posterior probabilities rounded to the closest integer \\
@@ -581,14 +581,18 @@ See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for s
     \end{itemize}
 
   \item HQ (Integer): Haplotype qualities, two comma separated phred qualities.
-  \item LAA (and LAD and LPL (*):
-  For callsets with a large number of samples, it is often the case that the majority of sites are not called and sites end up involving many alleles for which all the samples need to provide PL and AD.
-  This can cause the file-sizes to grow super-linearly with the number of samples.
+  \item LAA
+  In callsets with many samples, sites may grow to include numerous alternate alleles at the same POS.
+  Usually, few of these alleles are actually observed in any one sample, but each genotype must supply fields like PL and AD for all of the alleles---a very inefficient representation as PL's size is quadratic in the allele count.
+  Similarly, in rare sites, which can be the bulk of the sites, the vast majority of the samples are reference.
   To prevent this, one can choose to specify the allele depth and the genotype likelihood against a subset of ``Local Alleles''.
-  LAA is the (1-based) index into the list of alleles that are ALT for that variant.
+  LAA is the strictly increasing, 1-based index into ALT, pointing out the alternative alleles that are actually in-play for that sample.
   For example, if REF is G, ALT is A,C,T,\verb!<*>! and a genotype only has information about G, C, and \verb!<*>!, one can have LAA=[2,4] and thus LPL will be interpreted as pertaining to the alleles [G, C, \verb!<*>!] and not contain likelihood values for genotypes that involve A or T.
   It is implicit that REF is part of any ``local'' context, and it always has index 0, even if the genotype is compound HET.
+  Note that reordering might be required and care need to be taken to reorder LAD and LPL appropriately.
   LAA is required in order to interpret LAD and LPL.
+  \item LAD: See LAA
+  \item LPL: See LAA
   \item MQ (Integer): RMS mapping quality, similar to the version in the INFO field.
   \item PL (Integer): The phred-scaled genotype likelihoods rounded to the closest integer, and otherwise defined in the same way as the GL field.
   \item PP (Integer): The phred-scaled genotype posterior probabilities rounded to the closest integer, and otherwise defined in the same way as the GP field.
--- a/VCFv4.4.tex
+++ b/VCFv4.4.tex
@@ -482,9 +482,9 @@ See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for s
       GQ		& 1		& Integer	& Conditional genotype quality \\
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
-      LAA		& .		& Integer	& Local Alternate Alleles\footnotemark[1]\\
-      LPL		& .		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the Reference and the LAA alleles only.\footnotemark[1]\\
-      LAD		& .		& Integer	& Local Allele Depth for the reference and each of the local alleles\footnotemark[1]\\
+      LAA		& .		& Integer	& Local Alternate Alleles the 1-based index into the alternate alleles indicating which are relevant for the current sample \\
+      LAD		& .		& Integer	& Local Allele Depth for the reference and each of the local alternate alleles (see: LAA) \\
+      LPL		& .		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the Reference and the LAA alleles only (see LAA) \\
       MQ		& 1		& Integer	& RMS mapping quality \\
       PL		& G		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer \\
       PP		& G		& Integer	& Phred-scaled genotype posterior probabilities rounded to the closest integer \\
@@ -581,7 +581,7 @@ See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for s
     \end{itemize}
 
   \item HQ (Integer): Haplotype qualities, two comma separated phred qualities.
-  \item Local Alleles (*):
+  \item LAA (and LAD and LPL (*):
   For callsets with a large number of samples, it is often the case that the majority of sites are not called and sites end up involving many alleles for which all the samples need to provide PL and AD.
   This can cause the file-sizes to grow super-linearly with the number of samples.
   To prevent this, one can choose to specify the allele depth and the genotype likelihood against a subset of ``Local Alleles''.
--- a/VCFv4.4.tex
+++ b/VCFv4.4.tex
@@ -483,7 +483,8 @@ See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for s
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
       LAA		& .		& Integer	& Strictly increasing, 1-based indices into ALT, indicating which alternate alleles are relevant (local) for the current sample \\
-      LAD		& .		& Integer	& Local allele read depth for the reference and each of the local alternate alleles listed in LAA \\
+      LAD		& .		& Integer	& Read depth for the reference and each of the local alternate alleles listed in LAA \\
+      LGT		& .		& String 	& Genotype against the local alleles \\
       LPL		& .		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the reference and the local alternative alleles listed in LAA \\
       MQ		& 1		& Integer	& RMS mapping quality \\
       PL		& G		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer \\
@@ -587,11 +588,23 @@ See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for s
   Similarly, in rare sites, which can be the bulk of the sites, the vast majority of the samples are reference.
   To prevent this, one can choose to specify the allele depth and the genotype likelihood against a subset of ``Local Alleles''.
   LAA is the strictly increasing, 1-based index into ALT, pointing out the alternative alleles that are actually in-play for that sample.
-  For example, if REF is G, ALT is A,C,T,\verb!<*>! and a genotype only has information about G, C, and \verb!<*>!, one can have LAA=[2,4] and thus LPL will be interpreted as pertaining to the alleles [G, C, \verb!<*>!] and not contain likelihood values for genotypes that involve A or T.
+  LAD is the depth of the local alleles, LPL is subset of the PL array that pertains to the alleles that are REF or referred to by LAA, LGT is the genotype but referencing the local alleles rather than the global ones.
   It is implicit that REF is part of any ``local'' context, and it always has index 0, even if the genotype is compound HET.
+  For example, if REF is G, ALT is A,C,T,\verb!<*>! and a genotype only has information about G, C, and \verb!<*>!, one can have LAA=[2,4] and thus LPL will be interpreted as pertaining to the alleles [G, C, \verb!<*>!] and not contain likelihood values for genotypes that involve A or T.
+  In this case LGT=0/1 means that the sample is G/C.
+  GQ is still the genotype quality, even when the genotype is given against the local alleles.
   Note that reordering might be required and care need to be taken to reorder LAD and LPL appropriately.
-  LAA is required in order to interpret LAD and LPL.
+  LAA is required in order to interpret LAD, LPL, and LGT.
+
+  For example, these two lines are encoding the same information (some columns removed for clarity):
+
+  \begin{tabular}[l]{lllll}
+REF&  ALT&FORMAT&Alice&Bob\\
+G&A,C,T,\textless*\textgreater& LAA:LGT:LAD:LPL&  2,4:1/1:20,30,10:90,80,0,100,110,120 &3:0/1:15,25:40,0,80\\
+G&A,C,T,\textless*\textgreater& GT:AD:PL&  2/2:20,.,30,.,10:90,.,.,80,.,0,.,.,.,.,100,.,110,.,120&0/3:15,.,.,25,.:40,.,.,.,.,.,0,.,.,80,.,.,.,.\\
+\end{tabular}
   \item LAD: See LAA
+  \item LGT: See LAA
   \item LPL: See LAA
   \item MQ (Integer): RMS mapping quality, similar to the version in the INFO field.
   \item PL (Integer): The phred-scaled genotype likelihoods rounded to the closest integer, and otherwise defined in the same way as the GL field.
--- a/VCFv4.4.tex
+++ b/VCFv4.4.tex
@@ -482,6 +482,9 @@ See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for s
       GQ		& 1		& Integer	& Conditional genotype quality \\
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
+      LAA		& .		& Integer	& Local Alternate Alleles\footnotemark[1]\\
+      LPL		& .		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the Reference and the LAA alleles only.\footnotemark[1]\\
+      LAD		& .		& Integer	& Local Allele Depth for the reference and each of the local alleles\footnotemark[1]\\
       MQ		& 1		& Integer	& RMS mapping quality \\
       PL		& G		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer \\
       PP		& G		& Integer	& Phred-scaled genotype posterior probabilities rounded to the closest integer \\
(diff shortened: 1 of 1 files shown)


VCFv4.3.tex
```diff
@@ -406,12 +406,14 @@ \subsubsection{Genotype fields}
       GQ		& 1		& Integer	& Conditional genotype quality \\
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
+      LAA		& . 		& Integer	& Local Alternate Alleles\footnotemark[1]\\ 
+      LPL		& . 		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the Reference and the LAA alleles only.\footnotemark[1]\\
```
lbergelson: Shouldn't LPL be sorted after LAD?

VCFv4.3.tex
```diff
@@ -503,7 +505,15 @@ \subsubsection{Genotype fields}
   All phased genotypes that do not contain a PS subfield are assumed to belong to the same phased set.
   If the genotype in the GT field is unphased, the corresponding PS field is ignored.
   The recommended convention is to use the position of the first variant in the set as the PS identifier (although this is not required).
-\end{itemize}
+  \item LAA (and LAD and LPL (*):
+  For callsets with a large number of samples, it is often the case that the majority of sites are not called and sites end up involving many alleles for which all the samples need to provide PL and AD. 
+  This can cause the file-sizes to grow super-linearly with the number of samples.
```
mlin: Different take on the previous two sentences; feel free to use/recombine

> In callsets with many samples, sites may grow to include numerous alternate alleles at the same POS. Usually, few of these alleles are actually observed in any one sample, but each sample must supply fields like PL and AD for all of the alleles -- a very inefficient representation as PL's size is quadratic in the allele count.

VCFv4.3.tex
```diff
@@ -406,12 +406,14 @@ \subsubsection{Genotype fields}
       GQ		& 1		& Integer	& Conditional genotype quality \\
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
+      LAA		& . 		& Integer	& Local Alternate Alleles the 1-based index into the alternate alleles indicating which are relevant for the current sample.\\
```
mlin:
```suggestion
      LAA		& . 		& Integer	& Local Alternate Alleles the strictly-increasing, 1-based indices into the alternate alleles indicating which are relevant for the current sample.\\
```

VCFv4.3.tex
```diff
@@ -503,7 +505,15 @@ \subsubsection{Genotype fields}
   All phased genotypes that do not contain a PS subfield are assumed to belong to the same phased set.
   If the genotype in the GT field is unphased, the corresponding PS field is ignored.
   The recommended convention is to use the position of the first variant in the set as the PS identifier (although this is not required).
-\end{itemize}
+  \item LAA (and LAD and LPL (*):
```
cyenyxe: I would follow the existing of one item per field, otherwise it may be confusing to readers searching for LAD and LPL.

VCFv4.3.tex
cyenyxe: Or maybe:
```suggestion
      LAA		& . 		& Integer	& 1-based index indicating which alternate alleles are relevant (local) for the current sample\\
```

VCFv4.3.tex
```diff
@@ -406,12 +406,14 @@ \subsubsection{Genotype fields}
       GQ		& 1		& Integer	& Conditional genotype quality \\
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
+      LAA		& . 		& Integer	& Local Alternate Alleles the 1-based index into the alternate alleles indicating which are relevant for the current sample.\\
+      LAD		& . 		& Integer	& Local Allele Depth for the reference and each of the local alternate alleles (see: LAA).\\
```
cyenyxe:
```suggestion
      LAD		& . 		& Integer	& Read depth for the reference and each of the local alternate alleles listed in LAA\\
```

VCFv4.3.tex
```diff
@@ -406,12 +406,14 @@ \subsubsection{Genotype fields}
...
+      LPL		& . 		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the Reference and the LAA alleles only (see LAA).\\
```
cyenyxe:
```suggestion
      LPL		& . 		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the reference and the local alternate alleles listed in LAA\\
```

VCFv4.3.tex
mlin: SGTM but let's say the indices have to be strictly increasing; obviously that's a pedantic constraint but without it, the ordering of LPL could be unclear as well.

cyenyxe: Suggestion updated.

mlin: Dwelling on this further -- @yfarjoun @lbergelson -- I'm worried just slightly about confusion over the LPL ordering. Lets say the joint VCF has REF=A ALT=G,C,T and the sample GVCF has ALT=T,G. so LPL=1,3 if we require the indices to be in order, but this order is flipped with respect to the GVCF.

The merging algo then has to be really careful to [permute the GVCF PL values](https://github.com/dnanexus-rnd/GLnexus/blob/237e9bcbcd685445ab0aa65ecbc6c29cefe1b26e/src/diploid.cc#L16-L43) in just the right way to match that. In the wild, I'm nervous this is going to lead to difficult-to-detect bugs with subtle swapping of probability values. Did you have any experience with this in developing the idea? Should we permit the merger to write an out-of-order LAA (this might just shift the potential confusion to readers)? I don't have a better idea right now, it just seems to me like a significant pitfall at least warranting explication where LPL is discussed.

lbergelson: I agree that this is a concern.  The only actual implementation so far of a format with local alleles decided to store the allele string itself instead of the index because they didn't want to deal with the reordering problem.  So unordered [T, G] instead of [1, 3]. Move the difficulties from the merge step to the reading step in that now if you want to look up non-local values associated with an allele you have to do a search in the INFO allele values to find the corresponding index.  I

My thinking was that it was better to require the indexes as numbers and do the reordering, but I think there's a strong argument the other way.

mlin: Thanks, I'll spawn a background thread on this. BTW, if anyone knows a better formula than the following to go from PL index `gt` to the indices `(i,j)` of the constituent alleles, I would like to know. This is seriously the best I could come up with, holy sh**.

```
    unsigned j = (unsigned)((sqrt(8*gt+1)-1.0)/2.0);
    unsigned i = gt - j*(j+1)/2;
```

VCFv4.3.tex
pd3: Both LPL and LAD should be sorted after REF+LAA, the same way as REF+ALT define the order of PL and AD.

VCFv4.3.tex
pd3: I like @cyenyxe's suggestion, it solves a possible confusion. It also allows to determine the LPL ordering per allele-set, samples with the same set of local alleles are guaranteed to use the same order.

@mlin Note that the proposed formula is for the special case of diploid genotypes. In my experience, in practice it is sufficient to pre-compute the indexes using nested/recursive loops as suggested in the description of GL, section 1.6.2.

VCFv4.3.tex
mlin:
```suggestion
      LPL		& . 		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer for genotypes that involve the reference and the local alternate alleles listed in LAA, in the genotype order implied by LAA such that LPL would be a subsequence of PL were the latter present.\\
```

@lbergelson @pd3 following the previous discussion, I feel it's worth leaving no possible question about the LPL order -- here's my attempt, better suggestions welcome.

For GVCF merging, we're already responsible for carefully permuting the GVCF PL values into the target joint vector, the new wrinkle will be that each sample can have a different target. I think this is okay, it was just tempting to want to copy in the GVCF PL verbatim.

pd3: I think it's better to speak about ordering only and not require a strict subsequence: it should be possible to rescale the LPL likelihoods.

VCFv4.3.tex
yfarjoun: why do the indexes need to be strictly increasing? 

One of the main points is to lessen the burden on the creation of these pVCFs...so that when the alleles change the only change that needs to happen in the format field is the LAA. if we require strictly increasing order, that goes out the window.

VCFv4.3.tex
yfarjoun: should I leave the paragraph as is, and add entries for LAD and LPL pointing to LAA?

VCFv4.3.tex
pd3: If LAA are strictly increasing, LPL's are guaranteed to have the same order as 1) the original PLs and 2) samples with the same set of alleles.

Editing of VCFs is much less frequent operation than read-only processing, therefore it is more important to optimize the speed and convenience of the latter.

VCFv4.3.tex
jmarshall: The AD, ADF, ADR paragraph also groups related items into a single paragraph. Fortuitously in both cases the three entries are adjacent alphabetically.

(Though the whole LAA/LAD/LPL paragraph wants to move up to between HQ and MQ, and this label (`LAA (and LAD and LPL (*)`) wants less ‘and’ and better balanced parentheses!)

mlin:
Hi all, I was motivated to come back here after another collaborator just run aground on the [htslib int overflow](https://github.com/samtools/htslib/pull/820) when it tries to allocate a 2GB+ PL vector :smile: :sob: 

It seems like we are not far; the open items are

1. Must the LAA allele indices for each sample be strictly increasing? I advocate yes, to make reading LPL less tricky, but the trickiness burden is shifted to the writer.
2. Precise language for specifying the order of LPL values, if anyone agrees with me it's worthwhile to leave no doubt.
3. Move the item describing LAA, LAD, and LPL into alphabetical position within the itemized list. (+ misc wording and punctuation edits)

jmarshall:
I think this is a nice new feature, but it took me many minutes of very close reading to figure out just what the feature is! If I understand it correctly, it could perhaps be better described as:

* The table of reserved keys is just a brief taster / memory jogger of what each field is about, so: LAA List of relevant Local Alternate Alleles / LAD Read depth for each local allele / LPL Phred-scaled local genotype likelihoods rounded to the closest integer

* The text in the bulleted paragraph motivating this doesn't mention that the point of all this is to only list “interesting” alleles on a per-sample basis because each sample has a different small subset of alt alleles that are present for that sample. The text suggested in https://github.com/samtools/hts-specs/pull/434#discussion_r307968349 is rather better.

* It needs a complete example so you can see what is happening. Something like

```
… REF  ALT        …  FORMAT          SAMPLE1                       SAMPLE2
… G    A,C,T,<*>  …  GT:LAA:LAD:LPL  1/1:2,4:20,30,10:1,2,3,4,5,6  1/1:3:15,25:1,2,3
```

   thus demonstrating the different subsets per sample.

mlin:
> Similarly, in rare sites, which can be the bulk of the sites, the vast majority of the samples are reference. 

This is a good point; we should probably clarify how one should write LAA & LPL for samples reporting no evidence for any particular alternate allele.

VCFv4.3.tex
```diff
@@ -406,12 +406,15 @@ \subsubsection{Genotype fields}
       GQ		& 1		& Integer	& Conditional genotype quality \\
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
+      LAA		& . 		& Integer	& Strictly increasing, 1-based indies into ALT, indicating which alternate alleles are relevant (local) for the current sample \\
```
lbergelson: typo: *indies*

lbergelson:
Some misc things I've been thinking about while implementing this:

1.  ./. calls are tricky to convert to local alleles in a joined gvcf because we don't know which subset of alleles is relevant, in worst case you may have to include every allele in LAA or analyze the PL's to see which ones are not actually meaningful.
2.  Requiring LAA to be sorted is annoying when merging files. 
3.  Would it be a good idea to allow leaving out LAA in when the local alleles include all of the alleles at the site?  Benefits: potentially space saving, makes single sample files in local allele format cleaner.  Cons: easier to make mistakes when merging files.

VCFv4.3.tex
```diff
@@ -503,7 +506,41 @@ \subsubsection{Genotype fields}
...
+ 
+  \begin{tabular}[l]{llllll}
+POS &REF&  ALT&FORMAT&sample\\
+1&G&A,C,T,\textless*\textgreater& LAA:LGT:LAD:LPL&  2,4:1/1:20,30,10:90,80,0,100,110,120\\
+1&G&A,C,T,\textless*\textgreater& GT:AD:PL&  2/2:20,.,30,.,10:90,.,.,80,.,0,.,.,.,.,100,.,110,.,120\\
+2&A&C,G,T,\textless*\textgreater& LAA:LGT:LAD:LPL& 3:0/1:15,25:40,0,80\\
+2&A&C,G,T,\textless*\textgreater& GT:AD:PL&0/3:15,.,.,25,.:40,.,.,.,.,.,0,.,.,80,.,.,.,.\\
+3&C&G,T,\textless*\textgreater& LAA:LGT:LAD:LPL& 4:0/0:30,1:0,30,80\\
+3&C&G,T,\textless*\textgreater& GT:AD:PL& 0/0:30,.,.1:0,.,.,.,.,.,.,.,.,.,30,.,.,.,80\\
+4&G&A,T,\textless*\textgreater& LAA:LGT:LAD:LPL& :0/0:30:0\\
```
lbergelson: Typo here:  we either need to remove the LAA field and indicate that it can be left out if there are no non-ref local alleles, or we need to add a leading `.`

jmarshall:
Re (3) _Would it be a good idea to allow leaving out LAA in when the local alleles include all of the alleles at the site?_

Wouldn't sites where all samples include all the alleles just use GT/AD/PL rather than LAA/LGT/LAD/LPL? (And this would obviously be allowed as there is no text saying a file has to have all or none of its variant lines using LAA, and the default assumption would be that a mixture is fine.)

Or do you mean: when some samples include all the alleles and others don't, instead of writing `1,2,3,…,N` for LAA for the all-alleles samples perhaps allow some special value to be an abbreviation for that? e.g. `.` but frankly that would be an unexpected meaning for `.`…

h-2:
I was just reading through all of this, since I am implementing it for one of our tools, and I wondered: what's the point of `LGT` ? It doesn't save any space and just adds the indirection of parsing through LAA...

In our applications, the only point of doing this is to transform fields with number G (PL-values), although an argument could be made for transforming any/all fields with number A and R, as well. Is there a reason for only suggesting `AD` and not other A/R-fields? What is the point of `LGT`?

d-cameron:
My preference is to expand the scope of this and make it  generic over all `Type=R`/`A`/`G` fields with the local notation implictly defined by the presence of `LAA`^. That is, `PL`, `GT`, etc. are redefined from indexes w.r.t `ALT` to indexes w.r.t `ALT[LAA]` with a missing `LAA` being interpreted as the identity function.

In this manner, converting between the short and long-hand notations can be done not with `L`-prefixed fields, but with the fields themselves based on the presence/absence of `LAA`. This would actually simplify VCF merging as `LAA` would act as an indirection/remapping marker with the only change required for the `FORMAT` fields for each sample is the addition of the `LAA` field.

Another benefit of this approach is that is allows fields such as `PL` to preserve their original meaning.

^ `AI` (alt index) is probably a better name than `LAA` as no `L`-prefixed fields would be required

pd3:
@d-cameron That's a terrible idea, it would break existing tools. 

Also, I don't see how it would help with merging VCFs at all, or what do you mean by your comment that it would help preserve `PL` its original meaning?

If you want to go this route, please introduce a new set of tags, rather than change the semantics of existing ones.

olest:
Hi, 

we are using local alleles in DRAGEN gVCF Genotyper. More specifically, we write FORMAT/LPL, LAD, LGT and LAA in the multi-sample VCF output.

In particular FORMAT/LPL gives a significant space saving compared to FORMAT/PL which scales quadratically with the number of alleles at a variant site.

This becomes important for an msVCF representing a large cohort because multi-allelic sites with hundreds or even thousand of alleles will become more frequent.

In these cases, the msVCF line can exceed the size limit of the htslib parser as @mlin mentions above.

I would love to see this change to become part of the VCF spec. If there is anything needed to support this, please let me know.

VCFv4.3.tex
```diff
@@ -503,7 +506,41 @@ \subsubsection{Genotype fields}
   All phased genotypes that do not contain a PS subfield are assumed to belong to the same phased set.
   If the genotype in the GT field is unphased, the corresponding PS field is ignored.
   The recommended convention is to use the position of the first variant in the set as the PS identifier (although this is not required).
-\end{itemize}
+  \item LAA is a sorted list of n distinct integers, where 1 <= n <= $|ALT|$, giving the (1-based) indices within ALT of the alleles that are observed in the sample.
```
whitwham, line 509: **1 <= n <= $|ALT|$** converts to **1 ¡= n ¡= |ALT |** in the PDF.  Is this what was intended? (Not a notation I have seen before if so).

jmarshall:
Since this PR was started, v4.4 has become the canonical VCF version. We may wish to adjust this PR to add the new fields to the _VCFv4.4.tex_ document instead (or perhaps to both, but primarily to the current one, v4.4).

I have drafted PR #745 applying the local alleles changes to the VCFv4.4 document on top of current **master** and fixing some formatting etc. It is not intended to split the conversation from this PR, but to provide a preview PDF and demonstration of how these changes applies to current v4.4.

If @yfarjoun and the VCF maintainers agree, and wish to apply the local alleles additions to v4.4 instead of v4.3 (or perhaps to both), then we can force push the new PR's branch to **yfarjoun:yf_local_alleles**, close #745, and continue the conversation on this original PR.

VCFv4.3.tex
jmarshall: PR #745 corrects the formatting of the mathematics.

VCFv4.3.tex
jmarshall: I don't think this is a typo: the intention is that LAA is a 0-length list of integers (indicating that this record is homozygous ref for this sample and that none of the ALT alleles are in play), not an unknown MISSING value.

This may be a problem, as it's not clear that 0-length lists are valid syntax in VCF or BCF. Omitting LAA is probably not the right workaround, as presence of LAA is the simple indicator that local alleles are being used in a VCF record. Setting it to `.` (unknown) is somewhat misleading, as it **is** known…

jmarshall: This example contradicts the current text, which says “LAA is a sorted list of _n_ distinct integers, where _n_ ≥ 1 […]”. So the text will need adjustment if this example record is deemed valid.

daviesrob: `bcftools` converts the zero-length list into a missing value.  This comes from bcftools treating everything as BCF internally, and [BCF says zero-length lists are missing](https://samtools.github.io/hts-specs/VCFv4.4.pdf#subsubsection.6.3.3).

Possible fixes would be to say missing `LAA` means REF; to explicitly put 0 in `LAA` for REF-only; or use `GT` in these rows, but allow use of `LPL`, `LAD` etc. in the specific case that `GT` is REF-only.

pd3: It makes sense to use the missing value `.` and state it explicitly that it means REF. I can't think of a case where this could cause problems.

jkbonfield: I agree with Petr here.  While the BCF specification states that a zero length array is defined to be a MISSING array, hence `.`, we could add an explicit statement to LAA saying that missing LAAs means that it matches REF.  What we gain is a solution to this problem.  What we lose is the ability to store an unknown LAA list.  I'm unsure what the latter practically means.  Why would you need to explicitly state it's missing instead of just omitting it.  If it's missing then we're not using LAA (and if the PL, AD etc are missing we're not defining it that way either).

PS. Technically even if it's `.` this still contradicts "LAA is a sorted list of n distinct integers" as `.` isn't an integer.  However that's nit-picking and maybe somewhere else VCF already defines `.` as being an integer of unknown value?  I don't know what we do elsewhere when we have this technical disparty.  Is it just assumed that "." is a polymophic type that matches whatever the item wants it to be?

jmarshall: What the BCF spec says is that a type descriptor byte with ‘size’=0 denotes MISSING. It does not say what a type descriptor byte with ‘size’=15 followed by a typed Integer with value 0 denotes, but that would be a natural representation of a 0-length list. (The spec does say that ‘size’=15 implies the list length is ≥15, but the typed Integer is not biased so no meanings are described for typed Integer values <15.) Thus BCF does not provide a binding answer on whether 0-length lists are valid or representable in VCF/BCF.

But I'm not going to advocate blessing 0-length lists in VCF or BCF, especially not just for this, even though SAM found them useful for regularity (cf #135).

IMHO the clearest way to represent this corner case would be

> As a special case, the homozygous REF case in which no ALT alleles are observed in a particular sample is represented as LAA=[0].

(And we'd need to rework _n_ in the length maths: e.g. by dropping _n_ from the description of LAA, and adding “where _n_ is the number of **ALT** alleles listed in LAA” to the others.)

pd3: To expand on my previous comment: the logic of LAA is to define a list of alleles that can be referred to in subsequent localized fields. If only the reference allele is relevant, then it makes sense to provide the missing value `.`, in this sense it behaves in exactly the same way as what we write in REF and ALT columns. Note also the naming of the tag stands for "Local Alternate Alleles". By using `.` we are not introducing a new concept.

jkbonfield:
Background into this, that came up in Future of VCF.  All Of Us use VDS natively (Hail Variant Dataset).  I think this is likely the origin of this PR and #435.

https://support.researchallofus.org/hc/en-us/articles/14927774297620-The-new-VariantDataset-VDS-format-for-All-of-Us-short-read-WGS-data
https://hail.is/docs/0.2/vds/index.html

One difference there is VDS uses LA and not LAA,and LA includes 0 for REF.

jmarshall:
As discussed in today's meeting, this PR has been updated to apply to the v4.4 document instead (and some formatting is improved).

Some further work needed was discussed in the meeting:

* We may wish to apply this to an upcoming v4.5 draft instead;
* The new genotype fields are listed as `Number=.`, which may wish to be updated to `Number=P` (for some of them at least) once we have settled on a suitable definition of `P` (cf #705).

danking:
@h-2 
> I was just reading through all of this, since I am implementing it for one of our tools, and I wondered: what's the point of LGT ? It doesn't save any space and just adds the indirection of parsing through LAA...
> In our applications, the only point of doing this is to transform fields with number G (PL-values), although an argument could be made for transforming any/all fields with number A and R, as well. Is there a reason for only suggesting AD and not other A/R-fields?

We (Hail) also questioned the value of LGT until we discovered a footgun. Suppose:
1. You have a VCF that uses local alleles.
2. At locus chr1:1, the alternate allele array is `A,T,G` and there are two samples whose genotypes are: `0/3` and `1/2`.
3. You decide to remove the first sample.

It is not unreasonable to also remove the now unused alternate alleles from the alternate allele array, but now your `GT` is wrong. So, the options when filtering samples are:
1. Always keep the "dead" alternate alleles.
2. Remove the "dead" alternate alleles *and also* update any affected GTs.
3. Use LGT.

For Hail's purposes, using LGT seems best because it allows sample subsetting without O(N_SAMPLES_KEPT) edits. It also avoids the footgun of "whoopsie I changed the meaning of all my GTs!".

VCFv4.3.tex
danking: Parachuting in here a bit, so my apologies if I'm missing context, but defining LAA as an R-length array whose first element is always the number zero both avoids these nasal demons and avoids a branch [1] when connecting a local index into a global index.

Is it too late to consider an R-length LAA instead of an A-length one?


[1] Text processing cost probably dominates, but if you convert a VCF into some binary format without adjusting LAA, maybe this matters a bit.

danking:
@jmarshall 
> The new genotype fields are listed as Number=., which may wish to be updated to Number=P (for some of them at least) once we have settled on a suitable definition of P (cf https://github.com/samtools/hts-specs/issues/705).

Do I understand correctly that

> one for each allele specified in the GT

means that P is the ploidy of the GT? Concretely, `1/1`, `1/2`, and `2/2` all have P equal to 2?

I'd prefer three new numbers (all of which are distinct from the ploidy):

- LOCAL-A. One element for each allele present in LA minus one (to exclude the reference). For example: LEC.
- LOCAL-R. One element for each allele present in LA (including the reference). For example: LAD.
- LOCAL-G. One element for each possible genotype. If  $N_{LA}$ is the number of local alleles, LOCAL-G is the triangle number of $N_{LA}$. Concretely: if diploid, $\frac{1}{2}N_{LA} (N_{LA}+1)$, and, if haploid, $N_{LA}$.  For example: LPL.

I think `P` is sometimes equal to `LOCAL-A`, sometimes equal to `LOCAL-R`, and sometimes neither:

| GT | P | LOCAL-A | LOCAL-R |
| --- | --- | --- | --- |
| 0/0 | 2 | 0 | 1 |
| 0/1 | 2 | 1 | 2 |
| 1/2 | 2 | 2 | 3 |

d-cameron:
Based onto VCF 4.5. Please continue discussion on #758.

Issue #439, opened 2019-08-20
Please clarify if the INFO field can be left empty or the missing value should be used


eyherabh:
Section 1 of VCF 4.3 says that zero-length fields are not allowed, and that the period should be used instead, so it may seem redundant to state that in the description of the INFO field in section 1.6.1. However, the need for using a period (also referred as missing value) is mentioned there for every other non-required field (i.e. ID, QUAL, FILTER). Therefore, the fact that it is not explicitly stated for the INFO field left me under the impression that the INFO field can be left empty.
Could you please clarify that? Thanks

pd3:
INFO is the same, cannot be left empty. Only FORMAT and sample columns may be omitted when no samples are present.

eyherabh:
Thanks! Submitted clarification.

PR #440, opened 2019-08-27
Resolves issue #439 (INFO field clarificaton)

eyherabh:
Clarifies that the INFO field can be left empty, but the missing value must be used (issue #439).


VCFv4.3.tex
```diff
@@ -324,9 +324,9 @@ \subsubsection{Fixed fields}
   `0' is reserved and must not be used as a filter String.
   If filters have not been applied, then this field must be set to the MISSING value.
   (String, no white-space or semi-colons permitted, duplicate values not allowed.)
-  \item INFO --- additional information: (String, no semi-colons or equals-signs permitted; commas are permitted only as delimiters for lists of values; characters with special meaning can be encoded using the percent encoding, see Section~\ref{character-encoding}; space characters are allowed)
+  \item INFO --- additional information: (String, no semi-colons or equals-signs permitted; commas are permitted only as delimiters for lists of values; characters with special meaning can be encoded using the percent encoding, see Section~\ref{character-encoding}; space characters are allowed). If there is no additional information, then the MISSING value must be used.
```
lbergelson: This change looks good to me, but could you put the new sentence on a new line?  That's the newer formatting style we're using.  so that it makes it easier to see diffs in the tex.

eyherabh: Done!

eyherabh:
Does it look fine now?
By the way, thanks for the tip on diffs!

jmarshall:
There's a spurious whitespace change on the otherwise unchanged following line, and the added sentence would be better inside the parentheses.

Otherwise it's fine as far as it goes, but what this paragraph really wants is a complete rewrite to be more similar to the ALT section:

> INFO — additional information: Semicolon-separated list of additional information fields, or the MISSING value '`.`' if there are none. [**Then** dive into the details about semicolons not allowed _in values_ but may be percent-encoded, etc etc.]

yfarjoun:
I think that this can be back-proped into v4.2 as well, since it's a clarification.

VCFv4.2.tex
```diff
@@ -181,7 +181,7 @@ \subsubsection{Fixed fields}
...
-  \item INFO - additional information: (String, no white-space, semi-colons, or equals-signs permitted; commas are permitted only as delimiters for lists of values) INFO fields are encoded as a semicolon-separated series of short keys with optional values in the format: $<$key$>$=$<$data$>$[,data]. Arbitrary keys are permitted, although the following sub-fields are reserved (albeit optional):
+  \item INFO - additional information: (String, no white-space, semi-colons, or equals-signs permitted; commas are permitted only as delimiters for lists of values) If there is no additional information, then the MISSING value must be used. INFO fields are encoded as a semicolon-separated series of short keys with optional values in the format: $<$key$>$=$<$data$>$[,data]. Arbitrary keys are permitted, although the following sub-fields are reserved (albeit optional):
```
lbergelson, line 184: I think to match the rest of this list the statement about missing values should be at the end.   It looks like missing value is written in lowercase in this context as well.  

Maybe:

`If no keys are present, the missing value must be specified.`  

It's also a bit strange that the `()` description of what goes here is first while in the other list entries it's last.

jmmut:
I changed the wording to `If no keys are present, the missing value must be specified.`, and moved it a bit lower in the paragraph. 

I also made the change in 4.4. I agree we should probably restructure the parenthesis part in 4.4.

jmarshall:
IMHO this rewording remains confusing: the missing value must be specified _for what_? If we want to make this actually clear, what this paragraph really needs is a complete rewrite to be more similar to the ALT section:

> INFO — additional information: Semicolon-separated list of additional information fields, or the MISSING value '`.`' if there are none. [**Then** dive into the details about semicolons not allowed _in values_ but may be percent-encoded, etc etc.]

eyherabh:
@jmarshall Could you please clarify what you mean?

Both ALT and INFO are currently starting with a seemingly parallel structure, namely

<brief description> <values and formats> <not emtpy; use missing value> <something else>

I can see why this parallelism may seem rather forced, as opposed to immediately evident. However, I think that could be attributed to lack of structure, lack of separation of concerns, and different levels of complexity. 

For example, the description of the ALT field starts referring to the "alternate alleles", but by the third sentence it refers to the "Options". Both refer to the same entity and so it would be advisable to use the same subject. Then, formats are followed by their function, then followed by how utilities might behave, followed by more formats. The INFO field is more complex than the ALT field, requiring to describe key-value pairs as opposed to just values, and reserved names.

Having said that, the way it is written now is not terribly unclear to me. After receiding VCFs with empty INFO fields (hence containing two consecutive tabulators), I was looking forward to clarifying if that is VCF compliant, and the text now indicates that it is not and should have had a MISSING VALUE instead.

If you could clarify a little more your expectations/ideas, I could propose some rewrite for consideration.

jmmut:
@eyherabh I agree with you (and @jmarshall) that the INFO structure is understandably different. I didn't want to rewrite the whole thing but I think the proposed wording from jmarshall makes sense. I just added it. I think now it's a bit better regarding both concerns, let me know what you think.

For your case, note that at the beginning of section 1.6 it is already explained that 2 consecutive tabs is not allowed:
> 1.6    Data lines
> All data lines are tab-delimited with no tab character at the end of the line.  The last data line must end with a line separator.  In all cases, missing values are specified with a dot (‘.’).

So this PR is about making it clearer, really.

jmarshall:
I think what I mean is fairly clear, and the comment is unchanged from last September (https://github.com/samtools/hts-specs/pull/440#issuecomment-531151597). The rewrite I have in mind is now PR #507 and I see @jmmut has found my comment clear enough to make a similar update to this PR.

jmmut:
@jmarshall how shall we use your PR? that one and this one slightly conflict. Shall I cherry-pick your commit and put it here? I'll copy that to 4.4 too. I think that's the faster way to get all this done.

jmarshall:
The VCF maintainers have been presented with several proposals, and as always they should use their judgement as to which one or what combination of the proposed text changes leads to the most correct and understandable specification document.

The issue being addressed exists in all of 4.1, 4.2, 4.3, and the draft 4.4.

eyherabh:
@jmarshall As I said, your comment is and was unclear for it ignores the differences between both fields, and hence fails to explain how to constrain the many ways in which the same piece of text might be interpreted. As an example, see the vcfs produced by Beagle. 

@jmmut Thanks. However, as I mentioned in issue 439, "the need for using a period (also referred as missing value) is mentioned there for every other non-required field (i.e. ID, QUAL, FILTER). Therefore, the fact that it is not explicitly stated for the INFO field left me under the impression that the INFO field can be left empty."

PR #432, opened 2019-07-18
Update MAINTAINERS.md

raskoleinonen:
Proposal to add Rasko Leinonen as refget maintainer.

PR #441, opened 2019-08-29
Matt is no longer a maintainer

andrewyatz:
Removing Matt Laird as a maintainer of refget

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -37,7 +37,6 @@ Past VCF/BCF maintainers include Ryan Poplin and David Roazen.
 ### Refget
 
 * Andy Yates (@andrewyatz)
-* Matt Laird (@lairdm)
 * Rasko Leinonen (@raskoleinonen)
 
 [ga4gh-ff]:  https://www.ga4gh.org/howwework/workstreams/#lsg


jmarshall:
I did wonder whether this was intended in #432…

Does refget wish to record @lairdm's contribution via “Past refget maintainers include…”?

ohofmann:
> Does refget wish to record @lairdm's contribution via “Past refget maintainers include…”?

Very much so.

jmarshall:
Re-upping the discussion on this small PR, which doesn’t appear to have been considered when the PR was merged. (@andrewyatz)

andrewyatz:
Sorry John very right I missed that.

jmarshall:
There's certainly no need for commits or PRs doing an explicit revert. A follow-up commit is fine.

Your option 3 (a follow-up commit directly on **master** by the refget maintainers) is entirely in accordance with [MAINTAINERS.md](https://github.com/samtools/hts-specs/blob/master/MAINTAINERS.md#updating-the-specifications): it's both uncontroversial and already-discussed.

andrewyatz:
Thank you will do this now

jkbonfield:
Somewhere in the middle of that I stupidly remembered it's time I updated the CRAM pdf ;-)

If it needs redoing again that's such a trivial task it's a non-issue, but I assume we're not talking about actual rollbacks and are just having a bit more conversational commit history.

PR #444, opened 2019-09-16
[Meta] Codify the existing policy of generally squashing PRs

jmarshall:
Some text that spells out our de facto policy around how to merge PRs (AIUI at least). Mostly our PRs are a single commit that gets modified by a stream of meaningless “fix typo” and “respond to review” commits, so in the usual case that should all be squashed into one commit when the PR is merged.

However there are occasional cases of PRs that tell a story or otherwise have distinct non-trivial commits, so spell out that other treatment for such cases is appropriate.

The **hts-specs** repository currently has all three GitHub UI merge strategies (merge commit; squash; rebase) switched on. It doesn't seem to be possible to change the default; I think it defaults to whatever was selected previously (in the repo / by you / both). We could switch off some of the strategies (i.e. merge commit) and require such things to be done on the command line, but I'm not sure that's better than education + status quo.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -50,6 +50,10 @@ Larger changes should be proposed as pull requests so that they can be discussed
 (Even those with write access to the **samtools/hts-specs** repository should in general create their pull request branches within their own **hts-specs** forks.
 This way when the main repository is forked again, the new fork is created with a minimum of extraneous volatile branches.)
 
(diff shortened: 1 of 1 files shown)


lbergelson:
I know that no matter what the policy is, I will inevitably forget and hit merge with the default option at some point.

jmarshall:
As agreed in the meeting, I've switched off UI “Allow merge commits” on the basis that doing so will prevent most of the inevitable accidents. We'll see how this goes and can revisit if we want to…

PR #446, opened 2019-09-27
Permit AP_Delta in multi-ref slices

### LINKED ISSUE #431: CRAM AP_delta query
The CRAM spec states in the compression header "preservation map" that the AP key indicates whether the AP values are delta encoded: "true if AP data series is delta, false otherwise".

The main data series description for AP is: "if APDelta = true: 0-based alignment start delta
from the previous record. When the record is the first in the slice, its alignment start will be equal to that of the slice, so its alignment delta is 0. if APDelta = false: encodes the alignment start position directly".  Thus the description is couched entirely in terms of the APdelta field in the preservation map and doesn't make reference to multi-ref slices.

Elsewhere we state "The AP data series is delta encoded for reads mapped to a position-sorted slice containing data from a single reference, and as a normal integer value in all other cases."

I assume the logic here is it could involve a large negative jump when we get to the end of one reference and the start of the next.

However, in actuality, the AP series is a signed integer.  It works perfectly fine with negatives in there and can be decoded correctly by both Java and C implementations (I haven't checked JavaScript, but it's likely the same given it's just a decode of itf8).  Thus I wonder if this is a pointless requirement.  It harms compression considerably in certain scenarios, although possibly that's just my poor implementation. :-)  It doesn't really follow that we cannot have a delta for a saw-tooth pattern of positions and the limitation seems arbitrary.

Comments?  Would removing this actually be a main spec change?  it doesn't feel it given this is essentially freeing up a limitation and is completely backwards compatible.

jkbonfield:
This means AP_delta can become negative.  I have validated this
decodes fine in both htsjdk and htslib.  This is because AP is ITF8
and hence signed, like all other integers, so it would need explicit
code to forbid this (which obviously isn't in the implementations).
Hence the limitation is primarily one of an over-zealous specification.

The impact of this is for position-sorted multi-ref slices AP can
legally be stored efficiently.

Also clarified fields in the container compression header when in
multi-ref mode.

Fixes #431

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -450,10 +450,12 @@ itf8 & reference sequence id & reference sequence identifier  or\linebreak{}
 All slices in this container must have a reference sequence id matching this value.\tabularnewline
 \hline
 itf8 & starting position on the reference & the alignment start position or\linebreak{}
-0 for unmapped reads\tabularnewline
+0 if the container is multiple-reference
+or contains unmapped unplaced reads\tabularnewline
 \hline
 itf8 & alignment span & the length of the alignment or\linebreak{}
-0 for unmapped reads\tabularnewline
+0 if the container is multiple-reference
+or contains unmapped unplaced reads\tabularnewline
 \hline
 itf8 & number of records & number of records in the container\tabularnewline
 \hline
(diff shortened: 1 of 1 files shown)


cmnbroad:
👍 LGTM.

PR #447, opened 2019-09-27
Specify defaults to booleans in CRAM preservation map

### LINKED ISSUE #120: [CRAM] Default values for compression header preservation map
I've just noticed that the C CRAM implementation doesn't normally set RR (reference required), unless a referenceless cram is produce (eg scramble -x).  Prior to RR being added we always needed a reference, but now we can say otherwise.

The logical default value therefore is 1; we need a reference unless explicitly told otherwise.   However no default value is listed.

I think we could have the following defaults, or perhaps just state that all the bools default to true if not given?

RN: 1
AP: 1
RR: 1
SM <no default, or pick the one we nearly always use?>
TD: <none>

jkbonfield:
Fixes #120

I decided after consideration that SM and TD should be mandatory.  TD because TL data series is mandatory so we always look up in TD even when no tags are prevent, so we'd have to specify that blank TD as the default.  SM likewise; it's easier to make it mandatory than to specify the default value.

In both cases there appears to be no way in existing code to create CRAMs that does not add these fields.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -615,6 +615,8 @@ section\tabularnewline
 \hline
 \end{tabular}
 
+The boolean values are optional, defaulting to true when absent, although it is recommended to explicitly set them.  SM and TD are mandatory.
+
 \subsubsection*{Data series encodings}
 
 Each data series has an encoding. These encoding are stored in a map with byte[2] 


cmnbroad:
👍 All right, lets call this one done. I've also update my htsjdk branch to reflect this.

jkbonfield:
Are there any significant changes you need to make to the code?  My hope is we're codifying what happens in practice already.

Having said that, I see my own code defaults to zero for read-names included.  (AP_delta and RR default to 1.)

I've no idea where that code originated, or the apparently unused bits of code in htslib/io_lib for qs_included (MI, UI preservation keys) and unmapped_places (PI).   Those last 3 aren't even documented keys, and nothing happens with them other than setting variables.  I assume they were copied over from studying the original cramtools code as they appeared in my initial cram commit.

I have stumbled on my old SVN c-cram project though, that predated it arriving in io_lib.  I may have to go digging. :-)

cmnbroad:
Nothing substantial. The code was already always doing the right thing on write; and it assumed SM and TD would be present on read; I merely codified the latter by throwing an exception if SM or TD are  not detected on read, since otherwise bad things happen in downstream code.

jkbonfield:
Definitely code archaeology here! 

https://github.com/enasequence/cramtools/blob/dbdbe1c24effe06e3542e2ca2929a1ef53f4d1da/src/main/java/net/sf/cram/ReadWrite.java#L264-L284

I can't recall when they got ditched (they were absent in CRAM 2 spec, maybe even in 1.0), but the legacy apparently lives on in my code base.  I'll tidy up such relics.

jkbonfield:
For reference, I managed to find CRAM specifications prior to TeXifying lurking on the EBI web site.  These were written by Vadim in MS Word originally and I don't have the word version for anything other than 2.1 (which I obtained for purposes of converting it to TeX).

https://www.ebi.ac.uk/sites/ebi.ac.uk/files/groups/ena/documents/cram_format_1.0.1.pdf

https://www.ebi.ac.uk/sites/ebi.ac.uk/files/groups/ena/documents/cram_format_2_0.pdf

1.0 has all sorts of differences in the preservation map, while 2.0 is much closer to the one we know.  As I recall while writing C Cram I found lots of disparities between the code and the specification, which is what lead to a unifying of the two and the creation of the 2.0 specification document.  (As such CRAM 1.0 basically never existed in the form documented.)

At no point does it have comments on the default values though.  The ancient Java code does shed some light on this though.  The booleans are all specified with no assignments, so I assume they're defaulting to zero originally.   Practically speaking I think it makes no difference though as all known  implementations of CRAM writers have explicitly been setting these fields, hence a default has never mattered.

cmnbroad:
Yeah I suspect you're right, those are some interesting heirlooms though!

Issue #450, opened 2019-10-16
Clarify section 7.1 CRAM Header in first container


cmnbroad:
Section 7.1 of the CRAM spec describes how the  BAM file header is stored in the first container of a CRAM file, but the description doesn't seem to make sense (or at least seems awkward to implement), and as far as I can tell, doesn't  match what htsjdk or samtools actually do.

htsjdk uses a container with a single raw block that is padded, but not compressed. samtools appears to use two blocks, a gzipped block containing the header, with no padding, and a second raw block of (10000?) zeroes. Since gzipping a padded block doesn't actually result in much/any space actually being reserved, what samtools does seems to make more sense than what is spec'd. I'd propose to make htsjdk and the spec conform to what samtools appears to do, which is to use a second raw block to serve as the padding space.

> 7.1 CRAM header in the first container
> The first container in the CRAM file contains the BAM header in an optionally compressed block. Only gzip is allowed as compression method for this block. BAM header is terminated with \0 byte and any extra bytes in the block can be used to expand the BAM header. For example when updating @SQ records additional space may be required for the BAM header. It is recommended to reserve 50% more space in the CRAM header block than it is required by the BAM header.

cmnbroad:
Closed via #451.

PR #451, opened 2019-10-16
Clarify CRAM section 7.1 BAM header in first container.

cmnbroad:
Proposed clarification for section CRAM spec section 7.1. See issue https://github.com/samtools/hts-specs/issues/450.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -948,7 +948,7 @@ At present no tags are defined.
 
 \subsection{\textbf{Core data block}}
 
-A core data block is a bit stream (most significant bit first) consisting of one 
+A core data block is a bit stream (most significant bit first) consisting of data from one 
 or more CRAM records. Please note that one byte could hold more then one CRAM record 
 as a minimal CRAM record could be just a few bits long. The core data block has 
 the following fields:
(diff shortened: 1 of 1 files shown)


jkbonfield:
I think we should change the text to SAM header instead of BAM header, or better still to CRAM header with a separate paragraph stating that at present the CRAM header format is aligned to the SAM one.  Both BAM and CRAM are compressed representations of SAM and it's not helpful to couch CRAM in terms of BAM.  Additionally BAM actually does have an additional binary header component, so it's ambiguous.

The header is purely textual and matches the SAM specification (but I wouldn't want to paint ourselves into a corner - we may at some point wish to add a new `@xy` style meta-data item purely for use within CRAM).  

Regarding the padding, I explored several ideas when implementing this.  Prior to CRAM 3.0 I just expanded the block and terminated the string with much more than 1 nul.  As you point out that doesn't work for compressed headers.  Hence when we bumped to version 3 we also permitted more than one block to be present, so we could compress the text one and leave an uncompressed block for expansion purposes.  It looks like I add either 50% or 10K - whichever is smaller - but that's purely an implementation decision.  I don't object to mentioning it in a recommendation though.

I did also explore a third route, which was to increase the container size and simply have unallocated space after the block.  Eg a 10K container that holds a 5K compressed block.  I can't recall whether I ran into problems with it, but at some point I dismissed this scheme.

Agreed we should clarify how we want it to work in cram 3.0, which is simply to say we permit additional blocks which may be used as a way to reserve space for header expansion, permitting efficient header replacement.

PS.  I don't understand why we only permitted gzip for compressing the header.  Why not bzip2 for example?  One possible reason could perhaps be we were considering adding extra meta-data to explain codecs used.  This is sadly lacking.  A program that doesn't have lzma support could be happily decoding GBs of CRAM and then suddenly meet an lzma block.  There's no way for it to know it was permitted and may exist somewhere in the file.

CRAMv3.tex
```diff
@@ -477,15 +477,17 @@ \section{\textbf{Container header structure}}
...
-to reserve 50\% more space in the CRAM header block than it is required by the
-BAM header.
+The first container in a CRAM file contains the BAM header in an optionally
+compressed block. Only gzip is allowed as compression method for this block.
+The BAM header container does not include a compression header block.
+
+It is recommended to reserve 50\% more space in the BAM header container than
+is required for the BAM header by optionally padding the container with a second
+raw block consisting of all zeroes. This can be used to subsequently expand the header
+container in place, such as when updating @SQ records, while preserving the relative
```
jkbonfield: "while preserving the *absolute* offsets of all subsequent containers".

CRAMv3.tex
```diff
@@ -477,15 +477,17 @@ \section{\textbf{Container header structure}}
...
+\subsection{\textbf{BAM header in the first container}}
 
-The first container in the CRAM file contains the BAM header in an optionally
-compressed block. Only gzip is allowed as compression method for this block. BAM
-header is terminated with \textbackslash{}0 byte and any extra bytes in the
-block can be used to expand the BAM header. For example when updating @SQ
-records additional space may be required for the BAM header. It is recommended
-to reserve 50\% more space in the CRAM header block than it is required by the
-BAM header.
+The first container in a CRAM file contains the BAM header in an optionally
```
jkbonfield: Maybe something like:

> The first container in a CRAM file contains a textual header in a single block, optionally gzip compressed.
> This text header currently matches the SAM header specification.

From that point on we can refer to CRAM header instead of BAM header.  I can replace the images with tikz pic if you wish.

I say "single block" because that's how it is currently defined, but for future versions I want to amend this to "one or more blocks", or to permit blocks to be able to be more than 2GB in size.  I say this because people have already hit limits in both BAM and CRAM with *huge* headers due to millions of SQ lines.  If we keep blocks are 32-bit sizes, then we'd need to permit multiple blocks, with obvious concatenation properties to construct the header.

I guess it's such a trivially backwards compatible route (multi-blocks, not 64-bit sizes) that we could add that to CRAM 3.1 too.  If we chose that route in 3.1 then we'd also amend the next paragraph with "additional raw blocks" instead of "a second raw block".

CRAMv3.tex
cmnbroad: Oh yeah...Done.

cmnbroad:
I updated this based on comments (I **think** it follows what you suggested), and while I was at it, I added the end-of-file container to the file structure section, since it really should be there. The images do need to be updated (all this latex stuff is pretty new to me so if you think tikz is the way to go that sounds good to me).

jkbonfield:
Looks good.  I'll update the pictures and, if I may, push an extra commit onto your branch.

cmnbroad:
Thanks- feel free to update.

jkbonfield:
See my tikzification commit.

I had to move text around a little and reword some as the figure legends weren't particularly well placed and also seemed to be a bit repetitive, just duplicating the setence before.  Feel free to adjust and tweak as you see fit.

cmnbroad:
@jkbonfield I think these changes look pretty good- thanks for the tikz updates. Not sure what you mean by "the last figure hasn't yet been rewritten" ?

jkbonfield:
A few pages on there is another figure, in a different style but still a .png.

CRAMv3.tex
```diff
@@ -964,23 +964,104 @@ \subsection{\textbf{Core data block}}
...
+\draw[encodingArrow] (p3) + (1.7, 0.0) to (p8);
+\draw[encodingArrow] (p4) + (1.7, 0.0) to (p9);
+\draw[encodingArrow] (p5) + (1.7, 0.0) to (p10);
+
+\filldraw[draw=black, fill=white!10, drop shadow] (1.5, -1.5) ellipse (2.0 and 0.5);
+\filldraw[draw=black, fill=white!10, drop shadow] (1.5, -4.55) ellipse (1.8 and 0.5);
+\filldraw[draw=black, fill=white!10, drop shadow] (1.5, -6.1) ellipse (1.8 and 0.5);
+\filldraw[draw=black, fill=white!10, drop shadow] (1.5, -7.6) ellipse (1.8 and 0.5);
+
+\node [right=of p1, scale=0.7, text centered, text width=14em] {Core (Bit) Encodings (Huffman, RANS, etc.)};
```
jkbonfield: RANS isn't a bit encoding; it only applies to external blocks.  Instead "Huffman, Beta, etc." would be appropriate.

I'm also a little wary of the layout as it sort of implies specifics which the old picture didn't.  We previously had big fat arrows that spanned multiple items on the left and weren't well aligned with them.  The new one looks to have a deliberate one to one alignment with specific types of data.  While this is just an example, I'd be cautious not to hint that specific types are encoding in specific ways.

If  we wish to keep this image, we'll want to change the legend from things like "how a CRAM record _is_ distributed" to "how a CRAM record _may be_ distributed".

PS. More complex Tikz work than I've done so far - a good opener!   One of the reasons I didn't do this was because it'd mean spending more time, which I didn't have, learning Tikz. :-)

jkbonfield: The flip side to the above argument, if we are happy with a diagram showing an example rather than describing the basic concepts without implying anything concrete,  then we should perhaps be more bold and give a realistic example.

Eg using specific types of byte encodings (EXTERNAL,  BYTE_ARRAY_STOP, etc) on obvious numerics and strings.  We could then also add appropriate examples of compression codecs.  Eg BYTE_ARRAY_STOP on read identifiers to an external block compressed using bzip2,  EXTERNAL for quality scores compressed with rANS, and so on.  Maybe pick Bit flags, alignment start for HUFFMAN and BETA encodings.

We need to keep a clear separation of encoding (which may be byte based or bit based) and compression codec (as-is, gzip, bzip2, lzma, rans).  Especially as this is likely something we'll expand on in future.

cmnbroad: Ok, thx for the comments. Point taken on RANS, thats in the wrong place. Also, I had tried to make the left side of the arrows start from the `CRAM Record` box rather than the individual fields, and the right side end at the individual blocks, but perhaps thats too subtle. I'll do another pass.

cmnbroad:
@jkbonfield I took another shot at the image - see what you think.

jkbonfield:
> @jkbonfield I took another shot at the image - see what you think.

Looks great.  Thanks.  Gtg.

Issue #426, opened 2019-06-26
Clarification of CRAM external encodings and blocks


cmnbroad:
There are some inconsistencies in the CRAM spec around the use of the word  “external”, probably due to evolution of the spec over time. I’d like to make a PR to resolve them, but in some cases I’m unclear on the intent.

The text alternately refers to “the external  encoding” (singular), “external encodingS” (plural), and “external blocks". Its not always clear which encodings are considered "external"; whether “external” is actually a property of (one or more) encodings, or just of the containing block; and what if any restrictions there are on the relationship between encoding type and block type. A few quotes:

From “Section 3 Encodings”: 
> Encodings may have parameters of different types, for example _the external encoding_ has only a single parameter, integer id of the external blocks

Then later:
> _Each external encoding_ has an id parameter which must be one of the external block content ids.

From Section 8.1 “Block Content types”: 
> CORE is defined as “bit stream of all encodings except for _external_" (lower case)

From Section 8.1“Block Content ID”:
> There are two types of data blocks: core data blocks and external data blocks.The difference between core and external data blocks is that core data blocks consist of data series that are compressed using bit encodings while the external data blocks are byte compressed 

Elsewhere:

- There is an encoding called EXTERNAL that has a content id parameter.
- The BYTE_ARRAY_STOP encoding has an “external content ID” parameter
- The BYTE_ARRAY_LEN encoding does NOT have an external content ID parameter 
- There is a block content type called EXTERNAL_DATA, defined as containing “data produced by external encodings”

Presumably EXTERNAL, BYTE_ARRAY_STOP, and BYTE_ARRAY_LEN are all “external” encodings (?), and are used in EXTERNAL_DATA blocks, but are they restricted to EXTERNAL_DATA blocks ? Is it the case that the remaining encodings are intended to be “CORE” (is their use permitted in an external block )?

Htsjdk’s implementation is an interesting case. For tag values, it uses a BYTE_ARRAY_LEN encoding, which is a nested encoding, and splits the sub-encodings between core and external blocks by Huffman-encoding the array length into the core block (indeed, that is the ONLY thing htsjdk writes to the CORE block), and putting the array itself into a separate external block. If this is permitted, it it challenges the notion that an encoding can/should be categorized as either core or external.

Any insights appreciated.

jkbonfield:
It's a really confusing term and I don't know where it came from.

However I see there is potential for confusion.  Is "external" any block other than CORE, or any block  with encoding EXTERNAL?

[...]

PR #452, opened 2019-10-16
Distinguish between CRAM external encodings and EXTERNAL proper.

cmnbroad:
Minor clarification of how "external"is used (the smallest change I could make to satisfy https://github.com/samtools/hts-specs/issues/426).


jkbonfield:
Looks good to me.  Happy for it to be merged.

PR #454, opened 2019-10-17
Added BGISEQ as legal value for the PL tag

### LINKED ISSUE #178: Add 'BGISEQ' as a valid value of PL in @RG header record
As we all known, `BGISEQ-500` as an NGS platform have applicated in scientific research and clinic test. But in bam file's header recorder `@RG`, there is no proper option for `PL`, and cause some problem in the downstream analysis (e.g. GATK BQSR). So I advice add `BGISEQ` as a valid option

If you want to know more information of BGISEQ-500, there are tow demo data:
[BGISEQ-500 WGS demo FASTQ data](http://www.ebi.ac.uk/ena/data/view/PRJEB15427)  
[BGISEQ-500 RNA-seq demo FASTQ data](http://www.ebi.ac.uk/ena/data/view/PRJEB15426)

yfarjoun:
in the @RG SAM header line.

fixes #178

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -312,9 +312,9 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   & {\tt LB} & Library.\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
-  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values are}:
- {\tt CAPILLARY},  {\tt DNBSEQ} (MGI/BGI), {\tt HELICOS}, {\tt ILLUMINA}, {\tt IONTORRENT}, {\tt LS454}, {\tt ONT} (Oxford Nanopore), {\tt PACBIO} (Pacific Biotechnology), and {\tt SOLID}.
-  Tag should be omitted when the technology is not in this list, (though the {\tt PM} tag may still be present in this case) or is unknown.\\\cline{2-3}
+  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}:
+  {\tt CAPILLARY}, {\tt DNBSEQ} (MGI/BGI), {\tt HELICOS}, {\tt ILLUMINA}, {\tt IONTORRENT}, {\tt LS454}, {\tt ONT} (Oxford Nanopore), {\tt PACBIO} (Pacific Biotechnology), and {\tt SOLID}.
+  This field should be omitted when the technology is not in this list (though the {\tt PM} field may still be present in this case) or is unknown.\\\cline{2-3}
   & {\tt PM} & Platform model. Free-form text providing further details of the platform/technology used.\\\cline{2-3}
   & {\tt PU} & Platform unit (e.g., flowcell-barcode.lane for Illumina or slide for SOLiD). Unique identifier.\\\cline{2-3}
   & {\tt SM} & Sample. Use pool name where a pool is being sequenced.\\\cline{1-3}
@@ -1411,6 +1411,7 @@ at \url{https://github.com/samtools/hts-specs}.}
 \subsection*{1.6: 28 November 2017 to current}
 
 \begin{itemize}
+\item Add {\tt DNBSEQ} to the list of {\tt @RG PL} header tag values. (Apr 2020)
 \item Add {\tt @SQ TP} circular/linear topology header tag. (May 2019)
 \item\textbf{Restricted the allowable punctuation characters in reference sequence names} (in {\tt @SQ SN}, {\sf RNAME}, etc).
 The sets of characters allowed in {\tt @SQ SN} and {\tt @SQ AN} are now identical, which enlarges the previous {\tt AN} set. (Jan 2019)
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -312,8 +312,9 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   & {\tt LB} & Library.\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
-  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}:
-  {\tt CAPILLARY}, {\tt LS454}, {\tt ILLUMINA}, {\tt SOLID}, {\tt HELICOS}, {\tt IONTORRENT}, {\tt ONT}, and {\tt PACBIO}.\\\cline{2-3}
+  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}: 
+  {\tt BGI}, {\tt CAPILLARY}, {\tt LS454}, {\tt ILLUMINA}, {\tt SOLID}, {\tt HELICOS}, {\tt IONTORRENT}, {\tt ONT}, {\tt PACBIO}, and {\tt OTHER}. 
+  Nacent technologies that have not yet been added to the spec should use {\tt OTHER}. \\\cline{2-3}
   & {\tt PM} & Platform model. Free-form text providing further details of the
   platform/technology used.\\\cline{2-3}
   & {\tt PU} & Platform unit (e.g., flowcell-barcode.lane for Illumina or slide for SOLiD). Unique identifier.\\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -312,11 +312,10 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   & {\tt LB} & Library.\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
-  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}: 
-  {\tt BGI}, {\tt CAPILLARY}, {\tt LS454}, {\tt ILLUMINA}, {\tt SOLID}, {\tt HELICOS}, {\tt IONTORRENT}, {\tt ONT}, {\tt PACBIO}. 
-  Nacent technologies that have not yet been added to the spec, should omit the tag, and optionally put identifying information into the {\tt PM} tag.  \\\cline{2-3}
-  & {\tt PM} & Platform model. Free-form text providing further details of the
-  platform/technology used.\\\cline{2-3}
+  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values are}:
+  {\tt BGI}, {\tt CAPILLARY}, {\tt HELICOS}, {\tt ILLUMINA}, {\tt IONTORRENT}, {\tt LS454}, {\tt ONT}, {\tt PACBIO}, {\tt SOLID}.
+  Tag should be omitted when the technology is not in this list, (though the {\tt PM} tag may still be present in this case) or is unknown.\\\cline{2-3}
+  & {\tt PM} & Platform model. Free-form text providing further details of the platform/technology used.\\\cline{2-3}
   & {\tt PU} & Platform unit (e.g., flowcell-barcode.lane for Illumina or slide for SOLiD). Unique identifier.\\\cline{2-3}
   & {\tt SM} & Sample. Use pool name where a pool is being sequenced.\\\cline{1-3}
   \multicolumn{2}{|l}{\tt @PG} & Program. \\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -313,7 +313,7 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
   & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values are}:
- {\tt CAPILLARY},  {\tt DNB} (MGI/BGI's DNA NanoBalls), {\tt HELICOS}, {\tt ILLUMINA} (can be used for either 4- or 2-color chemistry, patterned or unpatterned flowcells), {\tt IONTORRENT}, {\tt LS454}, {\tt ONT} (Oxford Nanopore), {\tt PACBIO} (Pacific Biotechnology), and {\tt SOLID}.
+ {\tt CAPILLARY},  {\tt DNBSEQ} (MGI/BGI), {\tt HELICOS}, {\tt ILLUMINA}, {\tt IONTORRENT}, {\tt LS454}, {\tt ONT} (Oxford Nanopore), {\tt PACBIO} (Pacific Biotechnology), and {\tt SOLID}.
   Tag should be omitted when the technology is not in this list, (though the {\tt PM} tag may still be present in this case) or is unknown.\\\cline{2-3}
   & {\tt PM} & Platform model. Free-form text providing further details of the platform/technology used.\\\cline{2-3}
   & {\tt PU} & Platform unit (e.g., flowcell-barcode.lane for Illumina or slide for SOLiD). Unique identifier.\\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -314,7 +314,9 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
   & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}: 
   {\tt BGI}, {\tt CAPILLARY}, {\tt LS454}, {\tt ILLUMINA}, {\tt SOLID}, {\tt HELICOS}, {\tt IONTORRENT}, {\tt ONT}, {\tt PACBIO}, and {\tt OTHER}. 
-  Nacent technologies that have not yet been added to the spec should use {\tt OTHER}. \\\cline{2-3}
+  Nacent technologies that have not yet been added to the spec, should use {\tt OTHER}. 
+  In cases where the technology is unknown, the tag should be omitted.
+  Further details can be put into the {\tt PM} tag.  \\\cline{2-3}
   & {\tt PM} & Platform model. Free-form text providing further details of the
   platform/technology used.\\\cline{2-3}
   & {\tt PU} & Platform unit (e.g., flowcell-barcode.lane for Illumina or slide for SOLiD). Unique identifier.\\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -313,7 +313,7 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
   & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values are}:
-  {\tt BGI}, {\tt CAPILLARY}, {\tt HELICOS}, {\tt ILLUMINA}, {\tt IONTORRENT}, {\tt LS454}, {\tt ONT}, {\tt PACBIO}, {\tt SOLID}.
+ {\tt CAPILLARY},  {\tt DNB} (MGI/BGI's DNA NanoBalls), {\tt HELICOS}, {\tt ILLUMINA} (can be used for either 4- or 2-color chemistry, patterned or unpatterned flowcells), {\tt IONTORRENT}, {\tt LS454}, {\tt ONT} (Oxford Nanopore), {\tt PACBIO} (Pacific Biotechnology), {\tt SOLID}.
   Tag should be omitted when the technology is not in this list, (though the {\tt PM} tag may still be present in this case) or is unknown.\\\cline{2-3}
   & {\tt PM} & Platform model. Free-form text providing further details of the platform/technology used.\\\cline{2-3}
   & {\tt PU} & Platform unit (e.g., flowcell-barcode.lane for Illumina or slide for SOLiD). Unique identifier.\\\cline{2-3}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -313,7 +313,7 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
   & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values are}:
- {\tt CAPILLARY},  {\tt DNB} (MGI/BGI's DNA NanoBalls), {\tt HELICOS}, {\tt ILLUMINA} (can be used for either 4- or 2-color chemistry, patterned or unpatterned flowcells), {\tt IONTORRENT}, {\tt LS454}, {\tt ONT} (Oxford Nanopore), {\tt PACBIO} (Pacific Biotechnology), {\tt SOLID}.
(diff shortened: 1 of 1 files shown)


nh13:
Can we not have “other” and just not specify the tag?

yfarjoun:
In the call today there was a discussion discussion about this. The point here is to distinguish between not saying anything (like "unknown") and saying "other" which is stating that it is not one of the preceding values. 

You may disagree on the merits of being able to distinguish between these two states....

colinhercus:
How about adding MGI as well. See   https://en.mgitech.cn/

nh13:
@yfarjoun I like it, how about adding that if the platform is unknown, the tag should not be other?

jmarshall:
Re BGI vs BGISEQ: the existing list of values mostly reflects underlying technologies rather than manufacturers (though it's a bit of a mishmash), so I'd prefer going with what was requested — but that request was three years ago of unknown provenance, so hopefully a knowledgeable source may add further information to #178.

The list of values is currently more or less ordered by vintage/dominance rather than alphabetical. So if we're adding this new entry, either add it at the end or fully alphabetise the list.

jmarshall:
My problem with `OTHER` is that it's not something that anyone would ever choose to write in this field. “I know I'm using a Frobnitz-500, so why can't I put Frobnitz there?”

We list a well-known vocabulary in the spec so that tools can infer the underlying technology's error model etc without having to look for 18 different spellings/capitalisations of `ILLUMINA`/`HISEQ`/etc. That doesn't mean that this needs to be a fully-controlled vocabulary or that it's useful to validate that the `PL` value is one of those listed.

What if instead we open this up via the usual lowercase value mechanism:

> Value may be one of `CAPILLARY`, […], `PACBIO`, or a locally-defined lowercase keyword matching `/[a-z0-9]+/` [footnote: Contact the spec maintainers to add uppercase versions of new values that may be of general interest]. If the platform is unknown, the `PL` tag should be omitted.

Possibly the last sentence can be left out (as without the temptation of `OTHER` it should be obvious), as that's taken for granted for all these fields.

(This would also solve the BGI/MGI problem (FYI @colinhercus reports on #178 that they're actually the same thing): we could leave it off the list until further demand produces another request.)

colinhercus:
@jmarshall, From my experience HiSeq and NextSeq have very different sequencing error models so I'm not sure how helpful PL:ILLUMINA is!

jmarshall:
@colinhercus: Indeed. That's the historical reason for having a well-known PL but free-form PM, but it already broke down when we didn't differentiate 2-colour Illumina, alas — so you get to look for NextSeq/NovaSeq in PM after all…

jkbonfield:
I could go with upper/lowercase namespacing as an alternative to OTHER.

yfarjoun:
@jmarshall as long as we only allow certain values, there's merit in allowing "OTHER" for exactly the case where there's no valid value, but there's another reason: there are nascent technologies that are still stealthy enough that they would not want to declare what the technology is, but for the purposes of collaboration would like to say *something* in the PL field...

jmarshall:
By “for the purposes of collaboration” do you mean that Picard in fact does not allow you to omit PL to indicate that the platform is unknown or undisclosed?

Well…

yfarjoun:
@jmarshall what's the point of specifying in the spec what the VALID VALUES are, and then allowing sam files that have other values there?

yfarjoun:
@jkbonfield for some historical reason, htsjdk ignores case when comparing to the allowed values...so lower-casing isn't going to help...

jkbonfield:
> @jmarshall what's the point of specifying in the spec what the VALID VALUES are, and then allowing sam files that have other values there?

It at least means people don't invent other synonymous valid values when they see a term already in use.

I still think there is merit in uppercase vs lowercase.  I'm sure htsjdk could be modified to not complain about lowercase ones when it's doing the validation?  I don't think we need to worry about case insensitivity (infact it'd be a bonus to have it) as we're not proposing having both upper and lowercase tags with the same text meaning different things.  It's simply a way of pointing out to people that it's not yet "blessed" in the official dictionary.

jkbonfield:
> How about adding MGI as well. See https://en.mgitech.cn/

MGI is a subsidiary of BGI.   The "PL"atform field is an unfortunate mishmash of all sorts and it's a bit vague where a platform finishes and a model starts.  For example NovaSeq is a model and ILLUMINA is the platform despite it also being the company name.  Conversely SOLID is a platform, not ABI which was the company involved, perhaps because ABI were also ubiquitous with earlier sequencing platforms too (eg CAPILLARY based Sanger sequencing).

For MGI vs BGI, I can see benefits either way, but we only want one and not both.  If they're already starting to be populated in the wild then we should bless that one in this spec.  Otherwise my gut feeling is MGI but I don't object to BGI either.

jmarshall:
In fact, extant HTSJDK validates that PL is present and is one of the existing known values (see samtools/htsjdk@db13cd397ecd1b30fe1c7da1109cac3a34ac3b48) and since samtools/htsjdk@1e6d054d29127368d6bbfc6a414aa3ffe87411d3 does so case-insensitively (as an unexplained extension over the spec). So in fact HTSJDK is already accepting values outwith the spec.

As @jkbonfield replied and as I said in https://github.com/samtools/hts-specs/pull/454#issuecomment-543623293, the point of specifying the WELL-KNOWN VALUES but nevertheless allowing other values is that tools can depend on `SOLID` implying that the reads have Solid properties and similarly for other well-known values, but allowing other things to be specified. However I take your point that the world is full of bears of little brain who may miss this point and start writing `solid` or `solidish` or `illuminaseq`.

It is unfortunate that HTSJDK has allowed lowercase PL values equivalently to the well-known uppercase ones, but this does not itself prevent extending the value space to allow locally-defined lowercase values.

I remain opposed to adding `OTHER` as it is a meaningless additional defined value. To my mind, we have two sensible approaches to choose between:

1. (The status quo option.) Keep the list of valid values as is (adding `BGI`/`BGISEQ`/`MGI`/`DNBSEQ` separately as appropriate, as per #178), and explicitly allow the omission of PL and spell out what that indicates:

  > Platform/technology used to produce the reads. Valid values: CAPILLARY, LS454, ILLUMINA, SOLID, HELICOS, IONTORRENT, ONT, and PACBIO. Omitted for platforms not in this list (in which case PM can be used to indicate the platform [to human readers]) or when the platform is unknown.

   This would require HTSJDK changes to allow PL to be omitted, but there was never any justification in the spec for validating its presence anyway. (Can any Broadites look up PO-326's justification for validating PL presence in all cases? Presumably it's another case of “it _should_ be present in usual sequencing protocols, hence enforce”.)

2. (The complicated extensible option.) Valid values are the existing uppercase list or locally-defined lowercase keywords, as suggested in https://github.com/samtools/hts-specs/pull/454#issuecomment-543623293. This has the disadvantage of complexity compared to the very simple unambiguous status quo of “if the reads are known to be one of these well-known technologies, tools can easily detect this by the exact keyword in PL”.

Me, I now favour (1). It's basically equivalent to `OTHER` in its effect on implementations: either HTSJDK needs modification to allow `OTHER` or HTSJDK needs modification to allow omitted PL.

(1) has the advantage of not adding a meaningless keyword value to the spec, while still distinguishing between the unknown and other states (cf https://github.com/samtools/hts-specs/pull/454#issuecomment-543454950) via PM.

jmarshall:
Re https://github.com/samtools/hts-specs/pull/454#issuecomment-543638325, we should also discuss belatedly adding `ILLUMINA:2COLOR` (and perhaps `ILLUMINA:4COLOR`) or similar alongside the post-nextseq ambiguous `ILLUMINA`, so that centres can move to something that specifies which of the colour chemistry error model applies.

yfarjoun:
from call: split into two PRs, OTHER and BGISEQ

daviesrob:
A list of platforms and instrument models can be found in ftp://ftp.sra.ebi.ac.uk/meta/xsd/sra_1_5/SRA.common.xsd (part of the EBI's specification for XML metadata files that accompany data uploaded to the ENA).  Look for `<xs:complexType name="PlatformType">` to find the list of platforms.  Model names start from `<xs:simpleType name="type454Model">`.

jkbonfield:
A bit of a diversion.  IMO htsjdk/picard is perfectly entitled to emit warnings about fields it would prefer to have, and perhaps even have a mode where those warnings become hard failure (like -Werror on  compiler), but the default really should be to accept any legal file without *failure*.

Regarding this PR though, as @yfarjoun said above we'll remove OTHER for now and have that discussion separate, but the remaining issue is BGI vs MGI.  If anyone in the know has any key insight to this now is the time to raise it.  (We don't want both.)

jkbonfield:
> A list of platforms and instrument models can be found in ftp://ftp.sra.ebi.ac.uk/meta/xsd/sra_1_5/SRA.common.xsd (part of the EBI's specification for XML metadata files that accompany data uploaded to the ENA). Look for `<xs:complexType name="PlatformType">` to find the list of platforms. Model names start from `<xs:simpleType name="type454Model">`.

Thanks Rob.  That's useful and it's good to see the PL fields at least match up so we haven't forgotten something.

The list of choices for PM look quite thorough.  If we wish to add this to our spec it should probably be an appendix (unless it doesn't look *too* long as a footnote) and cite the source so any revisions can be discovered in case we're laggy on updating it.

lbergelson:
It looks like that list isn't entirely exhaustive, it looks like it's missing most BGI / MGI models.

kbergin:
Hi! I wanted to mention another option that could be considered for the PM field controlled vocabulary list. There exists sequencing platform standard terminology in EBI's OBI ontology here https://www.ebi.ac.uk/ols/ontologies/obi/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FOBI_0400103&viewMode=All&siblings=false

This is a public and controlled set of ontologies, and offers both specific names and IDs. New ontology fields can be requested as needed.

The ENCODE project uses the same Ontology for their metadata, for example here: https://www.encodeproject.org/files/ENCFF432YLJ/?format=json.

One downside is the naming seems to duplicate the PL field a bit. i.e. "Illumina HiSeq 2500"

jkbonfield:
That looks very out of date.  Maybe it should be populated with the information from the other EBI ftp link @daviesrob linked above.

Eg there's no ONT, no capillary sequencing instruments, only has 1 of the 4 PacBios, no BGI/MGI, etc.

I think we'd be better linking with the ENA ontology as they have a much more vested interest in keeping theirs up to date - they cannot accept data submissions without doing so.

raskoleinonen:
ENA shares accepted instrument models with the other two INSDC members: NCBI and DDBJ. As has been pointed our previously in the thread, they can be extracted from the SRA.common.xsd that is hosted, among other possible locations, here:

ftp://ftp.sra.ebi.ac.uk/meta/xsd/sra_1_5/SRA.common.xsd

PlatformType has all supported sequencing platforms. Instrument model enumerations are in the INSTRUMENT_MODEL element under each platform.

During submissions processing ENA also supports a number of synonyms for some of the instrument models, e.g. we could support both 'NextSeq 500' and 'Illumina NextSeq 500'. We would then map synonyms to the XML Schema enumerations.

I agree that it would be great to have these as an ontology somewhere to link to. If someone connects me with the right people I could maintain the list of instruments in an ontology as they get added to the SRA schema.

> It looks like that list isn't entirely exhaustive, it looks like it's missing most BGI / MGI models.

This is also true and the reason is that the INSDC has not been asked by submitters yet to support depositions from these models. Normally, INSDC adds instruments as needed by submitters, but we could look at changing this process slightly and add the instruments first in a ontology and then later, as needed, add them to the SRA.common.xsd.

kbergin:
Ah that's a great point - I'm definitely not an expert in this, I thought it'd be great to leverage an existing ontology, but I am reading that the main difference between ontology and controlled vocabulary is the relationships between terms, the management of the list, and how the information is expressed i.e. ontologies being in an ontology representation language. It's possible a representation language is not useful here?

I missed that the ones in the SRA field were pulled from an ontology, the person I asked at EBI about which ontology is used for SRA said that SRA uses a controlled vocabulary and that "INSDC works in a slightly different way for these things". She did say she's happy to have a call to explain more, but I didn't follow up.

In the end the main win is if there is another group who's job it already is to keep the list of terms up to date so this community doesn't have to, right?

yfarjoun:
I think that this PR is ready to merge.... @jmarshall ? @jkbonfield ?

jmarshall:
This comment hasn't been addressed (https://github.com/samtools/hts-specs/pull/454#issuecomment-543606559):

> The list of values is currently more or less ordered by vintage/dominance rather than alphabetical. So if we're adding this new entry [BGI], either add it at the end or fully alphabetise the list.

Spelling: Nascent; specification; prob no comma before “and optionally”; whitespace at ends of lines.

But for the “Nascent technologies” sentence I prefer the suggested sentence in #461 as it more blandly describes what is valid for the field without trying to describe (and so potentially tie down) the motivations of people wanting to omit the field.

jmarshall:
Also to record here something that I believe we've discussed during meetings: As far as I'm aware, we have not been supplied with or come up with any actual reasons for favouring any particular one of `BGI` / `BGISEQ` / `DNBSEQ` / `MGI` / etc, and we remain unaware of how people using these instruments currently refer to the technology (cf #178 discussions).

This PR has simply arbitrarily settled on `BGI` on the basis that (1) it was thought of first, (2) we want to choose (a single) something, and (3) the actual tag value chosen doesn't really matter technologically (though it probably somewhat does politically).

jkbonfield:
Agreed that the sentence in #461 is clearer.

With regards to BGI vs MGI vs MGISEQ, frankly it doesn't really matter hugely.  However we've got to pick something, so MGI or BGI both work equally well IMO.  I'd avoid a product name like MGISEQ as we can see with Illumina there are GA, HiSeq, MiSeq, NovaSeq, etc.  The only bit that hasn't changed is the company name, but they may well bring out another instrument *type* (*platform*).  Hence we'd have to name it then.

The various ontologies out there show what a total mess this is already.  EBI's contains "MiSeq", "NextSeq 500" and "Illumina NovaSeq 6000" so model number and/or company name appear and disappear in a rather adhoc manner.  We're already ahead of the game a bit here in that we separate out platform from model.  Whether to document a set of known terms for PM (while still permitting free form new ones) should probably be a topic of a subsequent PR.

SAMv1.tex
```diff
@@ -308,10 +308,10 @@ \subsection{The header section}
   & {\tt LB} & Library.\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
-  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}:
-  {\tt CAPILLARY}, {\tt LS454}, {\tt ILLUMINA}, {\tt SOLID}, {\tt HELICOS}, {\tt IONTORRENT}, {\tt ONT}, and {\tt PACBIO}.\\\cline{2-3}
-  & {\tt PM} & Platform model. Free-form text providing further details of the
-  platform/technology used.\\\cline{2-3}
+  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values are}:
+  {\tt BGI}, {\tt CAPILLARY}, {\tt HELICOS}, {\tt ILLUMINA}, {\tt IONTORRENT}, {\tt LS454}, {\tt ONT}, {\tt PACBIO}, {\tt SOLID}.
```
jkbonfield: The previous list was "a, b, c, and d" style.  This misses the and.

My thought is rather "meh!", but grammatically it's a bit better with "and".  Apart from that, thumbs up from me.

SAMv1.tex
```diff
@@ -308,10 +308,10 @@ \subsection{The header section}
   & {\tt LB} & Library.\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
-  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}:
-  {\tt CAPILLARY}, {\tt LS454}, {\tt ILLUMINA}, {\tt SOLID}, {\tt HELICOS}, {\tt IONTORRENT}, {\tt ONT}, and {\tt PACBIO}.\\\cline{2-3}
-  & {\tt PM} & Platform model. Free-form text providing further details of the
-  platform/technology used.\\\cline{2-3}
+  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values are}:
```
jmarshall: Other field descriptions use “Regular expression:” or “Valid values:” without “is” or “are”.

```suggestion
  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}:
```

SAMv1.tex
```diff
@@ -308,10 +308,10 @@ \subsection{The header section}
...
+  {\tt BGI}, {\tt CAPILLARY}, {\tt HELICOS}, {\tt ILLUMINA}, {\tt IONTORRENT}, {\tt LS454}, {\tt ONT}, {\tt PACBIO}, {\tt SOLID}.
+  Tag should be omitted when the technology is not in this list, (though the {\tt PM} tag may still be present in this case) or is unknown.\\\cline{2-3}
```
jmarshall: The comma is misplaced.

Personally I prefer just “Omitted when the […]” because it side-steps concerns about _must_ vs _should_ etc — it just _is_. But if you really want to put something more in front, it would gain from “This tag” or so; and “field” is better than “tag” because it clearly encompasses the whole `TAG:VALUE` item.

```suggestion
  This field should be omitted when the technology is not in this list (though the {\tt PM} field may still be present in this case) or is unknown.\\\cline{2-3}
```

jmarshall:
IMHO the recent news stories about MGI's new DNBSEQ developments show that `BGI` is increasingly not the right term to use. But I have had little success in encouraging people familiar with these instruments to come and comment here…

jkbonfield:
Good point.    I'm struggling to find MGISEQ mentioned anywhere on mgitech.cn, but I'm not sure if this is just product renaming or something new.  Older articles elsewhere talk about "MGISEQ-T7 powered by DNBSeq".  I also found [this](https://en.mgitech.cn/products/instruments_info/2/):

> Remark: DNBSEQ-G400 remain their original product names as MGISEQ-2000 respectively in China and some other overseas countries.

So this is what I feared with naming it MGISEQ and why that one wouldn't be my preferred choice - product names can and do change willy-nilly. :(

"Platform" is key here; the fundamental technology.  We could argue that DNBSEQ is more in tune with the chemistry than MGISEQ and more descriptive than MGI or BGI, but I'm not sure it matters provided we're consistent.  Does anyone know if this is fundamentally the same technology that was originally developed at BGI?  If so I'd stick with BGI and if at a later date MGI comes up with a radical new chemistry then we can distinguish it by adding MGI to the list.

colinhercus:
One of our users here in Malaysia has an MGISEQ2000  and as far as I know
it and all there other sequencers use DNA Nano Ballls (DNB) and two colour
sequencing. There is also one new very high capacity robotic sequencer
<https://www.technologyreview.com/s/615289/china-bgi-100-dollar-genome/>
just announced that while mechanically is very different (no flow cells)
the chemistry is probably similar.
I think there have been some changes in how they tag nucleotides with
phosphors on this new machine

jmarshall:
Thanks Colin. Perhaps we could consider a late bid for spelling this out as something like

> `CAPILLARY`, `DNBSEQ` (for BGI/MGI instruments), …, `ONT` (Oxford Nanopore), …

So all the manufacturer buzzwords are in the text but the platform is the keyword. And hints in parentheses like this would help me at least — I have to think about what `ONT` is every time I read this list!

jkbonfield:
I don't like DNBSEQ.  It's a product name and can change, as we've seen (it was formerly MGISEQ).  It'd be like adding NOVASEQ for example.

The others, with the exception of the earliest ones, are named after companies, which tends to suggest "MGI" for consistency.  Anything more specific belongs in the PM field.  Yes a company could come out with a radical different technology - eg if Illumina released a nanopore long read instrument - but we can deal with that if and when it happens and so far it's not been an issue.

[ I'm also thinking ideally we may want a CH free text field for chemistry, at least to indicate things like "no PCR" which may have a bigger impact on the error models than variance between two instrument models.  However that's probably a different PR. :-) ]

SAMv1.tex
```diff
@@ -312,10 +312,10 @@ \subsection{The header section}
   & {\tt LB} & Library.\\\cline{2-3}
   & {\tt PG} & Programs used for processing the read group.\\\cline{2-3}
   & {\tt PI} & Predicted median insert size.\\\cline{2-3}
-  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}:
-  {\tt CAPILLARY}, {\tt LS454}, {\tt ILLUMINA}, {\tt SOLID}, {\tt HELICOS}, {\tt IONTORRENT}, {\tt ONT}, and {\tt PACBIO}.\\\cline{2-3}
-  & {\tt PM} & Platform model. Free-form text providing further details of the
-  platform/technology used.\\\cline{2-3}
+  & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values are}:
+ {\tt CAPILLARY},  {\tt DNB} (MGI/BGI's DNA NanoBalls), {\tt HELICOS}, {\tt ILLUMINA} (can be used for either 4- or 2-color chemistry, patterned or unpatterned flowcells), {\tt IONTORRENT}, {\tt LS454}, {\tt ONT} (Oxford Nanopore), {\tt PACBIO} (Pacific Biotechnology), {\tt SOLID}.
```
jkbonfield: "DNB" is very nonspecific.  MGI I think acquired this from Complete Genomics, but the instruments are very different.  It's more like a data prep thing, which we tend to not describe here.  Eg rolling circle amplification, mate pair vs paired-end sequencing, etc.   I think it's not a good term for the PLatform field.

I also dislike all the extra verbiage which IMO isn't necessary here.

yfarjoun:
hmmm. I actually prefer technology names over company names...companies change (think solexa, and pacbio almost) but the technology stays the same...It's unfortunate that we have so many company names as valid values in PL, but we can start putting in the "correct" values as new technologies emerge, rather than simply asserting that everything from illumina is effectively the same....

yfarjoun:
I went with DNB, since it seems silly to add "SEQ" to all the SEQuencing platforms...

jkbonfield:
> hmmm. I actually prefer technology names over company names...companies change (think solexa, and pacbio almost) but the technology stays the same...It's unfortunate that we have so many company names as valid values in PL, but we can start putting in the "correct" values as new technologies emerge, rather than simply asserting that everything from illumina is effectively the same....

Maybe, but technology is shared across multiple vendors and often coupled with very different instruments.  DNB being an example of this.   Also what would you do if another Nanopore company came into the market?  ONT defines the instrument, but if we'd labelled them as a generic "Nanopore" it would then be confusing and misleading.

yfarjoun:
It never ceases to amaze me how difficult it is to keep everyone happy :-)

let's duke it out in the call.....

jkbonfield:
A quick summary from conference call, incase anyone is watching.

We agreed DNBSEQ would work.  It's the name of their whole suite of instruments so is close enough to "platform", and it's likely to be specific enough to that company that it's not going to be reused by anyone else (we never had a platform field for Complete Genomics anyway).  It also means if MGI comes up with a radically different technology then we haven't painted ourselves into a corner.  (Using the company name only was probably an error in the past and doesn't justify continued repetition.)

To that ilk there was some discussion on Illumina 2-colour vs 4-colour and whether they should be separated out, and also where is appropriate to note chemistry specific mods (eg PCR-free vs with PCR on the same instrument/model), but that'll be a topic for a future PR and shouldn't hold this up.

jmarshall:
I've taken the liberty of pushing a suggested minor typographic commit with fixes for the two remaining comments on https://github.com/samtools/hts-specs/pull/454/commits/bff98ccf95e1f74d9f6946748387708f7e5ccbfb. It also adds an entry to the version history appendix.

:+1: from me for squash'n'merge with these minor fixes or an equivalent. Thanks for persevering and finally dragging this over the finish line… :smile:

yfarjoun:
@jkbonfield your thumb?

jkbonfield:
> 
> @jkbonfield your thumb?

Apologies for missing that notification.  Yes it gets the thumbs up for me.  Thanks for the changes.

yfarjoun:
🎉

splaisan:
Thanks a lot @yfarjoun, great fix, will it be in the next release and when is this planned? (I prefer not to install the dev versions)

PR #457, opened 2019-11-20
Fix minor typo in CSI

r6eve:
- auxilary -> auxiliary


jmarshall:
Thanks. I've added this to some others that I had stashed after your last batch of spelling fixes and merged it as 39feb094ae003273b101e96a4ebfa83dd2572120.

PR #461, opened 2019-12-12
Clarify that RG-PL is omitted for unknown/other platforms (alternative to PL:OTHER)

jmarshall:
As a concrete proposal for an alternative to `@RG PL:OTHER`, here is a sentence clarifying what is indicated when `PL` is omitted (and thus also clarifying that omitting it is valid).

This is option (1) from <https://github.com/samtools/hts-specs/pull/454#issuecomment-553941441>. To my mind, this has equivalent expressiveness to the `OTHER` proposal but avoids adding a meaningless keyword to the specification.

See also previous discussion on #454.


yfarjoun:
included in #454 

I think that this can be closed.

jmarshall:
Yes, now subsumed into the original PR.

PR #463, opened 2019-12-12
CRAM: Clarify NM and MD creation

### LINKED ISSUE #453: Clarify NM and MD creation in CRAM
The CRAM decoder can compute MD and NM on the fly for files using reference based compression (either external or embedded).  Both Java and C implementations do this.  It appears I added code to do this in Feb 2013 https://github.com/jkbonfield/io_lib/commit/51c2610a726fccedb42e75b214161c1d0afa9daa and Vadim's predates that in Oct 2012 https://github.com/enasequence/cramtools/commit/181f75237e9b9a85cdc10acf09240c0997ca9e3f

Hence I assume I just copied Vadim's ideas here.

The problem though is any mention of MD and NM is totally absent from the specification itself.  It perhaps seemed trivial initially, but since then I've accrued various tweaks and bits of logic.  Eg:

-  If the tag dictionary already contains MD then we have a verbatim MD string encoded for that record.  Don't auto-generate it.  I did this to silence the whingers who complained CRAM wasn't round tripping correctly.  (it wasn't as it bug fixed MD strings that overhang the end of the reference but apparently that isn't desireable.)

- We have no way of knowing whether MD and NM were previously on the record.  Hence we either always create them or never create them based on command line options.  For V4 prototype I have a strategy of adding MDZ to tag dictionary for MD stored verbatim, no MD anything for do-not-generate and MD* for a request to generate an MD tag.

- There's no way of knowing whether MD is computed using the embedded reference vs the external reference, if both are present and differ!  This is the main reason for this ticket as I went to seek guidance and realised it was absent.

For this last point (this is V4 thinking again) I could consider an embedded consensus differing to the external reference and using the external ref for auto-MD generation so we don't have to store it verbatim when cons vs ref differs.  Or alternatively when we have embedded cons and no external ref, but have a literal MD tag (perhaps with appropriate flag, eg MDz instead of MDZ?) we use this as a way to learn how the actual reference differed to the embedded one so subsequent reads don't also need to explicitly store MD.  That's a little bit ghastly though! :)

Note RG also falls in this same category, but it's a bit different in construction.

Proposal for spec:

- Explicitly define that tags MD, NM may optionally be computed on the fly.  (Discussion point - does the embedded reference get used in preference to external if external is also avaialable?)
- Explicitly define that RG tag is computed from the RG data series if this is not set to -1.

jkbonfield:
Fixes #453 (mostly).

It doesn't cover the nuances I want for CRAM 4.0, but that will be a separate
commit.  It also glosses over the difference between embedded references and
external references.  We not currently state whether they must be identical
and this is a difference in the htslib and htsjdk implementations so it is
left undecided for now.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -1469,6 +1469,18 @@ Type is \texttt{c} (signed 8-bit integer), \texttt{C} (unsigned 8-bit integer), 
 
 For example a SAM tag \texttt{MQ:i} has name \texttt{MQ} and type \texttt{i} and will be decoded using one of MQc, MQC, MQs, MQS, MQi and MQI data series depending on size and sign of the integer value.
 
+Note some auxiliary tags can be created automatically during decode so can optionally be removed by the encoder.
+However if the decoder finds a tag stored verbatim it should use this in preference to automatically computing the value.
+
+The RG (read group) auxiliary tag should be created if the read group (RG data series) value is not $-1$.
+
+The MD and NM auxiliary tags store the differences (an edit string) between the sequence and the reference along with the number of mismatches.
+These may optionally be created on-the-fly during reference-based sequence reconstruction and should match the description provided in the SAMtags document.
+An encoder may decide to store these verbatim when no reference is used or where the automatically constructed values differ to the input data.
+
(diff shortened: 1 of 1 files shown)


cmnbroad:
:+1:

Issue #448, opened 2019-09-30
VCF Format: LOH regions and mosaic copy number regions


enwired:
My software group has been required to write VCF files containing, among other things, regions of LOH (Loss of Heterozygosity) and copy number regions.

I would like to know how to do this in the most standard way, if there is one.

We know that copy number variants are usually described this way

##FORMAT=<ID=CN,Number=1,Type=Integer,Description="Copy number genotype for imprecise events">

But we make copy number calls that are not always integer. For example "2.5" could indicate mosaic copy number where half of the sample has CN=2 and half has CN=3. (By itself, 2.5 is ambiguous. It could actually be any mixture that averages out to 2.5 copies. But this number represents all the information we have available.)

Can we change the FORMAT definition for CN to allow Type=Float ? Or should we create a different FORMAT for CNM (CN Mosaic) ? 

Is there a standard FORMAT for LOH regions ?

d-cameron:
I'm inclined to redefine as a Float for VCFv4.3.

cwhelan:
Genome STRiP writes an additional format field:

`##FORMAT=<ID=CNF,Number=1,Type=Float,Description="Estimate of fractional copy number">`

which is provided in addition to `CN`. For germline calling `CN` is the best estimate of absolute copy number and `CNF` can be used to determine if an event is mosaic.

In my opinion it's nice to have an integer FORMAT field to hold absolute copy number genotype, especially since `CNL`  is defined in the spec to store the likelihoods of each possible copy number. I'm not sure `CNQ` makes much sense if it `CN` is a float either.

enwired:
Thanks for your comments.  We are very near to completion on this product, so we barely had time to use your input.  We have decided to use "CNF" for non-integer (mosaic) copy number calls.

For LOH, we are doing this, but not sure it is the best thing:

    ##ALT=<ID=LOH,Description="Loss of heterozygosity region">
    #CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	HTCMAv2_P02_96F_A0002945_GT4_A12.rhchp
    1	46133947	.	N	<LOH>	.	.	SVTYPE=LOH;END=47243223;SVLEN=1109276;IMPRECISE	LOH	.

A VCF validator program complains that LOH is not a valid value for ALT, but several programs that we use are able to load the file with no problem.

d-cameron:
> several programs that we use are able to load the file with no problem.

[...]

PR #465, opened 2019-12-18
Improved Structural Variant Support

d-cameron:
Various drafts of improved Structural Variant support have been floating around for 2.5 years (see #231, #266) but never merged. I'm attempting to collate everything together in this PR but it is unclear as to what is in scope.

The following changes were agreed backed in 2017 by Cristina Yenyxe Gonzalez, Steve Huang, Daniel Cameron, Xuefang Zhao, Tobias Rausch, Tim Hefferon, John Lopez, Chris Whelan:

- Restrict SVTYPE to the 6 primitives
  - SVTYPE will be re-cast as a "basic primitive"; its distinction from EVENTTYPE will be made clear in the spec
  - SVTYPE will have the following closed controlled vocabulary: DEL, INS, DUP, INV, CNV, BND
  - No colons or subtypes are allowed in SVTYPE value
  - All SV VCF records must include a value for SVTYPE
- EVENTTYPE will be added to the spec. It will contain the "biological interpretation" of the variant
   - EVENTTYPE will have an open controlled vocabulary, to include: INS, MEI, ALU, L1, SVA, HERV, DEL, -MEI-, -ALU-, -L1-, -SVA-, -HERV-, INV, DUP, DELINS, CNV
- EVENT will continue to be used to specify identifiers to link together multiple VCF records
  - EVENT is currently discussed primarily in the graphical BND-notation section of the spec; text will be added elsewhere, in relevant context area(s) of the spec.
- CIPOS and CIEND will continue to be used as they have in the past: as the primary means of representing an interval within which a breakpoint is likely to fall.
  - The spec will be amended with examples, as needed, to illustrate the preferred usage of CIPOS and CIEND.
- New tags CIPOSPROB and CIENDPROB will be introduced to the spec as a means to indicate the level of confidence implied by CIPOS and CIEND, respectively.
  - Each CI*PROB value shall contain two (2) values separated by a comma. These values shall represent the proportion of a normal distribution expected to fall outside the values recorded in CIPOS. For example, CIPOS=(-50,100) and CIPOSPROB=(0,0.95) indicate the probability that the breakpoint lay more than 50 bp to the left of POS is zero, and the probability that it lay more than 100bp to the right of POS os 0.95.
- New tag HOMPOS will be introduced to the spec. Its value shall represent the coordinate on the given contig of the first basepair of the microhomology indicated by HOMSEQ.
  - The definition of HOMPOS will be: Position (relative to POS) of base pair identical micro-homology around event breakpoint. Note that length(HOMSEQ)=HOMLEN=HOMPOS[1]-HOMPOS[0]
- BND will be officially added to the SVTYPE controlled vocabulary, and it will be referenced consistently throughout the spec (unlike now). It will NOT be replaced by other terms that were discussed, such as TRA (for translocation) or ADJ (for adjacency).
- Examples of each type of event shall be drawn up and added to the spec, including one in "standard" notation and another in the equivalent "BND" notation. Tim and Daniel will draft these examples, respectively.

Additional changes that I'd like to see are:
- Either 1 SV per record, or SV fields counts standardised to handle multiple SV records
  - the latter was blocked by VCF not supporting list of lists
- A field to resolve STR expansion ambiguity (e.g. HOMSTRIDE)
- Sub-clonality support (for all variants)
- genotyping support for somatic SVs. There are two issues with the current specs: 
  - copy number != ploidy
  - maternal/paternal haplotypes are still meaningful for somatic SV, it's just that there can be many copies of each.
- Karyotype reconstruction
  - a 'next SV' field is sufficient
  - PSL could be adapted to handle this
- Explicit clarification around SVTYPE about what the claim is
  - currently both CNV and SV callers write DELs so it's unclear if a DEL is claim of a breakpoint adjacency, a segmental loss, or both

@lbergelson @yfarjoun @pd3 What's the best way forward from here that minimises the chances of us sitting on PRs for another 2 years?

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -585,14 +585,14 @@ For precise variants, END is $\mbox{POS} + \mbox{length of REF allele} - 1$, and
 \end{samepage}
 SVTYPE is a primitive basic data type and must be specified for all structural variants.
 Its value can usually be deduced from the REF and ALT fields, however despite this redundancy it is quite useful for filtering.
-For each VCF record, SVTYPE must be assigned exactly one of the following values:
+For each VCF record encoding a symbolic structural variant, SVTYPE must be assigned exactly one of the following values:
 \begin{itemize}
 \item DEL: Deletion relative to the reference
 \item INS: Insertion relative to the reference
-\item DUP: Region of elevated copy number relative to the reference
+\item DUP: Tandem duplication relative to the reference
 \item INV: Inversion of reference sequence
 \item CNV: Copy number variable region (multiallelic)
-\item BND: Breakend
+\item BND: Breakpoint or single breakend
 \end{itemize}
 
 \footnotesize
@@ -837,60 +837,6 @@ $2$ & \verb|a t c G C G C G a| & following the G base is an insertion of 2 bases
 \end{tabular}
 
 \subsection{Encoding Structural Variants}
-The following page contains examples of structural variants encoded in VCF, showing in order:
-\begin{enumerate}
-  \item A precise deletion with known breakpoint, a one base micro-homology, and a sample that is homozygous for the deletion.
-  \item An imprecise deletion of approximately 205 bp.
-  \item An imprecise deletion of an ALU element relative to the reference.
-  \item An imprecise insertion of an L1 element relative to the reference.
-  \item An imprecise duplication of approximately 21Kb. The sample genotype is copy number 3 (one extra copy of the duplicated sequence).
-  \item An imprecise tandem duplication of 76bp. The sample genotype is copy number 5 (but the two haplotypes are not known).
-\end{enumerate}
-
-\pagebreak
-\footnotesize
-\begin{landscape}
-\begin{verbatim}
-VCF STRUCTURAL VARIANT EXAMPLE
-
-##fileformat=VCFv4.3
-##fileDate=20180101
-##reference=GRCh38.p11
-##assembly=ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/release/sv/breakpoint_assemblies.fasta
-##INFO=<ID=BKPTID,Number=.,Type=String,Description="ID of the assembled alternate allele in the assembly file">
-##INFO=<ID=CIEND,Number=2,Type=Integer,Description="Confidence interval around END for imprecise variants">
-##INFO=<ID=CIPOS,Number=2,Type=Integer,Description="Confidence interval around POS for imprecise variants">
-##INFO=<ID=END,Number=1,Type=Integer,Description="End position of the variant described in this record">
-##INFO=<ID=HOMLEN,Number=.,Type=Integer,Description="Length of base pair identical micro-homology at event breakpoints">
-##INFO=<ID=HOMSEQ,Number=.,Type=String,Description="Sequence of base pair identical micro-homology at event breakpoints">
-##INFO=<ID=SVLEN,Number=.,Type=Integer,Description="Difference in length between REF and ALT alleles">
-##INFO=<ID=SVTYPE,Number=1,Type=String,Description="Type of structural variant">
-##ALT=<ID=DEL,Description="Deletion">
-##ALT=<ID=DEL:ME:ALU,Description="Deletion of ALU element">
-##ALT=<ID=DEL:ME:L1,Description="Deletion of L1 element">
-##ALT=<ID=DUP,Description="Duplication">
-##ALT=<ID=DUP:TANDEM,Description="Tandem Duplication">
-##ALT=<ID=INS,Description="Insertion of novel sequence">
-##ALT=<ID=INS:ME:ALU,Description="Insertion of ALU element">
-##ALT=<ID=INS:ME:L1,Description="Insertion of L1 element">
-##ALT=<ID=INV,Description="Inversion">
-##ALT=<ID=CNV,Description="Copy number variable region">
-##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
-##FORMAT=<ID=GQ,Number=1,Type=Integer,Description="Genotype quality">
-##FORMAT=<ID=CN,Number=1,Type=Integer,Description="Copy number genotype for imprecise events">
-##FORMAT=<ID=CNQ,Number=1,Type=Float,Description="Copy number genotype quality for imprecise events">
-#CHROM POS     ID        REF              ALT          QUAL FILTER INFO                                                               FORMAT       NA00001
-1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;END=2827708;HOMLEN=1;HOMSEQ=G;SVLEN=-14                 GT:GQ        1/1:14
-2       321682 .         T                <DEL>        6    PASS   SVTYPE=DEL;END=321887;SVLEN=-205;CIPOS=-56,20;CIEND=-10,62         GT:GQ        0/1:12
-2     14477084 .         C                <DEL:ME:ALU> 12   PASS   SVTYPE=DEL;END=14477381;SVLEN=-297;CIPOS=-22,18;CIEND=-12,32       GT:GQ        0/1:12
-3      9425916 .         C                <INS:ME:L1>  23   PASS   SVTYPE=INS;END=9425916;SVLEN=6027;CIPOS=-16,22                     GT:GQ        1/1:15
-3     12665100 .         A                <DUP>        14   PASS   SVTYPE=DUP;END=12686200;SVLEN=21100;CIPOS=-500,500;CIEND=-500,500  GT:GQ:CN:CNQ ./.:0:3:16.2
-4     18665128 .         T                <DUP:TANDEM> 11   PASS   SVTYPE=DUP;END=18665204;SVLEN=76;CIPOS=-10,10;CIEND=-10,10         GT:GQ:CN:CNQ ./.:0:5:8.3
-\end{verbatim}
-\end{landscape}
-\pagebreak
-\normalsize
-
 \subsection{Specifying complex rearrangements with breakends}
 
 An arbitrary rearrangement event can be summarized as a set of novel \textbf{adjacencies}. Each adjacency ties together $2$ \textbf{breakends}.
@@ -1415,6 +1361,191 @@ Example records are given below:
 \end{tabular}
 \end{flushleft}
 \normalsize
+\pagebreak
+
+\subsection{Structural Variants Examples}
+
+As with non-structural variants, there are multiple equivalent way to encode a structural variant.
+This section contains examples of the formats for which a structural variant can be encoded in VCF.
+
+\subsubsection{Simple Structural Variants}
+The following page contains examples of structural variants encoded in VCF, showing in order:
+\begin{enumerate}
+  \item A precise deletion with known breakpoint, a one base micro-homology, and a sample that is homozygous for the deletion.
+  \item An imprecise deletion of approximately 205 bp.
+  \item An imprecise deletion of an ALU element relative to the reference.
+  \item An imprecise insertion of an L1 element relative to the reference.
+  \item An imprecise duplication of approximately 21Kb. The sample genotype is copy number 3 (one extra copy of the duplicated sequence).
+  \item An imprecise tandem duplication of 76bp. The sample genotype is copy number 5 (but the two haplotypes are not known).
+\end{enumerate}
+
+\pagebreak
+\footnotesize
+\begin{landscape}
+\begin{verbatim}
+##fileformat=VCFv4.3
+##fileDate=20180101
+##reference=GRCh38.p11
+##assembly=ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/release/sv/breakpoint_assemblies.fasta
+##INFO=<ID=BKPTID,Number=.,Type=String,Description="ID of the assembled alternate allele in the assembly file">
+##INFO=<ID=CIEND,Number=2,Type=Integer,Description="Confidence interval around END for imprecise variants">
+##INFO=<ID=CIPOS,Number=2,Type=Integer,Description="Confidence interval around POS for imprecise variants">
+##INFO=<ID=END,Number=1,Type=Integer,Description="End position of the variant described in this record">
+##INFO=<ID=HOMLEN,Number=.,Type=Integer,Description="Length of base pair identical micro-homology at event breakpoints">
+##INFO=<ID=HOMSEQ,Number=.,Type=String,Description="Sequence of base pair identical micro-homology at event breakpoints">
+##INFO=<ID=SVLEN,Number=.,Type=Integer,Description="Difference in length between REF and ALT alleles">
+##INFO=<ID=SVTYPE,Number=1,Type=String,Description="Type of structural variant">
+##ALT=<ID=DEL,Description="Deletion">
+##ALT=<ID=DEL:ME:ALU,Description="Deletion of ALU element">
+##ALT=<ID=DEL:ME:L1,Description="Deletion of L1 element">
+##ALT=<ID=DUP,Description="Duplication">
+##ALT=<ID=DUP:TANDEM,Description="Tandem Duplication">
+##ALT=<ID=INS,Description="Insertion of novel sequence">
+##ALT=<ID=INS:ME:ALU,Description="Insertion of ALU element">
+##ALT=<ID=INS:ME:L1,Description="Insertion of L1 element">
+##ALT=<ID=INV,Description="Inversion">
+##ALT=<ID=CNV,Description="Copy number variable region">
+##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
+##FORMAT=<ID=GQ,Number=1,Type=Integer,Description="Genotype quality">
+##FORMAT=<ID=CN,Number=1,Type=Integer,Description="Copy number genotype for imprecise events">
+##FORMAT=<ID=CNQ,Number=1,Type=Float,Description="Copy number genotype quality for imprecise events">
+#CHROM POS     ID        REF              ALT          QUAL FILTER INFO                                                               FORMAT       NA00001
+1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;END=2827708;HOMLEN=1;HOMSEQ=G;SVLEN=-14                 GT:GQ        1/1:14
+2       321682 .         T                <DEL>        6    PASS   SVTYPE=DEL;END=321887;SVLEN=-205;CIPOS=-56,20;CIEND=-10,62         GT:GQ        0/1:12
+2     14477084 .         C                <DEL:ME:ALU> 12   PASS   SVTYPE=DEL;END=14477381;SVLEN=-297;CIPOS=-22,18;CIEND=-12,32       GT:GQ        0/1:12
+3      9425916 .         C                <INS:ME:L1>  23   PASS   SVTYPE=INS;END=9425916;SVLEN=6027;CIPOS=-16,22                     GT:GQ        1/1:15
+3     12665100 .         A                <DUP>        14   PASS   SVTYPE=DUP;END=12686200;SVLEN=21100;CIPOS=-500,500;CIEND=-500,500  GT:GQ:CN:CNQ ./.:0:3:16.2
+4     18665128 .         T                <DUP:TANDEM> 11   PASS   SVTYPE=DUP;END=18665204;SVLEN=76;CIPOS=-10,10;CIEND=-10,10         GT:GQ:CN:CNQ ./.:0:5:8.3
+\end{verbatim}
+\end{landscape}
+\pagebreak
+\normalsize
+
+\subsubsection{Equivalent Deletions}
+
+A 5bp deletion of the following form:
+
+\begin{verbatim}
+1234567890 position
+ATAGGTTCGC reference contig
+AT-----CGC variant
+\end{verbatim}
+
+\begin{verbatim}
+#CHROM    POS  ID          REF     ALT         QUAL FILTER INFO
+contig    2    del_indel   TAGGTT  T           .    .
+contig    2    del_indel2  TAGGTTC TC          .    .
+contig    1    del_indel3  ATAGGTT AT          .    .
+contig    3    del_indel4  AGGTTC  C           .    .
+contig    3    del_svtype  A       <DEL>       .    .      SVTYPE=DEL;SVLEN=-5;END=7
+contig    2    del_bnd_1   T       T[contig:8[ .    .      SVTYPE=BND;PARID=del_bnd_2
+contig    8    del_bnd_2   C       ]contig:2]C .    .      SVTYPE=BND;PARID=del_bnd_1
+\end{verbatim}
+
+By convention, only the reference base immediately preceding the deletion is reported.
+This makes $del\_indel$ preferable to $del\_indel2$, $del\_indel3$, or $del\_indel4$.
+
+$del\_bnd1$/$del\_bnd2$ are making a claim only about the presence of the breakpoint, not the copy number of the 'deleted' region.
+It is possible that this breakpoint forms part of a complex rearrangement and the 'deleted' is still present, in another location.
+
+\subsubsection{Equivalent Insertions}
+
+A 5bp insertion (CTCAG) of the following form:
+
+\begin{verbatim}
+123-----4567890 position
+ATA-----GGTTCGC reference contig
+ATACTCAGGGTTCGC variant
+\end{verbatim}
+
+Assuming a file $path\_to\_fasta\_containing\_asm.fasta$ contains the following:
+\begin{verbatim}
+>asm
+ACTCAG
+\end{verbatim}
+
+Then the following records all encode the same insertion:
+
+\begin{verbatim}
+##assembly=path_to_fasta_containing_asm.fasta
+#CHROM    POS  ID                         REF    ALT              QUAL FILTER INFO
+contig    3    ins_indel_representation1  A      ACTCAG           .    .    
+contig    4    ins_indel_representation2  G      CTCAGG           .    .    
+contig    3    ins_svtype_representation1 A      <INS>            .    .      SVTYPE=INS;SVLEN=5;END=3
+contig    4    ins_svtype_representation2 G      <INS>            .    .      SVTYPE=INS;SVLEN=5;END=4
+contig    3    ins_bnd_1                  A      ACTCAG[contig:4[ .    .      SVTYPE=BND;PARID=ins_bnd_2
+contig    4    ins_bnd_2                  G      ]contig:3]CTCAGG .    .      SVTYPE=BND;PARID=ins_bnd_1
+contig    4    ins_asm_full               G      <asm>            .    .      BKPTID=asm
+contig    3    ins_asm_1                  A      A[<asm>:2[       .    .      SVTYPE=BND;PARID=ins_asm_2
+contig    4    ins_asm_2                  G      ]<asm>:6]G       .    .      SVTYPE=BND;PARID=ins_asm_1
+\end{verbatim}
+
+Note the following:
+\begin{itemize}
+	\item The $indel$ and $svtype$ representations 1 and 2 corresponding to using the preceding and succeeding reference base respectively.
+	By convention, the preceding base is used as the anchor.
+	\item $ins\_svtype\_representation1$ and $ins\_svtype\_representation2$ are lossy representation as they do not include the inserted sequence.
+	\item $ins\_asm\_*$ require the external fasta file to reconstruct the variant.
+	\item $ins\_asm\_1$ starts at base 2 of the asm contig as $ins\_asm\_full$ encodes the ALT allele relative to the REF, and the REF includes the anchoring base.
+\end{itemize}
+\subsubsection{Equivalent Tandem Duplications}
+
+A 5bp tandem duplication (AGGTT) of the following form:
+
+\begin{verbatim}
+12-----34567890
+AT-----AGGTTCGC
+ATAGGTTCGGTTCGC
+\end{verbatim}
+
+\begin{verbatim}
+#CHROM    POS  ID                 REF    ALT          QUAL FILTER INFO
+contig    4    dup_indel_verbose  GGTTC  GGTTCGGTTC   .    .    
+contig    8    dup_indel_minimal  C      CGGTTC       .    .    
+contig    9    dup_indel_minimal2 C      GGTTCG       .    .    
+contig    4    dup_svtype         G      <DUP:TANDEM> .    .      SVTYPE=DUP;SVLEN=5;END=8
+contig    4    dup_bnd_1          G      ]contig:8]G  .    .      SVTYPE=BND;PARID=dup_bnd_2
+contig    8    dup_bnd_2          C      C[contig:4[  .    .      SVTYPE=BND;PARID=dup_bnd_1
+\end{verbatim}
+
+$dup\_bnd1$/$dup\_bnd2$ are making a claim only about the presence of the breakpoint, not the copy number of the 'duplication' region.
+It is possible that this breakpoint forms part of a complex rearrangement and the 'duplicated' region does not have elevated copy number.
+
+\subsubsection{Equivalent Inversions}
+
+A 4bp inversion of the following form:
+
+\begin{verbatim}
+1234567890
+ATAGGTTCGC
+ATAAACCCGC
+   ----
+\end{verbatim}
+
+\begin{verbatim}
+#CHROM    POS  ID         REF      ALT         QUAL FILTER INFO
+contig    4    inv_indel  GGTT     AACC        .    .    
+contig    4    inv_indel  AGGTT    AAACC       .    .    
+contig    4    inv_svtype G        <INV>       .    .      SVTYPE=INV;SVLEN=0;END=7
+contig    3    inv_bnd_1  A        A]contig:7] .    .      SVTYPE=BND;PARID=ins_bnd_2;EVENT=example_inv
+contig    7    inv_bnd_2  T        T]contig:3] .    .      SVTYPE=BND;PARID=ins_bnd_1;EVENT=example_inv
+contig    4    inv_bnd_3  G        [contig:8[G .    .      SVTYPE=BND;PARID=ins_bnd_4;EVENT=example_inv
+contig    8    inv_bnd_4  C        [contig:4[C .    .      SVTYPE=BND;PARID=ins_bnd_3;EVENT=example_inv
+\end{verbatim}
+
+\subsubsection{Copy Number Variants}
+
+Some methods identify changes in copy number.
+In such situations, SVTYPE should be specified as CNV as such methods do not make breakpoint claims.
+The deletion and duplication examples outlined above would be represented as follows:
+
+\begin{verbatim}
+#CHROM    POS    ID          REF    ALT    QUAL FILTER INFO
+contig    3      del_cnv     A      <DEL>  .    .      SVTYPE=CNV;SVLEN=-5;END=7
+contig    4      dup_cnv     G      <DUP>  .    .      SVTYPE=CNV;SVLEN=5;END=8
+\end{verbatim}
+
+Note that the ALT is specified as <DUP>, not <DUP:TANDEM>, as the location of the additional copy is not known.
 
 \pagebreak
 \section{BCF specification}
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -173,30 +173,53 @@ Genotype fields specified in the FORMAT field are described as follows:
 Possible Types for FORMAT fields are: Integer, Float, Character, and String (this field is otherwise defined precisely as the INFO field).
 
 \subsubsection{Alternative allele field format}
-Symbolic alternate alleles are described as follows:
+Symbolic alternate alleles specified in the ALT field are described as follows:
 \begin{verbatim}
 ##ALT=<ID=type,Description=description>
 \end{verbatim}
 
 \noindent \textbf{Structural Variants} \newline
-In symbolic alternate alleles for imprecise structural variants, the ID field indicates the type of structural variant, and can be a colon-separated list of types and subtypes.
-ID values are case sensitive strings and must not contain whitespace or angle brackets.
-The first level type must be one of the following:
-\begin{itemize}
-  \item DEL Deletion relative to the reference
-  \item INS Insertion of novel sequence relative to the reference
-  \item DUP Region of elevated copy number relative to the reference
-  \item INV Inversion of reference sequence
-  \item CNV Copy number variable region (may be both deletion and duplication)
-  \item BND Breakend
-\end{itemize}
 
-The CNV category should not be used when a more specific category can be applied. Reserved subtypes include:
-\begin{itemize}
-  \item DUP:TANDEM Tandem duplication
-  \item DEL:ME Deletion of mobile element relative to the reference
-  \item INS:ME Insertion of a mobile element relative to the reference
-\end{itemize}
+
+\begin{tabular}{l l}
+DEL  &  Deletion relative to the reference \\
+INS  &  Insertion relative to the reference \\
+DUP  &  Region of elevated copy number relative to the reference \\
+INV  &  Inversion of reference sequence \\
+CNV  &  Copy number variable region (multiallelic) \\
+BND  &  Breakend \\
+\end{tabular}\newline
+
+\noindent The list of first level symbolic alleles that represent structural variants is a closed set. Any symbolic alleles with a first level type other than one of those listed above are considered to be non-structural symbolic alleles.\newline
+
+\noindent Variant subtypes should be used to add specificity to the maximum extent supported by the data. Reserved subtypes include:
+\newline
+
+\begin{tabular}{l l}
+DEL:ME  &  Absence of mobile element present in the reference \\
+DEL:ME:ALU  &  Absence of Alu element present in the reference \\
+DEL:ME:LINE1  &  Absence of LINE1 element present in the reference \\
+DEL:ME:SVA  &  Absence of SVA element present in the reference \\
+DEL:ME:HERV  &  Absence of HERV element present in the reference \\
+\\
+INS:ME  &  Insertion of mobile element relative to the reference \\
+INS:ME:ALU  &  Insertion of Alu element relative to the reference \\
+INS:ME:LINE1  &  Insertion of LINE1 element relative to the reference \\
+INS:ME:SVA  &  Insertion of SVA element relative to the reference \\
+INS:ME:HERV  &  Insertion of HERV element relative to the reference \\
+\\
+DUP:TANDEM  &  Tandem duplication, in same orientation as original \\
+DUP:INV-BEFORE  &  Tandem duplication, inserted before original in inverted orientation \\
+DUP:INV-AFTER  &  Tandem duplication, inserted after original in inverted orientation \\
+\end{tabular}\newline \newline
+
+\noindent Variants should be written using the most precise type that can be determined by the variant caller. For example, if the insertion site of a new copy of a LINE1 element cannot be determined, it would be specified as a DUP of the originating LINE1 element. However if the new insertion site can be identified, the variant should be specified as INS:ME:LINE1 at the insertion site.\newline
+
+\noindent Note that the DUP type is not restricted to tandem duplications and that the DUP subtypes have different breakpoints than their parent type.\newline
+
+\noindent Variant subtypes are an open set and is it valid for a variant caller to define a subtype not specified in the above list. Such subtypes are considered to be instances of the first reserved subtype, with breakpoint and copy number semantics matching the reserved type. For example, INS:ME:PELEMENT is a valid caller-defined subtype, but INV:FOLDBACK is not since foldback inversions do not result in the same breakpoint and copy number changes as a simple inversion.\newline
+
+\noindent See examples below for further guidance.
 
 \bigskip
 
@@ -563,15 +586,14 @@ For precise variants, END is $\mbox{POS} + \mbox{length of REF allele} - 1$, and
 ##INFO=<ID=SVTYPE,Number=1,Type=String,Description="Type of structural variant">
 \end{verbatim}
 \normalsize
-\end{samepage}
-This key can be derived from the REF/ALT fields but is useful for filtering.
-The reserved values must be used for the types listed below:
+SVTYPE is a primitive basic data type and must be specified for all structural variants. Its value can usually be deduced from the REF and ALT fields, however despite this redundancy it is quite useful for filtering. For each VCF record, SVTYPE must be assigned exactly one of the following values:
 \begin{itemize}
   \item DEL: Deletion relative to the reference
-  \item INS: Insertion of novel sequence relative to the reference
+\item INS:  Insertion relative to the reference
   \item DUP: Region of elevated copy number relative to the reference
   \item INV: Inversion of reference sequence
-  \item CNV: Copy number variable region (may be both deletion and duplication)
+\item CNV:  Copy number variable region (multiallelic)
+\item BND:  Breakend
   \item BND: Breakend
 \end{itemize}
 
@@ -580,7 +602,13 @@ The reserved values must be used for the types listed below:
 ##INFO=<ID=SVLEN,Number=.,Type=Integer,Description="Difference in length between REF and ALT alleles">
 \end{verbatim}
 \normalsize
-One value for each ALT allele. Longer ALT alleles (e.g.\ insertions) have positive values, shorter ALT alleles (e.g.\ deletions) have negative values.
+SVLEN indicates the relative size of the ALT allele compared to the REF allele.\
+If ALT is longer than REF (e.g. insertions), SVLEN must be positive; if ALT is shorter than REF (e.g. deletions), SVLEN must be negative.
+Simple inversions will have SVLEN=0.
+The size of the inversion can be determined by the POS field and the END attribute.
+One SVLEN value for each ALT allele.
+One value for each ALT allele.
+Longer ALT alleles (e.g.\ insertions) have positive values, shorter ALT alleles (e.g.\ deletions) have negative values.
 
 \footnotesize
 \begin{verbatim}
@@ -830,8 +858,8 @@ The following page contains examples of structural variants encoded in VCF, show
 VCF STRUCTURAL VARIANT EXAMPLE
 
 ##fileformat=VCFv4.3
-##fileDate=20100501
-##reference=1000GenomesPilot-NCBI36
+##fileDate=20180101
+##reference=GRCh38.p11
 ##assembly=ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/release/sv/breakpoint_assemblies.fasta
 ##INFO=<ID=BKPTID,Number=.,Type=String,Description="ID of the assembled alternate allele in the assembly file">
 ##INFO=<ID=CIEND,Number=2,Type=Integer,Description="Confidence interval around END for imprecise variants">
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -184,9 +184,9 @@ Symbolic alternate alleles specified in the ALT field are described as follows:
 \begin{tabular}{l l}
 DEL  &  Deletion relative to the reference \\
 INS  &  Insertion relative to the reference \\
-DUP  &  Region of elevated copy number relative to the reference \\
+DUP  &  Tandem duplication relative to the reference \\
 INV  &  Inversion of reference sequence \\
-CNV  &  Copy number variable region (multiallelic) \\
+CNV  &  Copy number variable region \\
 BND  &  Breakend \\
 \end{tabular}\newline
 
@@ -207,15 +207,11 @@ INS:ME:ALU  &  Insertion of Alu element relative to the reference \\
 INS:ME:LINE1  &  Insertion of LINE1 element relative to the reference \\
 INS:ME:SVA  &  Insertion of SVA element relative to the reference \\
 INS:ME:HERV  &  Insertion of HERV element relative to the reference \\
-\\
-DUP:TANDEM  &  Tandem duplication, in same orientation as original \\
-DUP:INV-BEFORE  &  Tandem duplication, inserted before original in inverted orientation \\
-DUP:INV-AFTER  &  Tandem duplication, inserted after original in inverted orientation \\
 \end{tabular}\newline \newline
 
 \noindent Variants should be written using the most precise type that can be determined by the variant caller. For example, if the insertion site of a new copy of a LINE1 element cannot be determined, it would be specified as a DUP of the originating LINE1 element. However if the new insertion site can be identified, the variant should be specified as INS:ME:LINE1 at the insertion site.\newline
 
-\noindent Note that the DUP type is not restricted to tandem duplications and that the DUP subtypes have different breakpoints than their parent type.\newline
+\noindent Note that the DUP type is restricted to simple tandem duplications. More complex duplications should be specified using BND notation.\newline
 
 \noindent Variant subtypes are an open set and it is valid for a variant caller to define a subtype not specified in the above list. Such subtypes are considered to be instances of the first reserved subtype, with breakpoint and copy number semantics matching the reserved type. For example, INS:ME:PELEMENT is a valid caller-defined subtype, but INV:FOLDBACK is not since foldback inversions do not result in the same breakpoint and copy number changes as a simple inversion.\newline
 
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -2102,6 +2102,9 @@ Therefore, also \#\#PEDIGREE newly requires a unique ID.
 \item Chromosome names cannot use reserved symbolic alleles and contain characters used by breakpoints (Section~\ref{sec-contig-field}).
 \item IUPAC ambiguity codes should be converted to a concrete base.
 \item Symbolic ALTs for IUPAC codes.
+\item Made SVTYPE mandatory for SVs.
+\item Top-level SVTYPE values are explicitly restricted to the 6 defined in the specifications.
+\item Reserved some SVTYPE subtypes.
 \end{itemize}
 
 \subsection{Changes between BCFv2.1 and BCFv2.2}
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -217,7 +217,7 @@ DUP:INV-AFTER  &  Tandem duplication, inserted after original in inverted orient
 
 \noindent Note that the DUP type is not restricted to tandem duplications and that the DUP subtypes have different breakpoints than their parent type.\newline
 
-\noindent Variant subtypes are an open set and is it valid for a variant caller to define a subtype not specified in the above list. Such subtypes are considered to be instances of the first reserved subtype, with breakpoint and copy number semantics matching the reserved type. For example, INS:ME:PELEMENT is a valid caller-defined subtype, but INV:FOLDBACK is not since foldback inversions do not result in the same breakpoint and copy number changes as a simple inversion.\newline
+\noindent Variant subtypes are an open set and it is valid for a variant caller to define a subtype not specified in the above list. Such subtypes are considered to be instances of the first reserved subtype, with breakpoint and copy number semantics matching the reserved type. For example, INS:ME:PELEMENT is a valid caller-defined subtype, but INV:FOLDBACK is not since foldback inversions do not result in the same breakpoint and copy number changes as a simple inversion.\newline
 
 \noindent See examples below for further guidance.
 
@@ -603,7 +603,7 @@ SVTYPE is a primitive basic data type and must be specified for all structural v
 \end{verbatim}
 \normalsize
 SVLEN indicates the relative size of the ALT allele compared to the REF allele.\
-If ALT is longer than REF (e.g. insertions), SVLEN must be positive; if ALT is shorter than REF (e.g. deletions), SVLEN must be negative.
+If ALT is longer than REF (e.g.\ insertions), SVLEN must be positive; if ALT is shorter than REF (e.g.\ deletions), SVLEN must be negative.
 Simple inversions will have SVLEN=0.
 The size of the inversion can be determined by the POS field and the END attribute.
 One SVLEN value for each ALT allele.
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -586,15 +586,17 @@ For precise variants, END is $\mbox{POS} + \mbox{length of REF allele} - 1$, and
 ##INFO=<ID=SVTYPE,Number=1,Type=String,Description="Type of structural variant">
 \end{verbatim}
 \normalsize
-SVTYPE is a primitive basic data type and must be specified for all structural variants. Its value can usually be deduced from the REF and ALT fields, however despite this redundancy it is quite useful for filtering. For each VCF record, SVTYPE must be assigned exactly one of the following values:
+\end{samepage}
+SVTYPE is a primitive basic data type and must be specified for all structural variants.
+Its value can usually be deduced from the REF and ALT fields, however despite this redundancy it is quite useful for filtering.
+For each VCF record, SVTYPE must be assigned exactly one of the following values:
 \begin{itemize}
-  \item DEL: Deletion relative to the reference
-\item INS:  Insertion relative to the reference
-  \item DUP: Region of elevated copy number relative to the reference
-  \item INV: Inversion of reference sequence
-\item CNV:  Copy number variable region (multiallelic)
-\item BND:  Breakend
-  \item BND: Breakend
+\item DEL: Deletion relative to the reference
+\item INS: Insertion relative to the reference
+\item DUP: Region of elevated copy number relative to the reference
+\item INV: Inversion of reference sequence
+\item CNV: Copy number variable region (multiallelic)
+\item BND: Breakend
 \end{itemize}
 
 \footnotesize
@@ -607,8 +609,6 @@ If ALT is longer than REF (e.g.\ insertions), SVLEN must be positive; if ALT is 
 Simple inversions will have SVLEN=0.
 The size of the inversion can be determined by the POS field and the END attribute.
 One SVLEN value for each ALT allele.
-One value for each ALT allele.
-Longer ALT alleles (e.g.\ insertions) have positive values, shorter ALT alleles (e.g.\ deletions) have negative values.
 
 \footnotesize
 \begin{verbatim}


d-cameron:
See #448 for an example of how real-world tools are using fields in a non-compliant manner to work around the lack of proper subclonal support.

VCFv4.3.tex
```diff
@@ -563,15 +586,14 @@ \section{INFO keys used for structural variants}
 ##INFO=<ID=SVTYPE,Number=1,Type=String,Description="Type of structural variant">
 \end{verbatim}
 \normalsize
-\end{samepage}
```
jmmut: This line has to stay. The latex build is failing because there's an unmatched `\begin{samepage}`.

VCFv4.3.tex
```diff
@@ -563,15 +586,14 @@ \section{INFO keys used for structural variants}
...
-This key can be derived from the REF/ALT fields but is useful for filtering.
-The reserved values must be used for the types listed below:
+SVTYPE is a primitive basic data type and must be specified for all structural variants. Its value can usually be deduced from the REF and ALT fields, however despite this redundancy it is quite useful for filtering. For each VCF record, SVTYPE must be assigned exactly one of the following values:
 \begin{itemize}
   \item DEL: Deletion relative to the reference
-  \item INS: Insertion of novel sequence relative to the reference
+\item INS:  Insertion relative to the reference
```
jmmut: please, keep the previous whitespace pattern in these items.

VCFv4.3.tex
```diff
@@ -580,7 +602,13 @@ \section{INFO keys used for structural variants}
 ##INFO=<ID=SVLEN,Number=.,Type=Integer,Description="Difference in length between REF and ALT alleles">
 \end{verbatim}
 \normalsize
-One value for each ALT allele. Longer ALT alleles (e.g.\ insertions) have positive values, shorter ALT alleles (e.g.\ deletions) have negative values.
+SVLEN indicates the relative size of the ALT allele compared to the REF allele.\
+If ALT is longer than REF (e.g. insertions), SVLEN must be positive; if ALT is shorter than REF (e.g. deletions), SVLEN must be negative.
```
jmmut: the backlashes are needed to avoid a wider whitespace (I think it's interpreted as an end-sentence period)
```suggestion
If ALT is longer than REF (e.g.\ insertions), SVLEN must be positive; if ALT is shorter than REF (e.g.\ deletions), SVLEN must be negative.
```

d-cameron:
I've now stripped the bit where DUP subtypes defined different breakpoints than the root types. This badly breaks backwards compatability and can be better handled as part of the `EVENT`/`EVENTTYPE` PR.

VCFv4.3.tex
```diff
@@ -564,23 +583,28 @@ \section{INFO keys used for structural variants}
...
 \begin{itemize}
-  \item DEL: Deletion relative to the reference
-  \item INS: Insertion of novel sequence relative to the reference
-  \item DUP: Region of elevated copy number relative to the reference
-  \item INV: Inversion of reference sequence
-  \item CNV: Copy number variable region (may be both deletion and duplication)
-  \item BND: Breakend
+\item DEL: Deletion relative to the reference
+\item INS: Insertion relative to the reference
+\item DUP: Region of elevated copy number relative to the reference
```
jmmut: Is it intentional that the description for DUP is different here in SVTYPE than the one in symbolic ALTs? I would change it here too.

VCFv4.3.tex
```diff
@@ -564,23 +583,28 @@ \section{INFO keys used for structural variants}
...
+\item INV: Inversion of reference sequence
+\item CNV: Copy number variable region (multiallelic)
```
jmmut: Maybe this is a silly question, but what does "multiallelic" mean here? The same as "may be both deletion and duplication" to say that both copynumber 0 and greater than 1 is allowed? "multiallelic" doesn't look clear to me.

VCFv4.3.tex
```diff
@@ -173,30 +173,49 @@ \subsubsection{Individual format field format}
...
+DEL:ME:HERV  &  Absence of HERV element present in the reference \\
+\\
+INS:ME  &  Insertion of mobile element relative to the reference \\
+INS:ME:ALU  &  Insertion of Alu element relative to the reference \\
+INS:ME:LINE1  &  Insertion of LINE1 element relative to the reference \\
+INS:ME:SVA  &  Insertion of SVA element relative to the reference \\
+INS:ME:HERV  &  Insertion of HERV element relative to the reference \\
+\end{tabular}\newline \newline
+
+\noindent Variants should be written using the most precise type that can be determined by the variant caller. For example, if the insertion site of a new copy of a LINE1 element cannot be determined, it would be specified as a DUP of the originating LINE1 element. However if the new insertion site can be identified, the variant should be specified as INS:ME:LINE1 at the insertion site.\newline
```
tskir: I see a contradiction here:
* In the list above, `DUP` specifies a **tandem** duplication relative to the reference
* However, you also suggest using `DUP` when the insertion site of a new copy **cannot be determined**—i. e. it can be anywhere and not necessarily a tandem duplication

Josemi's comment about different DUP definitions here and in SVTYPE is also relevant here.

tskir: Additionally, immediately after it says that “the DUP type is restricted to simple tandem duplications”, adding to the confusion

VCFv4.3.tex
```diff
@@ -564,23 +583,28 @@ \section{INFO keys used for structural variants}
 \end{verbatim}
 \normalsize
 \end{samepage}
-This key can be derived from the REF/ALT fields but is useful for filtering.
-The reserved values must be used for the types listed below:
+SVTYPE is a primitive basic data type and must be specified for all structural variants.
+Its value can usually be deduced from the REF and ALT fields, however despite this redundancy it is quite useful for filtering.
```
tskir: I actually think that the redundancy of SVTYPE does more harm than good... For one, it causes the definitions of structural variant types to diverge between two sections in the specification (symbolic alleles and SVTYPE). If I'm correct in presuming that the definitions are supposed to be the same, then could we maybe drop this list entirely and instead say something like “The list of acceptable SVTYPE values is exactly the same as the top-level structural variant types described in (link to section above)”?

VCFv4.3.tex
```diff
@@ -173,30 +173,49 @@ \subsubsection{Individual format field format}
 Possible Types for FORMAT fields are: Integer, Float, Character, and String (this field is otherwise defined precisely as the INFO field).
 
 \subsubsection{Alternative allele field format}
-Symbolic alternate alleles are described as follows:
+Symbolic alternate alleles specified in the ALT field are described as follows:
```
cwhelan: I think that this section could use a little clarification or a preamble stating what the valid use cases for using symbolic alleles are -- to my reading it's a little unclear whether it's saying that the two enumerated cases below (Structural Variants and IUPAC ambiguity codes) are the only permitted uses for symbolic ALT alleles or whether other, arbitrarily defined, symbolic alts are allowable. What about `<NON-REF>`, which is allowed but discouraged according to section 5.5? Are symbolic alleles that represent markers for arbitrary but incompletely specified haplotypes allowed (I've seen these used with germline statistical phasing tools such as Beagle; they might or might not include structural variants but wouldn't have an SVTYPE)? I'm actually unsure as to whether these types of symbolic alleles are already disallowed, but I think we should state it clearly here either way.

VCFv4.3.tex
```diff
@@ -564,23 +583,28 @@ \section{INFO keys used for structural variants}
 \end{verbatim}
 \normalsize
 \end{samepage}
-This key can be derived from the REF/ALT fields but is useful for filtering.
-The reserved values must be used for the types listed below:
+SVTYPE is a primitive basic data type and must be specified for all structural variants.
```
tskir: Making SVTYPE mandatory raises another important question: what constitutes a structural variant? For example, a deletion can be any size from 1 bp to millions of bps. Does this mean that SVTYPE must now be specified for **all** deletions, including 1 bp? Or should a caller/user select a threshold for considering the variant “structural” and specifying SVTYPE in only that case? I think this needs to be clarified.

VCFv4.3.tex
cwhelan: Multiallelic could mean either the presence of both a simple deletion and duplication allele for the segment, or that there are duplication alleles with differing numbers of copies. For example, one allele might have a single tandem duplication of the segment (and therefore two copies of the reference segment) and another allele might have three or more copies. See for example https://www.ncbi.nlm.nih.gov/pubmed/25621458.

VCFv4.3.tex
cwhelan: I guess below you do say:

> Any symbolic alleles with a first level type other than one of those listed above are considered to be non-structural symbolic alleles.

Which indicates that non-structural symbolic alleles are allowed. But I still think it would be good to clarify their scope here in the header for the section as well.

VCFv4.3.tex
tskir: I mostly agree with @cwhelan's definition, although this needs to be reflected in the spec itself (not just this discussion)

VCFv4.3.tex
```diff
@@ -173,30 +173,49 @@ \subsubsection{Individual format field format}
...
+INS:ME  &  Insertion of mobile element relative to the reference \\
+INS:ME:ALU  &  Insertion of Alu element relative to the reference \\
+INS:ME:LINE1  &  Insertion of LINE1 element relative to the reference \\
+INS:ME:SVA  &  Insertion of SVA element relative to the reference \\
+INS:ME:HERV  &  Insertion of HERV element relative to the reference \\
+\end{tabular}\newline \newline
+
+\noindent Variants should be written using the most precise type that can be determined by the variant caller. For example, if the insertion site of a new copy of a LINE1 element cannot be determined, it would be specified as a DUP of the originating LINE1 element. However if the new insertion site can be identified, the variant should be specified as INS:ME:LINE1 at the insertion site.\newline
+
+\noindent Note that the DUP type is restricted to simple tandem duplications. More complex duplications should be specified using BND notation.\newline
```
cwhelan: I might add that there's also the option to use `<CNV>` for a duplication if you know it exists but aren't sure if it's tandem or not, or if you know that it's not tandem but aren't sure about the breakend structure.

VCFv4.3.tex
jmmut: From line 193 (the first paragraph after the list of ALT values):

 > Any symbolic alleles with a first level type other than one of those listed above are considered to be non-structural symbolic alleles.

The interpretation of that would be that a variant is structural if it uses a symbolic ALT allele from that list. For instance, a deletion of 1bp requires SVTYPE only if it uses ALT=<DEL>, not if it uses explicit alleles. I think that's ok.

VCFv4.3.tex
cwhelan: Right now this seems to indicate that if you know there's a duplication but aren't sure if it's tandem or not, you should use `SVTYPE=DUP` and have an `ALT` of `<CNV>`. Is that the intent?

VCFv4.3.tex
tskir: @jmmut Defining a structural variant based on whether it uses a symbolic allele is fine, but then it raises another problem: if SVTYPE is proposed to be used **when and only when** a structural variant symbolic allele is used, then SVTYPE appears to be completely redundant; unless it is expected that the type of a symbolic allele and the one in SVTYPE could be different.

Also, I've seen VCFs when alleles are specified with sequences but SVTYPE is still used. I'm not sure if we want to continue to allow that, but in any case, this needs to be described more explicitly

VCFv4.3.tex
cwhelan: Also, there are VCFs out there that use `<DUP>` to indicate regions of elevated copy number rather than tandem duplications. Perhaps we should call out a warning about interpreting legacy VCFs given this change.

cwhelan:
@d-cameron Added a few comments on places that I think could use clarifications -- sorry that I didn't make them sooner. Again, thanks for spearheading this change.

VCFv4.3.tex
jmmut: It is redundant. I think tools rely on it for filtering and it's easier for humans to search the SVTYPE and the symbolic ALT, specially for BND where the ALT can be the notation with `[` and `]`. At least that's what I can extract from discussions in this repo.

d-cameron:
Ok, there's quite a bit of discussion around `DUP` events. I think the real underlying issue is that we don't all agree on what a claim of a `DUP` is, and we already have an ecosystem in which different tools are making different claims.

Tools based on micro-array or copy number evidence report `DUP` as an a region of elevated copy number (typically +1 copy). They are making a claim about the number of copies of the duplicated region. Tools based on NGS report `DUP` when they find a breakpoint in the an orientation consistent with a tandem duplication.

In conclusion, the VCF specs don't actualy specify what symbolic structural variants actually mean, so different tools. Our options are:

1. Define DEL/DUP as breakpoint claims
2. Define DEL/DUP as CN claim
3. Define DEL/DUP as both breakpoint and CN claim
4. Add an additional field which is used clarify structural symbolic allele claims and existing calls remain ambiguous.

My preference is option 4.

Thoughts?

VCFv4.3.tex
tskir: Right. In any case, redundancy of SVTYPE needs to be clarified in the spec. Should it always exactly mirror the symbolic allele? Or can we have combinations, like a symbolic allele of `<CNV>` and an SVTYPE of `DEL`? In that case, what combinations of symbolic alleles and SVTYPE are allowed?

d-cameron:
Option 5. grandfather v4.2 or earlier, as ambiguous. Make an unambigous choice if the header is v4.3

cwhelan:
>Tools based on micro-array or copy number evidence report DUP as an a region of elevated copy number (typically +1 copy). They are making a claim about the number of copies of the duplicated region. Tools based on NGS report DUP when they find a breakpoint in the an orientation consistent with a tandem duplication.

Most integrated WGS/NGS germline SV calling pipelines that I've seen being developed for large scale studies include both PE/SR/ASM based calls that would have breakend support as well as depth based calls that don't. You just can't capture the breakpoints of a lot of germline CNVs with short reads, and I think we'll be getting lots of mixed VCFs in the future as integrated pipelines are more widely deployed.

> 4. Add an additional field which is used clarify structural symbolic allele claims and existing calls remain ambiguous.

Just to clarify, are you suggesting something like a SVBKPT INFO field, which, if present, clarifies that there is a breakend claim?

mbaudis:
Just chiming in here w/ opinion that tandem duplications & repeat expansions are *different* from CNVs & should be considered (a) separate entit(y|ies) (e.g. having annotations from sequence, times).

Otherwise liking the move towards resolving ambiguous annotations... But IMO a column for a second reference (chro) is needed, when keeping the columnar format :-)

rhdolin:
Doesn't PRECISE/IMPRECISE achieve what INFO.SVBKPT would? (Although I've seen examples that are both PRECISE and have CIPOS and CIEND, which I find confusing). 

And on a tangent, what is the likelihood of a Tabix index optionally supporting CIPOS/CIEND?

thefferon:
> But IMO a column for a second reference (chro) is needed, when keeping the columnar format :-)
@mbaudis, I am not clear on what you mean here. Please elaborate.

thefferon:
> Doesn't PRECISE/IMPRECISE achieve what INFO.SVBKPT would? (Although I've seen examples that are both PRECISE and have CIPOS and CIEND, which I find confusing).
@rhdolin, by "PRECISE" do you mean the examples had a PRECISE flag, or just that they **did not** have the IMPRECISE flag? If the former, I admit to find that confusing as well.

cwhelan:
@rhdolin
 
> Doesn't PRECISE/IMPRECISE achieve what INFO.SVBKPT would? (Although I've seen examples that are both PRECISE and have CIPOS and CIEND, which I find confusing).

Usually IMPRECISE is only used by callers that are using paired-end mapping signal to identify variants that show a read mapping signature but for which the exact breakpoint cannot be determined, either because there are no split reads or assembly-based evidence at the site, or the caller didn't look for it. This is different from a depth-based CNV caller, which is not making a claim about the breakpoint structure -- it's just reporting that an extra copy of a reference segment exists in the sample. The breakpoint structure of a CNV call could be much more complex than a simple tandem duplication -- imagine that segments from two chromosomes are duplicated, joined together, and inserted on a third chromosome. If the spec said something like, "if an SVTYPE=DUP variant is marked IMPRECISE, it is making a claim of a tandem duplication", it might have the same effect as the SVBKPT INFO field I was talking about above (or some other similar solution), but I think it would be a bit backwards from its original intent and hard to comprehend.

cwhelan:
@d-cameron I think my top preference from your list is your option 4: Keep DEL / DUP ambiguous but add an INFO field indicating whether there's a breakpoint claim of a simple deletion or tandem duplication (which can then be used with IMPRECISE and related fields).

I would also consider option 5 (as I understand it this is making DUP and DEL into breakpoint claims and requiring non-tandem DUPs to be written with SVTYPE=CNV) as long as the new version of the spec contained strong, clear language warning about the difference in the interpretation, particularly of DUP, from older VCFs. Do you know if any SV tools are currently writing VCFs marked 4.3? I would hate to change the interpretation of variants from tools that are in active use without changing the VCF version number.

Just to be clear, I support the goal of better codifying symbolic SV alleles so that it's easier to produce an less ambiguous sequence interpretation.

cwhelan:
One more comment on CNVs: cohort-based CNV callers can disentangle different copy number variable alleles even without breakpoint evidence. For example, in the case of overlapping duplications with clearly different boundaries (but for which we don't know the breakpoint structure): 

```
DUP1:          |-------------|
DUP2:                       |------------|
```
There are three "copy number variable regions" here, but you can sometimes use phasing and parsimonious modeling of allele frequencies to distinguish each sample's genotype for the DUP1 and DUP2 alleles. If we restrict non-breakpoint based claims to use the `<CNV>` alt allele, I think we'd have to change the description of the alt allele in the spec away from `Copy number variable region` to something more like `Allele that changes the copy number of the reference segment`. We'd probably also then need an additional INFO field to say in what direction and by how much the copy number changes for that particular allele. To me this lends support for my preference for your option 4 (allowing the use of `<DUP>` as the alts for these variants, and adding INFO fields for breakpoint support if they are tandem/simple deletions).

thefferon:
I suggest we adopt something like @cwhelan 's working definition of Option #5, perhaps: “DUP and DEL are breakpoint claims; non-tandem DUPs (i.e. strictly copy number based claims) should be written as SVTYPE=CNV.” However, there is a consequence:  moving dispersed DUPs to CNV requires broadening the definition of THAT category to include single-copy-gain dispersed dups.

The August 22, 2019 pdf of v4.3 uses the following definition for CNV:

“Copy number variable region (may be both deletion and duplication)”

or, as pointed out earlier in the current discussion (https://github.com/samtools/hts-specs/pull/465#pullrequestreview-347971907):

“Copy number variable region (multiallelic)”

A freshly-redefined ‘CNV’ would include all of the following variants:
•	regions in which both a duplication and a deletion have been observed
•	regions one would call “multiallelic” – that is, in which three or more alleles (reflecting discrete copy number increments) have been reported
•	the “new kid on the block”: dispersed duplications (for now let's limit these to increases of just one copy number relative to reference) 

#3, as defined above, can reasonably be called ‘biallelic’. In contrast, the very broad perception is that “SVTYPE=CNV” implies multiple alleles” (in fact, gnomAD's SV VCF uses 'MCNV' instead of 'CNV' - maybe this is a change we want to consider adopting in the spec). So the ‘CNV’ definition will need to be re-written to include the new kid on the block.

thefferon:
I think it may also be important to recognize, if only conceptually, that any called insertion can represent (depending on the nature and extent of the analysis involved) a dispersed duplication or a tandem duplication. Distinguishing among these possibilities requires an analysis of the insertion’s sequence content and immediate genomic context.

VCFv4.3.tex
```diff
@@ -173,30 +173,49 @@ \subsubsection{Individual format field format}
...
-  \item INS:ME Insertion of a mobile element relative to the reference
-\end{itemize}
+
+\begin{tabular}{l l}
+DEL  &  Deletion relative to the reference \\
+INS  &  Insertion relative to the reference \\
+DUP  &  Tandem duplication relative to the reference \\
+INV  &  Inversion of reference sequence \\
+CNV  &  Copy number variable region \\
+BND  &  Breakend \\
```
cwhelan: Is `<BND>` a valid first-level symbolic alt (which I think this language implies) according to this PR, or do breakend alts need to be written in the detailed format with square brackets, etc?

thefferon: @d-cameron may best be able to speak to this, but it is my understanding that the intention when using SVTYPE=BND is that the ALT column always contain the bracketed notation, e.g., "**]5:7576543]G**". @cwhelan, can you provide an (hypothetical) example of a variant in which SVTYPE=BND and ALT=<BND>? I am not sure what that would look like.

cwhelan: I'm not suggesting that it should be allowed (although this is something the current release of gnomAD does; I'm working to fix that for future releases right now); just trying to point out that unless I'm misreading something in the language in this PR, the specification would seem to allow it even after these changes are merged.

bhandsaker:
Hi, all,

Chris Whelan brought this discussion to my attention.
I want to share some of the use cases I currently implement in Genome STRiP and other software we develop/use in our lab.

From my perspective, it would be ideal if the VCF specification allowed these kinds of representations.
In general, I also think it is useful to allow the VCF format to be flexible enough to permit tool-specific extensions in a spec-compliant VCF.

1. SVTYPE. Although we could do this with a different tag, we currently use SVTYPE to default certain genotyping behavior.
As one example, if SVTYPE=DEL, then we will never assign INFO:CN > ploidy and all genotype likelihood calculations assume a bi-allelic variant where one allele is the reference and the other a deletion allele. This is in contrast to SVTYPE=CNV, which allows INFO:CN to be greater or less than ploidy.

2. For unphased copy number variants, we use <CNV> as the ALT allele. The code is also not picky about the ref allele, and in particular N is allowed (regardless of the actual reference base at POS). We use the <CNV> representation, in particular, when there is uncertainty about the true breakpoint (for example, if there are segmental duplications) or when the code does not want to make any assertion about the breakpoint location.

3. For phased copy number variants (or for "partitioned" but not phased variants), we use the notation 
 `<CN:n>` to represent the allelic copy number.
   For example, something like this (tabs changed to spaces to condense):

```
#CHROM POS  ID REF    ALT           QUAL FILTER INFO FORMAT SAMP1 SAMP2 SAMP3 SAMP4
chr1   1000 .  <CN:1> <CN:0>,<CN:2> .    .      .    GT:CN  0|0:2 0|1:1 2|2:4 1|2:2
```
  Older versions of the code used `<CNn>` (with no colon) as the allele representation, but this caused difficulties because strict interpretation of the specification said that all ALT alleles had to be defined in the header and this in turn required knowing in advance all of the alleles that might be encountered or writing the output file in two passes. The idea behind `<CN:n>` is that the allele type/template is defined once in the header, but n is variable.

4. For complex structures, such as the _C4_ locus (Sekar, 2016; Kamitaki, 2020 (under review)) and other examples, we represent the structural haplotypes as particular alleles with encodings of the structure representing the key biology.  For example, at _C4_, we previously defined ALT alleles like `<H_n_n_n_n_x>` for example `<H_2_1_1_1_B>`. The encoding is described in context (or in the description field).
   For _C4_, the encoding is the (allelic) copy number of total _C4_, _C4A_, _C4B_, the HERV element and an optional character suffix identifying a particular haplotype (among structurally equivalent, recurrent haplotypes that appear to have arisen independently in humans).
   As one additional example, we might label an allele as `<H_3_1_2_3_insCT>` representing a haplotype that carries 1 copy of C4AL, two copies of C4BL, and a common frame shift variant of potential phenotypic importance.

   While there are other representations one could choose, these were convenient, reasonably human readable (at least for us) and worked well with other tools, such as beagle, without requiring modifications to the downstream tools.

I would also like to say that while I don't expect such representations to be standardized to the point of interoperability or diagnostic use, I do think it is useful to allow the VCF format to have some flexiblity, enough to permit experimentation and innovation within the standard through the use of certain conventions, such as allele naming conventions like I describe above, which might only be understood by certain tools.

At the same time, I think it is good if it is easy tools encountering an unrecognized allele format to not make any special assumptions about it.

If VCF does not have this flexibility, then I think the alternative will to use other non-standard file formats, which I think leads to more file conversion, more friction trying to get tools to work together, more potential for errors, etc.

cwhelan:
It seems to me like @bhandsaker 's use cases shake out like this with respect to this pull request:

1. This is incompatible with the proposed changes since it can be applied to copy number variants without breakpoint claims. Either we'd have to revert to leaving SVTYPEs of DEL and DUP ambiguous with respect to breakpoint claims and encode that information in an INFO field (as in the option 4 above), or Genome STRiP would have to switch to using a different INFO tag to encode this logic (which doesn't seem like the end of the world to me personally, although @bhandsaker might have a different reaction, as long as we carefully explain the differences in semantics between VCF versions in the specs).

2. This seems fine with the PR language: `<CNV>` is an allowed top-level allele.

3. `<CN:x>` is not allowed by the current language in the PR, but maybe `CN` is worth adding as another type of top-level allele, for phased or "partitioned" copy number genotypes that are not making breakend claims.

4. It seems to me that arbitrary symbolic alleles can still be allowed as long as they are not given an `SVTYPE` (which would trigger the more rigorous parsing that this PR is trying to enable). The internal structure of this sort of allele shouldn't really be standardized in the spec.

For 3 and 4, it'd be very useful if the spec allowed "parameterized" symbolic alleles -- like `<CN:x>`, but where you don't have to enumerate every single possible value of `x` with its own header line. Maybe that's something to tackle in a different PR, though? 

@d-cameron @thefferon @bhandsaker Does this summary seem right to you? Thoughts?

d-cameron:
> It seems to me that arbitrary symbolic alleles can still be allowed as long as they are not given an SVTYPE (which would trigger the more rigorous parsing that this PR is trying to enable). The internal structure of this sort of allele shouldn't really be standardized in the spec.

><CN:x> is not allowed by the current language in the PR

<CN:x> is still allowed, it just has undefined semantics and is not considered a "non-structural symbolic allele". We just need to remove this phase and define unknown symbolic alleles as having implementation-dependent sematics.

> maybe CN is worth adding as another type of top-level allele, for phased or "partitioned" copy number genotypes that are not making breakend claims.

Allele-specific copy number is already supported purely by specifying the sample alt allele genotype. We're spreading the info required to correctly parse an event around quite a bit with such a change. We can already encode this information as follows:
```
#CHROM POS  ID REF    ALT           QUAL FILTER INFO FORMAT SAMP1 SAMP2 SAMP3 SAMP4
chr1   1000 .  N <CNV>,<CNV> .    .      CN=0,2    GT:CN  0|0:2 0|1:1 2|2:4 1|2:2
```

Except for the problem that, like many other SV field, the CN field only supports a single value thus only a single SV allele per record. If we fix that issue, and the fact that the spec defines the CN FORMAT field as sample CN, not ASCN (which make no sense since FORMAT is sample-independent), then there wouldn't be a need for <CN:x> ALTs.

> This is incompatible with the proposed changes since it can be applied to copy number variants without breakpoint claims. Either we'd have to revert to leaving SVTYPEs of DEL and DUP ambiguous

I quite like the idea of using SVTYPE for specifying the claim type. To make a copy number based del claim, you would write ALT=DEL, SVTYPE=CNV. It's still nott backwards compatible but it's less of a change than any of the other proposals. ALT=DEL, SVTYPE=CNV, would essentially be a generic claim of copy number loss, or a specific claim of ASCN=0 when genotyped. Similarly DUP/CNV would be defined generically as elevated copy number up to twice sample ploidy, or ASCN=2 when genotyped.

> it'd be very useful if the spec allowed "parameterized" symbolic alleles 

At a high level, I'm not a fan of encoding information into the ALT allele when it can already be specified elsewhere.

>we use the notation <CN:n> to represent the allelic copy number.

Technically using <CN:1> as REF is a voilation of the specs as REF must be `[ACGTN]+`. 1.6.1.4 specifies that the REF should be the base immediately preceding the variant, so the <CN:1>, whilst human readably and explicitly a CN call, is not syntactically valid and will result in a parsing failure if spec-compliant.

cwhelan:
@d-cameron Thanks for your answer. 

>CN:x is still allowed, it just has undefined semantics and is not considered a "non-structural symbolic allele". We just need to remove this phase and define unknown symbolic alleles as having implementation-dependent sematics.

Just to be 100% clear, what triggers having something be a "structural" vs "non-structural" symbolic allele? Is it just that the top level is one of the enumerated types in the SV section? Or is it triggered by having an SVTYPE on the variant, ie if you have an SVTYPE defined you need to use one of the closed set of top-level alleles? In the conversation about @jmmut said it should be the former, which is fine, but I think the spec should say so explicitly.

>Allele-specific copy number is already supported purely by specifying the sample alt allele genotype. We're spreading the info required to correctly parse an event around quite a bit with such a change. We can already encode this information as follows:

>`chr1   1000 .  N <CNV>,<CNV> .    .      CN=0,2    GT:CN  0|0:2 0|1:1 2|2:4 1|2:2`

I had actually forgotten about the INFO level CN tag -- I'm not actually sure what it's supposed to mean from reading the description in the current spec.. the copy number of the alternate allele? Does anyone use it? Maybe it might be better to rename it to `ASCN` rather than `CN` to avoid confusion with the FORMAT CN? 

This could work, but I don't really love having multiple identical alleles in the ALT field -- I think it really decreases human readability, and seems like it could be prone to bugs in code. I guess in practice the alleles could be renamed to `<CNV:0>, <CNV:2>` etc by individual tools for readability while keeping a top level allele in the closed set. 

>Technically using CN:1 as REF is a voilation of the specs as REF must be [ACGTN]+. 1.6.1.4 specifies that the REF should be the base immediately preceding the variant, so the CN:1, whilst human readably and explicitly a CN call, is not syntactically valid and will result in a parsing failure if spec-compliant.

Fair enough, that's a good point.

rhdolin:
Hi,

Most of the folks on this thread are YEARS ahead of me in understanding the nuances of SVs, so I hope folks can bear with me for a bit of a more basic comment... The biggest challenge I'm facing now, trying to extract SVs from VCFs for use in a clinical decision support pipeline, is the inconsistent use of INFO.SVTYPE, PRECISE/IMPRECISE, INFO.CIPOS, INFO.CIEND, FORMAT.GT, and FORMAT.CN. I'm afraid I don't remember where some of my sample VCFs originated, but I'm seeing cases where, for instance, SVTYPE=INS, GT=0/1, CN=2 AND SVTYPE=INS, GT=1/1, CN=2. I'm seeing cases where SVTYPE=CNV and there is no CN. I'm seeing cases that are PRECISE and include CIPOS/CIEND. etc.

What would be super helpful would be more guidance in the VCF spec around how each of these fields are to be used depending on SVTYPE. I'd like guidance on how CIPOS/CIEND overlap with dbVAR's inner/outer start/end. It would help to have a validation engine that allows me to determine if a particular variant caller is putting out conformant SV calls. Also, it would be useful if I could optionally include CIPOS/CIEND in tabix indexing so that I can more reliably extract all or potentially all SVs in a range of interest.

There you have my wish list :-)

thefferon:
A bit of feedback:

- I am in favor of making SVTYPE=DUP synonymous with 'tandem duplication'*. As reflected in the January 30, 2020 pdf of the spec.
- I am not convinced in the case of deletions, however. Is there a valid distinction, for the purposes of the spec, between a "breakpoint claim" loss and a "copy number claim" loss? It makes sense to me that we move read-depth-detected copy number gain from SVTYPE=DUP to SVTYPE=CNV – because you don't know the genomic location of the gain. But the same rationale does not apply to a read-depth-detected copy number loss – where we do know the location of the loss (albeit not to basepair resolution).

\* A tandem duplication can be:
* direct (same strand orientation as reference copy)
* inverted-upstream (opposite strand orientation, inserted upstream of reference copy)
* inverted-downstream (opposite strand orientation, inserted downstream of reference copy)
I am still concerned that the current spec can only describe the first case. Maybe this is not a big real-world problem; if that is the case then I am willing to drop the subject.

thefferon:
I hope I haven't missed the point in the above discussion, but I support the extant practice (and its explicit addition to the spec) of treating allele-specific copy number changes as SVTYPE=CNV and ALT=<CN=0>,<CN=1>,<CN=2>,<CN=3>... even if this means making 'CN=*' a member of the closed set of first-level ALT values. In my experience this model works, so I prefer to allow it unless it causes undue difficulties to tool developers or obscures data comprehension.

tskir:
I am afraid this PR is getting dangerously technical and the risk of it getting stuck for another two years is rising. I suggest that we focus on a minimal set of changes for the most basic SV types only and do not continue discussion of more complicated cases _in this PR._ Once it is merged, we can build on it with other PRs.

In terms of the actual changes, I propose the following approach. Much of it has already been said before, I am mostly just summarising.

## Completely decouple breakpoint and copy number claims
* **DEL, DUP, INS, INV, BND** would be strictly breakpoint claims. DUP would represent a simple tandem duplication with the same orientation to the reference.
* **CNV** would be a strictly copy number claim. For consistency, it would have to include non-multiallelic events, for example simple deletions and duplications, if they have not been detected up to breakpoint resolution. We can focus on further standartising CNVs in a later PR. (In fact, I would be willing to take this part of the work, since I also participate in the ELIXIR hCNV project.)

## Synchronise the lists of top-level types accepted by symbolic alleles and by SVTYPE
Currently the lists of six SV types are the same, but their meanings are slightly different in those two contexts, and the specification does not provide the allowed combinations of symbolic alleles and SVTYPEs, or their interpretations. This, for example, allows the existence of:
* `<CNV>` with `SVTYPE=DUP`;
* `<DUP>` with `SVTYPE=CNV`;
* `<CNV>` with `SVTYPE=CNV` and elevated copy number specified by `CN`.

This has got to be really confusing for a user because it is not clear if those combinations are valid and whether they are identical to each other. Additional confusion is brought about by the fact that `DUP` means a strictly tandem duplication in a symbolic allele context, but _any_ duplication when in SVTYPE context.

I suggest **exactly** synchronising the meaning of SVTYPE and the symbolic alleles. In fact, we could get rid of SVTYPE definitions entirely and just say that the list of allowable values is exactly equal to the top-level SV types for symbolic alleles, and that the value of SVTYPE must be always exactly equal to the top-level type of the structural symbolic allele.

@d-cameron, @jmmut, (and others), please let me know what you think about this. I have a number of comments and corrections for this PR (mostly wording and consistency things), but I don't want to introduce them before we have settled on a specific approach.

cwhelan:
@tskir Unless I'm misunderstanding, under your proposal a bi-allelic or rare duplication discovered by a read depth-based method would have to be represented by `SVTYPE=CNV`, an `ALT` of `<CNV>`, and `INFO/CN=2` or `FORMAT/CN=3` or both. Depending on your interpretation of what a 'breakend claim' is, the same might apply to a deletion discovered with a depth based method (since it has no resolved breakpoints), but with different `CN` values. My issue is that this gives no easy way to, for example, query the VCF file for all duplications, especially since interpreting the `CN` fields as deletions or duplications depends on knowing the expected ploidy of the reference contig. 

Please correct me if I'm wrong about this. If you are suggesting adding something to address this use case in a later PR, I get a little nervous about decoupling the two changes -- it seems like we'd leave the spec in a slightly broken state between the two PRs.

I think that allowing `SVTYPE` to vary from `ALT` (so that you can say `SVTYPE=DEL`; `ALT=<CNV>`) gives people flexibility they need to continue working with non-breakend-based CNV claims while separating them from breakend-based claims enough for downstream tools to know when they can try to construct an alternate sequence from them. I agree that it could be a little confusing but I'd recommend getting around that by adding detailed wording and examples to the spec document.

tskir:
@cwhelan I can see your point. How about this then?

* For breakpoint claims, top level of the symbolic structural allele type **must** be one of {`<DEL>`, `<INS>`, `<DUP>` (only tandem), `<INV>`, `<BND>`}. In this case, SVTYPE **must** match the top level of the symbolic structural allele exactly.
* In case breakpoints are unknown or not reported, symbolic allele **must** be `<CNV>`. In this case, SVTYPE **must** be one of exactly three types:
  + `SVTYPE=DEL` — copy number decrease compared to the reference;
  + `SVTYPE=DUP` — copy number increase compared to the reference (in this case the implied duplication is not necessarily tandem);
  +  `SVTYPE=CNV` — multiallelic copy number region of both increase and decrease of copy number compared to the reference.

In case we go with this approach, all of this needs to be explicitly and very carefully worded in the specification.

jmmut:
After reading multiple times this thread and the related ones, and also the current spec, I think the purpose of the current writing  :

## symbolic SV ALT : 

- DEL INS DUP INV CNV: read-depth claim. E.g. see the current wording for DEL: "Region of elevated copy number relative to the reference". CNV is the general category: DEL is equivalent to CNV/CN0, INS is CNV/CN1+ of new sequence, DUP is CNV/CN2+, etc. and the specific should be preferred.
- BND: breakend claim. Different to the other ALTs.

tskir:
@jmmut By current writing do you mean the state of the specification _before_ this PR? I doubt that all five types (DEL INS DUP INV CNV) were intended as read-depth claim, especially INS and INV because they cannot be inferred using the read depth alone. What is certain is that the current definitions are ambigous, and we want to clarify them (possibly changing the meanings compared to the original ones).

It is my understanding that in the previous calls & offline discussions we have already agreed to separate breakpoint claims from copy number claims; my proposal in the [comment above](https://github.com/samtools/hts-specs/pull/465#issuecomment-585242848) just suggests a formal way to do that.

jmmut:
(Sorry for the confusion, I misclicked and sent the message before it was ready, so I deleted it. This is the complete message.)

After reading multiple times this thread and the related ones, and also the current spec, I think the purpose of the current spec writing (before this PR) was:

## symbolic SV ALTs: 

- DEL INS DUP INV CNV: **read-depth claim**. E.g. see the current wording for DEL: "Region of elevated copy number relative to the reference". CNV is the general category: DEL is equivalent to CNV/CN0, INS is CNV/CN1+ of new sequence, DUP is CNV/CN2+, etc. and the specific should be preferred.
- BND: **breakpoint claim**. Different claim than the other ALTs.

With that, I could see that a read depth claim DUP, can use ALT=DUP and any SVTYPE (possibly DUP for simplicity), but SVTYPE becomes useful for a breakpoint claim where you use the breakend notation in ALT (to make a breakpoint claim) and put the SVTYPE=DUP. I'm not saying we should keep this as it is, I'm just trying to understand the history of the current writing and whether it can be clarified or needs a rewrite.

This is kind of similar to the last proposal by @tskir, where a read depth claim is specified with ALT=CNV and anything else is breakpoint claim.



If we go with the meaning I just explained, from tskir's comment I can see how it may not apply to INS and INV if those can not be identified by a read depth analysis (I'm no expert on that field), but I wonder if the main problem @d-cameron explained in the calls (about other callers misusing the ALT and SVTYPE fields) still applies? the **evidence** the callers have would be expressed in ALT (BND is breakpoint claim, anything else is read depth claim; or we can change to a similar combination as tskir's one), and the **interpretation** is expressed in SVTYPE.

One concern with that approach is if breakpoint claims are not a superset of read depth claims, and if it would make sense to be able to state both at the same time.

Also, tskir, how do you classify a non-tandem DUP with known location in your suggestion? With a breakend in ALT? is DUP:TANDEM unnecessary then?

tskir:
@jmmut You raised some very good points. I also had to re-read parts of the specification to address them.

> I think the purpose of the current spec writing (before this PR) was:
> symbolic SV ALTs:
> * **DEL INS DUP INV CNV:** read-depth claim. [...]
> * **BND:** breakpoint claim. Different claim than the other ALTs.

You were quite right to notice that **BND** is different from other symbolic allele types, and that the other types do not make breakpoint claims. However, those other types are _not_ read depth claims either in the specification. “Read depth” refers to a specific set of methods of (imprecisely) detecting increase and decrease in segment copy numbers. Rather, the specification makes the distinction between “precise” and “imprecise” calls, _regardless of the method of detection._ In section 1.4.5 “Alternative allele field format”, subsection “Structural Variants” starts:

> In symbolic alternate alleles for **imprecise structural variants,** the ID field indicates the type of structural variant...

That means that if you have an imprecise structural variant (meaning it has not been detected up to base pair resolution, using whatever method), you specify it using:
* **DEL** — for _any_ decreased copy number (for example, CN2 → CN0 and/or CN1)
* **DUP** — for _any_ increased copy number (for example, CN2 → CN3, CN4 and so on). Since the specification reserves the additional `DUP:TANDEM` subtype, this implies that the “regular” DUP can be **any** type of duplication, including dispersed, more than one additional copy inserted, different orientations to the reference, etc.
* **CNV** — only to be used when there are both DELs and DUPs in a single call
* **INS** — when a novel sequence is inserted (this is not necessarily related to copy number change, since the sequence is marked as novel)
* **INV** — when a portion of the reference sequence is inverted (this is also not related to copy number change)

The current wording of section 1.4.5 leaves it ambiguous what to do with precise structural variants — e. g. when you know coordinates of a huge deletion up to single nucleotide resolution. Some examples use the `INFO/IMPRECISE` key to indicate this, although this key is not officially reserved for this purpose.

> but SVTYPE becomes useful for a breakpoint claim where you use the breakend notation in ALT (to make a breakpoint claim) and put the SVTYPE=DUP

I'm not necessarily against doing it this way, but:
* BND notation is _in general_ used for complex rearrangements which do not necessarily fit into the five simple SV types discussed above. Hence, it most cases it would be impossible to assign a standard SVTYPE to a BND.
* The specification doesn't currently say a word about this, and actually all breakend examples in the current specification are using `SVTYPE=BND`. If we decide to allow specifying other SVTYPEs for BNDs, this needs to be explicitly clarified in the specification.
* Currently, filtering on `SVTYPE=BND` is the only simple way to find all BND records in the VCF. You  can't filter by symbolic allele, because BND are using a specific format for the ALT allele.

> Also, tskir, how do you classify a non-tandem DUP with known location in your suggestion? With a breakend in ALT? is DUP:TANDEM unnecessary then?

In light of the points you raised, I think I have an idea for a better proposal which would be much more consistent and also mostly compatible with the current specification version. I will post it shortly.

tskir:
Based on feedback from @cwhelan and @jmmut, I present to you:

# Consensus proposal, version 3

## Retain the same SV types for symbolic alleles; expand & clarify their definitions
The types currently present in the specification are just fine, but poorly defined (it took me and @jmmut a couple of days to understand their true indended meaning). Let's define them very explicitly:
* **DEL**: any copy number decrease compared to the reference (for example, CN2 → CN1 and/or CN0).
* **DUP**: any copy number increase compared to the reference (for example, CN2 → CN3 and/or CN4 and so on). By default this can refer to **any** type of duplication, including tandem or dispersed, adding one or several copies, same or different orientations to the reference.
  + **DUP:TANDEM** subtype: tandem duplication in the same orientation to the reference. This can also include more than one additional copy.
* **CNV** — only to be used when there are both DELs and DUPs in a single call (region where increased and decreased copy number is observed).
  + For **DEL, DUP** and **CNV** specific copy number changes can be specified using `INFO/CN` and `FORMAT/CN` tags.
* **INS** — when a novel sequence is inserted (this is not necessarily related to copy number change, since the sequence is marked as novel).
* **INV** — when a portion of the reference sequence is inverted (this is also not related to copy number change).
* **BND** — breakend notation, no changes to the current spec.

## Explicitly make SV calls of all types imprecise by default. Mark precise calls using CIPOS and CIEND
Again, this already looks like the way the current specification is intended to work, it's just not clear. Let's explicitly say that by default all structural symbolic alleles denote an **approximate** variant location with the start/end position as **best estimates.**

To make a “breakpoint claim”—that is, to specify that the start and/or end of a variant are known to single base resolution—existing CIPOS and/or CIEND fields **must be set to (0, 0) values.** (Alternatively, if people prefer, we could set up a special value, for example `CIPOS=PRECISE`, for this purpose, but I think double zero works just fine.) If CIPOS and CIEND are not specified, it must be assumed that the call coordinates are **not** precise, but uncertainties are not available or not reported.

## Exactly synchronise the lists of top level structural symbolic alleles and SVTYPE values
In this verison of the proposal, I'm back with my suggestion to completely synchronise the allowable values of the two lists. Since there will be already a way to discern exact (breakpoint) claims from inexact (e. g. read depth) claims, there is no need to mix the terms together between the symbolic alleles and the SVTYPE.

----

As far as I can see, this proposal addresses all concerns raised by @cwhelan:
* There will be a simple way to query the VCF for all duplications, whether they were discovered using breakpoints or read depth, because they will all have `ALT=DUP` (possibly with subtypes) and `SVTYPE=DUP`.
* There is a way to filter by precise and imprecise claims, using CIPOS/CIEND, without the need to mix different types in symbolic alleles and SVTYPEs.
* The spec will not be left in a broken state.

And the ones by @jmmut:
* This proposal retains as much backwards compatibility with the existing specification as possible; it only uses existing fields and is not introducing any new ones.
* It decouples breakpoint (“preciseness”) information from the actual variant type, allowing the users to specify a broad range of both precise and imprecise structural variant types.
* As for you question about specifying a non-tandem DUP with known insertion location: this will have to be done using BNDs; however, the current specification version also does not provide a way to do this, so there is no regression here. This new variant type can be possibly introduced in the future.

@d-cameron @cwhelan @jmmut Please let me know what you think.

cwhelan:
@tskir 

Your proposal is internally consistent but I think it misunderstands a couple of historical things and it moves away from the original intent of this PR, which is trying to figure out a way to distinguish between copy number and breakend-based claims so that downstream tools can figure out how to (at least partially) reconstruct the haplotype altered by the event, while not requiring every tool to write BND-formatted records (which, while complete, are extremely non-human readable and resistant to quick analyses). 

Historically and at the moment, both copy number-based calling methods and those based on read pair and split read mappings have used the same SVTYPEs and ALTs, leading to difficulty in interpretation. For example, the former class of tool might detect elevated copy number of a region and create an `SVTYPE=DUP`, `ALT=<DUP>` record. Read-pair based methods would detect read pairs mapping with a signature indicative of a tandem duplication, and use the same SVTYPE and ALT. 

IMPRECISE and CIPOS are used when the event is detected by read pairs or similar evidence that gives an indication of breakpoint structure, but the exact breakpoint coordinate is not known. Copy number based tools typically don't use those fields historically, since they are usually based on segmenting the genome arbitrarily into bins in which read depth is measured without any claim as to the breakend structure.

Your comment yesterday:

>For breakpoint claims, top level of the symbolic structural allele type must be one of {DEL,INS, DUP (only tandem), INV, BND}. In this case, SVTYPE must match the top level of the symbolic structural allele exactly.
>In case breakpoints are unknown or not reported, symbolic allele must be <CNV>. In this case, >SVTYPE must be one of exactly three types:
>SVTYPE=DEL — copy number decrease compared to the reference;
>SVTYPE=DUP — copy number increase compared to the reference (in this case the implied duplication is not necessarily tandem);
>SVTYPE=CNV — multiallelic copy number region of both increase and decrease of copy number compared to the reference.

Was a pretty good summary of the change in semantics that @d-cameron is trying to introduce into to the spec (with the backing of a group of us listed above who participated in calls organized by @thefferon several years ago). Rather than "In case breakpoints are unknown or not reported" I would say, "if the call makes no claim about breakpoint structure", however. In this case SVTYPE is providing the interpretation of the variant and the ALT can be used to derive breakpoint claims, as you stated earlier. I think that the changes in Daniel's PR are good and just need, as you said before, some careful and comprehensive wording and examples to flesh them out.

tskir:
@cwhelan 

Thank you for sharing your thoughts. I drafted proposal v3 following comments by @jmmut (while still taking into account your feedback on v1), and it seems to me that it better captures the original intent of the specification, introduces fewer potentially breaking changes, and is overall better and more consistent than v2. I am not _pushing_ for it, and if we have to go with v2 I will reluctantly agree, but let me try and defend v3 a bit.

I totally get the original intent of this PR and the necessity to separate different types of calls. Here's how v2 and v3 approach this using different means:

### Duplications _(not necessarily tandem)_
| Detection method | Breakpoint uncertainty | [Proposal v2](https://github.com/samtools/hts-specs/pull/465#issuecomment-585242848) | [Proposal v3](https://github.com/samtools/hts-specs/pull/465#issuecomment-585680211) |
| ---------------- | ---------------- | ---------------- | ---------------- |
| 1. Inferred from **read depth** | ✖ Unknown or not reported | **`ALT=⁠<CNV>`**     <br>`SVTYPE=DUP`<br> | `ALT=⁠<DUP>`     <br>`SVTYPE=DUP`<br> |
| 2. Inferred from split read mapping | ✖ Unknown or not reported | `ALT=⁠<DUP>`<br>`SVTYPE=DUP`<br>`CIPOS=`???<br>`CIEND=`??? | `ALT=<DUP>`<br>`SVTYPE=DUP` |
| 3. Inferred from **read depth** | **≈** Estimated e. g. using bin coordinates ★ | **`ALT=<CNV>`**<br>`SVTYPE=DUP`<br>`CIPOS=(-1042,5173)`<br>`CIEND=(-8324,2775)` | `ALT=<DUP>`<br>`SVTYPE=DUP`<br>`CIPOS=(-1042,5173)`<br>`CIEND=(-8324,2775)` |
| 4. Inferred from split read mapping | **≈** Estimated using analysis of adjacent reads | `ALT=<DUP>`<br>`SVTYPE=DUP`<br>`CIPOS=(-56,72)`<br>`CIEND=(-114,83)` | `ALT=<DUP>`<br>`SVTYPE=DUP`<br>`CIPOS=(-56,72)`<br>`CIEND=(-114,83)` |
| 5. Observed directly by long read sequencing | ✔ Known exactly | `ALT=<DUP>`<br>`SVTYPE=DUP`<br>`CIPOS=`???<br>`CIEND=`??? | `ALT=<DUP>`<br>`SVTYPE=DUP`<br>`CIPOS=(0,0)`<br>`CIEND=(0,0)` |

★ _You mentioned that tools based on read depth don't usually use CIPOS and CIEND because of how their algorithms work. However, it is still possible to estimate uncerntainties in those kinds of approaches. For instance, if a copy number increase is detected as starting in bin #N and ending in bin #M, I think it quite makes sense to use start and end coordinates of those boundary bins to infer CIPOS and CIEND. It makes a difference whether the bins were 500 nt sized or 1,000,000 nt._

### For comparison: CNV region (both deletions and duplications present)
| Detection method | Breakpoint uncertainty | [Proposal v2](https://github.com/samtools/hts-specs/pull/465#issuecomment-585242848) | [Proposal v3](https://github.com/samtools/hts-specs/pull/465#issuecomment-585680211) |
| ---------------- | ---------------- | ---------------- | ---------------- |
| 6. Region of CN increases and decreases inferred from **read depth** | ✖ Unknown or not reported | **`<CNV>`**<br>`SVTYPE=CNV` | **`<CNV>`**<br>`SVTYPE=CNV` |

From what I see, v3 has the following advantages over v2:
* Since v2 makes DUPs (and other SV types except for CNV) “breakpoint claims” by default, what does it mean we if specify a DUP at a given position and do _not_ specify CIPOS and CIEND? Does it mean a claim for exact breakpoint coordinates (case 5) or claim for inexact coordinates with unknown uncertainty (case 2)? In contrast, v3 creates a clear hierarchy for _all_ SV calls:
  + No CIPOS/CIEND → imprecise, uncertainties not estimated
  + Nonzero CIPOS/CIEND → imprecise, uncertainties given
  + Zero CIPOS/CIEND → exact call
* Symbolic allele types and SVTYPEs are perfectly synchronised in v3. No problems with DUP having different meanings in two different contexts.
* CNV only appears in situations where increase and decrease of copy number are mixed together in a single call (case 6), _as intended by the current specification wording._

If I understand correctly, your primary concern with v3 is that case 1 is indistinguishable from case 2, as well as case 3 from case 4. But I think that **this is exactly the point,** as they represent **the same events inferred by different methods.** By making CNV represent read depth claims in v2, we are tying that symbolic allele type to a specific detection method. I think it is much better when the calls themselves are completely agnostic to the detection technology, as implemented in v3 (and again, as appears to have been intended by the existing specification wording).

Having said that, I realise that in many cases there **is** a need to know how the call was produced—I 100% support that. But I think this is much better done by e. g. introducing a specific (possibly even non-optional or highly encouraged) INFO field with a list of standardised detection types, e. g.:
* READ_DEPTH
* SPLIT_READ
* LONG_READ_SEQUENCING
* (The list can be periodically amended—new CNV detection types do not come up all that often)

I believe someone mentioned this possibility during one of the recent calls (can't remember who unfortunately), but did not press this further. I think this is the best approach, as it encodes useful additional information in an INFO field rather than cramming it into the symbolic allele type.

@d-cameron @cwhelan @jmmut Please do let me know what you think.

d-cameron:
> Please do let me know what you think.

My concerns with proposal 3 are:

1. Redundancy and complexity

P3 requires 3 fields when we can disambiguate with 2. It still does not clarify what alt=<CNV>, svtype=<DUP> actually means.

> Symbolic allele types and SVTYPEs are perfectly synchronised in v3. No problems with DUP having different meanings in two different contexts.

This problem still exists in p3 - it just requires yet another field to determine what their meaning is. 

> Exactly synchronise the lists of top level structural symbolic alleles and SVTYPE values

It sounds very much like the difference between p2 and p3 is whether we reuse SVTYPE for disambiguation or deprecate SVTYPE and use a new field.

Less complex is always better than more complex when we're dealing with specifications. I don't think the utility of this additional field justifies the additional complexity.

2. New field vocabulary

The vocabulary must be either open or close. Both have issues. An open vocabulary would be required as new technologies will get introduced in the future but to unambigously interpret variants then the vocab needs to be closed. Realistically, new technology callers will just lie about their evidence and choose whatever pre-defined value corresponds to the claim being made. E.g. using your closed vocab, a microarray caller will just write READ_DEPTH, and a read-pair based caller will just write SPLIT_READ. This entirely defeats the purpopse of having tech type.

3. Technology agnotic

> I think it is much better when the calls themselves are completely agnostic to the detection technology.

I agree with this. VCF is currently technology agnotic and it should remain so. It shouldn't matter whether read depth or microarray probe intensity is used to define a CNV, merely that a CNV claim is being made. My issue with p3 is that it's not tech agnostic. You need to look up the tech source field to determine what claim is being made when a VCF has a DUP record.

----

> what does it mean we if specify a DUP at a given position and do not specify CIPOS and CIEND? 

I always interpreted an unspecificied confidence interval implicitly indicates an interval of [0,0] and my caller only writes CIPOS when the confidence interval is non-zero.

d-cameron:
> If I understand correctly, your primary concern with v3 is that case 1 is indistinguishable from case 2, as well as case 3 from case 4. But I think that this is exactly the point, as they represent the same events inferred by different methods.

Case 1 and 2 being indistinguishable is exactly my primary concern. Yes a simple tandem duplication looks the same for both case 1, and case 2, but case 1 and 2 are making different claims about the structure of the genome.

For a genome with segments ABC:
- case 1 is claiming the end of B is connected to the start of B.
- case 2 is claiming there are 2 copies of B

These are not claiming the same thing! A actual simple tandem duplication requires a) both of these claims to be true and b) there are no other claims interfere with B, c) there is reference allele support for an AB and/or AC transition.

Case 1 allows for ABBC or ABC / circular B
Case 2 allows for BABC, ABBC, ABCB, or ABC/circularB, and a whole host of most extreme interpretations such as AC/circularBB

And that's before we get into the possibility of missing or related events. case 1 could be part of a complex rearrangement and there may well be no copy number change of B at all.

To reiterate: **case 1 and 2 are making very different claims about the structure of the genome**. VCF needs to be able to unambigiously specify the actual claims being made.

tskir:
@d-cameron After your comments I now understand your issues with proposal v3 much better, and I see the way to fix them. Let's continue the discussion on this:

### 1. Technology type vocabulary
I agree that a new “detection technology” field with either an open or a closed vocabulary is a bad idea, for the reasons you described. Let's ditch it. The claim type can be specified without it—see below.

### 2. Position uncertainties
Since this PR deals with claim types, not uncertainty specification, we can clarify and standardise CIPOS handling in a future separate PR. It has no direct effect on proposal v2 vs. v3 discussion.

### 3. Specifying claim types
Your example with the A/B/C genome was very informative. I now see what you mean by different “claim types”. It looks to me that we actually have the same view of the situation, we've just been using different terms for the same thing.

> Case 1 and 2 being indistinguishable is exactly my primary concern. [...] These are not claiming the same thing.

The idea in my example was that cases 1 and 2 *do* represent and claim exactly the same thing. Note that the header for cases 1–5 is “Duplications **(not necessarily tandem)**”. The idea is that `DUP` in proposal v3 represents a general duplication, that is, increase in _copy number_ in the genome, in whatever location and in whatever orientation. So in your terms, this corresponds to a **copy number claim.**

For tandem duplications, in v3 you don't use `DUP`, you use the subtype `DUP:TANDEM` (see it defined here as a sub-bullet point in [proposal v3](https://github.com/samtools/hts-specs/pull/465#issuecomment-585680211)). So that would represent a **breakpoint claim.**

### 4. Redundancy and complexity of proposal v3

> P3 requires 3 fields when we can disambiguate with 2. It still does not clarify what alt=, svtype= actually means.

Without the technology vocabulary, the (updated) proposal v3 will also use only the two existing fields, ALT and SVTYPE.

It will clarify the meaning of specific types and subtypes of structural variants in the way I described above: for example, with `DUP` making a general copy number claim, a `DUP:TANDEM` a specific version of a breakpoint claim, etc.

Also, if we are allowed to be bold and deprecate SVTYPE, since it _does_ become redundant in this updated version of proposal v3, we can specify all information about the claim type using just **one** field, the symbolic allele type/subtype. I think it would the most elegant solution of all, utilising only the minimal number of existing fields, and having decent backwards compatibility.

@d-cameron @jmmut @cwhelan Please let me know if you have any additional feedback.

d-cameron:
Design goals for this PR are:

1. Retain as much backwards compatability with VCFv4.2 as possible

2. Be able to unambiguously determine if a call is making a CN claim, a breakpoint claim, or both.


Unfortunately, 1 and 2 are incompatible so we have to break something. Realistically, we're not going to have a large number of pre-4.4 VCFs around for a very long time. As much as I'd like to say that all breakpoint-based SV caller should report everything in BND notation, that's not going to happen.

> For tandem duplications, in v3 you don't use DUP, you use the subtype DUP:TANDEM (see it defined here as a sub-bullet point in proposal v3). So that would represent a breakpoint claim.

If DUP is a CN claim, DUP:TANDEM is a breakpoint claim, then how do you make a claim of an actual tandem duplication (ie CN + breakpoint)? T

I know it's very late in this discussion, but I'll just put this out as an option for dicussion:

**Proposal 3a: explicit specification of support type**

Proposal 3 has the fundamental problem that it fails design goal 2 (eg the DUP example above). This can be resolved with an additional `SVSUP` with allowable values of `CN`, `BP`, and `CNBP` (not wedded to theses names, feel free to propose something different). This approach fully satisfies the design goals but it does result in additional specifications complexity, (although in some senses, it's the simplest solution). It doesn't resolve the many way an event can be inconsistent (e.g. `SVTYPE=DEL ALT=<DUP>`) and even creates more (e.g. `ALT=N[chr:pos[`, `SVSUP=CN`) but this just requires verbiage in the specs, and more validator rules. It's a trade-off worth considering.

Practically, this approach has a few advantages. One that I particularly like is that downstream v4.4-aware SV event classifiers tools will already be dealing with the current ALT/SVTYPE mess, and it'd relatively simple to specify the pre-4.4 VCF out of band. E.g. `classify_events --svinput manta.vcf --cninput ascat.vcf`. Most existing VCFs contain either all CN or all BP claims. It's only DBs such as dbVar that have a mix of both.




`--svsup bp`

jmmut:
> If DUP is a CN claim, DUP:TANDEM is a breakpoint claim, then how do you make a claim of an actual tandem duplication (ie CN + breakpoint)?

For me this boils down to my initial confusion, where I thought that a breakpoint claim included the CN claim. After some of your explanations I understood how they are independent, and the class of CN+breakpoint is different than only breakpoint. For this, I see how v3 as we have discussed is inherently flawed as you could only specify CN or CN+breakpoint, and if we change v3 to allow the 3 claims, it will be basically the same as the changes in this PR.

The alternative (partially) done in this PR is still not clear to me about how to identify a CN, BP or CN+BP using only ALT+SVTYPE, but we can keep trying to make it clearer. At this point I'm a bit skeptical we can make it very clear. The clearest to me would be something similar to the PR in the sense that ALT=[INS,DEL,INV,DUP] is CN+BP; ALT=CNV is CN and ALT=BND is BP; and SVTYPE is used to explain the common type (INS,DEL,INV,DUP) if ALT=CNV and ALT=BND. SVTYPE=[BND,CNV] could be used for types that are more complex than the other 4 simple types. But this is problematic if someone wants to state CN+BP using the bracket notation (I don't know if this is a real problem or if there are other possible complications). For this last reason I think I prefer the newest option:

The third idea of "SVCLAIM with allowable values of CN, BP, and CNBP" has the clarity I was hoping to get from reorganising ALT and SVTYPE, but the biggest problem I see with that is that people will stick to only ALT+SVTYPE unless we make SVCLAIM required. It looks a bit of a pain, but I think it would be justified to make SVCLAIM required for SVs. It would still make sense to clarify (and simplify?) ALT+SVTYPE, but this becomes non-critical.

mbaudis:
`DUP` (CNV) and tandem duplications have no clear relation. `DUP:TANDEM` is *NOT* a subset of `DUP`, but a precisely described genomic alteration in contrast to an expression about a change in absolute or relative count of the number of alleles over a continuous genomic region.

A sane option is to just have `DUP`, `DEL` as the single statements for describing CNVs, to have the option for chaining them to flanking events by id (obviously, CNVs are flanked by breaks, fusions and this follows a graph approach).

There is IMO no good use case, in a VCF, to have something tagged as `CNV`. This tag shouldn't exist?!

Tandems, precisely defined indels should be just that. Users can decide themselves from which size on they parse them as CNVs, and how to do that.

Please keep (rather, make) it simple.

cwhelan:
@mbaudis Events need to have type CNV if they are measuring total copy number of a segment and you can't confidently say that there is only one type of alt allele (DEL or DUP) at the locus.. unless you have another proposal for how to represent that use case. 

@d-cameron The SVCLAIM idea seems good to me at first glance. Our integrated WGS pipeline (descended from the gnomAD SV pipeline) includes events called both by depth signal only, breakpoint signal, or both, in one final output vcf. We already track a value roughly equivalent to SVCLAIM; standardizing it in the spec would make sense to me.

VCFv4.3.tex
julesjacobsen: Has there been any further discussion about this? I'd like to propose a slight shift in the way things are defined which would mean that the current 'small' and 'structural' split can be removed/unified leading to a simpler and more robust API.

## Definitions

If we define 'variation' as 'a change relative to the reference sequence'. Then, in this context the term 'structural' can be further defined as 'where there is a change in length or arrangement relative to the reference sequence'. Under this definition SNV and MNV are not defined as structural. However, precisely defined, non-symbolic, insertions and deletions are defined a 'structural' regardless of their length. To fit with the existing spec there must be a distinction between 'implicit' and 'explicit' structural variation. Here 'implicit' would be a non-symbolic indel, precisely defined in the `REF` and `ALT` fields  and 'explicit' would include the extra fields used to define symbolic structural variants such as `SVTYPE`,`LEN/END` and `SVLEN`.

For example:

```tsv
#CHROM POS     ID        REF              ALT          QUAL FILTER INFO
1   1   .   A   .   .   PASS    # non-variation
1   1   .   A   T   .   PASS    # A SNV - this isn't a structural variation, but could have an implicit SVTYPE=SNV
1   1   .   A   TC  .   PASS    # an implicit `INS` here `SVTYPE` `LEN` and `SVLEN` can all be calculated from the REF and ALT alleles
1   1   .   AC  T   .   PASS    # an implicit `DEL` here `SVTYPE` `LEN` and `SVLEN` can all be calculated from the REF and ALT alleles
1   1   .   A  T   .   PASS    # an implicit `DEL` here `SVTYPE` `LEN` and `SVLEN` can all be calculated from the REF and ALT alleles
1   2827694 rs2376870   CGTGGATGCGGGGAC C   .   PASS    SVTYPE=DEL;LEN=15;SVLEN=-14 # an explicit, non-symbolic, precise deletion
2   321682  .   T   <DEL>   .   PASS    SVTYPE=DEL;LEN=206;SVLEN=-205;CIPOS=-56,20;CIEND=-10,62 # explicit, symbolic, imprecise deletion
```

So the only real difference here is that of the implicit and explicit declaration of the `SVTYPE`, `LEN` and `SVLEN`. Explicit types _must_ be used with symbolic alleles and *can* be used with non-symbolic alleles. 

All the other symbolic alleles follow the existing specifications. In order to maintain consistency between symbolic and non-symbolic types using the `SVTYPE` key the values `SNV` and `MNV` would need adding to the reserved set. Strictly `SVTYPE` ought to be renamed `TYPE` to be able to cover all the types of variation so applications can treat variation in a uniform manner, but its workable without doing this.

The upshot of this is a clean and consistent API in things like the HTSJDK where a client needs to call `VariantContext.getType()` and `getStructuralType()` depending on the return values. For example `INDEL` or `MIXED` in the case of `getType()` not being `SNV`/`MNV` or `null` being returned from `getStructuralType()` when the ALT allele isn't symbolic. Furthermore it is more in line with the way variation is reported in the HGVS, at least in the way they categorise some variation e.g. [Substitution](http://varnomen.hgvs.org/recommendations/DNA/variant/substitution/), [Deletion](http://varnomen.hgvs.org/recommendations/DNA/variant/deletion/), [Insertion](http://varnomen.hgvs.org/recommendations/DNA/variant/insertion/), [Inversion](http://varnomen.hgvs.org/recommendations/DNA/variant/inversion/), which marginally helps applications ingesting VCF and reporting findings in HGVS.

Does this sound workable or sheer lunacy?

DarioS:
How about 

> [It sometimes is difficult to determine the exact position of a break, generally because of homologies between the
> sequences being modified, such as in Figure 9. The breakend is then placed arbitrarily at the left-most position,
> and the uncertainty is represented with the CIPOS tag.](http://samtools.github.io/hts-specs/VCFv4.3.pdf)

versus

> [Left-aligning is the wrong choice for SVs.](https://github.com/PapenfussLab/gridss/issues/266#issue-509752376)

Might this rule change to centre-positioning in the next version of VCF? Discuss.

d-cameron:
Good catch, I didn't notice that bit of the specs.Left-aligning is the wrong choice for SVs because a) it can't be done for +/+ or -/- oriented breakpoints (left alignment of one side causes right alignment of the other), and b) for RP-based caller, left-alignment frequently results in a called position that is the least likely of the positions in the CIPOS interval.

SVs should be either centre-aligned or aligned to the most likely position.

VCFv4.3.tex
tskir: This is a very valid comment, though not strictly speaking related to SV improvements. I've created a separate issue https://github.com/samtools/hts-specs/issues/545 and will be marking this discussion as resolved.

VCFv4.3.tex
tskir: At this stage we're trying to fix the simple SV types before addressing BNDs, so I've separated these discussions into https://github.com/samtools/hts-specs/issues/546 and will resolve this thread

VCFv4.3.tex
tskir: @julesjacobsen Thank you for your comment! Since we're trying to merge this PR with minimal changes, I've separated it into a separate issue: https://github.com/samtools/hts-specs/issues/547

VCFv4.3.tex
tskir: This is obsolete in light of recent development & other PRs, will re-formulate and raise this again if needed

VCFv4.3.tex
tskir: Outside of scope of this PR, will raise as a separate issue if necessary

VCFv4.3.tex
tskir: In an attempt to un-stuck this PR and keep to the bare minimum, I've extracted this as a separate issue: https://github.com/samtools/hts-specs/issues/548, and will be resolving this discussion

VCFv4.3.tex
tskir: Extracted as a separate issue: https://github.com/samtools/hts-specs/issues/548

VCFv4.3.tex
```diff
@@ -173,30 +173,49 @@ \subsubsection{Individual format field format}
...
-\begin{itemize}
-  \item DUP:TANDEM Tandem duplication
-  \item DEL:ME Deletion of mobile element relative to the reference
-  \item INS:ME Insertion of a mobile element relative to the reference
-\end{itemize}
+
+\begin{tabular}{l l}
+DEL  &  Deletion relative to the reference \\
+INS  &  Insertion relative to the reference \\
+DUP  &  Tandem duplication relative to the reference \\
```
tskir, line 187:
```suggestion
DUP  &  Duplication relative to the reference \\
```

As discussed, DUP is going to be just a generic “duplication” and the details will be provided in a separate SVCLAIM field #517.

VCFv4.3.tex
tskir, line 214:
```suggestion
\noindent Variants should be written using the most precise type that can be determined by the variant caller.\newline
```

Looks like these examples contradict the approaches discussed in #517, so I suggest we remove them in this PR and then add more up-to-date examples as applicable.

VCFv4.3.tex
```diff
@@ -564,23 +583,28 @@ \section{INFO keys used for structural variants}
...
 \begin{itemize}
-  \item DEL: Deletion relative to the reference
-  \item INS: Insertion of novel sequence relative to the reference
-  \item DUP: Region of elevated copy number relative to the reference
-  \item INV: Inversion of reference sequence
-  \item CNV: Copy number variable region (may be both deletion and duplication)
-  \item BND: Breakend
+\item DEL: Deletion relative to the reference
+\item INS: Insertion relative to the reference
+\item DUP: Tandem duplication relative to the reference
```
tskir, line 592:
```suggestion
\item DUP: Duplication relative to the reference
```

Again, looks like this no longer applies and superseded by #517.

tskir (changes requested):
@d-cameron Here's the fresh round of reviews, as promised. (Sorry it comes at the latest possible moment before the deadline!)

Once the remaining comments are addressed, I think we can go ahead and merge this, and then tweak as necessary to align with subsequent PRs (such as the SVCLAIM one etc.)

I've also resolved several remaining discussion and extracted them into separate issues to keep the scope of this PR reasonable.

VCFv4.3.tex
mbaudis: I miss that "duplication" refers to a "any quantitative increase of the number of alleles compared to the reference genome, without indication about the physical location of the additional copies". I.e. tandems can be a subset, as can be e.g. extrachromosomal amplifications.

tskir: Agreed. That's an excellent phrasing as well, I've added it to the suggested change now

d-cameron:
#553 can we just kill SVTYPE for v4.4? We have to redefine it as `TYPE=A` anyway so v4.4 is guaranteed to have a breaking change to SVTYPE.

As it currently stands, I think the burden of having SVTYPE outweights the value it adds to the specifications and it should be removed. It can be entirely inferred from ALT (implementation could even expose a read-only SVTYPE in their APIs for backwards compatibiilty), and it's not that much greater a burden on the quick and dirty parsing `grep SVTYPE=DEL | wc -l` style analyses.

tskir:
@d-cameron Personally, I lean mostly in favour of killing SVTYPE, the reason being precisely its complete reduncancy. Implementations exposing a read-only attribute sounds like a great idea as well.

However, I really think we should separate the changes to avoid the PRs blocking each other. Do you think that, for the time being, we could rebaseline and merge this PR as it is, and then get back to SVTYPE discussions in #553?

d-cameron:
Current issues:
- need to redefine SVLEN as type-specific in light of #519, otherwise we can't represent the length of an `<INV>` record when there are multiple ALT alleles.
- SVTYPE (moved to #553)
- Update DEL/DUP definitions to refer to SVCLAIM field
- Rebaseline again

There's not really much left in this PR except for clarifications of reserved SV symbolic ALT alleles, and some additional examples. Should I move these into their own PRs and address there?

tskir:
@d-cameron Regarding this:

> There's not really much left in this PR except for clarifications of reserved SV symbolic ALT alleles, and some additional examples. Should I move these into their own PRs and address there?

Yes, I think moving each of those changes into a separate PR would work best.

PR #467, opened 2020-01-21
Show double quotes around ##ALT Description

jmarshall:
Unlike for the other headers, the section on `##ALT` does not show the `Description` field as having double quotes around it.

However §1.4.2 (Information field format) says Description must be surrounded by double-quotes and this can be construed as applying to Descriptions in all headers, all the other headers show it, all the examples of `##ALT` in the specification have double quotes, and tools such as bcftools and GATK use double quotes in the `##ALT` fields they output. So this is presumably just an omission in the spec.

Hat tip @lltw; see pysam-developers/pysam#881.

### DIFF
--- a/VCFv4.1.tex
+++ b/VCFv4.1.tex
@@ -89,7 +89,7 @@ Possible Types for FORMAT fields are: Integer, Float, Character, and String (thi
 Symbolic alternate alleles for imprecise structural variants:
 
 \begin{verbatim}
-##ALT=<ID=type,Description=description>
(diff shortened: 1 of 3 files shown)


jmarshall:
Thanks. Which merge button did you use? (I see GitHub has helpfully completely screwed up the commit message.)

jmmut:
I used the default one "squash and merge" but it asked for the commit message, so I copied the title and description in the PR. I chose the PR message because it was newer and more complete, but I didn't notice about the line breaks.

PR #469, opened 2020-02-03
EVENT/EVENTTYPE for SV classification

d-cameron:
This PR adds an `EVENTTYPE` field allowing multiple SV records to be grouped together and collectively classified as an event.

This allows for explicit separation of SV breakpoint and CN detection, and genomic rearrangement event classification. The specs currently lack a distinction between detection and classification.

### DIFF
--- a/VCFv4.4.draft.tex
+++ b/VCFv4.4.draft.tex
@@ -684,7 +684,37 @@ $MEINFO$ consists of successive quadruplets of records for each ALT allele.
 ##INFO=<ID=DBRIPID,Number=A,Type=String,Description="ID of this element in DBRIP">
 ##INFO=<ID=MATEID,Number=A,Type=String,Description="ID of mate breakends">
 ##INFO=<ID=PARID,Number=A,Type=String,Description="ID of partner breakend">
-##INFO=<ID=EVENT,Number=A,Type=String,Description="ID of event associated to breakend">
+##INFO=<ID=EVENT,Number=1,Type=String,Description="ID of associated event">
+##INFO=<ID=EVENTTYPE,Number=1,Type=String,Description="Type of associated event">
+\end{verbatim}
+\normalsize
+
+Whilst simple events such as deletions and duplications can be wholly represented by a single VCF record, complex rearrangements such as chromothripsis result in a large number of breakpoints.
+VCF uses the $EVENT$ field to group such related records together, and $EVENTTYPE$ to classify these events.
+All records with the same $EVENT$ value are considered to be part of the same event.
+
+The following $EVENTTYPE$ values are reserved and should be used when appropriate:
+
+\begin{itemize}
+  \item DEL - Deletion
+  \item DEL:ME - Deletion of mobile element with respect to the reference
+  \item INS - Insertion
+  \item INS:ME - Insertion of mobile element
+  \item DUP - Duplication
+  \item DUP:TANDEM - Tandem duplication
+  \item DUP:DISPERSED - Dispersed duplication
+  \item INV - Inversion
+  \item TRA - Translocation
+  \item TRA:BALANCED - Balanced inter-chromosomal translocation
+  \item TRA:UNBALANCED - Unbalanced inter-chromosomal translocation
+  \item CHROMOTHRIPSIS - Chromothripsis
+  \item CHROMOPLEXY - Chromoplexy
+  \item BFB - breakage fusion bridge
+  \item DOUBLEMINUTE - Double minute
+\end{itemize}
+The sematics of other $EVENTTYPE$ values is implementation-defined.
+\footnotesize
+\begin{verbatim}
 ##INFO=<ID=CILEN,Number=.,Type=Integer,Description="Confidence interval for the SVLEN field">
 \end{verbatim}
 \normalsize


d-cameron:
Rebaselined to VCFv4.4

VCFv4.4.tex
```diff
@@ -611,7 +611,38 @@ \section{INFO keys used for structural variants}
 ##INFO=<ID=DBRIPID,Number=1,Type=String,Description="ID of this element in DBRIP">
 ##INFO=<ID=MATEID,Number=.,Type=String,Description="ID of mate breakends">
 ##INFO=<ID=PARID,Number=1,Type=String,Description="ID of partner breakend">
-##INFO=<ID=EVENT,Number=1,Type=String,Description="ID of event associated to breakend">
+##INFO=<ID=EVENT,Number=1,Type=String,Description="ID of associated event">
```
tskir, line 614: Could you please also rebase this against `VCFv4.4.draft.tex`?

VCFv4.4.tex
```diff
@@ -611,7 +611,38 @@ \section{INFO keys used for structural variants}
...
+  \item INS - Insertion
+  \item INS:ME - Insertion of mobile element
+  \item DUP - Duplication
+  \item DUP:TANDEM - Tandem duplication
+  \item INV - Inversion
+  \item TRA - Translocation
+  \item TRA:BALANCED - Balanced inter-chromosomal translocation
+  \item TRA:UNBALANCED - Unbalanced inter-chromosomal translocation
+  \item CHROMOTHRIPSIS
+  \item CHROMOPLEXY
```
tskir, line 637:
```suggestion
  \item CHROMOTHRIPSIS - Chromothripsis
  \item CHROMOPLEXY - Chromoplexy
```

Let's add definition for consistency, even if they are trivial (for now; I'd like to see this expanded in the future).

tskir (changes requested):
Cool, I like where this is going. After rebasing, we could merge this one straight away as well. However, now that we have terms which keep appearing in multiple contexts, I think it is vital that we synchronise them specification-wide. This is to be done in subsequent PRs — see #549.

cwhelan:
The gnomAD call set and GATK-SV pipeline identify and attempt to resolve a number of classes of complex SVs observed in germline genomes. These are detailed in figure 2 of the gnomAD paper (https://www.nature.com/articles/s41586-020-2287-8); I'll copy them below with the major class and possible subclasses we identify:

Paired-duplication inversion (dupINVdup)
Paired-deletion inversion (delINVdel)
Paired-deletion/duplication inversion (delINVdup, dupINVdel)
Deletion-flanked inversion (delINV, INVdel)
Insertion with insertion site deletion (dDUP-iDEL, INS-iDEL)
Duplication-flanked inversion (dupINV, INVdup)
Dispersed duplication (dDUP)

In the gnomAD callset, the median sample is genotyped as having 37 of these events. In the current release these events are represented by out-of-spec `CPX` events with custom INFO tag descriptors of the segments involved. I suggest considering adding these classes to the list of reserved EVENTTYPES, since they are part of a widely used reference data set and pipeline.

cwhelan:
Two other types to consider:

- Head-to-head and tail to tail inverted duplications.
- DUP-TRP/INV-DUP structures (eg https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3235474/)

I'm not sure where to draw the line at what should be a custom type and what should be reserved in the specification, though, so if you think these suggestions are too specific let me know. It might be nice to have an example in the spec of how a custom EVENTTYPE can be defined by tool developers who come up with their own new categories of events.

VCFv4.4.tex
```diff
@@ -611,7 +611,38 @@ \section{INFO keys used for structural variants}
 ##INFO=<ID=DBRIPID,Number=1,Type=String,Description="ID of this element in DBRIP">
 ##INFO=<ID=MATEID,Number=.,Type=String,Description="ID of mate breakends">
 ##INFO=<ID=PARID,Number=1,Type=String,Description="ID of partner breakend">
-##INFO=<ID=EVENT,Number=1,Type=String,Description="ID of event associated to breakend">
+##INFO=<ID=EVENT,Number=1,Type=String,Description="ID of associated event">
+##INFO=<ID=EVENTTYPE,Number=1,Type=String,Description="Type associated event">
```
cwhelan, line 615: Typo: this should be "Type of associated event"

cwhelan:
I would like to suggest that the text clarify or specify what types of records can be linked together using an EVENT/EVENTTYPE. The wording here talks about linking "a large number of breakpoints" and "related records". The examples in 5.4.6 and 5.4.7 show the use of EVENT to link together multiple BND records. Should it be legal to include non-BND SV records in an EVENT? 


Here's an example of where this might be useful. For a dispersed duplication, one possible representation would be a group of records that include BNDs showing the novel adjacencies, combined with a CNV record showing the duplicated sequence's elevated copy number (using the SVCLAIM=D suggested by https://github.com/samtools/hts-specs/pull/517).  For example, for an insertion of a copy of the segment chr2:5000-6000 at chr1:1000, assuming that SVCLAIM becomes part of the spec and we allow "dDUP" as an EVENTTYPE for dispersed duplications we might have (I think I've got these BNDs right but go easy on me if I messed it up):

```
chr1    1000    BND1    A     A[chr2:5000[    99    PASS    SVTYPE=BND;EVENT=1;EVTENTYPE=dDUP 
chr1    1001     BND2    C   ]chr2:6000]C    99    PASS    SVTYPE=BND;EVENT=1;EVTENTYPE=dDUP 
chr2    5000   BND3   G    ]chr1:5000]G   99   PASS SVTYPE=BND;EVENT=1;EVTENTYPE=dDUP 
chr2    5000   CNV   G    <CNV>   99   PASS SVTYPE=CNV;SVCLAIM=D;CN=2;END=6000;EVENT=1;EVTENTYPE=dDUP 
chr2    6000   BND3   T    T[chr1:1001[   99   PASS SVTYPE=BND;EVENT=1;EVTENTYPE=dDUP 
```

This would have the benefit of allowing someone using a browser or programmatically scanning for duplicated intervals for some analysis to find the CNV record via its overlap with a region or gene of interest and determine that chr2:5000-6000 is duplicated. Is this legal? Should it be? If not, I suggest that we make it explicit that only BND records are allowed to be linked by an EVENT (or possibly an EVENT with an EVENTTYPE in the closed set we are talking about here). I'm aware that you can also put a CN on the BND records themselves, but I've often found this to be difficult to interpret -- perhaps if that's the way to go then we could add some more clarification about how to interpret them in this sort of scenario or provide some examples.

d-cameron:
> Should it be legal to include non-BND SV records in an EVENT?

Yes. A `DEL SVCLAIM=J` record is making the same claim as a BND record so there's no reason that these can't be part of an event.

Conceptually, I'm treating a an event is the structural transformation from one genomic configuration to another so, in theory, all records pertaining to that transformation should share the same EVENT. This is somewhat problematic for CNV records as they are aggregate records. If you have CN=4, and a +1 CN from eventA and a +1 CN from eventB, what do we label the CNV record? Do we add both events? Do we split it out into two CN+1 records and give each their own EVENT?

I didn't want this complexity to be a barrier to usage of these fields so I've left it up to the implementation to choose what they want to do.

d-cameron:
On a similar note, @cwhelan do you know of any structural variant ontologies that would be useful to us? If there's something out there that actually handles most of our events then it'd make sense to incorporate it but I haven't found any that are relevant enough to be useful for us.

d-cameron:
@cwhelan do you see instances of INV with flanking duplications? You have instances of ABCD -> A^(CBA)^D with A being both duplicated and inserted, but do you see ABCD -> AA^(CB)^D where A is duplicated, but not actually inverted? Or do you just treat these as two separate but adjacent events since they're non-overlapping?

d-cameron:
Would it make sense to treat `dDUP-iDEL` and `INS-iDEL` as a cut&paste at the target site? These seem similar to mobile element translocation (copy/number & cut/paste respectively) with the change that there is a deletion at the insertion site.

Given that LINE insertions can have deletions at the insertion site.


More generally, what are the sizes of the flanking del/dups for your complex events? I frequently see associated insertions/deletions at both inversion and ME translocations. Perfectly clean breakpoints on both sides are the exception rather than the norm in my cancer analysis so I'm wondering whether it's appropriate to just treat these as subtypes of inversion and translocations. If the del/dup lengths are small (<50bp) then I'd say to treat them as subtypes but if they're large, then they're probably their own thing.

cwhelan:
> @cwhelan do you see instances of INV with flanking duplications? You have instances of ABCD -> A^(CBA)^D with A being both duplicated and inverted, but do you see ABCD -> AA^(CB)^D where A is duplicated, but not actually inverted? Or do you just treat these as two separate but adjacent events since they're non-overlapping so you don't treat them as a complex event inversion subtype?

We don't specifically look for these in our pipeline, since the breakpoints wouldn't overlap. If they exist we'd just emit them as separate adjacent events. I don't know that anyone has specifically looked for this type of event in our data, but the impression of people on our team is that this isn't a class of event we've seen, or at least seen often enough to be aware of it as a distinct type of germline variation and want to classify as a complex event automatically.

cwhelan:
> More generally, what are the sizes of the flanking del/dups for your complex events? I frequently see associated insertions/deletions at both inversion and ME translocations. Perfectly clean breakpoints on both sides are the exception rather than the norm in my cancer analysis so I'm wondering whether it's appropriate to just treat these as subtypes of inversion and translocations. If the del/dup lengths are small (<50bp) then I'd say to treat them as subtypes but if they're large, then they're probably their own thing.

The DUP/DEL intervals in the CPX events we're describing here are all, I believe, larger than 50bp, and some range into the hundreds of kb, so these go beyond the "microdeletion" and "microduplication" range. For example, in the gnomAD v2 call set's `dupINVdup` events, the smallest duplicated interval is 316bp and the largest is over 800kb. Our complex event resolution logic works by combining raw SV calls from a variety of tools (PE/SR callers and depth-based CNV-callers), and so these CPX events are the results of combining different INV, DEL, DUP, and CNV calls produced by those tools into single complex structures based on read-depth re-genotyping into plausible structures. We probably do miss small micro-deletions and duplications at, for example, simple INV events in our callset, but the focus of our complex event resolution code is on these larger structures (which I would think have distinct mutational processes behind them as compared to microdels and dups.)

cwhelan:
> A DEL SVCLAIM=J record is making the same claim as a BND record so there's no reason that these can't be part of an event.

Does this mean that a DEL, `SVCLAIM=D` record shouldn't be included in an event? As another example, our dupINVdup events (where ABC becomes A^(CBA)C) can be represented with only two breakpoint pairs. However, it would be nice to add DUP records, with `SVCLAIM=D` since they are not simple tandem duplications, to indicate that segments A and C have actually been duplicated as part of this event. This would go a long way towards making the records more useful for downstream analyses -- for example in a common disease association analysis where you are looking for genes that have a high burden of dosage changes across many samples in a multi-sample VCF. Including these segments would make that process a lot easier, since you wouldn't have to add intermediate steps to go back to re-infer the copy number of every segment based on the breakends.

d-cameron:
>Does this mean that a DEL, SVCLAIM=D record shouldn't be included in an event?

If definitely can, my point was that CN claims get messy when there is aneuploidy and multiple overlapping events. Not really an issue for germline analysis, but is if you're looking at chromothripsis+breakage-fusion-bridge or ecDNA (where you find DELs/DUPs on only some of the copies of a double minute).

For a dupINVdup, I'd do something like:
```
chr	101	did001	N	<dupINVdup>	END=400;EVENT=did001;EVENTTYPE=INV:DUP:DUP
chr	101	did001_1_bp1	N	[[chr:301[[N	MATEID=did001_1_bp2;EVENT=did001;EVENTTYPE=INV:DUP:DUP
chr	101	did001_2	N	<DUP>	END=200;SVCLAIM=D;EVENT=did001;EVENTTYPE=INV:DUP:DUP
chr	200	did001_3_bp1	N	N]]chr:400]]	MATEID=did001_3_bp2;EVENT=did001;EVENTTYPE=INV:DUP:DUP
chr	301	did001_1_bp2	N	[[chr:101[[N	MATEID=did001_1_bp1;EVENT=did001;EVENTTYPE=INV:DUP:DUP
chr	301	did001_4	N	<DUP>	END=400;SVCLAIM=D;EVENT=did001;EVENTTYPE=INV:DUP:DUP
chr	400	did001_3_bp2	N	N]]chr:200]]	MATEID=did001_3_bp1;EVENT=did001;EVENTTYPE=INV:DUP:DUP
```
We have the two inversion breakpoints, two CN duplications claims since the B and D segments are amplified.
The umbrella  `<dupINVdup>` record has undefined semantics from a spec perspective, but provides value as an implementation-specific term for GnomAD that allows for the high-level definition of the bounds of the event for downstream users that don't care about the specific details of exactly what the rearrangement looks like.

cwhelan:
Thanks for the example @d-cameron , this looks good! 

I found it a little hard to figure out from the text exactly what is allowed in terms of combinable events under a single EVENT -- especially since the examples of complex events in section 5.4 only use BND records. In particular, Section 5.4.6 talks about EVENT IDs like this, "As mentioned previously, a single rearrangement event can be described as a set of novel adjacencies," and uses only BND records in the example. 

So perhaps some wording could be added to this PR making it explicit that any records can be combined would help, ie "Records joined together under a single EVENT ID can have any combination of alternate allele types, including breakends and symbolic alternate alleles". 

It would also be good to update sections 5.4.6 and 5.4.7 to use these EVENTTYPEs in their examples (if they aren't being already worked on in a different PR that I'm missing).

tskir (dismissed):
Thank you @cwhelan and @d-cameron for very insightful discussions!

I have two thoughts about this PR:
1. The basic mechanics of `EVENT` / `EVENTTYPE` seem very sensible and useful, as they are currently described;
2. It does make sense to support some/all of the more complex types raised by @cwhelan, because, as mentioned, this could be useful for the downstream processing. However, deliberations on this could take a lot of time and expand the scope of this PR significantly.

Hence, I believe the best way to proceed with this PR is to merge it, essentially, as is to create the basic boilerplate for the EVENTTYPE; and then address the more complex types raised by @cwhelan in a separate follow-up PR(s).

@d-cameron Could you please rebaseline this PR to VCF 4.4 draft only and resolve the merge conflicts? Currently this includes identical changes to both 4.3 and 4.4 draft, but we certainly don't want to add this to 4.3 since this is quite a major change.

d-cameron:
Rebased, 4.4 only with basic set of `EVENTTYPE`s ready for merging.

lbergelson (approved):
Kiril's comments were addressed and this is ready to merge.

PR #472, opened 2020-02-19
Proposal: Clarify that the htsget specification text takes precedence over OpenAPI

jmarshall:
The specification text is the original document, while the OpenAPI description is a later simplified contribution from a third party. If there are discrepancies between the two, then it is the OpenAPI description that is incorrect and needs clarifying.

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -23,6 +23,7 @@ Explicitly this API does NOT:
 * Provide a way to discover the identifiers for valid ReadGroupSets --- clients obtain these via some out of band mechanism
 
 This protocol specification is accompanied by a [corresponding OpenAPI description](pub/htsget-openapi.yaml). OpenAPI is a language-independent way of describing REST services and is compatible with a number of [third party tools](http://openapi.tools/).
+(Note: if there are differences between this text and the OpenAPI description, this specification text is definitive.)
 
 # Protocol essentials
 


jeromekelleher:
Seems like a reasonable clarification to me.

jmtcsngr:
Wouldn't it be better to fix the OpenAPI implementation to match whatever is in the spec? We will have a issue if it is not possible to express in OpenAPI language what the spec has. Probably @brainstorm can help us evaluating if it is possible.

andrewyatz:
I think the point here is to accept that the two specs will drift and when that happens the right way to view this is the text is definitive. That doesn't mean don't fix the OpenAPI spec but it means unobserved discrepancies are resolved. In an ideal world there would be a way to create one from the other.

jmarshall:
Of course the intention would be to fix identified discrepancies in the OpenAPI description — hence “and needs clarifying” above — but the point is to tell readers which one takes precedence in the meantime. i.e. :+1: what @andrewyatz said.

This is a pretty standard disclaimer to have when you have two parallel texts that are intended to be equivalent. I meant to add such text when I merged #385, but I guess I forgot about it and just copied refget's OpenAPI paragraph. Refget might like to consider whether they need a similar note.

andrewyatz:
Agreed @jmarshall and #473 has been opened to add the same text

brainstorm:
Thanks @jmtcsngr for mentioning, this is the first time I see "later simplified contribution from a third party".... happy to tackle the gaps.

If something cannot be expressed let's discuss what it is. In general I prefer programmatic, precise and strict definitions than loose text prone to misinterpretation and unnecessary variability between implementations ;)

brainstorm:
Other than [this comment in igv.js](https://github.com/igvteam/igv.js/issues/1061#issuecomment-586364835) are there other discrepancies people in this thread and elsewhere noticed?

Shall I open another issue to collect them?

jeromekelleher:
> Shall I open another issue to collect them?

Seems like a good idea, thanks @brainstorm.

In the mean time, I don't think anyone has objected to having this clarification (some drift between two descriptions of the same spec is inevitable). Perhaps you could merge if you agree @mlin?

brainstorm:
Please redistribute to the appropriate channels: https://github.com/samtools/hts-specs/issues/475

jmarshall:
@mlin: Jerome and I think the added sentence is a no-brainer that should be merged immediately, but we won't do so without you approving it. Please take a look (and hit the approve button :smile:), as just having this PR open seems to be itself causing some misplaced angst and confusion…

jmarshall:
Merged on the basis that @mlin was supportive when this PR was initially raised during the [meeting on February 19th](https://docs.google.com/document/d/1SC9TAukP06A44rqGFTYllIJDyZM6ZATRxv0QFaAqvqk/edit#heading=h.i3nbyvrstrao).

PR #473, opened 2020-02-20
Refget spec text takes precedence over OpenAPI.

andrewyatz:
If there is divergence between refget's text and the OpenAPI, the text is definitive. This is a copy of #472

### DIFF
--- a/refget.md
+++ b/refget.md
@@ -34,7 +34,7 @@ Explicitly this API does NOT:
 
 ## OpenAPI Description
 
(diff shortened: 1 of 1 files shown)


jmarshall:
FWIW :+1: from me. After the conversation at last week's meeting, IMHO you should feel free to exercise your prerogative as refget maintainer to just merge this.

If in future the text and OpenAPI are generated from each other / a common source, or if the situation changes in some other way, the sentence can be removed again. In the meantime, it provides useful clarity.

andrewyatz:
Agreed and will be brought in

PR #476, opened 2020-03-03
Change BAM and BAI length/count fields to uint32_t (PR#460 alternative)

jmarshall:
Revisit PR #460 (see conversation there), adding notes in the existing _Value_ table columns indicating length/count fields whose maximum values are still (well) within the `int32_t` range convenient for e.g. Java arrays. This clarifies that, while the fields are naturally described as `uint32_t`, they are not required to be implemented using such a type if that is inconvenient in your implementation language.

OTOH this PR explicitly raises the `l_text` limit from 2GiB to 4GiB:

* HTSlib already handles 4GiB of BAM header text.

* HTSJDK represents SAM/BAM/etc headers as a collection of separate headers so no representation change is needed to deal with this `l_text` range increase. However _BAMFileReader/Writer.java_'s `readHeader()` and `writeHeader()`  currently operate on the full header text as a `String` so would require some rearranging internally to deal with files with such large headers.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -36,6 +36,8 @@
 \newcommand*{\cclass}[1]{{\rm\sf :#1:}}
 \newcommand*{\caret}{\textsuperscript{$\wedge$}}
 
+\newcommand*{\memlimited}{\textcolor{gray}{\footnotesize\it limited}}
+
 \begin{document}
 
 \input{SAMv1.ver}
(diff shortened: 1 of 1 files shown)


jkbonfield:
I'm not convinced by "_memory_", although I'm struggling to think of a better term.  Memory seems weak, as I may (will!) have well over 4Gb spare to play with.  Perhaps _limited_?

I also don't like having some fields with their maximum sizes being displayed in the Value column, which for now is for fixed values "BAM\1" or default values "[-1]", given we have many other fields where the limits are written in the description.  e.g. "Reference sequence ID, −1 ≤ refID < n ref" or "−1 ≤ next refID < n ref".  We ought to be consistent here.

I think it's also actually useful to be able to state for example that `0 <= n_ref < 2^{31}` as it emphasises that a BAM file with no references is still valid.

Why is `block_size` 31-bit, or at least _memory_ marked?  Does Java also load it into a string?  it would seem an unobvious choice given the binary nature of the data.  Agreed it's highly unlikely to be large.  Is that the message here, that a typical file won't trigger priblems?  if so neither memory nor limited are the correct term to use, but it's not what should be in a spec either.  Maybe I don't understand all of Java's woes here.  Is it simply unable to have any block of memory more than 2Gb in total, and thus cannot even read a large block of data without splitting into multiple memory allocations?

If I had to choose, I think for brevity sake and that an extra column may be unwieldy, I'd put the two 2^{31} limits into the description field.  Note this is also formatted poorly as the exponent is overflowing the cell bounding slightly (although the same was already happening in footnotes).

jmarshall:
> Why is `block_size` 31-bit, or at least _memory_ marked?

It's memory marked because you're gonna load it into memory. The thinking was: headers are 1-set-per-file so mostly anything goes, but for read records once you've got a hundred or more in memory at once, that's starting to be a lot of memory. However most things do mostly just stream one at a time…

You're quite right that the memory constraint is the minor consideration; the main thing for these field sizes is simply the reality of the data represented. I'll carry on thinking about how to express that (and improve this overall, as discussed), even if it's just “Note A” in the table with the explanation in the note.

jmarshall:
Changed _memory_ to _limited_, and removed the lifting of `l_text` to 4GB.

jkbonfield:
Thanks for the update.  I like it and am happy for this to go in as is.

jmarshall:
@yfarjoun? I'd like to merge this this week if possible.

yfarjoun:
sorry for the holdup :+1: from me.

jmarshall:
Thanks all; merged now (and _SAMv1.pdf_ rebuilt with this and `@RG-PL:DNBSEQ`).

I have the 4 GiB `l_text` proposed change stashed separately and might raise it as its own PR one day — but there's no hurry :smile:

PR #479, opened 2020-03-04
Refget version 2.0.0 bump

andrewyatz:
This is a placeholder PR request for the version 2.0.0 bump in refget and a place to accumulate discussion around the changes. To note the major changes are:

- Integration of service-info (change of schema)
- Preferring the ga4gh identifier proposed by VRS over TRUNC512 (backwards compatible)
- Moving code examples out into notebooks and scripts
- Version bump to 2.0.0 due to the change of schema brought on by service-info


jmarshall:
Apropos the service-info part of this (which replaces the older service-info response with the more-detailed JSON in [service-info as ratified by GA4GH](https://github.com/ga4gh-discovery/ga4gh-service-info)): TASC maintains a [list of the [service-info] type values for GA4GH APIs](https://github.com/ga4gh/TASC/tree/master/service-info) and there is at present a PR ga4gh/TASC#9 to add more entries to the list.

FYI your TASC representative's position is that refget does not currently support (standard) service-info so it is premature to add refget to the list at present. _(Please :+1: or :-1: this comment so we can gauge the refget working group's support for this position.)_

When this PR has been merged and refget then does support standard service-info, we should then raise a TASC PR to add refget to their list (as described in their process for adding to the list).

pub/refget-openapi.yaml
```diff
@@ -195,33 +190,34 @@ components:
...
+                description: Maximum length of sequence which may be requested using start and/or end query parameters
+              algorithms:
+                type: array
+                items:
+                  type: string
+                  description: An array of strings listing the digest algorithms that are supported
+                  example:
+                  - md5
+                  - ga4gh
+                  - trunc512
```
tcezard, line 223: Maybe we should remove `trunc512` from the example since it's deprecated

refget.md
```diff
@@ -103,21 +103,24 @@ Authorization: Bearer [access_token]
 The policies and processes used to perform user authentication and authorization, and the means through which access tokens are issued, are beyond the scope of this API specification. GA4GH recommends the use of the OAuth 2.0 framework ([RFC 6749](https://tools.ietf.org/html/rfc6749)) for authentication and authorization.
 
 ## Checksum calculation
-The supported checksum algorithms are `MD5` (a 32 character HEX string) and a SHA-512 based system called `TRUNC512` (a 48 character HEX string, see later for details). Servers MUST support sequence retrieval by one or more of these algorithms, and are encouraged to support all to maximize interoperability. To provide CRAM Reference Registry compatibility an implementation must support MD5.
+The supported checksum algorithms are `MD5` (a 32 character HEX string) and a SHA-512 based system called `ga4gh` (a base64 URL-safe string, see later for details). Servers MUST support sequence retrieval by one or more of these algorithms, and are encouraged to support all to maximize interoperability. An older algorithm called `TRUNC512` existed in version 1.0.0 of refget but is now deprecated in favour of the GA4GH sequence checksum string.
```
tcezard, line 106:
```suggestion
The recommended checksum algorithms are `MD5` (a 32 character HEX string) and a SHA-512 based system called `ga4gh` (a base64 URL-safe string, see later for details). Servers MUST support sequence retrieval by one or more of these algorithms, and are encouraged to support all to maximize interoperability. An older algorithm called `TRUNC512` existed in version 1.0.0 of refget but is now deprecated in favour of the GA4GH sequence checksum string.
```

refget.md
```diff
@@ -368,16 +378,40 @@ Array of strings listing the versions of this API supported by the server.
 GET /sequence/service-info
```
tcezard, line 379: Shouldn't we take the opportunity of a major version change to move the service-info at the top level ?
I think this is what's implied in the [explanation](https://github.com/ga4gh-discovery/ga4gh-service-info#how-do-i-describe-a-service-implementing-multiple-specifications) of how multiple services should be described.

refget.md
```diff
@@ -568,30 +491,29 @@ Key to generating reproducible checksums is the normalisation algorithm applied
...
 
 ### Checksum Choice
 
 MD5 provides adequate protection against hash collisions occurring from sequences. However the consequence of a sequence derived hash collision appearing would be catastrophic as two or more sequences with different content would report to be the same entity.
 
-The VMC, Variant Modelling Collaboration, is a complementary GA4GH effort to model genomic variation based on deviations from a reference sequence. Part of their work was to explore hashing algorithms. We have defined a new checksum algorithm based on [previous work and analysis from VMC](https://docs.google.com/document/d/12E8WbQlvfZWk5NrxwLytmympPby6vsv60RxCeD5wc1E/edit#heading=h.2nwv6g36f686), based around the SHA-512 algorithm.
+The VRS, Variation Representation Specification, is a complementary GA4GH effort to model genomic variation based on deviations from a reference sequence. Part of their work was to explore hashing algorithms. We have adopted the [checksum algorithm from VRS v1](https://vr-spec.readthedocs.io/), based around the SHA-512 algorithm.
```
tcezard, line 503: If we are pointing to VRS v1.0 I suggest we change the url to https://vrs.ga4gh.org/en/1.0/ or https://vrs.ga4gh.org/en/1.0/impl-guide/computed_identifiers.html#truncated-digest-sha512t24u if we want to be very precise.
```suggestion
The VRS, Variation Representation Specification, is a complementary GA4GH effort to model genomic variation based on deviations from a reference sequence. Part of their work was to explore hashing algorithms. We have adopted the [checksum algorithm from VRS v1](https://vrs.ga4gh.org/en/1.0/impl-guide/computed_identifiers.html#truncated-digest-sha512t24u), based around the SHA-512 algorithm.
```

tcezard:
Also, We should consider collecting multiple PRs for version 2.0.0. So maybe this one should be merged to a `release_2.0` branch.

pub/refget-openapi.yaml
```diff
@@ -166,8 +154,15 @@ components:
               type: string
               description: >-
                 Truncated, to 48 characters, SHA-512 checksum of the reference
-                sequence encoded as a HEX string
+                sequence encoded as a HEX string. No longer a preferred serialisation of the SHA-512
               example: 959cb1883fc1ca9ae1394ceb475a356ead1ecceff5824ae7
+            ga4gh:
+              type: string
+              description: >-
+                A ga4gh identifier used to identify the sequence. This is a SHA-512 base64 url encoded
```
jmarshall, line 162: To be easily understood, this should be referred to as “base64url-encoded” with a link to RFC 4648 §5.

```suggestion
                A ga4gh identifier used to identify the sequence. This is a base64url-encoded
```

refget.md
```diff
@@ -368,16 +378,40 @@ Array of strings listing the versions of this API supported by the server.
...
+  "refget": {
+    "circular_supported": true,
+    "subsequence_limit": 0,
+    "algorithms": [
+      [
+        "md5",
+        "ga4gh",
+        "trunc512"
+      ]
+    ]
```
tcezard, line 408:
```suggestion
    "algorithms": [
      "md5",
      "ga4gh",
      "trunc512"
    ]
```

pub/refget-openapi.yaml
andrewyatz: Removed

refget.md
andrewyatz: Done in commit 12cbba1ceaed8dd71307a9ae71e397e0ce2d690e

refget.md
andrewyatz: Done

pub/refget-openapi.yaml
andrewyatz: Done

refget.md
andrewyatz: Done

andrewyatz:
All problematic commits have been resolved and the bad PR has now gone. Source tree indicates latest discussions.

andrewyatz:
New question to be raised. The naming authorities work includes a number of capitalised names e.g. `INSDC`. CURIEs normally use lowercase e.g. `insdc` or `ensembl`. I believe we should reflect this change to lowercasing and also mention the use of identifiers.org as a source of good prefixes (but obviously not exhaustive since any prefix can be generated).

andrewyatz:
Please ignore the c888522 commit on here. It was a mis-push as I tried to double check everything was working correctly with the branch and I was up to date in what I had to do. Apologies

pub/refget-openapi.yaml
```diff
@@ -159,7 +159,8 @@ components:
             ga4gh:
               type: string
               description: >-
-                A ga4gh identifier used to identify the sequence. This is a base64url-encoded encoded
+                A ga4gh identifier used to identify the sequence. This is a SHA-512 
+                [base64 url encoded](https://datatracker.ietf.org/doc/html/rfc4648#section-5)
```
jmarshall, line 163: Thanks for the link, but the text was mostly better prior. “SHA-512” already appears later in the sentence; the name of the encoding defined in RFC4648 §5 is `base64url` (no space — this is unrelated to url-encoding, i.e. percent encoding, so best not to risk confusion with that).

pub/refget-openapi.yaml
```diff
@@ -159,7 +159,8 @@ components:
             ga4gh:
               type: string
               description: >-
-                A ga4gh identifier used to identify the sequence. This is a base64url-encoded encoded
+                A ga4gh identifier used to identify the sequence. This is a SHA-512 
+                [base64 url encoded](https://datatracker.ietf.org/doc/html/rfc4648#section-5) 
                 string taking the 1st 24bits from the SHA-512 digest (sha512t24u). This is the preferred method
```
jmarshall, line 164: The first 24 **bytes**.

It's also not clear here  that `sha512t24u` refers to the entire SHA-512 (`sha512`) truncated to 24 bytes (`t24`) base64url-encoded (`u`) pipeline.

refget.md
```diff
@@ -514,7 +514,7 @@ MD5 provides adequate protection against hash collisions occurring from sequence
...
 
-The algorithm performs a SHA-512 digest of a sequence and creates a base64url encoding of the first 24 bytes of the digest. Analysis performed by VRS suggests this should be sufficient to avoid message collisions (see the VRS documentation for more details). Should a message collision occur within this scheme then the number of bytes retained from the SHA-512 checksum will be increased.
+The algorithm performs a SHA-512 digest of a sequence and creates a base64url encoding ([using RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648#section-5)) of the first 24 bytes of the digest. Analysis performed by VRS suggests this should be sufficient to avoid message collisions (see the VRS documentation for more details). Should a message collision occur within this scheme then the number of bytes retained from the SHA-512 checksum will be increased.
```
jmarshall, line 517:
```suggestion
The algorithm performs a SHA-512 digest of a sequence and creates a base64url encoding ([as defined in RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648#section-5)) of the first 24 bytes of the digest. Analysis performed by VRS suggests this should be sufficient to avoid message collisions (see the VRS documentation for more details). Should a message collision occur within this scheme then the number of bytes retained from the SHA-512 checksum will be increased.
```

pub/refget-openapi.yaml
andrewyatz: Good point. Changing this back

pub/refget-openapi.yaml
andrewyatz: Eeep. Yes bytes. Okay

jmarshall: Slightly astonished neither I nor anyone else had noticed that before! :smile:

jmarshall:
From https://github.com/samtools/hts-specs/pull/479#discussion_r786941363:

> Shouldn't we take the opportunity of a major version change to move the service-info [to] the top level? I think this is what's implied in the [explanation](https://github.com/ga4gh-discovery/ga4gh-service-info#how-do-i-describe-a-service-implementing-multiple-specifications) of how multiple services should be described.

My reading of that explanation is that it specifies the opposite and the service-info endpoint should remain at `/sequence/service-info`.

In this case `foo` is `sequence`, the root is at `https://server.example.com/sequence` and hence to my mind it is saying that service-info should be at `https://server.example.com/sequence/service-info`.

So IMHO this needs to be revisited and PR's 12cbba1ceaed8dd71307a9ae71e397e0ce2d690e needs to be considered again.

andrewyatz:
I believe this was our original interpretation but at least my interpretation is that a service would be bound by a sub-namespace and since `/sequence` is a specified part of the API wouldn't this mean an implementation more like `https://server.example.com/refget/service-info`?

jmarshall:
That makes sense as an explanation, but I have not seen any expectation or hint in the spec that the endpoint might be or ought to be under `…/refget/…`.

I've always read the “configurable HTTP(S) endpoint” stuff as meaning the main endpoint might be somewhere like `https://server.example.com/pub/andysgroup/projectx/sequence/<id>`, if that made sense for Example Inc's webspace organisation.[^1]

Certainly that's the approach that htsget has taken, with `…/reads/service-info` and `…/variants/service-info`.

[^1]: And I think other parts of GA4GH have been surprised that refget and htsget have this idea of configurable endpoints as appropriate for your organisation at all. Seems to me that they've expected that when we write `GET /sequence/<id>` or `GET /reads/<id>` or etc then that really is exactly the entire path part after the hostname… 🤷

andrewyatz:
I mean on your sub-point I would be horrified if this was my only server on a domain :grinning:. Do you think it's best to reach out to discovery about this since this is the big change that's forcing the v2 bump

andrewyatz:
That's true re: change in format of service-info. That would still be a breaking change. Let's see if we can invoke discovery into this. @jmarshall please see if you think I've got the gist of the conversation right.

@mcupak, @mpanik, @mbaudis as co-developers of service info we'd like a bit more clarification over the intended position of `service-info` in a service. We want to get this right once and for all, and part of this PR involves a move. It boils down to options:

1. `/sequence/service-info`. Refget's current solution where `service-info` is a child 
2. `/service-info`. The v2.0.0 proposal where `service-info` is a peer of `sequence`

The refget currently treats `/sequence` as a namespace for the API. We can see positives & negatives to this system as it means we now have APIs which respond to URLs like `/sequence/{id}` and they need to know `/sequence/service-info` is a special endpoint. Versus not treading on the toes of the URL structure if they were peers of each other.

Some guidance would be really appreciated here.

mcupak:
As Michael said, we see service-info implemented at the top level of APIs.

To elaborate a bit more, since this API has a single endpoint, I think it's your decision how you want `/sequence` to be treated:
- If you see adding other top level endpoints besides `/sequence` to the API in the future, I'd use top-level `/service-info`.
I think it's generally better for APIs not to prescribe their own prefix so that the implementors can choose where to root them in their system and avoid awkward patterns, this option is cleaner since it doesn't conflict with IDs, and keeps the options open for more operations/endpoints in the future.
- If `/sequence` is meant to be a prescribed prefix for every implementation by design, as suggested above, and there will never be other endpoints, then `/sequence` is effectively the root and `/sequence/service-info` makes sense.
I can see how you might want this, e.g. if the use case is for this API to be implemented together with other APIs also defining their prefixes, such as htsget, and you know this will only ever involve ID lookups. It's definitely unusual though and has disadvantages.

mbaudis:
+1 for @mcupak 's detailed explanation.

jmarshall:
@mcupak: This is precisely the scenario that I asked about nearly three years ago.

Htsget serves reads (BAM/CRAM files) and variants (VCF files), both of which are used in conjunction with an associated reference genome. Refget serves reference genomes. IMHO it would be very natural for a research group to serve all three — sequencing data, variants, and the associated references — from the same host, especially if their organisation prefers not to have many distinct publicly-accessible hostnames.

Particularly if this is their “server host”, separate from their `www.example.com` text-based web site, then these three categories of data will likely be served from the default top-level URLs:

* `https://datahost.example.com/reads/<id>`
* `https://datahost.example.com/variants/<id>`
*  `https://datahost.example.com/sequence/<id>`

In this case, there will likely be separate htsget (backing the `/reads/…` and `/variants/…` endpoints) and refget (backing the various `/sequence/…` endpoints) server programs running, with some Apache/nginx/etc server program in front dispatching requests by URL to the two other programs as appropriate.

What is your recommendation for the URL(s) that clients should make requests to to get service-info data?

If your recommendation is `https://datahost.example.com/service-info` :— Is it possible for a single JSON-formatted service-info response to represent information about more than one service?

mcupak:
@jmarshall What makes the most sense to me is:
- `https://datahost.example.com/htsget/reads/<id>`
- `https://datahost.example.com/htsget/variants/<id>`
- `https://datahost.example.com/htsget/service-info`
- `https://datahost.example.com/refget/sequence/<id>`
- `https://datahost.example.com/refget/service-info`

Where `htsget` and `refget` parts are not prescribed by the specs and implementers can choose different prefixes.

The unusual situation here is that we have 1 bundle/repo containing 2 GA4GH APIs containing 3 capabilities. Service-info is intentionally tied to APIs. If these 3 capabilities are so similar and often implemented together, I naively wonder if it makes sense for them to be parts of the same API (with optional endpoints).

jmarshall:
1. If you care to read the [htsget specification](https://samtools.github.io/hts-specs/htsget.html#ga4gh-service-info), you will see that the service-info JSON contains reads-/variants-specific information, so separate `…/reads/service-info` and `…/variants/service-info` endpoints are specified.

2. Requiring a `…/htsget/…` or `…/refget/…` component in the path is news to me, and is not mentioned in any of these specifications as far as I can see. Are you suggesting that the trio of `https://datahost.example.com/{reads,variants,sequence}/<id>` URLs in my previous comment, which exactly reproduce the paths shown in the specs and are the natural simple forms that most people would choose in the absence of other constraints, should be forbidden?

3. This is an example. I would think that users might want to make any number of GA4GH services available on the same hostname, particularly in the (not uncommon) case where their organisation wishes to limit the number of distinct publicly-accessible hostnames or setting up a publicly-accessible server host is otherwise administratively non-trivial.

    In short, I do not see this situation as unusual. It would be unfortunate if service-info was so inflexible as to not support it.

jmarshall:
> Neither option is forbidden or not compliant from service-info perspective. You should go with whatever you think is best for your API […] It's definitely common to have multiple services on the same host or generally base path. In that case, I'd expect different services to have different service-info endpoints

Great, we are in agreement! In that case I'll repeat the question. If some third-party users deploy htsget and refget, and set up their endpoint URLs as follows (as they think it is best for their purposes, and which you say is compliant from a service-info perspective), what is Discovery's recommendation for the URLs to be used as the service-info endpoints?

* `https://datahost.example.com/reads/<id>`
* `https://datahost.example.com/variants/<id>`
*  `https://datahost.example.com/sequence/<id>`

andrewyatz:
Okay so after reading all of this and discussing in the call today we think the right thing to do is move it back to `/sequence/service-info` as this does not necessitate namespacing the entire API and maintains the URL structure of refget v1. The group that met felt what's been written here should move into more central docs in the service-info spec. There were concerns of diversifying APIs by having specs that approach the problem in different ways but also felt we should do what's right for refget.

mcupak:
Thanks, makes sense to me. I filed this https://github.com/ga4gh-discovery/ga4gh-service-info/issues/72.

refget.md
juliangehring, line 1: Can the `md5` return value in `metadata` also be null (same as with `ga4gh`)? According to the specs above, a server can support either checksum and leaves the option open that `md5` is not support ("Servers MUST support sequence retrieval by one or more of these algorithms [...]").

andrewyatz: Good point checking the spec we have inconsistency between `md5` and `ga4gh`. Makes sense [the md5 line changes](https://github.com/andrewyatz/hts-specs/blob/feature/2.0.0/refget.md?plain=1#L255) to

```
md5, if the server does not support md5 the value will be <code>null</code>.
```

Thoughts from other refget people? Also will post this in the refget Slack channel for visibility

tcezard: Thanks for bringing that up. we should indeed treat `md5` and `ga4gh` the same way and I actually think we should explicitly allow the field to be missing when the algorithm is not supported.

juliangehring, line 1: If a sub-sequence is requested, the response must only contain the specified sub-sequence.

Is this a must since the server can decide to ignore the Range header and return the full sequence?

If a sub-sequence is requested via a Range header, the response must be one of 206 and only contain the specified sub-sequence, be 200 and contain the entire sequence (thus ignoring the Range header) [...]

juliangehring: I just spotted this bit in the int "Alternative Checksum Algorithms" section:
> Refget implementations MUST support the MD5 identifier space and SHOULD support the ga4gh identifier.

That would indicate that md5 can't be null after all.

juliangehring, line 1: > subsequence_limit: An integer giving the maximum length of sequence which may be requested using start and/or end query parameters.

From a discussion with @andrewyatz, the `subsequence_limit` seems to apply to both query and header. I might be good to also mention the Range header here.

andrewyatz: Ah yes. I think this is us pushing to make sure each instance supports CRAM by default since CRAM doesn't sniff a refget by default

andrewyatz: Good point. It's more if the server supports ranges then it must support it but we accepted that some instances just don't do this right.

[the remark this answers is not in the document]
andrewyatz: You're fine to redirect all types of request. This was trying to define what the redirection code we wanted to use was

andrewyatz: Yes I believe it did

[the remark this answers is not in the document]
andrewyatz: This is stated in the protocol essentials

> HTTP response may include a 3XX response code and Location header redirecting the client to retrieve sequence data from an alternate location as specified by [RFC 7231](https://tools.ietf.org/html/rfc7231), clients SHOULD be configured follow redirects. 302, 303 and 307 are all valid response codes to use.

andrewyatz:
To note this is going up to steering committee on Monday. If successful we will merge this branch in and 2.0 will be released.

refget.md
jkbonfield: It's something that could be changed in a future CRAM release, but they've been proving hard to get out the door!

tcezard: I tried to clarify this in https://github.com/andrewyatz/hts-specs/pull/9/commits/3b0ba58b86a8047c2bbd6701ba71bc6ec57cac70 but I have not changed the requirement for have `md5` specified.

tcezard: Thank you for the suggestion, I've clarified this point in https://github.com/andrewyatz/hts-specs/pull/9/commits/459683bcf142e527b416d461aafaeb7041296e69

Issue #260, opened 2017-11-02
GP now in linear scale?


eyherabh:
Contrary to VCF4.1 and VCF4.2, which specify GP in phred scale, VCF4.3 now specifies it in linear scale. Could you explain me the rationale behind this change and why wouldn't be better just to let anyone use any of the scales (phred, linear or log10) provided is indicated in the metadata? Thank you very much.

lbergelson:
@eyherabh This seems to have changed because it was inconsistent with the labelling of other fields like GL/PL.  This now matches the scaling of GL, and the phred scaled version of this is PP.  PP was never added as a reserved tag, but it was adopted by GATK (I don't know if other tools emit it). We should probably add it as a reserved tag.

PR #480, opened 2020-03-04
Adding PP tag to vcf 4.3 spec.

lbergelson:
* The PP tag is to GP as PL is to GL
* It seems like we should have added PP when we changed the definition of GP to float.
* Reworded slightly because a local reviewer read the term 'precisely as' as a comment on the precision of representation and not as a description of how the field was defined.

* This came up as part of trying to understand https://github.com/samtools/hts-specs/issues/260.  GATK switched from using GP to PP at that time and it seems like an oversight to not add the tag to the spec.


lbergelson:
@jmmut Good catch, that was a dumb mistake.  I fixed it now I think, and added a line to the changelog.

PR #484, opened 2020-03-23
Clarify the existing rules for the `@HD` header

jmarshall:
Clarify the current rules: having an `@HD` line in your SAM headers is optional, but when present it must come first.

I might have just pushed this to **master**, but presenting this as a PR to give @jkbonfield and @yfarjoun a chance to vet the description of `@HD` as _File-level metadata_ (rather than the current meaninglessly-generic _The header line_!).

Okay to merge? I'd hope we can turn this one around in a day :smile:

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -242,7 +242,7 @@ meanings can be avoided.}
   \endfirsthead
   \cline{1-3}
   \endhead
-  \multicolumn{2}{|l}{\tt @HD} & The header line. The first line if present. \\\cline{2-3}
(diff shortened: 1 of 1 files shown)


yfarjoun:
I think that this PR is fine as a clarification, though I'd love to discuss making it required if a header is present...(at least from some version and on)

jmarshall:
Yep, although the ambiguity is pretty tenuous given that the introduction says “Each file may optionally specify the version being used via the `@HD VN` tag” and the Recommended Practice section recommends not omitting it. Anyway, it's good to have it locally clearer.

@yfarjoun: So rename your issue and we can continue having that discussion there.

yfarjoun:
👍

PR #487, opened 2020-04-02
Fix Typo in VCF spec

lbergelson:
* fix a typo in the vcf 4.3/4.4 spec

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -95,7 +95,7 @@ Characters with special meaning (such as field delimiters `\verb|;|' in INFO or 
 
 \subsection{Data types}
 Data types supported by VCF are: Integer (32-bit, signed), Float (32-bit IEEE-754, formatted to match one of the regular expressions \verb|^[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?$| or \verb"^[-+]?(INF|INFINITY|NAN)$" case insensitively),%
-\footnote{Note Java's {\tt Double.valueOf} is particular about capitalisation, so additional code is needed to parse all VCF infinite/NaN values.}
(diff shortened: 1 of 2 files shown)


jmarshall:
That's not a typo, it's a trans-Atlantic difference.

jmmut:
is there any policy on whether using british or american spellings? https://en.wikipedia.org/wiki/Capitalization
As non-native I have to put special attention not to mix them.

jmarshall:
If there was a policy, it would be in [MAINTAINERS.md](https://github.com/samtools/hts-specs/blob/master/MAINTAINERS.md) or would at least be generally known amongst the group of maintainers.

The policy I've been basically following is consistency with nearby text for particularly significant words (so `color` for SAM's CM/CQ/CS tags) but otherwise (thus for incidental words that, like this one, are just part of the prose) authors should feel free to use their local natural spellings and refrain from over-correcting other authors' locally-correct spellings.

Other policies would require one side of the Atlantic to kowtow to the other side (both as readers and authors) and require the authors on one side to write in an unfamiliar dialect. For our little specs, IMHO the consistency gained would be immaterial. Think of it as adding local colour to otherwise dull prose… :smile:

lbergelson:
@jmarshall Huh, I didn't know... I'll have to teach it to my auto-spell check...

jmarshall:
That is the flaw in my argument: that policy does somewhat require authors on both sides to have semi-encyclopaedic knowledge of what the differences are :smile:

I was curious which side of the divide Canada tended towards, and unsurprisingly the answer appears to be “it depends”. [This is an interesting list](http://www.lukemastin.com/testing/spelling/cgi-bin/database.cgi?action=rules) though _citation needed_ perhaps on his ‘although the “-ize” form is actually considered to be more correct even in Britain nowadays’ claim.

lbergelson:
Yes, in this case the offense was due to ignorance rather than any sense of American spelling superiority.

PR #492, opened 2020-05-04
Clarify percent encoding

### LINKED ISSUE #252: Potentially ambiguous interpretation of character encoding
I find the section 1.2 of VCFv4.3.pdf ambiguous about the encoding of characters with special meaning. Specifically, that section requires that characters with special meaning be encoded using capitalized percentage encoding, but neither seems to say whether that encoding is required only within what fields, nor whether such encoding is restricted to fields or can be used interchangeably with UTF-8 encoding when the special characters are used in their special meaning.
That is, could you please clarify me whether the standard does actually prevent, for example, two columns be separated by "%09"  (three characters) as opposed to '\t' (one character)?
Thank you very much.
Best regards..

jmmut:
This is the best wording I could think of, and still not great. closes #252.

On a more general topic (regarding the PR migration to 4.4), the next commands are simple and worked for me to copy the changes in the last commit from VCFv4.3.tex to VCFv4.4.tex:
```
git diff HEAD^ -- VCFv4.3.tex > my_patch
patch -p1 VCFv4.4.tex  my_patch
```

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -77,7 +77,7 @@ UTF-8 is a multi-byte character encoding that is a strict superset of 7-bit ASCI
 As a result, most software that processes VCF files does not have to be aware of the possible presence of multi-byte UTF-8 characters.
 Note that non-printable characters U+0000--U+0008, U+000B--U+000C, U+000E--U+001F are disallowed.
 Line separators must be CR+LF or LF and they are allowed only as line separators at end of line.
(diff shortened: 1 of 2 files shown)


jmmut:
agree. my second candidate is `literally`, which is technically better, but I'm not sure it will be understood better.

lbergelson:
maybe just drop the extra qualifier and make it "Some characters have a special meaning when they appear."

PR #495, opened 2020-05-14
fix htsget OpenAPI repetition

jb-adams:
@mlin @jmarshall @brainstorm @daviesrob 

Partial solution to fixing OpenAPI repetition. ( issue #489 )

1. Responses
Added a generic `Error` Schema, and an Error schema for each error code outlined in the htsget specification text. These error 'subclasses' inherit from the base `Error` schema through the `allOf` property. These errors are then grouped in `#/components/responses` according to their error code, which are referenced in each API route.

Important to Note: The master branch OpenAPI spec currently has a **misalignment** with the spec markdown, as the OpenAPI error code responses do not prescribe a JSON response with `error` and `message` properties nested under `htsget`. This PR aims to correct this by defining the Error object.

2. Parameters
`class`, `referenceName`, `start`, `end` were moved to `#/components/parameters` and referenced by each endpoint. The description for `id` was made generic and also moved to components. `format` is kept as 2 distinct parameters, as they have different enum constraints for reads or variants.

I was unsure what to do with `fields`, `tags`, and `notags`. What do `tags` refer to in a reads context, and in a variants context? In a BAM/CRAM context, I thought they referred to the two-letter SAMtags the client wanted (ie. column 12 and after), but the examples do not reflect this. Would appreciate clarification on this.

3. Schemas
I attempted to model inheritance and polymorphism of certain schemas by relegating common properties to a parent schema, and having child schemas 'inherit' parent properties via the `allOf` property. This is seen in the parent `htsgetResponse`, and its children, `htsgetResponseReads` and `htsgetResponseVariants`. Polymorphism was needed because the allowed `format` values are different for reads and variants.
Similarly, there is `htsgetUrl`, `htsgetUrlReads` and `htsgetUrlVariants`. The child schemas contain different example values, which is important for the example objects in Swagger UI.

### DIFF
--- a/pub/htsget-openapi.yaml
+++ b/pub/htsget-openapi.yaml
@@ -36,7 +36,7 @@ paths:
         - $ref: '#/components/parameters/endParam'
         - in: query
           name: fields
-          description: A comma separated list of SAM fields to include. By default, i.e., when fields is not specified, all fields will be included
+          description: A comma-separated list of SAM fields to include. By default, i.e., when fields is not specified, all fields will be included
           example: QNAME,RNAME
           required: false
           schema:
@@ -55,18 +55,19 @@ paths:
             type: string
         - in: query
           name: tags
-          description: A comma separated list of tags to include. By default, i.e., when tags is not specified, all tags will be included
+          description: A comma-separated list of tags to include. By default, i.e., when tags is not specified, all tags will be included
           example: MD,NM
           required: false
           schema:
             type: string
         - in: query
           name: notags
-          description: A comma separated list of tags to exclude. By default, i.e., when notags is not specified, no tags will be excluded
+          description: A comma-separated list of tags to exclude. By default, i.e., when notags is not specified, no tags will be excluded
           example: OQ
           required: false
           schema:
             type: string
+
       responses:
         200:
           description: results matching criteria
@@ -100,9 +101,11 @@ paths:
           $ref: '#/components/responses/500InternalServerErrorResponse'
         default:
           $ref: '#/components/responses/500InternalServerErrorResponse'
+
       security:
         - htsget_auth:
           - read:genomic_reads
+
   /variants/{id}:
     get:
       summary: Gets the variants from a pre-indexed id
@@ -125,18 +128,17 @@ paths:
         - $ref: '#/components/parameters/endParam'
         - in: query
           name: tags
-          description: A comma separated list of tags to include, by default all.
-          example: cancer,melanoma
+          description: A comma-separated list of tags to include, by default all.
           required: false
           schema:
             type: string
         - in: query
           name: notags
-          description: A comma separated list of tags to exclude, default none.
-          example: normal,skin
+          description: A comma-separated list of tags to exclude, default none.
           required: false
           schema:
             type: string
+
       responses:
         200:
           description: results matching criteria
@@ -170,9 +172,11 @@ paths:
           $ref: '#/components/responses/500InternalServerErrorResponse'
         default:
           $ref: '#/components/responses/500InternalServerErrorResponse'
+
       security:
         - htsget_auth:
           - read:genomic_variants
+
 components:
   schemas:
     htsgetResponse:
@@ -193,6 +197,7 @@ components:
               type: string
               description: MD5 digest of the blob resulting from concatenating all of the payload data
               example: 8a6049fad4943ff4c6de5c22df97d001
+
     htsgetUrl:
       type: object
       required:
@@ -209,6 +214,7 @@ components:
           description: Indicates whether the header or body of the requested file will be retrieved by this url
           example: body
           enum: [header, body]
+
     htsgetUrlReads:
       allOf:
         - $ref: '#/components/schemas/htsgetUrl'
@@ -218,11 +224,13 @@ components:
               headers:
                 Authorization: Bearer xxxxx
                 Range: bytes=65536-1003750
+
     htsgetUrlVariants:
       allOf:
         - $ref: '#/components/schemas/htsgetUrl'
         - example:
             - url: https://example.com/sample1234/run1.vcf
+
     Error:
       type: object
       description: Generic error response object
@@ -326,6 +334,7 @@ components:
                   example: InternalServerError
                 message:
                   example: An internal server error occurred
+
   parameters:
     idPathParam:
       in: path
@@ -371,6 +380,7 @@ components:
         type: integer
         format: int64
         minimum: 0
+
   securitySchemes:
     htsget_auth:
       type: oauth2
@@ -381,6 +391,7 @@ components:
           scopes:
             read:genomic_reads: read access to genomic reads
             read:genomic_variants: read access to genomic variants
+
   responses:
     400BadRequestResponse:
       description: Bad request
--- a/pub/htsget-openapi.yaml
+++ b/pub/htsget-openapi.yaml
@@ -29,7 +29,7 @@ paths:
         - in: "query"
           name: fields
           description: A list of fields to include, such as QNAME, FLAG, RNAME, etc...
-          example: "fields=QNAME,RNAME"
+          example: "QNAME,RNAME"
           required: false
           schema:
             enum:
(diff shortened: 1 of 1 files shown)


brainstorm:
Thanks @jb-adams for this refactor! I agree with all your changes and corrections over my first-pass OA3 spec.

The examples for fields/tags/notags should be corrected and adapted, I think that your intuition is right w.r.t SAMtags and the examples on the prose/free-text spec can also be amended accordingly.

Great job!

daviesrob:
As requested, I've added a fix for the broken `fields`, `tags` and `notags` examples for the `/reads/{id}` endpoint here.  I haven't done anything about `/variants/{id}` mainly as I'm not entirely sure what `tags` and `notags` refer to in that context.

mlin:
Thanks @daviesrob! Agree the meaning of `tags` is ~undefined for variants currently. Perhaps we should just remove it from the variants endpoint description for the time being, otoh, it's not *too* harmful to leave in.

shaikh-rashid:
+1

pub/htsget-openapi.yaml
```diff
@@ -309,16 +178,184 @@ components:
...
+        enum: [VCF, BCF]
+    classParam:
+      in: "query"
+      name: class
+      description: Request different classes of data (such as header or body).
+      example: "header"
+      required: false
+      schema:
+        type: string
+        enum: [header, body]
```
jmarshall, line 330: The only defined value for this field is `header`.

```suggestion
        enum: [header]
```

jmarshall:
+1 to the general idea of reducing repetition, but note that it's also useful to keep the YAML text readable for those editing it or choosing to view it raw.

So I think it would be better to keep the unshared items — readsFormatParam, variantsFormatParam, 200OkReadsResponse, htsgetResponseReads, 200OkVariantsResponse, htsgetResponseVariants — inline rather than abstracting them as $refs, just as `fields`, `tags`, etc have been left inline.

Now that you're spelling out the allowed values in an `enum`, the description doesn't need to repeat that information: so something like this for the two format parameters:

```yaml
- in: "query"
  name: format
  description: Desired file format
  example: BAM
  required: false
  schema:
    type: string
    enum: [BAM, CRAM]
    default: BAM
```

pub/htsget-openapi.yaml
```diff
@@ -20,55 +20,16 @@ paths:
...
-        - in: "query"
-          name: end
-          description: The end position of the range on the reference, 0-based exclusive.
-          example: "99999"
-          required: false
-          schema:
-            type: integer
-            format: int64
-            minimum: 0
+        - '$ref': '#/components/parameters/idParam'
```
jmarshall, line 23: It would be good to indicate in the ref name that this is a different kind of query from the others (it's the only `in: path` one), e.g.,

```suggestion
        - '$ref': '#/components/parameters/idPathParam'
```

daviesrob:
This generally looks much better than the original, and seems to fix a few problems.

There may be an issue with the 400 error response.  Looking [at the branch on ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/jb-adams/hts-specs/htsget/fix-openapi-repetition/pub/htsget-openapi.yaml#operation/searchReadId) it's showing the JSON for the 404 error and not any of the 400 options.  I think that it's being picked up from the [`/components/schemas/Error`](https://github.com/samtools/hts-specs/blob/a168e7ac/pub/htsget-openapi.yaml#L199) section.  I don't know if this is an actual problem with our source, or the viewer not handling it correctly.

jb-adams:
Pushed another commit for this branch

Edits requested from @jmarshall 

> The only defined value for this field is `header`.

This is fixed for `classParam`

> So I think it would be better to keep the unshared items — readsFormatParam, variantsFormatParam, 200OkReadsResponse, htsgetResponseReads, 200OkVariantsResponse, htsgetResponseVariants — inline rather than abstracting them as $refs, just as `fields`, `tags`, etc have been left inline.

Removed these definitions from `#/components` in favor of inline definitions

> Now that you're spelling out the allowed values in an `enum`, the description doesn't need to repeat that information: so something like this for the two format parameters:
> 
> ```yaml
> - in: "query"
>   name: format
>   description: Desired file format
>   example: BAM
>   required: false
>   schema:
>     type: string
>     enum: [BAM, CRAM]
>     default: BAM
> ```

This format has been used to describe acceptable `format` param values separately for `reads` and `variants` 

> It would be good to indicate in the ref name that this is a different kind of query from the others (it's the only `in: path` one), e.g.,

`idParam` has been renamed to `idPathParam`

Edits requested from @daviesrob 

> There may be an issue with the 400 error response. Looking [at the branch on ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/jb-adams/hts-specs/htsget/fix-openapi-repetition/pub/htsget-openapi.yaml#operation/searchReadId) it's showing the JSON for the 404 error and not any of the 400 options. I think that it's being picked up from the [`/components/schemas/Error`](https://github.com/samtools/hts-specs/blob/a168e7ac/pub/htsget-openapi.yaml#L199) section. I don't know if this is an actual problem with our source, or the viewer not handling it correctly.

This has been corrected to render correctly with ReDoc. `BadRequest` is a special case in that it uses `oneOf` to refer to one of 3 options: `UnsupportedFormat`, `InvalidInput`, `InvalidRange`. ReDoc was not pulling example values from any of these but instead pulling from the base `Error` object. To correct, I've removed all example values from the base object, only providing examples in the "subclasses". The 3 schemas get picked up by ReDoc and can be flipped through via dropdown.

I also evaluated how the specification gets rendered via ReDoc and added some additional descriptions where they weren't popping up before. I believe I got them all. This also required some minor repetition of the `urls` array property across `reads` and `variants`, as having this property be defined in the base `htsgetResponse` class and subclasses led to examples not being rendered. 

Lastly, I removed many extraneous quotations (', ") for keys and values, as they aren't required for the YAML

jmarshall:
> Lastly, I removed many extraneous quotations (', ") for keys and values, as they aren't required for the YAML

Mixing in changes like this (in the very same commit!) makes reviewing the substantive changes much harder. In future please make trivial changes like this in at least a separate commit or ideally an entirely separate “minor tidy up” PR.

mlin:
Thanks @jb-adams @jmarshall @daviesrob!

@brainstorm what do you think of the further improvements here? (Appealing to you as you have the most context for the original!)

brainstorm:
Thanks @mlin,

I've been following the thread and I think that all suggestions have great attention to detail, the resulting spec is much more compact, I like it! :)

The raw yaml spec reads much better and the improved examples, messages and security definitions make the overall UX/DX pretty good (`npm install -g @stoplight/prism-cli`):

Screen Shot 2020-09-01 at 12 36 46 pm

I'd say merge and let developers work with it 👍🏻

brainstorm:
Is there anything else pending with this one? Shall we merge it?

jmarshall:
This is a substantial improvement on the status quo. IMHO okay to merge once at least the first two of the following are addressed:

* Hyphenate “comma-separated” (5 instances).

* As noted on https://github.com/samtools/hts-specs/pull/495#issuecomment-654998555 it's not entirely clear what `tags` and `notags` are intended to refer to for the variants endpoint. What is clear is that it is some combination of `INFO` and genotype keys and perhaps others (`FILTER`?), but not arbitrary values like `melanoma` or `skin`. As you're already touching these lines, it would be best to just remove these two `example:` lines for the time being.

* What is the reason for removing the blank lines between major sections? Does this make a difference for any of the tools parsing this file? Removing them certainly negatively impacts the readability of the raw text.

jb-adams:
From @jmarshall :

> * Hyphenate “comma-separated” (5 instances).

hyphens have been added

> * As noted on [#495 (comment)](https://github.com/samtools/hts-specs/pull/495#issuecomment-654998555) it's not entirely clear what `tags` and `notags` are intended to refer to for the variants endpoint. What is clear is that it is some combination of `INFO` and genotype **keys** and perhaps others (`FILTER`?), but not arbitrary **values** like `melanoma` or `skin`. As you're already touching these lines, it would be best to just remove these two `example:` lines for the time being.

removed these two `example` lines

> * What is the reason for removing the blank lines between major sections? Does this make a difference for any of the tools parsing this file? Removing them certainly negatively impacts the readability of the raw text.

re-added spacing lines between sections as they appear in the `master` branch. Also added spacing lines between some new sections created in this PR

PR #496, opened 2020-05-18
missing FORMAT lists can be . or .,.,. (closes #229)

### LINKED ISSUE #229: VCF: Should missing values for Number=2 be `.` or `.,.`?
When the cardinality of a INFO or FORMAT field is greater than 1, should a missing be represented with just 1 dot? or several dots separated by commas to match the Number? This is the first example on the spec:
```
20 14370 rs6054257 G A 29 PASS NS=3;DP=14;AF=0.5;DB;H2 GT:GQ:DP:HQ 0|0:48:1:51,51 1|0:48:8:51,51 1/1:43:5:.,.
```
which suggests that there should be several dots by commas, but there's no explicit mention to it, and it was introduced before 4.1, and it has never been updated.

If this is the case, I suggest changing the explanation of missing values at "... For example if the FORMAT is GT:GQ:DP:HQ then `0/0:.:23:23,34`" to be on a field with cardinality > 1. So it could be:

> For example if the FORMAT is GT:GQ:HQ:DP then `0/0:42:.,.:23` indicates that HQ is missing

Because otherwise, external tools might put single dots. For example, I just tested `bcftools merge` and it merged these vcfs:
```
##fileformat=VCFv4.3
##FORMAT=<ID=HQ,Number=2,Type=Integer,Description="HQ">
##FORMAT=<ID=GT,Number=1,Type=String,Description="GT">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	HG97
1	100	rs180734498	C	T	100	PASS	AC=1	GT	0|0
```
```##fileformat=VCFv4.3
##FORMAT=<ID=HQ,Number=2,Type=Integer,Description="HQ">
##FORMAT=<ID=GT,Number=1,Type=String,Description="GT">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	HG00096
1	100	rs180734498	C	T	100	PASS	AC=1	GT:HQ	0|0:3,6
```
into this one:
```
##fileformat=VCFv4.3
##FILTER=<ID=PASS,Description="All filters passed">
##FORMAT=<ID=HQ,Number=2,Type=Integer,Description="HQ">
##FORMAT=<ID=GT,Number=1,Type=String,Description="GT">
##contig=<ID=1>
##bcftools_mergeVersion=1.5+htslib-1.5
##bcftools_mergeCommand=merge -O v -o /tmp/comma-missing.merged.vcf /tmp/comma-missing.vcf.gz /tmp/comma-missing.2.vcf.gz; Date=Thu Jul 20 10:13:30 2017
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	HG00096	HG97
1	100	rs180734498	C	T	100	PASS	.	GT:HQ	0|0:3,6	0|0:.
```
As you can see, the last sample has a `.` for HQ, which is Number=2. I don't see a big advantage in either option of putting commas or not, but the spec should be clear on that, IMO.

jmmut:
This change is my interpretation of @pd3 comment at https://github.com/samtools/hts-specs/issues/229#issuecomment-316908769

I did this change only for FORMAT because if an INFO field is missing, you can just not write it. In FORMAT you are forced to write it unless it's the last field.

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -397,6 +397,7 @@ Additional Genotype keys can be defined in the meta-information, however, softwa
 
 If any of the fields is missing, it is replaced with the MISSING value.
 For example if the FORMAT is GT:GQ:DP:HQ then $0\mid0:.:23:23,34$ indicates that GQ is missing.
+If a field contains a list of missing values, it can be represented either as a single MISSING value (`.') or as a list of missing values (e.g. `.,.,.' if the field was Number=3).
 Trailing fields can be dropped, with the exception of the GT field, which should always be present if specified in the FORMAT field.
 
 
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -397,7 +397,7 @@ Additional Genotype keys can be defined in the meta-information, however, softwa
 
 If any of the fields is missing, it is replaced with the MISSING value.
 For example if the FORMAT is GT:GQ:DP:HQ then $0\mid0:.:23:23,34$ indicates that GQ is missing.
-If a field contains a list of missing values, it can be represented either as a single MISSING value (`.') or as a list of missing values (e.g. `.,.,.' if the field was Number=3).
+If a field contains a list of missing values, it can be represented either as a single MISSING value (`.') or as a list of missing values (e.g.\ `.,.,.' if the field was Number=3).
 Trailing fields can be dropped, with the exception of the GT field, which should always be present if specified in the FORMAT field.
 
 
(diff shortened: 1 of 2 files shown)


VCFv4.3.tex
```diff
@@ -397,6 +397,7 @@ \subsubsection{Genotype fields}
 
 If any of the fields is missing, it is replaced with the MISSING value.
 For example if the FORMAT is GT:GQ:DP:HQ then $0\mid0:.:23:23,34$ indicates that GQ is missing.
+If a field contains a list of missing values, it can be represented either as a single MISSING value (`.') or as a list of missing values (e.g. `.,.,.' if the field was Number=3).
```
jmarshall, line 400:
```suggestion
If a field contains a list of missing values, it can be represented either as a single MISSING value (`.') or as a list of missing values (e.g.\ `.,.,.' if the field was Number=3).
```

jmmut: good catch

PR #499, opened 2020-05-25
Fix the description of BAI metadata pseudo-bin file offsets

### LINKED ISSUE #498: Inconsistency between SAM spec and htslib .bai index generation
I'm not sure if this is a misunderstanding on my part, but there seems to be a mismatch between the description of pseudo-bins at 37450 in the spec https://samtools.github.io/hts-specs/SAMv1.pdf and what is written out by htslib.

The spec states in section 5.2 that unmapped beg/end should contain the virtual file offset of the start/end of *placed unmapped* reads.  It doesn't specify exactly what "unmapped according to its FLAG" means, but I inferred that it means FLAG bit 0x04 (segment unmapped) is not set.

In hts_idx_push and hts_idx_finish, it seems that the META_BIN is populated with information covering *all* reads, not just placed unmapped ones.  The effect is that the pseudo-bin at 37450 is always describing the virtual file offsets for all reads placed against the reference.

jmarshall:
Ever since the metadata pseudo-bins were added to the spec in #2, their virtual file offset entries have been misdescribed as locating the **placed unmapped reads** (which are not contiguous but spread throughout the reads on the chromosome), contrary to how they are computed by both HTSlib and HTSJDK and to Heng Li's original [description on the samtools-devel mailing list](https://sourceforge.net/p/samtools/mailman/message/25690499/):

> […] each chr has a bin 37450 which currently has two records […]
> The first record keeps the start and end file offset of the entire chr (chunk_beg[0]=chr_start_off, chunk_end[0]=chr_end_off).

Fixes #498 — see additional background there.

Also mention in _CSIv1.tex_ that metadata pseudo-bins appear in CSI too, with the same contents as BAI, and generalise their bin number calculation from BAI's 37450. Clarify CSI “depth”, which is more _max_level_ than _number-of-levels_.

### DIFF
--- a/CSIv1.tex
+++ b/CSIv1.tex
@@ -36,6 +36,14 @@
 \end{tabular}}
 \end{table}
 
+\noindent
+The following functions generalise those given in the SAM specification for a BAI-style binning scheme.
+Note that in CSI \textit{depth} refers to the scheme's maximal depth, i.e., the level number of the scheme's smallest bins, and the single-bin level spanning the entire coordinate range is level 0.
+Hence the BAI-style binning scheme, with six levels in total, is represented by $\mbox{\sf min\_shift} = 14, \mbox{\sf depth} = 5$.
+
+CSI index files may contain metadata pseudo-bins for each reference sequence, with the same contents as BAI pseudo-bins.
+In CSI, the pseudo-bins have bin number $\mbox{\sf bin\_limit}(\mbox{\sf min\_shift}, \mbox{\sf depth}) + 1$.
+
 {\footnotesize
 \begin{verbatim}
 /* calculate bin given an alignment covering [beg,end) (zero-based, half-close-half-open) */
@@ -46,6 +54,7 @@ int reg2bin(int64_t beg, int64_t end, int min_shift, int depth)
         if (beg>>s == end>>s) return t + (beg>>s);
     return 0;
 }
+
 /* calculate the list of bins that may overlap with region [beg,end) (zero-based) */
 int reg2bins(int64_t beg, int64_t end, int min_shift, int depth, int *bins)
 {
@@ -55,6 +64,12 @@ int reg2bins(int64_t beg, int64_t end, int min_shift, int depth, int *bins)
         for (i = b; i <= e; ++i) bins[n++] = i;
     }
     return n;
+}
+
+/* calculate maximum bin number -- valid bin numbers range within [0,bin_limit) */
+int bin_limit(int min_shift, int depth)
+{
+    return ((1 << (depth+1)*3) - 1) / 7;
 }
 \end{verbatim}
 }
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -1307,7 +1307,7 @@ saved.
 \footnotetext{The number of chunks in a single bin is effectively limited by available memory and in any case is typically a maximum of some thousands.}
 
 The index file may optionally contain additional metadata providing a summary
-of the number of mapped and unmapped read-segments per reference sequence,
+of the number of mapped and placed unmapped read-segments per reference sequence,
 and of any unplaced unmapped read-segments.\footnote{By \emph{placed unmapped
 read} we mean a read that is unmapped according to its {\sf FLAG} but whose
 {\sf RNAME} and~{\sf POS} fields are filled in, thus ``placing'' it on a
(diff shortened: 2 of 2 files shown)


jmarshall:
Also if you do the arithmetic about how many bins there are on each level (1 + 8 + 64 + 512 + 4096 + 32768), you confirm that the last BAI bin is numbered 37448 (when the first bin is numbered 0). So the ‘claim’ in the SAM spec is easily confirmed to be accurate :smile:

Leaving 37449 unused is a historical accident, as [noted here on samtools-help](https://sourceforge.net/p/samtools/mailman/message/36157993/). It wasn't ‘we’ who apparently accidentally skipped one — this was back in the days when one person was implementing things in samtools without review or discussion.

(It's pseudo-bin 37450, first chunk and second chunk. There's no bin 37451 either. Also I suppose 37449 not being valid means the recently added `n_bin` constraint (SAMv1 §5.2) could be one tighter, but it's only an upper bound and one tighter would be more confusing…)

jkbonfield:
Ah sorry, yes it's <37451 mentioned.  I was just scanning and failed to read properly. :)

I realised of course this "we" was before our time.  I just meant we as in the group of people who have ever been maintainers of the spec.

Anyway, happy with the changes.

yfarjoun:
what about CSIv2.pdf? does it also need fixing? also, why _are_ there two versions?

jmarshall:
CSIv2 is [marked as draft](https://samtools.github.io/hts-specs/) and has never been implemented. See also https://github.com/samtools/hts-specs/pull/321#issuecomment-394808164. It should really just be removed until the ideas get revisited.

jmarshall:
Rebased, and version history entry changed to June.

I'd like to merge this before the next meeting. It's just waiting for an okay from @lbergelson — transferred from Yossi at the last meeting on the basis that you know more about indexes and are better placed to confirm that the new description matches HTSJDK behaviour (implemented I think in _BAMIndexMetaData.java_'s `recordMetaData()`).

CSIv1.tex
```diff
@@ -36,6 +36,14 @@
 \end{tabular}}
 \end{table}
 
+\noindent
+The following functions generalise those given in the SAM specification for a BAI-style binning scheme.
+Note that in CSI \textit{depth} refers to the scheme's maximal depth, i.e., the level number of the scheme's smallest bins, and the single-bin level spanning the entire coordinate range is level 0.
+Hence the BAI-style binning scheme, with six levels in total, is represented by $\mbox{\sf min\_shift} = 14, \mbox{\sf depth} = 5$.
```
lbergelson, line 42: This is extremely helpful to include here.  Thank you.

SAMv1.tex
```diff
@@ -1328,13 +1328,17 @@ \subsection{The BAI index format for BAM files}
   \hline
   {\sf bin} & Magic bin number & {\tt uint32\_t} & 37450 \\\hline
   {\sf n\_chunk} & \# chunks & {\tt uint32\_t} & 2 \\\hline
-  {\sf unmapped\_beg} & (Virtual) file offset of the start of placed unmapped reads & {\tt uint64\_t} & \\\hline
-  {\sf unmapped\_end} & (Virtual) file offset of the end of placed unmapped reads & {\tt uint64\_t} & \\\hline
+  {\sf ref\_beg} & (Virtual) file offset of the start of reads placed on this reference& {\tt uint64\_t} & \\\hline
+  {\sf ref\_end} & (Virtual) file offset of the end of reads placed on this reference& {\tt uint64\_t} & \\\hline
   {\sf n\_mapped} & Number of mapped read-segments for this reference & {\tt uint64\_t} & \\\hline
   {\sf n\_unmapped} & Number of unmapped read-segments for this reference & {\tt uint64\_t} & \\\hline
```
lbergelson, line 1334: Should this be clarified as "Number of placed unmapped read-segments for this reference"?

lbergelson: It's also a bit weird that we say "reads" in some of these lines and "read-segments" on others when we mean the same thing.

jmarshall:
Rebased, and version history entry changed to July.

Thanks for the thoughtful review, Louis. The placed/unplaced terminology is explained in footnote 36 on the same page as this pseudobin table — does this suffice? You are quite right in your comments about `n_unmapped`: it would be good to rework that description to include ‘placed’, and what all these fields are probably strictly referring to is _alignment records_ rather than reads/segments (consider when a read has two mappings to the same chromosome and appears in two alignment records). And it's currently intentionally vague about exactly where the `ref_end` file offset points to (it's the byte after that last alignment record).

I plan to merge this as is and address those editorial concerns in a followup.

lbergelson:
@jmarshall You're right.  I looked for something like that but somehow missed it even though it was right there. The footnote seems fine.

Issue #177, opened 2016-11-18
Value of NM tag when ambiguous bases are involved


yfarjoun:
The definition of the NM tag is 
  > Edit distance to the reference, including ambiguous bases but excluding clipping.

The Tags file doesn't edify much more. It find it unclear, in particular, what does "including ambiguous codes" mean? 

I can understand it in two different ways:
A:
> "taking into account that an ambiguous code is always an edit distance away from anything" 

B: 
>"if a read base matches the pattern of the reference the edit distance is 0, else 1"

Here's a table of examples with the two possible interpretations:
 
\#  | Read | Ref | NM (A) | NM (B) | comments
----|----|----|----|----|---
1| A | A | 0| 0| clearly..
2| A| a | 0| 0| since bases as case-insensitive
3| A |G |1 | 1| clearly
4| A | g| 1 | 1 | base-insensitive
5| A| N | 1 | 0|
6| N| N | 1 | 0|
7 | N |A | 1| 1 | Since the read is less specific than the reference? not sure
8 | A| M |1 | 0 | M is A or C
9 | M| A | 1 | 1 | again, not sure
10 | M|R | 1 | 1| R is A or G...not sure here
10|M |M| 1 |0|

A discussion would be great, and when we reach an understanding, I'll be happy to submit a PR for the spec as well.

colinhercus:
Hi Yossi,


Thanks for bringing this up. It would be great to have a clear definition.

Can you also consider case where read has a base with quality 2 Perr = 0.63
which is pretty much the equivalent to an N.

Kind Regards, Colin

jkbonfield:
I totally agree with the need for a clear definition.

My personal preference is simple exact match only (case insensitive).  (Arguably N in seq vs N in reference is a corner case, as we know the base *is* incorrect, just unknown in both cases.  Samtools already has special code dealing with N.)  Otherwise it gets too complex.  What if we have R (AG) in SAM seq and V (ACG) in ref? Is that a match?  N technically matches everything, so is N in reference always a match to seq?  The list of cases is long, so best simply avoided.

I tried comparing samtools vs picard here, but alas couldn't get SetNmAndUqTags working.  Is there another way for it ito compute NM?   Samtools does appear to treat any textual mismatch (or an N N match) as a mismatch.  Conversely R in seq and R in ref is considered a match.  It also treats T vs U as a mismatch.

yfarjoun:
I lean towards exact matches for non-ambiguous bases: R in both read and
[...]

PR #500, opened 2020-05-26
Add MD string definition clarifying how MD is constructed

### LINKED ISSUE #485: Clarify MD tag description
@jmarshall and I were having a discussion on pysam about the MD tag: https://github.com/pysam-developers/pysam/issues/895

The MD regex is clear that we need zeros between adjacent mismatches, or after an insertion.  This makes the MD tag alone able to describe the reference sequence.  Nonetheless, for SAM, we most often have a cigar, so the extra zeros I do not believe are needed in this case. For example, `10A5^AC6` is underspecified if a CIGAR is not given, since a cigar `10=1X5=1D1X6=` or `10=1X5=2D6=` could both "work" with the MD tag.  The former should be `10A5^A0C6` (with the `0` between the deletion and next mismatch), and the latter should be `10A5^AC6`.  With a CIGAR (even using `M` instead of `=` or `X`) it is clear. 

What do folks think about MD tags that do not have the zero padding?

jmarshall:
Fixes #485 — see that issue for much background discussion.

The behaviour described for the more exotic CIGAR operators is derived from that of the `samtools calmd` command and (equivalently) the MD computation in HTSlib's _cram_encode.c_.

This documents the status quo, which requires a `0` between otherwise adjacent mismatch and/or deleted bases.

There would be something to be said for relaxing this and I toyed with adding a footnote along the lines of

> It would be possible to consult the CIGAR string to glean the length of deletions and not need zeros. Reader may wish to do this to additionally accept abbreviated MD strings; writers should output the zeros.

However as discussed on the associated issue, rather than doing that it might be better to revisit the whole question and invent a new more easily parsed tag field (cf minimap2's `cs` tag, though that is too underspecified to be adopted as is).

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -87,7 +87,7 @@ or
   {\tt IH} & i & Query hit total count \\
   {\tt LB} & Z & Library \\
   {\tt MC} & Z & CIGAR string for mate/next segment \\
-  {\tt MD} & Z & String for mismatching positions \\
+  {\tt MD} & Z & String encoding mismatched and deleted reference bases \\
   {\tt MF} & ? & Reserved for backwards compatibility reasons \\
   {\tt MI} & Z & Molecular identifier; a string that uniquely identifies the molecule from which the record was derived \\
   {\tt MQ} & i & Mapping quality of the mate/next segment \\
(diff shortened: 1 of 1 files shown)


SAMtags.tex
```diff
@@ -182,15 +182,29 @@ \subsection{Additional Template and Mapping data}
...
+This can enable variant calling without requiring access to the entire original reference.
+
+The {\tt MD} string consists of the following items, concatenated without additional delimiter characters:
+\begin{itemize}
+\item \verb"[0-9]+", indicating a run of reference bases that are identical to the corresponding {\sf SEQ} bases;
+\item \verb"[A-Z]", identifying a single reference base that differs from the {\sf SEQ} base aligned at that position;
+\item \verb"\^[A-Z]+", identifying a run of reference bases that have been deleted in the alignment.
+\end{itemize}
+
+As shown in the regular expression above, numbers alternate with the other items.
```
jkbonfield, line 196: I was initially writing a long comment on how the three separate regular expressions don't show this, and then noticed the un-edited part of the text with the full regexp.

Maybe edit this to "As shown in the full regular expression above".

jmarshall: Good point. I'll make it “complete regular expression above” or something. Initially I had the bullet points as “Numbers …” / “Single bases …” / “`^`+string of bases …” rather than regexes so didn't notice this problem.

yfarjoun:
This seems fine to me, but I feel woefully ill-fitted to say since I have never used MD...

nh13:
I think this footnote is wholly appropriate and in fact many folks output such MDs (although non-compliant): 

> It would be possible to consult the CIGAR string to glean the length of deletions and not need zeros. Reader may wish to do this to additionally accept abbreviated MD strings; writers should output the zeros.

I think it would be great to include and not lose this knowledge. I agree a better tag would be nice and a great follow up discussion, but I don’t why it precludes this very minor footnote that is just a comment on something folks already do. I’ll vote for including to but defer to the maintainers.

jmarshall:
Yossi: Thanks. Nor had I until the pysam conversation with Nils :smile:

Nils: Personally I'm inclined to agree, perhaps more in the style of a “known incompatibilities” footnote. However there's some _citation needed_ here: do you have examples of programs outputting non-compliant MD strings like this?

nh13:
I’ll try to find time in the next few days to re-examine my sources. Myexamination of bwa and htsjdk shows they output zeros. 

I think my original use case is generating synthetic alignment records, where I hand build the MD string. I personally find for small unit tests synthetic records are easier to maintain than storing lots of static SAM files. Perhaps a solution is to make generating the MD automated from a read and reference versus hand crafting the MD. Regardless, as someone who has contributed, albeit very sparingly, to each of the major SAM libraries, I think this footnote is extremely helpful to understand the MD and it’s intent.

jmarshall:
Not sure exactly what we agreed for this one at the meeting, but it has approval from all SAM editors — so merged as it adds a definition of the status quo to the hopelessly incomplete example that was previously all there was.

Further conversation on whether/how to relax the `0` thing on #485 (which can be reopened if needed) and on whether/in what way to dis-recommend using `MD` at all on #505.

jmarshall:
BTW this description is still incomplete — it glibly says “run of reference bases that are identical to the corresponding SEQ bases” without describing what **identical** means. Cf NM's text enumerating considerations re ambiguity codes in SEQ/REF and `=` in SEQ.

jkbonfield:
Indeed.  From recollection htslib treats N vs N as a mismatch, on the grounds that N isn't a real base so it is always incorrect.  There was a whole discussion on NM here: #177 #308.  However that's NM and not MD.

As I recall we also discussed whether the two should explicitly be tied together, so +1 to NM means +1 element to MD, or whether MD should purely be able generating the reference from the sequence and so can permit a more relaxed approach.  I *think* we settled on the latter, but I can't find that statement anywhere.

Regardless the tables of combinations may be useful to refer to.

PR #503, opened 2020-05-28
Removing the confusing and incomplete CSIv2 pdf

lbergelson:
The CSIv2 spec is incomplete and causes continual confusion.  We should remove the pdf until someone actually resurrects this project.


jmarshall:
Also removed the link in _index.md_ and pointed at the **CSIv2** branch in the comment message (even though branch names in git are ephemeral :smile:).

I think the consensus at today's meeting was that this PR may be merged to **master** immediately. Thanks for preparing it.

lbergelson:
👍 Good catch.

jkbonfield:
I'm not a VCF maintainer, but my recollection was that removing the PDF was agreed upon.  We may at some point also consider renaming the CSIv2.tex to CSIv2.draft.tex to reinforce the unofficial nature of it, but I'll let the maintainers thrash that one out.

PR #504, opened 2020-05-28
Adding Vcf 4.4 draft spec to the index.md.

lbergelson:
I didn't realize this was a thing that needed updating when I added the vcf 4.4 pdf.


lbergelson:
@jmarshall Since your edit to my last PR made me aware of this, could you take a look?  No good deed goes unpunished!

jmarshall:
I thought the plan was not to publicise this 4.4 spec until it is (more) finalised? The source LaTeX has been added to **master** because that makes PRs and editing easier, but that doesn't mean that the PDF needs to be checked in or links like this one added at this stage. (Actually I thought there was a plan to add it with a filename like _VCFv4.4draft.tex_ and rename it at the end, but that hasn't eventuated.)

I guess the trade-off is between how easy it is for non-involved people to look at the draft text vs. how obvious it is that 4.4 is an early non-canonical draft (even before the reader opens the PDF and sees the watermarks). It seems to me that the VCF editors should either

* Have the PDF checked in, add this link in _index.md_ (that looks fine), and also add text to the VCF paragraph of _README.md_ explaining that `**[VCFv4.4.tex](VCFv4.4.pdf)**` (but using a reference-style link like the others) is a still-in-flux early draft. Then it's as obvious to readers of <http://samtools.github.io/hts-specs/> as it is to readers of <https://github.com/samtools/hts-specs> that it's a draft, even before they look at the PDF contents.

or 

* Do none of those three things. (@hts-specs-bot will still generate previews on PRs but the PDF wouldn't be otherwise immediately available. Readers of the web site will be unaware of this draft, and readers of the hts-specs repo view will see just the _.tex_ filename.)

TL;DR — if you're going for the first option, this PR should add text and a link to _README.md_ too.

d-cameron:
I was under the impressed that the PDF should exist, with it's giant front page warning message.

If we're adding to the index, then yes, we should add to the readme as well, both with disclaimer text about this being a draft under continual active revision.

jmarshall:
Fair enough — but **why** are you under that impression? What is the plan about making this draft easily available for reading now, and to what extent do you still plan to make major sweeping changes? The PDF may have been added by inertia rather that by considering the pros and cons.

For comparison, the last PDF spec added, _crypt4gh_, did not add a PDF until the final merge (it was a somewhat different scenario in not having _crypt4gh.tex_ on **master** until the final merge either) and had a large watermark on every page during its development:

```latex
\usepackage{draftwatermark}
…
\SetWatermarkAngle{315}
\SetWatermarkText{PRELIMINARY}
\SetWatermarkScale{4}
```

lbergelson:
Well, this sparked more discussion than I anticipated... 
I added the PDF because I assumed that if we had a draft spec people would want to be able to read it easily and see diffs / formatting etc.  Installing latex can be a bear and I suspect many people who want to comment on the spec probably don't want to deal with that.

I've renamed the files to include the word *draft* and added a watermark which seems like a good idea.   I also updated the readme.

yfarjoun:
needs a rebase.

PR #511, opened 2020-06-19
Correct CRAM MF bit 2: it means unmapped, not mapped.

jkbonfield:
The pseudocode in CRAMv3.tex was already correct.

This flaw appears to date all the way back to the initial CRAM1.0 pdf document:

https://www.ebi.ac.uk/sites/ebi.ac.uk/files/groups/ena/documents/cram_format_1.0.1.pdf

Implementation-wise, the initial commit for the CRAMtools Java implementation of CRAM (enasequence/cramtools@6576802) had (Oct 28 2012):

```
        public byte getMateFlags() {
                if (mateFlags == null) {
                        byte b = 0;
                        b |= mateMapped ? 1 : 0;
                        b <<= 1;
                        b |= mateNegativeStrand ? 1 : 0;
                        mateFlags = new Byte(b);
                }
                return mateFlags;
        }
```

7 days later, commit enasequence/cramtools@f0cf055 changed that with:

```
        public byte getMateFlags() {
                if (mateFlags == null) {
                        byte b = 0;
-                       b |= mateMapped ? 1 : 0;
+                       b |= mateUmapped ? 1 : 0;
                        b <<= 1;
                        b |= mateNegativeStrand ? 1 : 0;
                        mateFlags = new Byte(b);
```

This is shortly before the first release of CRAM (indeed the commit message for above is "cram2bam, almost there").  It appears the spec was never updated so has been wrong since day 1.

### DIFF
--- a/CRAMv2.1.tex
+++ b/CRAMv2.1.tex
@@ -964,7 +964,7 @@ The following flags are defined for each CRAM read record:
 0x10 &  & SEQ being reverse complemented\tabularnewline
 \hline
 0x20 & calculated* or stored in the mate's info & SEQ of the next segment in the 
-template being reversed\tabularnewline
+template being reverse complemented\tabularnewline
 \hline
 0x40 &  & the first segment in the template\tabularnewline
 \hline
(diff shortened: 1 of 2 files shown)


jmarshall:
Nice catch, and makes it line up better with the “BAM bit flags (BF data series)” table on p14.

It's not so critical, but while you're nearby you could also apply `s/reversed/reverse complemented/` to 0x20 on that table  — looks like this was copied from SAM prior to e78064fc638475d7b85821d7fb36db64bcea458b.

jkbonfield:
Done.  It's a bit embarassing to think it's been there all that time!  I caught a huge number of similar things during the 1.0 to 2.0 CRAM update while writing my implementation.  Every time I had to go to the code to figure out why things didn't work I noted it down.  I guess that's one I missed.   I wonder how many others are lurking :-)

PR #512, opened 2020-06-24
CRAM validation / test files

jkbonfield:
The files are numbered such that a natural sort order will mean the earlier files may be decodable without yet being able to decode any of the later files.  Therefore this may be considered as a roadmap for development of new CRAM decoders.

Not yet complete, but obviously better than what was there before. :-)
I made the PR so people can provide feedback while this is progressing.


jkbonfield:
Finally beaten the .md file into some reasonable state.

I think I'm happy with this as a starting point for CRAM 3.0.  CRAM 3.1 would just be a few new codecs included (which are tested outside of CRAM in their own repository already), so it'd be easy enough to add them when we're ready.  What I don't have are failure tests, but they're not as vital as the "passed" ones.

I've included pre-made CRAMs and the samtools decoded versions of their SAM files.  We probably want something similar for BAM too, but I'm all worn out in specs testing land atm.

tskir (approved):

tskir:
As discussed on the File Formats call 2021-04-01 between @jmarshall, @lbergelson, @tcezard and @tskir, this PR **will be merged in its current form** in four weeks (2021-04-29), unless there are strong objections.

PR #468, opened 2020-02-03
Allow multiple SVs per record

d-cameron:
5Although the example in 5.4.3 has multiple SV alt alleles, many of the SV INFO fields definitions support only a single SV per record. This PR updates these fields to allow multiple SVs per record.

PR #517, opened 2020-07-13
Added SVCLAIM field

d-cameron:
Would like to add examples of how a CNV and SV caller would report transposition events (both replicative and conservative), as well some more complicated ones such as CN-neutral chromothripsis.

Waiting on #491 for actual files.

Should we be putting more examples directly into the VCF document itself? I'm of the opinion that the specs themselves should be clean, but come with a relatively extensive set of example stand and edge case VCFs.


VCFv4.4.tex
```diff
@@ -625,6 +625,20 @@ \section{INFO keys used for structural variants}
 \end{verbatim}
 \normalsize
 
+\footnotesize
+\begin{verbatim}
+##INFO=<ID=SVCLAIM,Number=1,Type=String,Description="Claim made by the structural variant call. Valid values are CN, BP, CNBP for copy number, breakpoint and both respectively.">
+\end{verbatim}
+\normalsize
+
+This field disambuates the claim being made by variant caller for DEL, INS, DUP structural variants.
```
jmmut, line 634:
```suggestion
This field disambiguates the claim being made by variant caller for DEL, INS, DUP structural variants.
```

VCFv4.4.tex
```diff
@@ -625,6 +625,20 @@ \section{INFO keys used for structural variants}
...
+CN indicates the claim is a copy number claim and the call has been made based on a measure of DNA abundance of the called region.
+BP indicates the claim is a breakpoint claim and the call has been made based on the detection of a non-reference DNA adjacency.
+CNBP indicates the claim is both a copy number and breakpoint claim.
+Records with SVTYPE of CNV are implicitly copy number claims. SVCLAIM is optional for these records but, if specified, cannot be BP.
+Records with SVTYPE of BND or INV are implicitly breakpoint claims. SVCLAIM is optional for these records but, if specified, cannot be CN.
```
jmmut, line 639: > if specified, cannot be CN

does this prevent to state a CNBP claim with breakend notation? how is it expected to state for instance that a translocation has CN=1? (in the sense that it's not a deletion nor a duplication, but just a sequence that moved to another place)

d-cameron: It's not intended to. I'll reword it to make this explicit.

VCFv4.4.tex
```diff
@@ -625,6 +625,20 @@ \section{INFO keys used for structural variants}
...
+\footnotesize
+\begin{verbatim}
+##INFO=<ID=SVCLAIM,Number=1,Type=String,Description="Claim made by the structural variant call. Valid values are CN, BP, CNBP for copy number, breakpoint and both respectively.">
+\end{verbatim}
+\normalsize
+
+This field disambuates the claim being made by variant caller for DEL, INS, DUP structural variants.
+CN indicates the claim is a copy number claim and the call has been made based on a measure of DNA abundance of the called region.
+BP indicates the claim is a breakpoint claim and the call has been made based on the detection of a non-reference DNA adjacency.
+CNBP indicates the claim is both a copy number and breakpoint claim.
```
lbergelson, line 637: Would this be better formatted as a 3 element list?

tskir, line 637: @d-cameron @jmmut I suggest including a short explanation about why this is necessary. It took me several weeks of discussions to understand this very complex topic, so a regular VCF user might be confused otherwise.

I also rewrite the claim types as a list as suggested by @lbergelson.

```suggestion

The distinction between different claim types is made necessary by the fact that many structural variants affect both the overall copy number of a region and its adjacency structure, but often only one of those effects is being directly observed. It is important to specify which one, because if this is not done, other events in a genome may make the call ambiguous.

For example, a “conventional” deletion usually both decreases the total copy number of a region and creates an adjacency between two sequence locations, which are the breakpoints of that deletion. However, if a deletion is accompanied by a compensatory event elsewhere in the genome, it may only cause changes in the adjacency structure, but not the overall copy number. On the contrary, combinations of certain complex events in a genome may cause the overall copy number decrease, but not the linear excision of a sequence fragment.

To address this, three claim types are defined:

\begin{itemize}
  \item CN (copy number) claim indicates that the call has been made based only on a measure of DNA abundance of the called region, with no evidence to support changes in breakpoint structure.
  \item BP (breakpoint) claim indicates that the call has been made based on the detection of a non-reference DNA adjacency, with no evidence to support overall changes in DNA abundance.
  \item CNBP indicates that there is evidence for both DNA abundancy and adjacency changes, which are consistent with each other and suggest the structural variant of the type being reported.
\end{itemize}
```

VCFv4.4.tex
tskir: I think this CN (copy number claim) is meant to be different from the other CN (number of copies) which is the INFO/FORMAT tag. Actually, maybe we want to change _this_ CN to something else to avoid the confusion?

tskir:
I also wanted to comment on this:

> Would like to add examples of how a CNV and SV caller would report transposition events (both replicative and conservative), as well some more complicated ones such as CN-neutral chromothripsis.
>
>Waiting on #491 for actual files.

I think that (once the comments are addressed) we can go ahead and merge this PR, and the example files can be added in a later PR. It helps to try to reduce the number of PRs opened simultaneoulsy.

VCFv4.4.tex
lbergelson: That's a good point.  Would changing the nomenclature from copy number -> abundance and breakpoint -> novel adjacency distinguish it better?

tskir: “Abundance" + “adjacency” sounds nice, although I'm not sure how to best abbreviate those. Presumably we need two/three letter acronyms, plus something to refer to the "both claims at once" case

VCFv4.4.tex
jmmut: +1000 except that the style used for quotes in the spec is different:
```
a ``conventional'' deletion
```

 https://github.com/samtools/hts-specs/blob/master/VCFv4.3.tex#L37

tskir: @jmmut Thanks a lot! I updated the suggestion.

My last hands on experience with LaTeX was in 2013, so I was certain I would mix up some formatting things :)

d-cameron:
- Needs Type=A for consistency with #468
- Further expand on @tskir's suggestion, with DUP example giving multiple interpretations of 
- Abundance and adjacency are nice as they're clear and not overloaded. `ADJ` is obvious. Abundance not so much. `AB`, ABD`, `ABU`? Would quantity/`QTY` work or is that going too far towards a nice acronyn for the wrong word?
  - What about 1 letter abbreviations for 'J' for adjacency (aka break Junction), and abundance 'A' (or would `C`/copies work better)?

VCFv4.4.tex
tskir: @d-cameron @jmmut @lbergelson 

> Abundance and adjacency are nice as they're clear and not overloaded. ADJ is obvious. Abundance not so much. AB, ABD, ABU? Would quantity/QTY work or is that going too far towards a nice acronyn for the wrong word? What about 1 letter abbreviations for 'J' for adjacency (aka break Junction), and abundance 'A' (or would C/copies work better)?

The main problem with ADJ + ABU/ABD is that they start with the same letter and require some mental effort to distinguish. I think “adjacency” + “quantity” is the perfect word pair, and can be translated into either of those choices:
1. **ADJ + QTY + BOTH**
1. ADJ + QTY + ADJQTY
1. A + Q + B
1. A + Q + AQ

To me, the first one looks like it's the most elegant and readable.

VCFv4.4.tex
```diff
@@ -625,6 +625,41 @@ \section{INFO keys used for structural variants}
 \end{verbatim}
 \normalsize
 
+\footnotesize
+\begin{verbatim}
+##INFO=<ID=SVCLAIM,Number=A,Type=String,Description="Claim made by the structural variant call. Valid values are CN, BP, CNBP for copy number, breakpoint and both respectively.">
```
jmmut, line 630: update to D, J and DJ. also, the verbatim formatting in the pdf makes this to go out of the page. I would just break it in two lines, as the alternative is to put the lines vertically as in other places of the pdf.

VCFv4.4.tex
```diff
@@ -625,6 +625,41 @@ \section{INFO keys used for structural variants}
...
+
+To resolve any ambiguity in the interpretation of deletion and duplication symbolic alleles, three claim types are defined:
+
+\begin{itemize}
+  \item D (abun\textbf{\emph{d}}ance / read \textbf{\emph{D}}epth) claim indicates that the call has been made based only on a measure of DNA abundance of the called region, with no evidence to support changes in breakpoint structure. This includes indirect claims of abundance made using SNV variant allele frequency.
+  \item J (ad\textbf{\emph{j}}acency / break \textbf{\emph{J}}unction) claim indicates that the call has been made based on the detection of a non-reference DNA adjacency, with no evidence to support overall changes in DNA abundance.
+  \item DJ indicates that there is evidence for both DNA abundancy and adjacency changes, which are consistent with each other and suggest the structural variant of the type being reported.
+\end{itemize}
+CNV alleles are implicitly abundance claims (SVCLAIM=D).
+SVCLAIM is optional but, if specified, must be D, DJ, or the missing value ``.'', which is equivalent to D.
```
jmmut, line 657: I would be ok merging this, but notice how you say contradicting things here and two lines below, without explaining that those sentences only apply to CNV, and "INV, INS, breakpoint, and single breakend alleles" respectively. In the pdf it looks like a single paragraph.

VCFv4.4.tex
tskir: I support breaking into two lines. IMO having the lines run vertically is terribly unreadable and should be avoided.

VCFv4.4.tex
```diff
@@ -625,6 +625,41 @@ \section{INFO keys used for structural variants}
...
+In contrast, a breakpoint based duplication call requires that the end of the duplicated region be adjacent to the start of the duplicated region.
+This rules out events such as dispersed duplication or retrotransposition, but is consistent with simple tandem duplications as well as double minutes.
+Both deletion and duplication breakpoint calls can form part of more complex events that result in no copy number change in the deleted/duplicated region.
+
+To resolve any ambiguity in the interpretation of deletion and duplication symbolic alleles, three claim types are defined:
+
+\begin{itemize}
+  \item D (abun\textbf{\emph{d}}ance / read \textbf{\emph{D}}epth) claim indicates that the call has been made based only on a measure of DNA abundance of the called region, with no evidence to support changes in breakpoint structure. This includes indirect claims of abundance made using SNV variant allele frequency.
+  \item J (ad\textbf{\emph{j}}acency / break \textbf{\emph{J}}unction) claim indicates that the call has been made based on the detection of a non-reference DNA adjacency, with no evidence to support overall changes in DNA abundance.
```
tskir, line 653: IMO a combination of italic+bold looks a bit weird and overaccentuates those letters. I would suggest instead using either just bold without italic, or (in my view even better) simply underlining them.

VCFv4.4.tex
```diff
@@ -625,6 +625,41 @@ \section{INFO keys used for structural variants}
...
+A copy number based duplication call does not make any claims on where the duplicated region is in the genome.
+It could be a simple tandem duplication, a dispersed duplication, retrotransposition, or even form part of extra-chromosomal DNA such as a double minute.
+In contrast, a breakpoint based duplication call requires that the end of the duplicated region be adjacent to the start of the duplicated region.
+This rules out events such as dispersed duplication or retrotransposition, but is consistent with simple tandem duplications as well as double minutes.
+Both deletion and duplication breakpoint calls can form part of more complex events that result in no copy number change in the deleted/duplicated region.
+
+To resolve any ambiguity in the interpretation of deletion and duplication symbolic alleles, three claim types are defined:
+
+\begin{itemize}
+  \item D (abun\textbf{\emph{d}}ance / read \textbf{\emph{D}}epth) claim indicates that the call has been made based only on a measure of DNA abundance of the called region, with no evidence to support changes in breakpoint structure. This includes indirect claims of abundance made using SNV variant allele frequency.
```
tskir, line 652: Hmm, actually a good point bringing up SNV frequencies. I wonder if this should be part of “D” claim, or moved out to be a separate “F” (frequency) claim? The problem with frequencies is that they don't even provide an explicit depth measure, only ratios of depths of different alleles

VCFv4.4.tex
tskir: Yes, I agree, and I would suggest reformatting this last paragraph as a list, starting it with something like “Rules of applying SVCLAIM field are different depending on symbolic allele type”, and then list the cases one by one

VCFv4.4.tex
```diff
@@ -625,6 +625,41 @@ \section{INFO keys used for structural variants}
...
+  \item J (ad\textbf{\emph{j}}acency / break \textbf{\emph{J}}unction) claim indicates that the call has been made based on the detection of a non-reference DNA adjacency, with no evidence to support overall changes in DNA abundance.
+  \item DJ indicates that there is evidence for both DNA abundancy and adjacency changes, which are consistent with each other and suggest the structural variant of the type being reported.
+\end{itemize}
+CNV alleles are implicitly abundance claims (SVCLAIM=D).
+SVCLAIM is optional but, if specified, must be D, DJ, or the missing value ``.'', which is equivalent to D.
+INV, INS, breakpoint, and single breakend alleles are implicitly adjacency claims (SVCLAIM=J).
+SVCLAIM is optional but, if specified, must be J, DJ, or the missing value ``.'', which is equivalent to J.
```
ctsa, line 659: @d-cameron - What is "DJ" supposed to mean for the balanced SVs listed here? Isn't this only meant to be well defined for DEL & DUP?

VCFv4.4.tex
```diff
@@ -625,6 +625,41 @@ \section{INFO keys used for structural variants}
...
+Both deletion and duplication breakpoint calls can form part of more complex events that result in no copy number change in the deleted/duplicated region.
+
+To resolve any ambiguity in the interpretation of deletion and duplication symbolic alleles, three claim types are defined:
+
+\begin{itemize}
+  \item D (abun\textbf{\emph{d}}ance / read \textbf{\emph{D}}epth) claim indicates that the call has been made based only on a measure of DNA abundance of the called region, with no evidence to support changes in breakpoint structure. This includes indirect claims of abundance made using SNV variant allele frequency.
+  \item J (ad\textbf{\emph{j}}acency / break \textbf{\emph{J}}unction) claim indicates that the call has been made based on the detection of a non-reference DNA adjacency, with no evidence to support overall changes in DNA abundance.
+  \item DJ indicates that there is evidence for both DNA abundancy and adjacency changes, which are consistent with each other and suggest the structural variant of the type being reported.
+\end{itemize}
+CNV alleles are implicitly abundance claims (SVCLAIM=D).
```
ctsa, line 656: What does CNV allele mean here? Is this meant to represent DEL/DUP only. If so, SVCLAIM=J should also be valid. If not, why aren't DEL/DUP itemized in this list?

VCFv4.4.tex
```diff
@@ -625,6 +625,41 @@ \section{INFO keys used for structural variants}
...
+Both deletion and duplication breakpoint calls can form part of more complex events that result in no copy number change in the deleted/duplicated region.
+
+To resolve any ambiguity in the interpretation of deletion and duplication symbolic alleles, three claim types are defined:
+
+\begin{itemize}
+  \item D (abun\textbf{\emph{d}}ance / read \textbf{\emph{D}}epth) claim indicates that the call has been made based only on a measure of DNA abundance of the called region, with no evidence to support changes in breakpoint structure. This includes indirect claims of abundance made using SNV variant allele frequency.
+  \item J (ad\textbf{\emph{j}}acency / break \textbf{\emph{J}}unction) claim indicates that the call has been made based on the detection of a non-reference DNA adjacency, with no evidence to support overall changes in DNA abundance.
+  \item DJ indicates that there is evidence for both DNA abundancy and adjacency changes, which are consistent with each other and suggest the structural variant of the type being reported.
```
ctsa, line 654: Given the example case below, I might have a CNV DEL call with breakpoints refined by SV breakpoints that do not represent a conventional deletion. Can the language here clarify whether or not this is "DJ"? I assume it is not meant to be, but the term "adjacency changes" here is fairly open to interpretation.

image

VCFv4.4.tex
tskir: Will be addressed in #550

VCFv4.4.tex
tskir: My past self kept bringing up some obscure things which are clearly out of scope for this PR. Apologies. Closing this discussion

VCFv4.4.tex
```diff
@@ -625,6 +625,41 @@ \section{INFO keys used for structural variants}
 \end{verbatim}
 \normalsize
 
+\footnotesize
```
tskir, line 628: As with the other PRs, please rebase against `VCFv4.4.draft.tex`

VCFv4.4.tex
tskir, line 630: CN, BP, CNBP should be changed to D, J, DJ, as well as their respective explanations

tskir (changes requested):
I already reviewed this in the past, so only minor changes remaining from my point of view.

And also could you please address three comments by @ctsa?

VCFv4.4.tex
d-cameron: INV DJ is acceptable as it's a claim of 2 breakpoints and no CN change.

VCFv4.4.tex
d-cameron: itemised list

VCFv4.4.tex
d-cameron: Added clarification of the relationship between SVCLAIM and complex events.

tskir (approved):

Issue #425, opened 2019-06-26
VCF spec should highlight INFO/END's use in indexing


jmarshall:
The VCF specification's §1.6.1's Table 1 simply describes END as

    END    1    Integer    End position (for use with symbolic alleles)

and describes its value in more detail later in §3 for structural variants:

> For precise variants, END is POS + length of REF allele − 1, and the for imprecise variants the corresponding best estimate.

Readers would not guess from this that END, when present, provides the size of the interval that the variant is located in — and both htslib and htsjdk use it when constructing indexes for VCF/BCF files. So in fact it is vital that this field, if present, is correct and is used with the intended meaning.

The specification should highlight END's use in indexing. Ideally the specification should have a more detailed description of END's value.

jmarshall:
For example, gnomAD's SV files use END for their own purposes with a different meaning and value. It was recently noticed that this breaks `bcftools view` with a query region (hat tip @lindenb): see [this twitter thread](https://twitter.com/yokofakun/status/1143451289130590209), [this biostars posting](https://www.biostars.org/p/386345/), and macarthur-lab/gnomad_browser#140.

cwhelan:
There has been some discussion about fixing this as part of https://github.com/samtools/hts-specs/pull/266

jmarshall:
@cwhelan: In particular, this comment of yours: https://github.com/samtools/hts-specs/pull/266#discussion_r282908254. I wish I had realised its significance better at the time!

jmarshall:
Delly has also been using END like this for a number of years; see dellytools/delly#159.

PR #519, opened 2020-07-17
Add LEN INFO field as a replacement for END

lbergelson:
As discussed on the calls this is PR to deprecate the END field in VCF 4.4 and replace it with a new LEN field.  This would be a breaking change that requires special support in vcf libraries.  The benefits are that it should be easier to use and less error prone.  It should also compress better although as an INFO field this isn't a huge deal.   It matches the proposed RBS field better than END does as well.

LEN may not be the right name, I'm happy to change it.

### DIFF
--- a/VCFv4.4.draft.tex
+++ b/VCFv4.4.draft.tex
@@ -373,9 +373,10 @@ Fixed fields are:
 	CIGAR		& A		& String	& Cigar string describing how to align an alternate allele to the reference allele \\
 	DB		& 0		& Flag		& dbSNP membership \\
 	DP		& 1		& Integer	& Combined depth across samples \\
-	END		& 1		& Integer	& End position on CHROM (used with symbolic alleles; see below) \\
+	END		& 1		& Integer	& End position on CHROM (deprecated; prefer LEN; see below) \\
 	H2		& 0		& Flag		& HapMap2 membership \\
 	H3		& 0		& Flag		& HapMap3 membership \\
+	LEN		& 1		& Integer 		& Length of variant (used with symbolic alleles; see below) \\
 	MQ		& 1		& Float		& RMS mapping quality \\
 	MQ0   		& 1		& Integer	& Number of MAPQ == 0 reads \\
 	NS		& 1		& Integer	& Number of samples with data \\
@@ -387,11 +388,19 @@ Fixed fields are:
 
 \begin{itemize}
 \renewcommand{\labelitemii}{$\circ$}
-\item END: End reference position (1-based), indicating the variant spans positions POS--END on reference/contig CHROM.
-Normally this is the position of the last base in the REF allele, so it can be derived from POS and the length of REF, and no END INFO field is needed.
-However when symbolic alleles are used, e.g.\ in gVCF or structural variants, an explicit END INFO field provides variant span information that is otherwise unknown.
+\item LEN / END: Fields used to explicitly specify the end position of the variant on the reference/contig CHROM.
+Normally this is the position of the last base in the REF allele, so it can be derived from POS and the length of REF, and no additional INFO field is needed.
+However when symbolic alleles are used, e.g.\ in gVCF or structural variants, an additional explicit field provides variant span information that is otherwise unknown.
+These fields are used to compute BCF's {\tt rlen} field (see~\ref{BcfSiteEncoding}) and are important when indexing VCF/BCF files to enable random access and querying by position.
 
-This field is used to compute BCF's {\tt rlen} field (see~\ref{BcfSiteEncoding}) and is important when indexing VCF/BCF files to enable random access and querying by position.
+  \begin{itemize}
+   \item END: End reference position (1-based), indicating the variant spans positions POS--END on reference/contig CHROM.
+   END is deprecated and has been replaced with LEN.
+   \item LEN: The length of the variant's alignment to the reference indicating the variant spans positions POS--(LEN - 1) on reference/contig CHROM.
+  \end{itemize}
+
+At most one of LEN or END should be present on each variant line.
+If both are present they must be consistent with each other.
 \end{itemize}
 
 \subsubsection{Genotype fields}
@@ -569,10 +578,10 @@ When a key reflects a property of a single alt allele (e.g.\ SVLEN), then when t
 \begin{verbatim}
 ##INFO=<ID=IMPRECISE,Number=0,Type=Flag,Description="Imprecise structural variation">
 ##INFO=<ID=NOVEL,Number=0,Type=Flag,Description="Indicates a novel structural variation">
-##INFO=<ID=END,Number=1,Type=Integer,Description="End position of the variant described in this record">
+##INFO=<ID=LEN,Number=1,Type=Integer,Description="Length of the variant described in this record">
 \end{verbatim}
 \normalsize
-For precise variants, END is $\mbox{POS} + \mbox{length of REF allele} - 1$, and the for imprecise variants the corresponding best estimate.
+For precise variants, LEN is \mbox{length of REF allele}, and the for imprecise variants the corresponding best estimate.
 
 \footnotesize
 \begin{verbatim}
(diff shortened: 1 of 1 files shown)


d-cameron:
We already have `SVLEN` but that won't work for `<INV>` events since `SVLEN=0` for those. I would have thought having `LEN=100;SVLEN=0` would be *more* confusion to tool authors and end users than keeping `END`.

VCFv4.4.draft.tex
```diff
@@ -387,11 +388,19 @@ \subsubsection{Fixed fields}
...
-This field is used to compute BCF's {\tt rlen} field (see~\ref{BcfSiteEncoding}) and is important when indexing VCF/BCF files to enable random access and querying by position.
+\item LEN / END: Fields used to explicitly specify the end position of the variant on the reference/contig CHROM.
+Normally this is the position of the last base in the REF allele, so it can be derived from POS and the length of REF, and no additional INFO field is needed.
+However when symbolic alleles are used, e.g.\ in gVCF or structural variants, an additional explicit field provides variant span information that is otherwise unknown.
+These fields are used to compute BCF's {\tt rlen} field (see~\ref{BcfSiteEncoding}) and are important when indexing VCF/BCF files to enable random access and querying by position.
+
+  \begin{itemize}
+   \item END: End reference position (1-based), indicating the variant spans positions POS--END on reference/contig CHROM.
+   END is deprecated and has been replaced with LEN.
+   \item LEN: The length of the variant's alignment to the reference indicating the variant spans positions POS--(LEN - 1) on reference/contig CHROM.
```
jmmut, line 399: > variant's alignment to the reference

Isn't it simpler to say 'LEN is the length of the REF allele' (as you do below)? as anecdote, my team had a funny/worrysome convention in some code that the variant's length is `max(length(REF), length(ALT))`

VCFv4.4.draft.tex
```diff
@@ -387,11 +388,19 @@ \subsubsection{Fixed fields}
...
+  \end{itemize}
+
+At most one of LEN or END should be present on each variant line.
+If both are present they must be consistent with each other.
```
jmmut, line 403: - here's a constraint
- if you break the constraint, at least be nice

maybe the second part can be removed?

VCFv4.4.draft.tex
```diff
@@ -872,12 +881,12 @@ \subsection{Encoding Structural Variants}
...
-1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;END=2827708;HOMLEN=1;HOMSEQ=G;SVLEN=-14                 GT:GQ        1/1:14
-2       321682 .         T                <DEL>        6    PASS   SVTYPE=DEL;END=321887;SVLEN=-205;CIPOS=-56,20;CIEND=-10,62         GT:GQ        0/1:12
-2     14477084 .         C                <DEL:ME:ALU> 12   PASS   SVTYPE=DEL;END=14477381;SVLEN=-297;CIPOS=-22,18;CIEND=-12,32       GT:GQ        0/1:12
-3      9425916 .         C                <INS:ME:L1>  23   PASS   SVTYPE=INS;END=9425916;SVLEN=6027;CIPOS=-16,22                     GT:GQ        1/1:15
-3     12665100 .         A                <DUP>        14   PASS   SVTYPE=DUP;END=12686200;SVLEN=21100;CIPOS=-500,500;CIEND=-500,500  GT:GQ:CN:CNQ ./.:0:3:16.2
-4     18665128 .         T                <DUP:TANDEM> 11   PASS   SVTYPE=DUP;END=18665204;SVLEN=76;CIPOS=-10,10;CIEND=-10,10         GT:GQ:CN:CNQ ./.:0:5:8.3
+1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;LEN=15;2827708;HOMLEN=1;HOMSEQ=G;SVLEN=-14                 GT:GQ        1/1:14
```
jmmut, line 884:
```suggestion
1      2827694 rs2376870 CGTGGATGCGGGGAC  C            .    PASS   SVTYPE=DEL;LEN=15;HOMLEN=1;HOMSEQ=G;SVLEN=-14                 GT:GQ        1/1:14
```

jmmut:
regarding the SVLEN, yes, it's unfortunate that SVLEN means "length **difference**". Not sure what to do about this TBH.

VCFv4.4.draft.tex
lbergelson: The first is a suggestion not a constraint.  It could be changed to a constraint though.

VCFv4.4.draft.tex
lbergelson: The problem with saying  it that way is that for gvcf ref-blocks the written REF allele will have length 1.  I'm open to changing the wording though.

Oh no!  That's a bad but easy to make mistake.

lbergelson:
So this came out of an item that was in our backlog of things to do.  I assumed it had originated in the future of vcf group to match the RBS treatment of end position but I honestly don't know where the initial plan had come from.  

I'm lukewarm on this change because it breaks things with a questionable benefit, but I thought it had backing from the future group.  Maybe I'm wrong about that...

It is a fairly significant improvement to compressibility of the end position in gvcf files but compressing single sample gvcfs isn't that big of a deal compared to multisample gvcfs which this doesn't really help. It also makes fewer fields dependent on the exact position which is nice, but changes to the start position will still usually require updating LEN so it's kind of a wash on usability.  

On the whole I think this is a better representation than END but not a massive improvement so it may not be worth changing given that END already exists.

This is more generic than SVLEN since it would also be used for gvcf blocks ( which SVLEN is inappropriate to use for).

daviesrob:
It was suggested because historically people have made mistakes using `END`, which has caused various problems (for example, #425 and samtools/bcftools#1154). So the idea was to replace it with something that is easier to use.

d-cameron:
VCFv4.4 redefines `END` from `1` to `A` copies to fix the issue that we can't encode multiple SVs in a single record so we already have an (at least partially) breaking change in 4.4. Looking at the bigger picture, we also have an example library, and a VCF 'strict' specifications in the works. `strict` should take care of the more egregious misuses of `END` (such as when it's before `POS`) through validation failures and hopefully tool authors will make extensive use of our library of example VCFs as they get created. Given that multiple popular SV callers have been misusing `END`, maybe we should consider this change. I'll bring this issue up at the next future of VCF meeting.

d-cameron:
If we're going to make a breaking change that deprecates `END`, we might as well do it properly and deprecate `SVTYPE` and `SVLEN` as well. That information is already in the ALT and `END`/`LEN` fields anyway.

Having thought about it, I'm warming to this approach. It essentially makes VCFv4.4 a notational reset for symbolic SV and lets us make changes that we've been reticent to make since they're not backwards compatible.

d-cameron:
So we have multiple interpretations of what `End position of the variant described in this record` actually means since there can be multiple variants described in the record. We can interpret it as:
1) record-level end of ref allele/scope of record
2) alt-level end of alt allele.

There were many SV fields needing changing from 1 per record, to 1 per alt in my 4.4 PR. I included END.

So, is END an indexing scoping field for the entire record, or an ALT level field?

If it's the former, we need to explicitly state that it needs to be set to the max scoping of the longest ALT, (and that symbolic SV alleles have SVLEN as the authorative size).

If it's the latter, then we'd need `Number=A` and the effective record length for indexing purposes `max(END)`. This seems like a it could be nicer for allele filtering and subsetting (since you don't need to work out what ALT is driving END nor have to recompute what the new END is supposed to be if an ALT is removed.

jkbonfield:
Given retrieval mechanisms in place are for an entire record rather than a single ALT, and that END is used for index computation, I don't see how END can mean anything other than the union of all alts - ie the entire record.

lbergelson:
We've decided that this change is not a worthwhile breaking change.  The END tag will remain the way it is, and we will redefine SVLEN to be able to represent the length of multiple SV's as a site.

lbergelson:
Closing this because it's more trouble than it's worth.

PR #521, opened 2020-07-22
Compute BIN for zero-consumption CIGARs as for unmapped reads

jmarshall:
Like placed unmapped reads, "mapped" reads with POS set but with unorthodox CIGARs that consume no reference bases would naturally span 0 bases on the reference. We artificially consider unmapped reads to have length 1 for BIN calculation and indexing purposes, so do the same thing for these zero-consumption reads.

See samtools/samtools#1240 and PR samtools/htslib#1063 for context. These are unorthodox alignments used to represent something other than mapping to a reference (if a mapping matches no reference bases at all, on what basis was POS chosen rather than flagging the record as unmapped?), so this change does not affect typical mapped data.

However even when data doesn't particularly mean anything in an alignment context, there's nothing stopping it appearing in a BAM file and it's good for tools to be able to do the expected thing with it. So this is a similar case to unmapped reads, and IMHO it's worth making a similar special case for this similar case.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -1084,7 +1084,7 @@ Section~\ref{sec:recommended-practice}) and reads whose {\sf CIGAR} strings
 consume no reference bases at all, the alignment is treated as
 being of length one.
 Note unmapped reads with {\sf POS} $0$ (which becomes $-1$ in BAM) therefore
-use $\mbox{\sf reg2bin}(-1, 0)$ which is $4680$.
+use $\mbox{\sf reg2bin}(-1, 0)$ which is computed as $4680$.
 
 \subsubsection{N\_CIGAR\_OP field}\label{sec:ncigar}
 With 16 bits, {\sf n\_cigar\_op} can keep at most 65535 CIGAR
@@ -1345,6 +1345,14 @@ The following functions compute bin numbers and overlaps for a BAI-style
 binning scheme with $6$~levels and a minimum bin size of $2^{14}$~bp.
 See the CSI specification for generalisations of these functions designed for
 binning schemes with arbitrary depth and sizes.
+
+When these functions are called with regions representing unplaced
+unmapped reads, e.g., $\mbox{\sf reg2bin}(-1, 0)$, they involve operations
+such as \verb|(-1)>>14| which are undefined or implementation-defined in
+some programming languages.
+They must be implemented as if these operations use the common
+two's-complement semantics: $\mbox{\sf reg2bin}(-1, 0) = 4680$ and
+$\mbox{\sf reg2bins}(-1, 0, \ldots)$ returns $[\,0, 0, 8, 72, 584, 4680\,]$.
 
 {\small
 \begin{verbatim}
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -1077,13 +1077,14 @@ those marked ``\textit{limited}'' are limited by available memory and the practi
 \subsubsection{BIN field calculation}\label{sec:bin-field}
 {\sf BIN} is calculated using the {\sf reg2bin()} function in
 Section~\ref{sec:code}.
-For mapped reads this uses {\sf POS-1} (i.e.,~0-based left position) and the
+For mapped reads this uses $\mbox{\sf POS}-1$ (i.e.,~0-based left position) and the
 alignment end point using the alignment length from the {\sf CIGAR} string.
-For unmapped reads (e.g.,~paired end reads where only one part is mapped, see
-Section~\ref{sec:recommended-practice}) the alignment is treated as
-being length one.
+For unmapped reads (e.g.,~paired-end reads where only one part is mapped, see
+Section~\ref{sec:recommended-practice}) and reads whose {\sf CIGAR} strings
+consume no reference bases at all, the alignment is treated as
+being of length one.
 Note unmapped reads with {\sf POS} $0$ (which becomes $-1$ in BAM) therefore
-use {\sf reg2bin(-1, 0)} which is $4680$.
+use $\mbox{\sf reg2bin}(-1, 0)$ which is $4680$.
 
 \subsubsection{N\_CIGAR\_OP field}\label{sec:ncigar}
 With 16 bits, {\sf n\_cigar\_op} can keep at most 65535 CIGAR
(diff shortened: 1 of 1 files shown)


jkbonfield:
We were discussing yesterday in our Samtools meeting, so nice timing.  If you wish for queries to return a read then it is also important that it is in the correct bin, so this makes sense.

+1 from me.

jmarshall:
Removed the sentence about this never representing an orthodox mapping, as per the MSA scenario discussed at last month's meeting — now described in the commit message:

> This change does not affect typical mapped data, though it would affect a read mapped entirely within an insertion at a particular reference position that has been located there via multiple sequence alignment with other nearby overlapping reads.

yfarjoun:
I don't understand why this is needed. my reading of reg2bin will result in some bin even if start==end. Why do we need a special case for this?

jmarshall:
We have long had a special case for unmapped reads, which do not consume any reference bases. This extends the special case to other scenarios of reads that do not consume any reference bases.

The purpose of this special case (for both unmapped reads and this new category of reads) is to make queries work even for such reads that are mapped at POS = 1. See the first few comments on samtools/samtools#1240 (then recoil in horror from @daviesrob's 11 May comment onwards).

jkbonfield:
> 
> 
> I don't understand why this is needed. my reading of reg2bin will result in some bin even if start==end. Why do we need a special case for this?

Specifically the issue is for reads that are mapped entirely to an insertion.  The number of reference bases consumed by such reads is zero, but the mapping coordinate of such a read is considered 1 past the end - ie all bases in the sequence are actually prior to POS.

If you want it pulled out in queries, then it must be given a length of 1 bp so it can be placed in the correct bin as very occasionally it'll be on a bin boundary with SEQ in the prior bin and POS in the next bin!  Yes, it's ugly and bad design, but we are where we are...

jmarshall:
Two new facets became apparent during this month's meeting:

* This change is in fact not really important for `reg2bin` and bin calculations at all — where the special case for unmapped or these reads is actually used is in another phase of interval queries, in checking that a putative intersecting read returned by walking the bin structure really does intersect the query interval. It's discussed in this spec section because the computation should be the same for querying and bin calculation, but the real problem is that the specification does not properly describe to this level of detail how to use bins to answer queries.

* The POS=0 `reg2bin(-1, 0)` special case needs rewording — at present it at least appears to depend on evaluating `(-1) >> 14` etc, which is not strongly defined in various programming languages.

jmarshall:
Added text resolving the `(-1) >> 14` issue.

Note the duplicated `0` returned by `reg2bins`! The appendix's definition for this function does this for some proper begin/end values too, so needs to be compared to the versions actually used in implementations…

SAMv1.tex
```diff
@@ -1345,6 +1346,14 @@ \subsection{C source code for computing bin number and overlapping bins}\label{s
 See the CSI specification for generalisations of these functions designed for
 binning schemes with arbitrary depth and sizes.
 
+When these functions are called with regions representing unplaced
+unmapped reads, e.g., $\mbox{\sf reg2bin}(-1, 0)$, they involve operations
+such as \verb|(-1)>>14| which are undefined or implementation-defined in
+some programming languages.
+They must be implemented as if these operations use the common
+two's-complement sematics: $\mbox{\sf reg2bin}(-1, 0) = 4680$ and
```
jkbonfield, line 1354: "sematics" -> "semantics" typo.

jkbonfield (commented):
Approve of change, barring 1 trivial typo.  Happy for it to be merged with that fix.

However minutes of last meeting cryptically refer to "one minor aspect to be updated".

jmarshall:
Thanks — typo fixed, and I added a hint about the previous interpretation to the history entry. The ”one minor aspect to be updated” minutes entry had me worried for a moment, but it was simply carried forward from the previous actions and refers to the remaining TODO which is now recorded separately as #539.

jkbonfield:
When I reviewed this originally, it all made sense, but I was specifically thinking about 0-size ranges and the desire to extend them to be a minimum of 1 base.  That all makes sense.  Or perhaps the reg2bin wrap-around change occurred after the review. 

However, for reads starting at position 0 (-1 in 0-base coords), with hindsight I think the correction to reg2bin is wrong, for the following reasons.  I haven't looked at the details myself, but am repeating the outcome from our weekly Samtools meetings.  These are a summary of Rob's findings - @daviesrob  please correct me if I'm remembering incorrect.

1. Apparently it's not what samtools does, it moves -1 to 0 before indexing so the sequences end up in a left-edge bin instead.
2. Htsjdk does allocate the far right bin as per the spec, but queries won't look in it so it's a moot point.
3. Similarly htslib also can't query this bin.  (Clearly the query needs fixing)
4. It's just illogical!  We're basically blessing an integer overflow, rather than "doing the right thing".  I assume we did this as it's what code appeared to already be doing (htsjdk), but with hindsight if nothing does the same wrap-around when doing range queries then it wasn't worth "blessing" this bug in the spec.

As a consequence, we're pretty sure the vast majority of software just isn't working.  It wasn't working before this spec change, and since it it's still broken.  Maybe it did work at the time of change?  It seems odd that we changed the spec this way if they weren't functioning at the time anyway.

Maybe we should have fixed it differently, but water under the bridge.

However if it is infact totally broken right now in mainstream implementations, then we lose nothing by changing the specification to be simpler, avoiding undefined wrap-around behaviour as a bonus.  So my thoughts for an alternate fix:

- State that coordinates <0 are treated as 0.  Hence no wrap-around.
- Coordinates must have a finite size, so a zero-width object will be unit width.  Hence -1 to -1 becomes 0 to 1 inclusive. 100-100 becomes 100-101, etc.
- We need to state that both left-edge bins and wrapped-around bins (eg 4680) are searched for by range queries, as we don't know which generation of index will be used.  (And right now we're have both being actively created already)

Thoughts?

daviesrob:
Actually, I've now traced the `reg2bin(-1, 0)` behaviour further back, to PR #2, no less.  I don't think anything has yet explained what you're supposed to do with the linear index for this case - but storing the data in the first entry is really the only solution that makes sense.

I would add that this problem doesn't affect SAM (the specification of which also describes the BAI index) as the only reads with position -1 are unmapped and so don't have a reference name and don't appear in the index.  It does affect formats like VCF which use BAI-like indexes, referring to the SAM specification for details, and also allow positions before the start of the reference.

jkbonfield:
AFAIK the only reason people want VCF pos 0 is to mark things that occur prior to the ref start.  So just shift the start by 1 and the problem becomes vastly simpler!  It feels like a failing of the GRC to not leave room for "before the start of my reference" as a coordinate by starting with an N.   Or am I hopelessly naive here?  Of course "wouldn't start from here" applies, so it's not a helpful observation.

Practically speaking though, we need to know categorically what each implementation is doing for both writing and querying.  Are there actually any fully working implementations out there?

If yes, changes would break them.
If no, then we shouldn't feel any pressure to maintain status quo if there is a good reason for change.

PR #522, opened 2020-07-22
SAM: explicitly limit the TLEN field to the "primary alignment".

jkbonfield:
Imagine the following scenario:

align

R1 and R2 form a proper pair in chromosome 1.  I refer to them as R1(p) and R2(p) sometimes before.  R2 also has a secondary alignment in another chromosome, sometimes referred to as R2(s).

The definitions from the spec as it currently stands:

> \item[Multiple mapping]
  The correct placement of a read may be ambiguous, e.g., due to repeats.  In
  this case, there may be multiple read alignments for the same read.  One of
  these alignments is considered primary.  All the other alignments have the
  secondary alignment flag set in the SAM records that represent them.  All the

>   \item For each read/contig in a SAM file, it is required that one and only
        one line associated with the read satisfies \mbox{`{\sf FLAG} {\tt \& 0x900
        == 0}'}. This line is called the \emph{primary line} of the read.

So the secondary R2 is not considered primary.  That is crystal clear.

Now let's look at PNEXT and RNEXT:

> \item {\sf RNEXT}: Reference sequence name of the primary alignment of the NEXT read in the
  template. [...]
\item {\sf PNEXT}: 1-based Position of the primary alignment of the NEXT read in the template.  [...]

So they only ever point to a primary alignment.  R1(p) and R2(p) point to each other, while R2 secondary points back to R1 primary.  (This would even be required if there is an R1 secondary mapping adjacent to R2 secondary forming a complete second template alignment, but that's another battle!)

TLEN however has no mention about primary alignments:

> \item {\sf TLEN}: signed observed Template LENgth. If all segments are
  mapped to the same reference sequence, the absolute value of TLEN
  equals the distance between the mapped end of the template and the
  mapped start of the template, inclusively [...]

It just talks about _all segments_.   This is a bit confusing.  A read is what comes off the sequencing instrument, which may contain one or more segments.  Thus for our purposes we start with two segments (read 1 and read 2).   Segment however isn't defined to be the original sequence data, but just a contigous sequence or sub-sequence.  If we have a chimeric alignment then I assume we may have multiple segments, although it confusingly switches to start talking about records instead.  However to me this implies a segment is an alignment and not the source material, and thus we can have as many segments as there are alignment records. 

Under that interpretation, our secondary alignment gives us 3 segments in play.  R1(p), R2(p) and R2(s).

Now you can see the problem.  TLEN is only valid when all segments match to the same chromosome, which they do not.  The fact we had a secondary alignment elsewhere has invalidated TLEN even for the two primary alignments.

My fix therefore is to change the wording to "If all segments in the primary alignment are"...

This doesn't however do much to explain what the TLEN field should be for the secondary alignment, but this is covered later on with:

>   It is set as 0 for a single-segment template or when the
  information is unavailable (e.g., when the first or last segment
  of a multi-segment template is unmapped or when the two are mapped
  to different reference sequences).

Arguably that could be expanded to make it explicit by adding to the "or" clause with "or this alignment record is non-primary".

This does however forbid TLEN from appearing in any secondary template pairs, but like I say that's a separate battle and we already have problems there with RNEXT / PNEXT being dictated to only ever point to primary alignments.  I'm sort of resigned to never managing to resolve that one as the desire for any review is zero. :/

Comments?

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -562,8 +562,9 @@ Each bit is explained in the following table:
   0 when the information is unavailable. This field equals {\sf POS} at the primary line of
   the next read. If {\sf PNEXT} is 0, no assumptions can be made on
   {\sf RNEXT} and bit 0x20.
-\item {\sf TLEN}: signed observed Template LENgth. If all segments are
-  mapped to the same reference sequence, the absolute value of TLEN
+\item {\sf TLEN}: signed observed Template LENgth.
+  For primary reads where the primary alignments of all reads in the template
+  are mapped to the same reference sequence, the absolute value of TLEN
   equals the distance between the mapped end of the template and the
   mapped start of the template, inclusively (i.e., $\mbox{end}-\mbox{start}+1$).%
   \footnote{Thus a segment
@@ -1438,6 +1439,7 @@ at \url{https://github.com/samtools/hts-specs}.}
 \subsection*{1.6: 28 November 2017 to current}
 
 \begin{itemize}
+\item Clarify the meaning of TLEN when secondary alignments are present. (May 2021)
 \item Bin calculation changed for alignment records whose CIGAR strings consume no reference bases: like unmapped records, they are considered to have length one (rather than zero). (Jan 2021)
 \item Correct the description of index pseudo-bins, which previously stated that {\sf ref\_beg}/{\sf ref\_end}, then named {\sf unmapped\_beg}/{\sf unmapped\_end}, include only placed unmapped reads. (Jul 2020)
 \item Add {\tt DNBSEQ} to the list of {\tt @RG PL} header tag values. (Apr 2020)


jkbonfield:
Oops, I just noticed this is an almost exact duplicate of https://github.com/samtools/hts-specs/issues/48

I guess it just goes to show how hard it is to get this fixed! lol.  6 years and no resolution.  However this is a PR and that was an issue.  Maybe that makes it one step closer. :-)

yfarjoun:
this is an improvement as it brings us to the implementations....

yfarjoun (approved):

jmarshall:
I agree with the direction*, but I think the wording needs more work.

[* Sigh. Really I too agree with @tfenne's comment on #48, but this PR has the advantage of being a small achievable step…]

(Firstly RNEXT/PNEXT talk about _reads_ but TLEN talks about _segments_. The read/segment distinction is super-confusing and not well explained in §1.2's glossary. I have a vague memory that segments were introduced to allow for sequencing technologies that turn reading/recording on and off as the molecular machinery proceeds along a strand (e.g. PacBio's strobe reads…?) but who knows; and who knows what if any relationship reads vs. segments might have with chimeric/multiple mapping…

Anyway, I don't want to divert this PR with trying to address that — just pointing out that one day we need to check every use of the word _segment_ in the SAM spec… :smile:)

The main problem with just adding “in the primary alignment” is that in §1.2 _primary alignment_ is a quality of one of the alignment records for a **read**, but here we're trying to use it to apply to a bunch of the alignment records for the whole **template**. So for someone reading the text in isolation, what we're trying to add here is unclear.

Maybe something like:

> For primary reads where the primary alignments of all reads[/segments/argh] in the template are mapped to the same reference sequence, […]

Also the commit message's “So for example […]” sentence is missing something between ‘with’ and ‘for’.

jmarshall:
Also this constitutes a rule change, so we should consider a SAM Version History entry outlining the change.

jkbonfield:
I did raise the whole read / segment / record debacle in my write up in the first post of this PR.

I don't know what it should say given the rather vagueness of it all, hence my change was to do the minimal thing possible.  However yes it probably meant reads all along.

As for requiring explicit versioning, my reasoning was we're not actually changing the *intention* of the spec, which is that TLEN is only valid for primary alignments.  I believe the current wording was written before supplementary / secondary alignments so it was valid (and identical with our intentions).  Since supplementaries / secondaries were added, literally no one has implemented TLEN in this manner.  It's totally broken.  I don't believe it's really a rule change, change a change of text to match the unwritten rule which never actually changed.

jkbonfield:
Thanks for the suggested rewording.  I've worked it in and also added to the version history and fixed the commit message.  (I've added you as a co-author given the new wording.)

One thought is whether we should be explicit when it comes to the value of TLEN for the non-primary case.  Do we just leave it unworded, as it is now, or do we explicitly want to state it is aligner-implementation defined for the non-primary records?  This is a big can of worms, so I'm happy with what we've got for step 1 at least.

colinhercus:
Wow, not sure this is right.

I think supplementary and secondary alignments that are mapped as a proper
pair should still have TLEN.

Cases in point are supplementary alignments to alt-scaffolds and decoy
sequences.

Secondary alignments where multiple pairs of alignments are reported for a
read. Some HLA typing programs use TLEN as part of their bayesian model.

colinhercus:
I just checked SAMv1 and Novoalign is not conforming to the standard for
RNEXT & PNEXT either.

If we map a pair to multiple locations the RNEXT and PNEXT are not from the
primary alignment but from the mate of secondary or supplementary
alignments when pair are "properly mapped"

A lot of downstream analysis tools using Novoalign BAM files rely on this
so I think we'll stick with this and make a note in our documentation.

jkbonfield:
> Wow, not sure this is right. I think supplementary and secondary alignments that are mapped as a proper pair should still have TLEN. Cases in point are supplementary alignments to alt-scaffolds and decoy sequences. Secondary alignments where multiple pairs of alignments are reported for a read. Some HLA typing programs use TLEN as part of their bayesian model.

This is what I was referring to as "_whether we should be explicit when it comes to the value of TLEN for the non-primary case_".

I fully agree that TLEN has been overly restrictive in the past.  I did try to fix this many years ago, but it just got quagmired in different views and got nowhere.  Hence this fix is at least a small incremental improvement which I hope no one will disagree with.   The disagreements, like this, are likely to be that it doesn't go far enough.

Specifically as it stands now,  PNEXT and RNEXT are both defined in terms of the next primary alignment.  TLEN however has no exclusion stating how it works on primary alignments.  It's just a blanket "all". 

> If all segments are mapped to the same reference sequence ...

That's clearly wrong as "all segments" includes secondary alignments, which should have no bearing on the value of TLEN for the primary alignments.  Not a single aligner I checked worked that way.  Most followed the "if all primary segments are mapped" model, but some like Novoalign are more "if all segments in this full template alignment are mapped..." argument, where a complete secondary alignment of both ends may occur on another chromosome as a logically self-consistent alternate alignment.  IMO it's totally correct in that scenario to use TLEN sensibly.  However that's not the purpose of this PR (and note to do so you'd need to reword PNEXT and RNEXT too). 

This PR clarifies the case for primary alignments by stating explicitly that TLEN on primary alignments is based on the mapping location of primary alignments (ie not secondary, glossing over the issue of whether supplementary alignments can be part of a primary alignment: IMO yes, but some people disagree with that interpretation).  It's not explicitly stating TLEN is invalid on secondary alignments.

So the question (for this PR) is what should we do for the case of non-primary alignments.   The change here is a minimal one, in that it doesn't explicitly discuss secondary alignments simply because the existing spec also doesn't discuss them.

There is a vague part later on which states:

> It is set to 0 for a single-segment template or when the information is unavailable (e.g. ...).

The examples there don't explicitly rule out Novoalign's implementation.  My interpretation is that it's covering the case of primary alignments that don't both match on the same strand.  I suspect this text was originally written before secondary alignments were added to the spec.

Obviously I would love for the whole PNEXT / RNEXT / TLEN trio to be clarified much more, but I've been there and got the scars.  The community just isn't up for fixing that, or at least wasn't back then.  If you want to tackle it then please make a PR with a suggested rewording.

jkbonfield:
> 
> I just checked SAMv1 and Novoalign is not conforming to the standard for RNEXT & PNEXT either. If we map a pair to multiple locations the RNEXT and PNEXT are not from the primary alignment but from the mate of secondary or supplementary alignments when pair are "properly mapped" A lot of downstream analysis tools using Novoalign BAM files rely on this so I think we'll stick with this and make a note in our documentation.

I totally agree.

I raised this back in 2014 in Issue #48, but we got nowhere with a resolution.  See https://github.com/samtools/hts-specs/issues/48#issuecomment-649361323 and the PDF linked from there for a bunch of examples.  Novoalign is doing, IMO, the correct and logical thing here.  However sadly it is technically forbidden by the wording of PNEXT and RNEXT.

My own personal preference would simply be to relax these and state they are aligner specific.  It's something we should have fixed many years ago, but the longer it's been left the harder it is to reword it now.  About the best thing we can do is to acknowledge the difference of opinion, permit the existing implementations to be valid, and maybe give guidance in an appendix (with eg similar pictures to the ones shown in my RFC) as to where you'd want PNEXT on a secondary to point back to the primary and where you'd want it to point to another secondary.

As I said above though, that's a different, more wide ranging, PR.  For purposes of not causing this minor amendment to just stall, I decided to bite off something tiny and test the waters with that first!

colinhercus:
Hi James,

I can see you tried before.

"I understand that some aligners may produce e.g. a single logical
alignment for R1 and then multiple for R2, but I think this is a bad idea"
      Novoalign will do this but only if none of the R2 made a proper pair
with R1. I think it's okay but maybe not useful.

"So, if you accept what I've described above, then I think defining what to
store in RNEXT/PNEXT and TLEN becomes simpler and perhaps even relatively
simple. I would suggest something like the following:

   1. RNEXT/PNEXT should contain the information for the single
   non-supplementary alignment record of the *other* read in the read pair
   from the same logical alignment"

I have a slight problem with this. Novoalign uses supplementary alignments
for alt-scaffolds and may map R1/2 as a proper pair. Perhaps I should have
made them secondary alignments but I used supplementary as variant callers
often ignore secondary alignments and I wanted to call variants on the
alt-scaffolds. It would be nice to have a separate flag bit for
alt-scaffold alignments.

jkbonfield:
Indeed the supplementary alt thing is a bit misleading.

Anyway, getting back on track for this PR specifically, if you can think of better wording for clarifying how TLEN works on the non-primary case then please suggest it.  At present it's not ideal, but if we do explicitly cover non-primary alignments then we'd need to be sure that it's (for now) consistent with the existing PNEXT/RNEXT statements while not being overly specific and wanting a change in phase 2 (if it ever materialises).

colinhercus:
"we'd need to be sure that it's consistent with the existing PNEXT/RNEXT"
makes having a logical definition of TLEN for supplementary and secondary
proper pairs less than moot.

With proposed TLEN definition and current RNEXT/PNEXT it makes no sense to
have proper pair alignments other than for the primary alignment. So maybe
you should think about adding that to the spec's as well.

Bit 2 0x2 each segment properly aligned according to the aligner.* Does not
apply to secondary and supplementary alignments.*

colinhercus:
I haven't looked at what BWA does for multi-mapped reads. I can see from
the discussion that RNEXT&PNEXT on secondary alignments are set to the
primary alignment.

Does bwa ever set 0x2 Proper Pair on secondary or Supplementary alignments?

If it doesn't then RNEXT&PNEXT definition could be changed for proper pairs
while remaining consistent with the way BWA works when 0x2 is not set.

jkbonfield:
Note I'm explicitly NOT trying to make this a rehash of #48 as I beat my head against that wall repeatedly and never got any traction.  Too many people wanting too many conflicting things or just dead against change.

Yes the problems still exist, but this PR is about the minimal fix to the current spec.  The current spec simply forbids PNEXT/RNEXT in proper-pair aligned secondary pairs.  I don't like that, but that's what it currently says.  Hence this change is to bring things into alignment so all 3 fields are singing from the same song sheet.  This should be trivial to do.

Fixing it properly will take much longer, and as mentioned I've been here before and frankly don't have the will power to try herding this particular flock of cats twice!

jkbonfield:
> 
> I haven't looked at what BWA does for multi-mapped reads. I can see from the discussion that RNEXT&PNEXT on secondary alignments are set to the primary alignment. Does bwa ever set 0x2 Proper Pair on secondary or Supplementary alignments? If it doesn't then RNEXT&PNEXT definition could be changed for proper pairs while remaining consistent with the way BWA works when 0x2 is not set.

I didn't check flags, but as reported in https://github.com/samtools/hts-specs/issues/48#issuecomment-648773081 bwa handles secondary alignments as always pointing back to the primary, even when the secondary sequences could align as a pair.

Anyway, for full pnext/rnext/tlen renegotiations, I suggest the conversion is taken to that issue.

jmarshall:
Force-pushed to avoid a conflict. I'm happy with the wording but again haven't done anything for an illustration. IMHO we should apply this now and then revisit an illustration and what to do about rules for non-primary alignment records separately.

jkbonfield:
This doesn't cope with what TLEN means for secondary / supplementary alignments, but it's a step forward in at least making it clearer and correct for the primary case.   Given how long anything to do with TLEN takes to get done, I'm all for chipping away at the edges of it one line at a time so we make *some* progress! :-)

Approved.

PR #533, opened 2020-12-03
Fix PacBio name

amwenger:
PacBio is "Pacific Biosciences" not "Pacific Biotechnology"


jmarshall:
Indeed it is (as per [e.g. here](https://www.pacb.com/company/about-us/)). Apologies to PacBio.

PR #537, opened 2020-12-17
strawman attempt at adding array of booleans

yfarjoun:
This is just a first draft, once the wording is finalized, I intend to:

- [ ] modify VCFv4.4 as well
- [ ] add and entry to the change logs

The point of this change is to allow an INFO (or FORMAT) field to contain an arrays of boolean values. 

it maintains the old behavior so that a Number of "0" would indicate that the presence of the field is "true" and its absence, "false". but it also allow all the other Number options, and in that case, 0 will indicate "false" while 1 will indicate "true".

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -151,7 +151,10 @@ There are also certain special characters used to define special cases:
   \item . (dot): The number of possible values varies, is unknown or unbounded.
 \end{itemize}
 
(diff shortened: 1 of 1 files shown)


d-cameron:
It's a bit unclear on what the values can/should be. I presume an array of 3 booleans would be `0,1,1` with 0=FALSE, 1=TRUE, but the PR should explicitly state that no other values other than 0 or 1 are allowed.

PR #542, opened 2021-02-01
Add subtitles noting VCF v4.1 & v4.2 have been superseded by a later version

jmarshall:
These older specifications currently have no mention of the later VCF specification that supersedes them. This adds such a note on the title page.

The October 2015 date comes from c8194c65a10b9487ab5b65be0ec75a6f11b3005b.

### DIFF
--- a/VCFv4.1.tex
+++ b/VCFv4.1.tex
@@ -7,7 +7,9 @@
 
 \begin{document}
 \input{VCFv4.1.ver}
-\title{The Variant Call Format (VCF) Version 4.1 Specification}
+\title{The Variant Call Format (VCF) Version 4.1 Specification
+\normalsize\\[2ex]
+(Superseded by the \href{http://samtools.github.io/hts-specs/VCFv4.2.pdf}{VCF v4.2} and \href{http://samtools.github.io/hts-specs/VCFv4.3.pdf}{v4.3} specifications)}
 \date{\headdate}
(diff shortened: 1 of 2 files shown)


tskir:
Thanks John, looks good to me

jmarshall:
Thanks Kirill, that's good quick executive action :+1: :smile:

That [starts the clock](https://github.com/samtools/hts-specs/blob/master/MAINTAINERS.md#rationale) on updating the PDFs. Probably pushing a collected PDF update after the meeting later this week will be fine.

tskir:
I agree, doing a joint update after the meeting sounds the most sensible.

I was also wondering if we could just teach @hts-specs-bot to automatically update PDFs in the master branch whenever there are changes? Sounds like it could keep track of the clock by itself and implement this “PDFs are at most one week old” policy

jmarshall:
Previously briefly discussed in https://github.com/samtools/hts-specs/pull/328#issuecomment-408589576 onwards. Having previews on PRs has been a great success, but IMHO having robots automatically making commits has a different set of tradeoffs and would not be worthwhile: too complex, and not flexible enough.

tskir:
I personally think [the future is here](https://twitter.com/gabro27/status/1173547934132178944?lang=en) in terms of how much autonomy can be granted to the robots; but I can see your point as well.

PR #553, opened 2021-04-05
Deprecate SVTYPE

d-cameron:
SVTYPE is redundant with with ALT.

This PR deprecates SVTYPE for VCFv4.4.

The field would have to have been redefined as TYPE=A anyway, and 4.4 is our best chance of cleaning up SV handling.

### DIFF
--- a/VCFv4.4.draft.tex
+++ b/VCFv4.4.draft.tex
@@ -177,7 +177,7 @@ Genotype fields specified in the FORMAT field are described as follows:
 
 Possible Types for FORMAT fields are: Integer, Float, Character, and String (this field is otherwise defined precisely as the INFO field).
 
-\subsubsection{Alternative allele field format}
+\subsubsection{Alternative allele field format} \label{altfield}
 Symbolic alternate alleles are described as follows:
 \begin{verbatim}
 ##ALT=<ID=type,Description="description">
@@ -582,7 +582,7 @@ For precise variants, END is $\mbox{POS} + \mbox{length of REF allele} - 1$, and
 \end{samepage}
 
 This field has been deprecated due to redundancy with ALT.
-Refer to the ALT field for set of valid symbolic structural variant alleles.
+Refer to section \ref{altfield} for the set of valid ALT field symbolic structural variant alleles.
 
 \footnotesize
 \begin{verbatim}
@@ -921,12 +921,12 @@ Notice how the ALT field expresses the orientation of the breakends.
 \vspace{0.3cm}
 \begin{tabular}{ l l l l l l l l }
 \#CHROM &POS & ID & REF & ALT & QUAL & FILTER & INFO \\
-$2$ & $321681$ & bnd\_W & G & G$]17$:$198982]$ & $6$ & PASS &  \\
-$2$ & $321682$ & bnd\_V & T & $]$13:123456$]$T & 6 & PASS &  \\
-$13$ & $123456$ & bnd\_U & C & C$[$2:321682$[$ & 6 & PASS &  \\
-$13$ & $123457$ & bnd\_X & A & $[$17:198983$[$A & 6 & PASS &  \\
-$17$ & $198982$ & bnd\_Y & A & A$]$2:321681$]$ & 6 & PASS &  \\
-$17$ & $198983$ & bnd\_Z & C & $[$13:123457$[$C & 6 & PASS &  \\
+$2$ & $321681$ & bnd\_W & G & G$]17$:$198982]$ & $6$ & PASS & . \\
+$2$ & $321682$ & bnd\_V & T & $]$13:123456$]$T & 6 & PASS & . \\
+$13$ & $123456$ & bnd\_U & C & C$[$2:321682$[$ & 6 & PASS & . \\
+$13$ & $123457$ & bnd\_X & A & $[$17:198983$[$A & 6 & PASS & . \\
+$17$ & $198982$ & bnd\_Y & A & A$]$2:321681$]$ & 6 & PASS & . \\
+$17$ & $198983$ & bnd\_Z & C & $[$13:123457$[$C & 6 & PASS & . \\
 \end{tabular}
 
 \subsubsection{Inserted Sequence}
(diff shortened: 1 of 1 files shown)


VCFv4.4.draft.tex
```diff
@@ -1257,10 +1248,10 @@ \subsubsection{Sample mixtures}
...
 \begin{tabular}{ l l l l l l l l l l l }
 \#CHROM & POS & ID & REF & ALT & QUAL & FILTER & INFO & FORMAT & Blood & TumorSample \\
-2 & 321681 & bnd\_W & G & G$]2:421681]$ & 6 & PASS & SVTYPE=BND;MATEID=bnd\_U & GT:CNADJ & 0:1 & 1:1 \\
-2 & 321682 & bnd\_V & T & $[2:421682[$T & 6 & PASS & SVTYPE=BND;MATEID=bnd\_X & GT:CNADJ & 0:1 & 1:1 \\
-13 & 421681 & bnd\_U & A & A$]2:321681]$ & 6 & PASS & SVTYPE=BND;MATEID=bnd\_W & GT:CNADJ & 0:1 & 1:1 \\
-13 & 421682 & bnd\_X & C & $[2:321682[$C & 6 & PASS & SVTYPE=BND;MATEID=bnd\_V & GT:CNADJ & 0:1 & 1:1 \\
+2 & 321681 & bnd\_W & G & G$]2:421681]$ & 6 & PASS & ;MATEID=bnd\_U & GT:CNADJ & 0:1 & 1:1 \\
+2 & 321682 & bnd\_V & T & $[2:421682[$T & 6 & PASS & ;MATEID=bnd\_X & GT:CNADJ & 0:1 & 1:1 \\
+13 & 421681 & bnd\_U & A & A$]2:321681]$ & 6 & PASS & ;MATEID=bnd\_W & GT:CNADJ & 0:1 & 1:1 \\
+13 & 421682 & bnd\_X & C & $[2:321682[$C & 6 & PASS & ;MATEID=bnd\_V & GT:CNADJ & 0:1 & 1:1 \\
```
tskir, line 1254: Here and in multiple other locations, an extra semicolon is present. E.g. `;MATEID=bnd\_U`, which should be `MATEID=bnd\_U`.

VCFv4.4.draft.tex
```diff
@@ -1172,7 +1163,7 @@ \subsubsection{Uncertainty around breakend location}
 This partner relationship can be specified explicitly with the tag PARID=bnd\_X in the VCF line for breakend V and PARID=bnd\_Y in the VCF line for breakend U, and vice versa.
 
 \subsubsection{Single breakends}
-We allow for the definition of a breakend that is not part of a novel adjacency, also identified by the tag SVTYPE=BND.
+We allow for the definition of a breakend that is not part of a novel adjacency, also identified by the tag .
```
tskir, line 1166: Removal of `SVTYPE=BND` now means that the sentence ends abruptly.

VCFv4.4.draft.tex
```diff
@@ -1004,10 +995,10 @@ \subsubsection{Large Insertions}
...
 \begin{tabular}{ l l l l l l l l }
 \#CHROM & POS & ID & REF & ALT & QUAL & FILTER & INFO \\
-$13$ & $123456$ & bnd\_U & C & C$[<$ctg$1>:229[$ & 6 & PASS & SVTYPE=BND \\
-$13$ & $123457$ & bnd\_V & A & $]<$ctg$1>:45]$A & 6 & PASS & SVTYPE=BND \\
-$<$ctg$1>$ & 1 & bnd\_X & A & $]<$ctg$1>:329]$A & 6 & PASS & SVTYPE=BND \\
-$<$ctg$1>$ & 329 & bnd\_Y & T & T$[<$ctg$1>:1[$ & 6 & PASS & SVTYPE=BND \\
+$13$ & $123456$ & bnd\_U & C & C$[<$ctg$1>:229[$ & 6 & PASS &  \\
+$13$ & $123457$ & bnd\_V & A & $]<$ctg$1>:45]$A & 6 & PASS &  \\
+$<$ctg$1>$ & 1 & bnd\_X & A & $]<$ctg$1>:329]$A & 6 & PASS &  \\
+$<$ctg$1>$ & 329 & bnd\_Y & T & T$[<$ctg$1>:1[$ & 6 & PASS &  \\
```
tskir, line 1001: Here and in several other locations, I understand the INFO should now contain `.` because there is no data

VCFv4.4.draft.tex
```diff
@@ -580,16 +580,9 @@ \section{INFO keys used for structural variants}
...
-  \item DEL: Deletion relative to the reference
-  \item INS: Insertion of novel sequence relative to the reference
-  \item DUP: Region of elevated copy number relative to the reference
-  \item INV: Inversion of reference sequence
-  \item CNV: Copy number variable region (may be both deletion and duplication)
-  \item BND: Breakend
-\end{itemize}
+
+This field has been deprecated due to redundancy with ALT.
+Refer to the ALT field for set of valid symbolic structural variant alleles.
```
tskir, line 585:
```suggestion
This field has been deprecated due to redundancy with ALT.
Refer to the ALT field for the set of valid symbolic structural variant alleles.
```

tskir (changes requested):
Definitely agree with the changes. SVTYPE becomes especially problematic when a single record contains multiple SVs of different types, so we should definitely get rid of it. Left several purely technical comments.

VCFv4.4.draft.tex
jmarshall: Dumb question: Does “ALT field” here refer to column 5 (so it's talking about breakend notation and/or…?) or is it talking about an INFO or FORMAT tag (possibly newly defined in another PR? If so, which?)?

tskir: @d-cameron to confirm, but I understand this to refer specifically to column 5. Indeed, it always contains either the variation type (DEL/DUP/INV/...), or the breakend notation, which makes it possible to always infer the type

jmarshall: Ah, okay. That would be clearer with a cross-reference to §1.4.5 (“Alternative allele field format”).

tskir: Valid point, @d-cameron do you think you could add it to this PR as well?

lbergelson (approved):
It looks like all the comments have been addressed.

PR #562, opened 2021-05-04
Moved the no-EOF test cram from passed to failed.

### LINKED ISSUE #561: test/cram: 0000_empty_noeof.cram is not valid
[CRAMv3.0 § 9 End of file container](https://samtools.github.io/hts-specs/CRAMv3.pdf#End%20of%20file%20container) states a final EOF container is required:

> A special container is used to mark the end of a file or stream. It is required in version 3 or later.

Therefore, [`test/cram/3.0/passed/0000_empty_noeof.cram`](https://github.com/samtools/hts-specs/blob/64c8759705ada74196198c4ff95ce23a1ec3b3e5/test/cram/3.0/passed/0000_empty_noeof.cram) should not be considered a valid CRAMv3 file.

jkbonfield:
As reported by Michael Macias, CRAM 3.0 makes EOF containers mandatory so the file is invalid.  I chose to keep it as it forms part
of the narrative for building up a CRAM one feature at a time, but document the failed status in the README.

Fixes #561


jkbonfield:
I did think about adding test data with internal empty containers, as I've done before with BAM testing (although we've yet to add any BAM test data to this repository).  The BAM spec explicitly states that the EOF marker must be at the end of file in order to sybolise EOF.  However CRAM has no such statement (which is likely my fault).

Therefore I'm wondering quite what this means.  Theoretically people could interpret the spec as implying a CRAM file with an appended other-format file is valid as the CRAM parser must stop once it hits the EOF container.  I doubt it's what we intended though, and note earlier versions of `samtools cat` did erroneously include the EOF containers, hence giving internal EOF markers.

Is this something we should fix?

Issue #551, opened 2021-03-01
Formally define GA4GH and MD5 checksums in VCF v4.4


tskir:
As discussed during GA4GH Connect 2021 calls:

@d-cameron

> Just checked VCFv4.3: MD5 isn't even formally defined as a ##contig header field. We should add both ga4gh & md5 into the specs for 4.4 

> md5 is shown in the example VCFs included in the specs, and the specs mention MD5 in S1.4.7, but it's not actually formally defined (e.g. header name, how the md5 hash is string encoded, etc 

@jkbonfield 

> Agreed, it needs a tight definition - uppercasing, white space removal, what to do about out of range chars, etc.

andrewyatz:
Hope we can refer to refget's spec as the reference for that final point @jkbonfield and if refget's spec is insufficient we will update accordingly

ohofmann:
Thank you for starting the discussion, @tskir - very much in support of this and would love to see this integrated with the refget services.

jkbonfield:
I'm hoping the definition of what's valid and how to deal with invalid chars is compatible between CRAM and RefGet (I'm sure they must be given the ancestry), which will therefore serve as the logical starting point for VCF.

Down stream, I think it's maybe worth trying to enforce more for VCF 4.4.   Right now, not only does it not require any checksum or assembly information, even the contig lines themselves are purely optional!  I can see it may help for rapid hacking around, but we're past those days and should really focus on data provenance for future spec versions.

lbergelson:
A way to refer to a reference collection in 1 line instead of needing thousands of contig lines would make requiring it a much easier pill for most users to swallow.

yfarjoun:
Isn't that almost have that specc.ed out in refget?

andrewyatz:
It's getting there :) @yfarjoun

PR #566, opened 2021-05-12
Clarify the the source of the contig in VCF

tcezard:
Very small PR to clarify that the MD5 for contigs are calculated from the sequence and not the whole file
This is not addressing #551 but should clarify what is already stated

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -223,7 +223,7 @@ The structured \texttt{contig} field must include the ID attribute and can inclu
 the following ones reserved:
 \begin{itemize}
   \item length: the length of the sequence
-  \item md5: MD5 checksum of the sequence as defined in the Sam specification v1\footnote{See Reference MD5 calculation
+  \item md5: MD5 checksum of the sequence as defined in SAM's {\tt @SQ-M5} field\footnote{See Reference MD5 calculation
       section in \href{https://samtools.github.io/hts-specs/SAMv1.pdf}{\tt SAM Format Specification}.} Briefly, the digest
       is calculated excluding all characters outside of the inclusive range 33 (`\char33') to 126 (`\char126')
       and all lowercase characters converted to uppercase. The MD5 digest is calculated as described in
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -224,8 +224,8 @@ the following ones reserved:
 \begin{itemize}
   \item length: the length of the sequence
   \item md5: MD5 checksum of the sequence as defined in the Sam specification v1\footnote{See Reference MD5 calculation
-      section in \href{https://github.com/samtools/hts-specs/blob/master/SAMv1.pdf}{\tt SAMv1.pdf}.} Briefly, the digest
-      is calculated including all characters outside of the inclusive range 33 (`\char33') to 126 (`\char126').
+      section in \href{https://samtools.github.io/hts-specs/SAMv1.pdf}{\tt SAMv1.pdf}.} Briefly, the digest
+      is calculated excluding all characters outside of the inclusive range 33 (`\char33') to 126 (`\char126')
       and all lowercase characters converted to uppercase. The MD5 digest is calculated as described in
       \href{https://tools.ietf.org/html/rfc1321}{\sl RFC 1321} and presented as a 32 character lowercase hexadecimal number.
   \item URL: tag to indicate where the sequence can be found
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -224,7 +224,7 @@ the following ones reserved:
 \begin{itemize}
   \item length: the length of the sequence
   \item md5: MD5 checksum of the sequence as defined in the Sam specification v1\footnote{See Reference MD5 calculation
-      section in \href{https://samtools.github.io/hts-specs/SAMv1.pdf}{\tt SAMv1.pdf}.} Briefly, the digest
+      section in \href{https://samtools.github.io/hts-specs/SAMv1.pdf}{\tt SAM Format Specification}.} Briefly, the digest
       is calculated excluding all characters outside of the inclusive range 33 (`\char33') to 126 (`\char126')
       and all lowercase characters converted to uppercase. The MD5 digest is calculated as described in
       \href{https://tools.ietf.org/html/rfc1321}{\sl RFC 1321} and presented as a 32 character lowercase hexadecimal number.
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -232,10 +232,11 @@ the following ones reserved:
 \end{itemize}
 
 For example:
+{\scriptsize
 \begin{verbatim}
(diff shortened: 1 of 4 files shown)


VCFv4.3.tex
```diff
@@ -219,7 +219,7 @@ \subsubsection{Assembly field format}
 \subsubsection{Contig field format}
 \label{sec-contig-field}
 It is recommended for VCF, and required for BCF, that the header includes tags describing the contigs referred to in the file.
-The structured \texttt{contig} field must include the ID attribute and typically includes also sequence length, MD5 checksum, URL tag to indicate where the sequence can be found, etc.
+The structured \texttt{contig} field must include the ID attribute and typically includes also sequence length, MD5 checksum of the sequence as defined in the \href{https://github.com/samtools/hts-specs/blob/master/refget.md}{Refget specification}, URL tag to indicate where the sequence can be found, etc.
```
tskir, line 222: Generally the direction of these changes looks good to me! I'm especially puzzled as to why examples had all the same length and hash to start with. Thanks for noticing & correcting.

I have slight concerns about linking to an external specification (Refget); nothing that can't be addressed though. In particular, I suggest:
* Explicitly specifying the version of the specification when linking (1.0.0 in this case);
* Linking to a permalink / frozen commit, not a live branch;
* Linking to a specific section of the page (`#checksum-calculation`);
* Making the specification a footnote instead of an inline link.

And finally, I think it's a good idea to also just duplicate the actual guidance here in the specification text as well (while still keeping the footnote linking to Refget), especially seeing it's just one sentence long:

> When calculating the checksum for a sequence, all non-base symbols (\n, spaces, etc) must be removed and then uppercase the rest. The allowed alphabet for checksum calculation is uppercase ASCII (`0x41`-`0x5A` or `A-Z`).

This makes things perfectly explicit while still keeping the provenance / alignment with where this guidance originally comes from.

jmarshall:
An MD5 checksum in a VCF `##contig` header will have come from the corresponding `M5` field of a SAM/BAM/CRAM `@SQ` header.

The SAM specification (currently §1.3.2, “Reference MD5 calculation”) describes the MD5 calculation in more detail than the refget spec does. The two algorithms **differ** in whether wacky non-alphabetical characters are included in the string to be hashed. (In general such characters mostly shouldn't appear in FASTA files so the differences are moot. But consider the possibility of `-` gap characters and the following—) In particular, SAM's string specifically may include `*` padding characters; Refget's string vaguely says “all non-base symbols (\n, spaces, etc) must be removed” and precisely says `A-Z` are the only characters to be checksummed.

So I would suggest that the VCF spec should refer to the **SAM spec** (not refget) and/or duplicate the salient information itself. (The duplicated information could be shorter: just the brief formal description together with a pointer directing readers to the SAM spec for examples and details, rather than duplicating the whole thing.)

The SAM and CRAM specs have a convention of referring to other specifications by title along with a footnote providing links and URLs. (See e.g. SAMv1.pdf footnotes 19, 23, and 37 and SAMtags.pdf footnotes 1 and 2.)

jmarshall:
VCFv4.3 §1.4.7 “Contig field format” refers to MD5 checksums but doesn't tell you how they're supposed to be specified in the `##contig` header. You have to look at non-normative examples elsewhere in the specification to figure out that the tag to use is probably `md5=<hash>`.

tcezard:
@tskir @jmarshall Thank you for the review.
I've copied some detail of the SAM specification and explicitly named the attributes as they were in the examples. 
I stopped short of describing other attributes (assembly, name) that are commonly used because they were not described before and I didn't to add more information, only clarify what was already there. 
This said I think we should consider if other attributes should be described in a separate PR.

VCFv4.3.tex
```diff
@@ -219,10 +219,21 @@ \subsubsection{Assembly field format}
 \subsubsection{Contig field format}
 \label{sec-contig-field}
 It is recommended for VCF, and required for BCF, that the header includes tags describing the contigs referred to in the file.
-The structured \texttt{contig} field must include the ID attribute and typically includes also sequence length, MD5 checksum, URL tag to indicate where the sequence can be found, etc.
+The structured \texttt{contig} field must include the ID attribute and can includes additional optional attributes with
+the following ones reserved:
+\begin{itemize}
+  \item length: the length of the sequence
+  \item md5: MD5 checksum of the sequence as defined in the Sam specification v1\footnote{See Reference MD5 calculation
+      section in \href{https://github.com/samtools/hts-specs/blob/master/SAMv1.pdf}{\tt SAMv1.pdf}.} Briefly, the digest
```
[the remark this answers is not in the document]
tcezard: Should I use the `https` instead ?

VCFv4.3.tex
```diff
@@ -219,10 +219,21 @@ \subsubsection{Assembly field format}
...
+      is calculated excluding all characters outside of the inclusive range 33 (`\char33') to 126 (`\char126')
+      and all lowercase characters converted to uppercase. The MD5 digest is calculated as described in
+      \href{https://tools.ietf.org/html/rfc1321}{\sl RFC 1321} and presented as a 32 character lowercase hexadecimal number.
+  \item URL: tag to indicate where the sequence can be found
+\end{itemize}
+
 For example:
 \begin{verbatim}
-##contig=<ID=ctg1,length=81195210,URL=ftp://somewhere.org/assembly.fa,...>
+##contig=<ID=ctg1,length=81195210,URL=ftp://somewhere.org/assembly.fa,md5=f126cdf8a6e0c7f379d618ff66beb2da,...>
```
jmarshall, line 236: This now disappears off the edge of the page.
(Also I'd suggest taking the opportunity to change the URL to use [`example.org`](https://en.wikipedia.org/wiki/Example.com).)

VCFv4.3.tex
```diff
@@ -219,10 +219,21 @@ \subsubsection{Assembly field format}
 \subsubsection{Contig field format}
 \label{sec-contig-field}
 It is recommended for VCF, and required for BCF, that the header includes tags describing the contigs referred to in the file.
-The structured \texttt{contig} field must include the ID attribute and typically includes also sequence length, MD5 checksum, URL tag to indicate where the sequence can be found, etc.
+The structured \texttt{contig} field must include the ID attribute and can includes additional optional attributes with
+the following ones reserved:
+\begin{itemize}
+  \item length: the length of the sequence
+  \item md5: MD5 checksum of the sequence as defined in the Sam specification v1\footnote{See Reference MD5 calculation
```
jmarshall, line 226:
```suggestion
  \item md5: MD5 checksum of the sequence as defined in SAM's {\tt @SQ-M5} field\footnote{See Reference MD5 calculation
```

jmarshall:
We think most problems with this have been addressed. Agreed at Feb 1st meeting that this should be merged on 2022-02-08 if no further issues are raised.

jmarshall:
The one week time limit suggested at the last meeting has long since expired, but only VCF maintainers are really authorised to push to _VCFv*.tex_. @lbergelson as the only active VCF maintainer, could you please either (a) review this PR; (b) merge it as per the suggested time limit having ended without further comments; or (c) authorise (by commenting here) someone else (e.g. me) to merge it on the VCF maintainers' behalf.

Issue #569, opened 2021-05-19
Switch from `rerun.sh` to `latexmk`


michaelmhoffman:
The `latexmk` command that comes with TeX Live seems to fulfill the same purpose as `rerun.sh` and is robust to a lot more edge cases, and probably quite a bit more future-proof.

michaelmhoffman:
For example, `rerun.sh` does not seem to rerun TeX enough times for the acronym package to work.

jmarshall:
Various maintainers have been reluctant to use `latexmk` in the past, because they did not have it installed. It may well be time to revisit this… 🤷

Starting from an empty `new/` directory, both `rerun.sh` and `latexmk` run latex the same number of times (namely 3) and produce the same acronym-related warnings on the final run:

```
LaTeX Warning: Hyper reference `acro:ASCII' on page 6 undefined on input line 374.
LaTeX Warning: Hyper reference `acro:ASCII' on page 6 undefined on input line 374.
LaTeX Warning: Hyper reference `acro:regex' on page 6 undefined on input line 377.
LaTeX Warning: Hyper reference `acro:regex' on page 6 undefined on input line 377.
```

Is this what you're seeing, or do you experience a scenario in which `rerun.sh` and `latexmk` produce different results?

I suspect those warnings are some unfortunate interplay between hyperrefs and acronym commands…

michaelmhoffman:
After reverting commit c773dcf93830b05d38b21e6f65887c9d07a49d17 to restore the `changepage` package (in an incorrect attempt to get this working on the CircleCI system that doesn't have `changepage`), `rerun.sh` is now working for me. I don't know why it wasn't working before—it's probably either because I was using it correctly or something else.

Will figure out how to remove those warnings for #570.

I'd still suggest using `latexmk` for the reasons above but it might not be a blocker for #570.

jkbonfield:
FYI latexmk seems to work for me, assuming I used it correctly.  Is there a specific document which you need comparing?

michaelmhoffman:
If you do this, you will also need to remove `new/*.fls` and `new/*.fdb_latexmk` as part of the `mostlyclean` target, as done in #570.

jmarshall:
Assuming there are no maintainers or contributors who still do not have access to `latexmk`, there is really no good reason not to be using it, as it's the standard well-known tool for this job.

[...]

PR #570, opened 2021-05-19
Add BEDv1 specification

michaelmhoffman:
Add a Browser Extensible Data (BED) specification.

- [BEDv1.pdf](https://github.com/michaelmhoffman/hts-specs/releases/download/20210923/BEDv1.pdf) draft 2021-09-23 revision 2481c79
- ~[BEDv1.pdf](https://github.com/michaelmhoffman/hts-specs/releases/download/20210831/BEDv1.pdf) draft 2021-08-31 revision 2a5cd5d~
- ~[BEDv1.pdf](https://github.com/michaelmhoffman/hts-specs/releases/download/20210830/BEDv1.pdf) draft 2021-08-30 revision af943fe~
- ~[BEDv1.pdf](https://github.com/michaelmhoffman/hts-specs/releases/download/20210628/BEDv1.pdf) draft 2021-06-28 revision 1caa7c6~

### DIFF
--- a/BEDv1.tex
+++ b/BEDv1.tex
@@ -1,0 +1,393 @@
+\documentclass[11pt]{article}
+\usepackage[T1]{fontenc}
+
+\usepackage[letterpaper,margin=1in]{geometry}
+
+\usepackage{acronym}
+\usepackage{amsmath}
+\usepackage{booktabs}
+\usepackage[flushmargin,hang]{footmisc}
+\usepackage{microtype}
+\usepackage{newverbs}
+\usepackage{tablefootnote}
+\usepackage{tabularx}
+\usepackage{todonotes}
+\usepackage[hyperfootnotes=false]{hyperref} % doesn't work in tabulars as currently set
+\usepackage{footnotehyper}
+\usepackage[strict]{changepage}
+\usepackage[binary-units=true]{siunitx}
+\usepackage[mark]{gitinfo2}
+
+\hypersetup{colorlinks=true,
+  linkcolor=blue,
+  filecolor=magenta,
+  urlcolor=blue,
+  pdfinfo={githash=\gitHash}}
+
+\definecolor{cverbbg}{gray}{0.93}
+
+\title{The Browser Extensible Data~(BED) format}
+\author{Jeffrey Niu, Danielle Denisko, Michael M.~Hoffman}
+\date{\today}
+
+\setlength{\emergencystretch}{\hsize}
+\setlength{\footnotemargin}{1em}
+
+\interfootnotelinepenalty=1000000
+\makesavenoteenv{tabularx}
+
+\newcolumntype{L}{>{\raggedright\arraybackslash}X}
+
+% eliminate passive voice warnings
+% chktex-file -3
+
+\begin{document}
+\makeatletter
+\renewcommand*{\AC@hyperlink}[2]{#2} % do not hyperlink acronyms
+\makeatother
+
+
+\frenchspacing
+
+\maketitle
+
+\acused{ASCII}
+
+\section{Specification}
+
+\Ac{BED} is a whitespace-delimited file format, where each~\textbf{file} consists of one or more~\textbf{line}s.\footnote{``Frequently Asked Questions: Data File Formats.'' \ac{UCSC} Genome Browser FAQ, \url{https://genome.ucsc.edu/FAQ/FAQformat.html}}
+Each~\textbf{line} describes discrete genomic~\textbf{feature}s by physical start and end position on a linear~\textbf{chromosome}.
+The file extension for the \ac{BED} format is~\texttt{.bed}.
+
+\subsection{Typographic conventions}
+
+This document uses the following typographic conventions:
+
+\vspace{2ex}
+
+\noindent
+\begin{tabularx}{\textwidth}{r L L}
+  \toprule
+  Style & Meaning & Examples \\
+  \midrule
+  Bold & Terms defined in subsections~\ref{sec:terms}--\ref{sec:lines} & \textbf{chromosome}{\quad}\textbf{file} \\
+  Sans serif & Names of~\textbf{field}s & \textsf{chrom}{\quad}\textsf{chromStart}{\quad}\textsf{chromEnd} \\
+  Fixed-width & Literals or \ac{regex}es\footnote{POSIX/IEEE~1003.1--2017 Extended Regular Expressions, for the ``C'' locale.
+                \emph{IEEE Standard for Information Technology---Portable Operating System Interface~(POSIX) Base Specifications}, IEEE~1003.1--2017, 2017} & \texttt{.bed}{\quad}\texttt{grep}{\quad}\texttt{[[:alnum:]]+}{\quad}\texttt{ATCG} \\
+  \bottomrule
+\end{tabularx}
+
+\subsection{Terminology and concepts}\label{sec:terms}
+\begin{description}
+\item[0-start, half-open coordinate system:]
+  A coordinate system where the first base starts at position~0, and the start of the interval is included but the end is not.
+  For example, for a sequence of bases~\texttt{ACTGCG}, the bases given by the interval~[2,~4) are~\texttt{TG}. % chktex 9
+
+\item[BED$n$:]
+  A~\textbf{file} with the first $n$~\textbf{field}s of the \ac{BED} format.
+  For example, \textbf{BED3}~means a~\textbf{file} with only the first 3~\textbf{field}s; \textbf{BED12}~means a~\textbf{file} with all 12~\textbf{field}s.
+
+\item[BED$n$+:]
+  A~\textbf{file} that has $n$~\textbf{field}s of the \ac{BED} format, followed by any number of~\textbf{field}s of custom data defined by a user.
+
+\item[BED$n$+$m$:]
+  A~\textbf{file} that has a custom tab-delimited format starting with the first $n$~\textbf{field}s of the \ac{BED} format, followed by $m$~\textbf{field}s of custom data defined by a user.
+  For example, \textbf{BED6+4}~means a~\textbf{file} with the first 6~\textbf{field}s of the \ac{BED} format, followed by 4~user-defined~\textbf{field}s.
+
+\item[block:]
+  Linear subfeatures within a~\textbf{feature}.
+  Usually used to designate exons.
+
+\item[chromosome:]
+  A sequence of nucleobases with a name.
+  In this specification, ``chromosome'' may also describe a named scaffold that does not fit the biological definition of a chromosome.
+  Often, chromosomes are numbered starting from~\texttt{1}.
+  There are also often sex chromosomes such as~\texttt{W}, \texttt{X}, \texttt{Y}, and~\texttt{Z}, mitochondrial chromosomes such as~\texttt{M}, and possibly scaffolds from an unknown chromosome, often labeled~\texttt{Un}.
+  The name of each chromosome is often prefixed with~\texttt{chr}.
+  Examples of chromosome names include~\texttt{chr1}, \texttt{21}, \texttt{chrX}, \texttt{chrM}, \texttt{chrUn}, \texttt{chr19\_KI270914v1\_alt}, and~\texttt{chrUn\_KI270435v1}.
+
+\item[feature:]
+  A linear region of a~\textbf{chromosome} with specified properties.
+  For example, a~\textbf{file}'s~\textbf{feature}s might all be peaks called from ChIP-seq data, or transcript.
+
+\item[field:]
+  Data stored as non-tab text.
+  All~\textbf{field}s are 7-bit US \ac{ASCII}.
+
+\item[file:]
+  Sequence of one or more~\textbf{line}s.
+
+\item[line:]
+  String terminated by a~\textbf{line separator}, in one of the following classes.
+  Either a~\textbf{data line}, a~\textbf{comment line}, or a~\textbf{blank line}.
+  Discussed more fully in~\autoref{sec:lines}
+
+\item[line separator:]
+  Either carriage return, line feed, or carriage return followed by line feed.
+  The same \textbf{line separator} must be used throughout the \textbf{file}.
+\end{description}
+
+\subsection{Lines}\label{sec:lines}
+
+\subsubsection{Data lines}
+
+Data lines contain \textbf{feature}~information.
+A data line is composed of~\textbf{field}s separated by whitespace.
+The whitespace must match the \ac{regex}~\texttt{[[:space:]]+}\footnote{\texttt{[[:space:]]} includes the following characters: space, form-feed, newline, carriage-return, tab, and vertical-tab}.
+
+\subsubsection{Comment lines and blank lines}
+
+Both comment lines and blank lines provide no~\textbf{feature} data.
+
+Comment lines start with~\texttt{\#} with no whitespace beforehand.
+A~\texttt{\#} appearing anywhere else in a line is treated as~\textbf{feature} data, not a comment.
+
+Blank lines consist entirely of whitespace.
+Both comment and blank lines may appear as any line in a~\textbf{file}, at the beginning, middle, or end of the file.
+They may appear in any quantity.
+
+\subsection{\acs{BED} fields}
+
+Each~\textbf{data line} contains between~3 and 12~whitespace-delimited~\textbf{field}s.
+The first 3~\textbf{field}s are mandatory, and the last 9~\textbf{field}s are optional.
+In optional~\textbf{field}s, the order is binding---if 1~\textbf{field} is filled, then all previous~\textbf{field}s must also be filled.
+However, \textbf{BED10} and \textbf{BED11} are prohibited.
+
+In a \ac{BED}~\textbf{file}, each~\textbf{data line} must have the same number of~\textbf{field}s.
+The positions in \ac{BED}~\textbf{field}s are all described in the~\textbf{0-based, half-open coordinate system}.
+
+\begin{adjustwidth}{-0.5in}{-0.5in}
+  \noindent
+  \begin{tabularx}{\linewidth}{r l l l L}
+    \toprule
+    Col & Field & Type & Regex or range & Brief description \\
+    \midrule
+    1 & \textsf{chrom} & String & \texttt{[[:alnum:]\_]\{1,255\}}{\footnotemark} & \textbf{Chromosome} name \\
+    2 & \textsf{chromStart} & Int & $[0, 2^{64}-1]$ & \textbf{Feature} start position \\
+    3 & \textsf{chromEnd} & Int & $[0, 2^{64} -1]$ & \textbf{Feature} end position \\
+    4 & \textsf{name} & String & \texttt{[{\textasciicircum}{\textbackslash}t]\{0,255\}} & \textbf{Feature} description \\
+    5 & \textsf{score} & Int & $[0, 1000]$ & A numerical value \\
+    6 & \textsf{strand} & String & \texttt{[-+.]} & \textbf{Feature} strand \\
+    7 & \textsf{thickStart} & Int & $[0, 2^{64}-1]$ & Thick start position \\
+    8 & \textsf{thickEnd} & Int & $[0, 2^{64}-1]$ & Thick end position \\
+    9 & \textsf{itemRgb} & Int,Int,Int & \texttt{(}$[0, 255], [0,255], [0,255]$\texttt{) | 0} & Display color \\ % chktex 9
+    10 & \textsf{blockCount} & Int & $[0, \textsf{chromEnd}-\textsf{chromStart}]${\footnotemark} & Number of \textbf{block}s \\
+    11 & \textsf{blockSizes} & List[Int] & \texttt{([[:digit:]]+,)\{\textsf{blockCount}$-1$\}[[:digit:]]+,?}{\footnotemark} & \textbf{Block} sizes \\
+    12 & \textsf{blockStarts} & List[Int] & \texttt{([[:digit:]]+,)\{\textsf{blockCount}$-1$\}[[:digit:]]+,?} & \textbf{Block} start positions \\
+    \bottomrule
+  \end{tabularx}\label{sec:table}
+  \footnotetext[5]{\texttt{[[:alnum:]\_]} is equivalent to the \ac{regex} \texttt{[A-Za-z0-9\_]}. % chktex 8
+      It is also equivalent to the Perl extension \texttt{[[:word:]]}.}
+  \footnotetext[6]{\textsf{chromEnd}-\textsf{chromStart} is the maximum number of~\textbf{block}s that may exist without overlaps.}
+  \footnotetext{For example, if~$\textsf{blockCount} = 4$, then the allowed \ac{regex} would be~\texttt{([[:digit:]]+,)\{3\}[[:digit:]]+,?}}
+\end{adjustwidth}
+
+\subsection{Coordinates}
+\begin{enumerate}
+\item \textsf{chrom}: The name of the~\textbf{chromosome} or scaffold where the~\textbf{feature} is present.
+  Limiting only to word characters only, instead of all non-whitespace characters, makes \ac{BED}~\textbf{file}s more portable to varying environments which may make different assumptions about allowed characters.
+  The name must be between~1 and 255~characters long, inclusive.
+
+\item \textsf{chromStart}: Start position of the~\textbf{feature} on the~\textbf{chromosome} or scaffold.
+  \textsf{chromStart}~must be an integer greater than or equal to~0 and less than the total number of bases of the~\textbf{chromosome} to which it belongs.
+  If the size of the~\textbf{chromosome} is unknown, then \textsf{chromStart}~must be less than or equal to~$2^{64} - 1$, which is the maximum size of an unsigned 64-bit integer.
+
+\item \textsf{chromEnd}: End position of the~\textbf{feature} on the~\textbf{chromosome} or scaffold.
+  \textsf{chromEnd}~must be an integer greater than or equal to the value of~\textsf{chromStart} and less than or equal to the total number of bases in the~\textbf{chromosome} to which it belongs.
+  If the size of the~\textbf{chromosome} is unknown, then \textsf{chromEnd}~must be less than or equal to~$2^{64} - 1$, the maximum size of an unsigned 64-bit integer.
+\end{enumerate}
+
+\subsection{Simple attributes}
+\begin{enumerate}
+  \setcounter{enumi}{3}
+
+\item \textsf{name}: String that describes the~\textbf{feature}.
+  The name must be~0 to 255~non-tab characters.
+  The name must not be empty or contain whitespace, unless all fields in file are delimited exclusively using single tab characters.
+  A visual representation of the \ac{BED} format may display the name next to the~\textbf{feature}.
+
+\item \textsf{score}: Integer between~0 and~1000, inclusive.
+  If the~\textbf{feature} has no score information, then~\texttt{0} should be used as a default value.
+  A visual representation of the \ac{BED} format may shade features differently depending on their score.
+
+\item \textsf{strand}: Strand that the~\textbf{feature} appears on.
+  The strand may either refer to the~\texttt{+}~(sense or coding) strand or the~\texttt{-}~(antisense or complementary) strand.
+  If the~\textbf{feature} has no strand information or unknown strand, then a dot~(\texttt{.}) must be used.
+\end{enumerate}
+
+\subsection{Display attributes}
+\begin{enumerate}
+  \setcounter{enumi}{6}
+
+\item \textsf{thickStart}: Start position at which the~\textbf{feature} is visualized with a thicker or accented display.
+  This value must be an integer between~\textsf{chromStart} and~\textsf{chromEnd}, inclusive.
+  There is no specified default value for~\textsf{thickStart}.
+
+\item \textsf{thickEnd}: End position at which the~\textbf{feature} is visualized with a thicker or accented display.
+  This value must be an integer greater than or equal to~\textsf{thickStart} and less than or equal to~\textsf{chromEnd}, inclusive.
+  In \ac{BED} files with fewer than 7~\textbf{field}s, the whole~\textbf{feature} has thick display.
+  In \textbf{BED7+}~files, to achieve the same effect, set \textsf{thickStart}~equal to~\textsf{chromStart} and \textsf{thickEnd}~equal to~\textsf{chromEnd}.
+  If this~\textbf{field} is not specified but \textsf{thickStart}~is, then the entire~\textbf{feature} has thick display.
+  There is no specified default value for~\textsf{thickEnd}.
+
+\item \textsf{itemRgb}: A triple of integers that determines the color of this~\textbf{feature} when visualized.
+  The triple is three integers separated by commas.
+  Each integer is between~0 and~255, inclusive.
+  To make a~\textbf{feature} black, \textsf{itemRgb}~may be a single~\texttt{0}, which is visualized identically to a~\textbf{feature} with \textsf{itemRgb} of \texttt{0,0,0}.
+\end{enumerate}
+
+\subsection{Blocks}
+\begin{enumerate}
+  \setcounter{enumi}{9}
+
+\item \textsf{blockCount}: Number of~\textbf{block}s in the~\textbf{feature}.
+  \textsf{blockCount}~must be an integer greater than 0.
+  \textsf{blockCount}~is mandatory in~\textbf{BED12+}~files.
+  Null or empty~\textsf{blockCount} are not allowed, because~\textsf{blockSizes} and~\textsf{blockStarts} rely on~\textsf{blockCount}.
+  A visual representation of the \ac{BED} format may have blocks appear thicker than the rest of the~\textbf{feature}.
+
+\item \textsf{blockSizes}: Comma-separated list of length~\textsf{blockCount} containing the size of each~\textbf{block}.
+  There must be no spaces before or after commas.
+  There may be a trailing comma after the last element of the list.
+  \textsf{blockSizes}~is mandatory in \textbf{BED12+} files.
+  Null or empty~\textsf{blockSizes} is not allowed, because \textsf{blockStarts}~cannot be verified without~\textsf{blockSizes}.
+
+\item \textsf{blockStarts}: Comma-separated list of length~\textsf{blockCount} containing each \textbf{block}'s~start position, relative to~\textsf{chromStart}.
+  There must not be spaces before or after the commas.
+  There may be a trailing comma after the last element of the list.
+  Each element in~\textsf{blockStarts} is paired with the corresponding element in~\textsf{blockSizes}.
+  Each \textsf{blockStarts}~element must be an integer between~0 and~$\textsf{chromEnd} - \textsf{chromStart}$, inclusive.
+  For each couple~$i$ of~$(\textsf{blockStarts}_i, \textsf{blockSizes}_i)$, the quantity~$\textsf{chromStart} + \textsf{blockStarts}_i + \textsf{blockSizes}_i$ must be less or equal to \textsf{chromEnd}.
+  These conditions enforce that each~\textbf{block} is contained within the~\textbf{feature}.
+  The first~\textbf{block} must start at~\textsf{chromStart} and the last~\textbf{block} must end at~\textsf{chromEnd}.
+  Moreover, the~\textbf{block}s must not overlap.
+  The list must be sorted in ascending order.
+  \textsf{blockStarts}~is mandatory in~\textbf{BED12+} files.
+  Null or empty~\textsf{blockStarts} is not allowed.
+\end{enumerate}
+
+\section{Examples}
+
+\subsection[title]{Example BED6 file from the \acs{UCSC} Genome Browser FAQ\footnote{``Frequently
+    Asked Questions: Data File Formats.'' \ac{UCSC} Genome Browser FAQ,
+    \url{https://genome.ucsc.edu/FAQ/FAQformat.html}}}\label{sec:example-bed6}
+
+\begin{verbatim}
+chr7  127471196  127472363  Pos1  0  +
+chr7  127472363  127473530  Pos2  0  +
+chr7  127473530  127474697  Pos3  0  +
+chr7  127474697  127475864  Pos4  0  +
+chr7  127475864  127477031  Neg1  0  -
+chr7  127477031  127478198  Neg2  0  -
+chr7  127478198  127479365  Neg3  0  -
+chr7  127479365  127480532  Pos5  0  +
+chr7  127480532  127481699  Neg4  0  -
+\end{verbatim}
+
+\subsection{Example BED12 file from the \acs{UCSC} Genome Browser FAQ}
+\begin{verbatim}
+chr22 1000 5000 cloneA 960 + 1000 5000 0 2 567,488, 0,3512
+chr22 2000 6000 cloneB 900 - 2000 6000 0 2 433,399, 0,3601
+\end{verbatim}
+
+The~\textbf{block}s in this example satisfy the required constraints.
+The first~\textbf{block} starts at~\textsf{chromStart} since the first~\textsf{blockStarts} element is~0.
+The last~\textbf{block} ends at~\textsf{chromEnd} since the last~\textbf{block} starts at position 4512~(1000+3512) with size~488, and therefore ends at position 5000~(4512+488).
+
+\section{Recommended practice for the \acs{BED} format}
+
+\subsection{Mandatory fields}
+\begin{itemize}
+\item \textsf{chrom}: The name of each~\textbf{chromosome} should also match the names from a reference genome, if applicable.
+  For example, in the human genome, the chromosomes may be named~\texttt{chr1} to \texttt{chr22}, \texttt{chrX}, \texttt{chrY}, and~\texttt{chrM}.
+  Names should be consistent within a~\textbf{file}.
+  For example, one should not use both~\texttt{17} and~\texttt{chr17} to represent the same~\textbf{chromosome} in the same~\textbf{file}.
+\end{itemize}
+
+\subsection{Optional fields}\label{sec:optional}
+\begin{itemize}
+\item \textsf{name}: If a feature has no name, then a dot~(\texttt{.}) should be used.
+  Names should avoid using the space character even if the file is exclusively delimited with single tab characters because parsers may interpret a space as a delimiter.
+
+\item \textsf{itemRgb}: Eight or fewer colors should be used as too many colors may slow down visualizations and are difficult for humans to distinguish.\footnote{``Frequently
+    Asked Questions: Data File Formats.'' \ac{UCSC} Genome Browser FAQ,
+    \url{https://genome.ucsc.edu/FAQ/FAQformat.html}}
+
+\end{itemize}
+
+\subsection{User-defined fields}
+
+Custom data \textbf{fields} may contain any non-tab 7-bit US \ac{ASCII} character.
+Definitions of a custom \ac{BED} format should restrict the type of each \textbf{field} to the extent possible.
+Each custom \textbf{field} should contain either one of the following data types or a comma-separated list of values of the same type:
+
+\noindent
+\begin{tabularx}{\textwidth}{r L}
+  \toprule
+  Type & Definition \\
+  \midrule
+  Integer & String representation of 64-bit signed integer\footnote{\emph{IEEE 754--1985 IEEE Standard for Binary Floating-Point Arithmetic.} IEEE 754--1985, 1985} \\
+  Unsigned & String representation of 64-bit unsigned integer\footnotemark[10] \\
+  Float & String representation of 64-bit floating point number\footnotemark[10] \\
+  Character & One character, other than tab \\
+  String & One or more characters, other than tab \\
+  \bottomrule
+\end{tabularx}
+
+This specification does not contain a means for interchanging custom \ac{BED} format definitions.
+The AutoSQL format\footnote{Kent, W.~James.
+  (2000) ``AutoSQL.''
+  \url{https://hgwdev.gi.ucsc.edu/~kent/exe/doc/autoSql.doc}} provides one method for defining custom \ac{BED} formats in a separate file.
+
+\subsection{Sorting}
+\Ac{BED} \textbf{file}s should be sorted by~\textsf{chrom}, then by~\textsf{chromStart} numerically, and finally by~\textsf{chromEnd} numerically.
+\textsf{chrom} may be sorted using any scheme (such as lexicographic or numeric order), but all lines with the same~\textsf{chrom} value should occur consecutively.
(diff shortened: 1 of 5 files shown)


michaelmhoffman:
This fails the CircleCI build because I am using the `footnotehyper` package to get footnotes to work well in a table. I'm able to build all PDFs fine on my Ubuntu 20.10 system (which includes TeX Live 2020). It's not present on the Ubuntu 18.04.5 (Tex Live 2015) container used on CircleCI. I was using Ubuntu 20.04 LTS until recently and I think that would also work if you want to stick to an LTS version. Would you be willing to upgrade?

jmarshall:
The existing documents try to be somewhat conservative in their package requirements. It's no so much about what is available on the CI's installation but what is available on maintainers' and interested contributors' installations, and avoiding making it unnecessarily difficult for people to build the PDFs locally. e.g. @jkbonfield tends to have an elderly installation :smile: — James, do you have `footnotehyper` available?

jmarshall:
Re gitinfo2: on the inside, this hook script and `genversion.sh` are very similar. The significant difference is that `genversion.sh` identifies the most recent commit affecting the particular document's source files (47aaedf6b7bd3a033191b68433ab4b6e761882ce), while gitinfo2 identifies the most recent commit overall (i.e., `HEAD`). This is less than great in a repository containing more than one specification: regenerated gitinfo2-using PDFs will have spurious differences in their recorded git hashes and datestamps due to unrelated changes in other documents.

Hence I'd recommend just going with the existing `genversion.sh` infrastructure.

[BTW on my complete and fully updated TeX Live (macTeX) 2021 installation, I got `Environment adjustwidth undefined`. Is this missing a `\usepackage[strict]{changepage}`?]

jkbonfield:
This fails to build with me:

```
------------
Running 'lualatex  -recorder -output-directory="new"  "BEDv1.tex"'
------------
This is LuaTeX, Version 1.0.4 (TeX Live 2017/Debian) 
 restricted system commands enabled.
(./BEDv1.tex
LaTeX2e <2017-04-15>
(using write cache: /nfs/users/nfs_j/jkb/.texlive2017/texmf-var/luatex-cache/ge
neric)(using read cache: /var/lib/texmf/luatex-cache/generic /nfs/users/nfs_j/j
kb/.texlive2017/texmf-var/luatex-cache/generic)
luaotfload | main : initialization completed in 0.160 seconds
Babel <3.18> and hyphenation patterns for 1 language(s) loaded.
(/usr/share/texlive/texmf-dist/tex/latex/base/article.cls
Document Class: article 2014/09/29 v1.4h Standard LaTeX document class
(/usr/share/texlive/texmf-dist/tex/latex/base/size11.clo
luaotfload | db : Font names database not found, generating new one.
luaotfload | db : This can take several minutes; please be patient.(save: /nfs/users/nfs_j
/jkb/.texlive2017/texmf-var/luatex-cache/generic/fonts/otl/lmroman10-regular.lu
a)(save: /nfs/users/nfs_j/jkb/.texlive2017/texmf-var/luatex-cache/generic/fonts
/otl/lmroman10-regular.luc)))
(/usr/share/texlive/texmf-dist/tex/latex/base/fontenc.sty
(/usr/share/texlive/texmf-dist/tex/latex/base/t1enc.def)
(/usr/share/texmf/tex/latex/lm/t1lmr.fd))
(/usr/share/texlive/texmf-dist/tex/latex/geometry/geometry.sty

... snip ...

(/usr/share/texlive/texmf-dist/tex/latex/siunitx/siunitx-abbreviations.cfg)
(/usr/share/texlive/texmf-dist/tex/latex/siunitx/siunitx-binary.cfg)
! Undefined control sequence.
l.57 \Ac
      {BED} is a whitespace-delimited file format, where each~\textbf{file} ...
```

I get the same failure with pdflatex too though so it's not a lualatex thing.

Is this some capitalization issue?  I see `\ac{foo}` everywhere except at the start of sentences where it is `\Ac{foo}`, yet in the file the only definition I see is for '\AC'.  I thought latex was case sensitive so I don't understand this.

Replacing them with the lowercase version gives me a PDF that builds, but neither xpdf nor evince can open it.  Doing a make clean and rebuilding then gave me an error bizarrely, so I'm really not sure what's going on.

```
! LaTeX Error: Environment adjustwidth undefined.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.158 \begin{adjustwidth}
                       {-0.5in}{-0.5in}
```


I'm also getting gitinfo warnings:

```
Package gitinfo2 Warning: I can't find the file '.git/gitHeadInfo.gin'.
(gitinfo2)                All git metadata has been set to '(None)'.

```

This is all on an Ubuntu 18.04 machine with the standard installed latex packages.  I don't have admin rights on any of our machines, but can ask to get packages installed.

jkbonfield:
I just noticed John's comment - yes adding `\usepackage[strict]{changepage}` cures it.  Evince can even view it now. :-)

jmarshall:
`\Ac` is a for-the-start-of-sentences variant of `\ac`, added in acronym 1.42 in Nov 2019. Presumably your TeX installation's edition of _acronym.sty_ is too old to have it.

The gitinfo warnings are because you haven't installed the hook into your git repository or otherwise run the `post-xxx-sample` script. I don't think we should be requiring people to know how to set up repository hooks, but I left that out of my previous comment because I thought I had already flogged the gitinfo horse quite a bit…

jkbonfield:
Agreed IMO we shouldn't use gitinfo for the same reasons you mentioned.  It's not as useful as the existing infrastructure there.

Our servers are running Ubuntu 18.04.5 (18.0 was released in 2018, with 18.04.5 in Aug 2020).  It's a Long Term Support release, supported up to 2023.   I could download a newer acronym.sty and install locally, but IMO if we have dependencies on specific versions then we should include them in the hts-specs respository itself so it builds cleanly on all "current" OS distributions, or provide a "bootstrap.sh" or similar that downloads and caches them automatically.

However in this case, the only time that command was used was on BED and UCSC acronyms, both starting with a capital letter, so we'd be better off just using the lowercase variant to avoid the recency dependency.

michaelmhoffman:
My plans:

- [ ] re-add the `changepage` package (removed in an incorrect attempt to get this working on CircleCI)
- [ ] switch from using `gitinfo2` to using `genversion.sh` for revision info
- [ ] remove use of `\Ac` for capitalization
- [ ] fix the `acronym` missing references issue described by @jmarshall in #569.

I have added some strikethrough styling to the pull request description to reflect planned changes.

jmarshall:
Re the `\Ac` thing, you could add something like the following to your preamble and carry on using both `\Ac` and `\ac` as appropriate to keep your linter happy:

```latex
% work around outdated acronym.sty packages
\providecommand*{\Ac}[1]{\ac{#1}}
```

(TIL about `\providecommand` — very handy in just this circumstance!)

michaelmhoffman:
I believe I've fixed the discussed issues. The CircleCI build image still doesn't have `footnotehyper` though.

```
! LaTeX Error: File `footnotehyper.sty' not found.

Type X to quit or <RETURN> to proceed,
or enter new name. (Default extension: sty)

Enter file name: 
! Emergency stop.
<read *> 
   
l.17 \usepackage
              [strict]{changepage}
 264 words of node memory still in use:
   2 hlist, 1 rule, 42 glue_spec, 1 write, 1 dir, 1 pdf_colorstack nodes
   avail lists: 2:17,3:1,6:2,7:1,9:2,12:1
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on BEDv1.log.
Latexmk: Missing input file: 'footnotehyper.sty' from line
  '! LaTeX Error: File `footnotehyper.sty' not found.'
Latexmk: Errors, so I did not complete making targets
Collected error summary (may duplicate other messages):
  pdflatex: Command for 'pdflatex' gave return code 256
Latexmk: Use the -f option to force complete processing,
 unless error was exceeding maximum runs of latex/pdflatex.
Makefile:40: recipe for target 'new/BEDv1.pdf' failed
make: *** [new/BEDv1.pdf] Error 12

Exited with code exit status 2

CircleCI received exit code 2
```

@jkbonfield @jmarshall Do you want to test that this works on your local TeX installation?

Also, we have yet to get to diff generation and if the build is still broken there, I would consider setting things up so that building a latexdiff PDF does not result in a build failure. Latexdiff doesn't understand everything and there are occasionally things that require manual intervention.

jmarshall:
That builds locally for me, though the table of acronyms is badly formatted (I haven't investigated yet).

I have started looking into updating the CI installation, and have been having similar thoughts about having a borked latexdiff not cause complete failure…

michaelmhoffman:
The formatting of the acronym list in the revised version is greatly improved.

simonbrent:
The spec document states:

> line separator: Either carriage return, line feed, or carriage return followed by line feed. The same
line separator must be used throughout the file.
...
1.3.1 Data lines
Data lines contain feature information. A data line is composed of fields separated by whitespace.
The whitespace must match the regex [[:space:]]+3
....
3
[[:space:]] includes the following characters: space, form-feed, newline, carriage-return, tab, and vertical-tab

It does not make sense that fields can be separated by newline or carriage-return, when those are the designated line separators.
I also think there should be consistency of language here: One refers to line feed and the other to newline (also "carriage return" and "carriage-return") - if the line separator must be one of "\r", "\n", or "\r\n", I think it should be stated as such, rather than leaving it a bit more open to interpretation, given that "line feed" can be taken to mean "\r\n", which would make "carriage return followed by line feed" = "\r\r\n"

simonbrent:
The spec states:

> chromEnd: End position of the feature on the chromosome or scaffold. chromEnd must be
an integer greater than or equal to the value of chromStart

In a 0-based, half-open coordinate system, shouldn't end be **greater** than start? Otherwise conversion to a 1-based system would result in a feature with a end greater than its start.

fkokocinski:
>  A data line is composed of fields separated by whitespace.

I know it would invalidate some existing files, but if we have the chance to specify tab "\t" as a separater going forward it would make the format much more stable, i.e. safer to parse, in my opinion.

pdl:
> Though lines may use any kind of whitespace as a delimiter between fields, a single tab (\t) should
be used. This is because almost all tools support tabs while some tools do not support other kinds
of whitespace. Also, whitespace within the name field may be used only if the field delimiter is tab
throughout the file.

How is a parser supposed to determine that "the field delimiter is tab throughout the file", if other white space is permitted as a separator?

andrewyatz:
> How is a parser supposed to determine that "the field delimiter is tab throughout the file", if other white space can be parsed as either a separator or as field content? This would lead to ambiguous parsing and also implies the entire file must be read before parsing begins.
> 

From experience, if you look at programs such as bedToBigBed there are explicit command line switches to enforce the use of tabs. Without that kind of change or some form of metadata payload in the bed file (out of scope as this spec is defined) then any whitespace must be assumed

JspSrs:
Regarding topic 3.2, on RGB color coding, I think a good point would be to ad an additional line advising to avoid a Red/Green color scheme.
Something like: _itemRgb: Eight or fewer colors should be used as too many colors may slow down visualizations
and are difficult for humans to distinguish.8_  Avoid using a simple Red/Green color scheme for positive and negative, but add Yellow or Blue, as in proper scientific color use.

The reason for this is the many colorblind individuals (mostly males, but including females). Off note, colorblindness is quite a spectrum of color nuances due to simple polymorphisms to seeing dar-red and dark-green as the same kind of greyish. And for many years avoiding simple Red/Green use is supported by important journals, including Nature (i.e see their tutorials on article writing and artwork)

A line on using transparency to reduce color intensity seems worthwhile also, when establishing a formal standard. Unfortunately, GA4GH/ClinGen uses Red for deletions and Blue for gains in copy number visualization, while Blue often stands for 'colder' and 'less', while Red often is 'more intens' and 'fire'.

JspSrs:
Note for lsg-coordinator@ga4gh.org: the domain is not accepting email. While this lsg-coordinator is the mailto link in the GA4GH call for Public Comment on the BED file format.

_ Hello j.saris@erasmusmc.nl,

_We're writing to let you know that the group you tried to contact (LSG-Coordinator) may not exist, or you may not have permission to post messages to the group. A few more details on why you weren't able to post:

 * You might have spelled or formatted the group name incorrectly.
 * The owner of the group may have removed this group.
 * You may need to join the group before receiving permission to post.
 * This group may not be open to posting.

If you have questions related to this or any other Google Group, visit the Help Center at https://eur01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fsupport.google.com%2Fa%2Fga4gh.org%2Fbin%2Ftopic.py%3Ftopic%3D25838&amp;data=04%7C01%7Cj.saris%40erasmusmc.nl%7C6a89f5c950b84cc4720408d92b59af03%7C526638ba6af34b0fa532a1a511f4ac80%7C0%7C1%7C637588486455819461%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C2000&amp;sdata=%2FCVTUGF3kZ40hlrRZ3Xy4Vio8qLmWdrF55kr0mAIuiI%3D&amp;reserved=0.

Thanks,

ga4gh.org admins__

michaelmhoffman:
> I know it would invalidate some existing files, but if we have the chance to specify tab "\t" as a separater going forward it would make the format much more stable, i.e. safer to parse, in my opinion.

@fkokocinski Thank you for your suggestion. I agree it would be nice, but we designed this version of the specification to formalize the existing BED description, which clearly allows other whitespace as a delimiter. There are probably many extant BED files, parsers, and generators that use non-tab whitespace, and it is undesirable for a "v1" specification to declare them non-conforming years after their creation according to a sensible reading of the previous description.

The "Recommended practice for the BED format" section *recommends* that a single tab be used as this presents many advantages.

michaelmhoffman:
@pdl Thank you for considering this specification so carefully. Some replies below:

> Getting a formal spec for BED would be great and it's really great to have this document as a basis for it. In order to achieve its purpose, the spec will need to address three concerns:
> 
>     1. How to represent features as BED
>     2. How to determine if a file is conformant
>     3. How to parse a conformant file
> 
> At the moment the spec feels like it's more descriptive than normative, i.e. that it is indicating the diversity in how information is represented in the files that currently exist and are interpreted by existing tooling, rather than establishing expectations for predictable behaviour which you'd need in order to write tooling in the future.

The spec is normative in that it clearly indicates that some files are valid and that others are invalid. I understand why you would say it feels more descriptive, in that, we are quite a bit more flexible about what is valid than what the drafter of a greenfield specification might allow. This is to avoid declaring files and implementations incompatible with this first version of the specification years after they were created, when those files and implementations were created according to reasonable interpretations of the previous BED description.

> The spec should absolutely be informed by a corpus of real files and hopefully this will be used to underpin the spec with a suite of tests as other specs in this repo are, but I suspect at least some existing documents might be found to be inconsistent with such a spec.

We have created test files for many corner cases as a separate project. We hope to release a preprint on it later this summer.

> > Though lines may use any kind of whitespace as a delimiter between fields, a single tab (\t) should
> > be used. This is because almost all tools support tabs while some tools do not support other kinds
> > of whitespace. Also, whitespace within the name field may be used only if the field delimiter is tab
> > throughout the file.
> 
> How is a parser supposed to determine that "the field delimiter is tab throughout the file", if other white space can be parsed as either a separator or as field content? This would lead to ambiguous parsing and also implies the entire file must be read before parsing begins.

We wrote the spec to indicate whether *files* are valid or not, and does not guarantee an unambiguous parsing. I know this is undesirable but it is difficult to avoid without declaring invalid existing files and implementations created under a reasonable interpretation of the previous description. The recommendation is to use tabs only for new files.

Parsers could infer that the delimiter is tab only from the first line, and can then report an error if they run into a line that cannot be parsed later. Unfortunately I believe this could lead to some BED files considered valid under this specification not being parsed. The only way to guarantee that this will happen while still enabling the advantages of using tabs is to use out-of-band information declaring that the only delimiter is tab, as @andrewyatz points out.

> Related to this, the handling of null values seems inconsistently specified and documented - this should be clearer:
> 
>     * It seems like the first three columns must be present and not empty, but I'm not sure this is explicit

The current text says `Each~\textbf{data line} contains between~3 and 12~whitespace-delimited~\textbf{field}s.
The first 3~\textbf{field}s are mandatory, and the last 9~\textbf{field}s are optional.` Do you think this is explicit enough?

>     * A null value for score it seems should be encoded as 0 - the 'default value' according to the spec. It's not clear if 0 is always supposed to be interpreted as being of null content, or whether 'default value' means that parsers should assume that empty string means 0. It's not clear whether scores are arbitrary or whether scores should be scaled to the range 1..1000, and if so whether features with a score of 1000 are the most 'prominent'.

Will address the default value issue better. The semantics of `score` are out-of-band information and not discussed in this spec.

michaelmhoffman:
Thank you everyone for all your helpful comments! @niujeffrey can you please draft the following changes by making a pull request against michaelmhoffman/hts-specs

- [ ] change the field separator to not include newline or carriage return ([thanks](#issuecomment-857452438) @simonbrent)
- [ ] "line feed" -> "newline" ([thanks](#issuecomment-857452438) @simonbrent)
- [ ] "carriage-return" -> "carriage return" ([thanks](#issuecomment-857452438) @simonbrent)
- [ ] "form-feed" -> "form feed"
- [ ] "vertical-tab" -> "vertical tab"
- [ ] define "newline" on first use as `'\n'` ([thanks](#issuecomment-857452438) @simonbrent)
- [ ] clarify that when `chromEnd` is equal to `chromStart` represents a zero-length feature, which is a feature between two bases such as an insertion. `chromStart=0`, `chromEnd=0` represents an insertion before the first nucleotide of a chromosome ([thanks](#issuecomment-857460372) @simonbrent)
- [ ] specify that fields, including custom fields, can only be empty when a single tab is used as the delimiter ([thanks](#issuecomment-857475922) @pdl)
- [ ] specify that field data is ASCII printable characters only---the range `'\x20'` to `'\x7e'` ([thanks](#issuecomment-857475922) @pdl)
- [ ] `name`: change regex so that it cannot be empty ([thanks](#issuecomment-857475922) @pdl)
- [ ] `score`: clarify that `0` should be used in BED5+ files where a `score` attribute of features would be uninformative ([thanks](#issuecomment-857475922) @pdl)
- [ ] `strand`:
  - [ ] clarify that `.` is the default value, and that a parser should treat BED5 files as if they have `strand=.`
  - [ ] explicitly specify that it cannot be empty in a BED6+ file ([thanks](#issuecomment-857475922) @pdl)
- [ ] `thickEnd`: clarify that the field is not specified but `thickStart` is means BED files that are not BED8+ ([thanks](#issuecomment-857475922) @pdl)
- [ ] change "null or empty" to only "empty" ([thanks](#issuecomment-857475922) @pdl)
- [ ] add recommendation to use colorblind-friendly color schemes, and especially to avoid red-green color schemes ([thanks](#issuecomment-857788414) @JspSrs)
- [ ] sorting:
  - [ ] add that arbitrary orderings of `chrom` are allowed as long as all lines with the same `chrom` value occur consecutively ([thanks](#issuecomment-857955521) @ZhenyuZ)
  - [ ] specify that multiple features with the same `chrom`, `chromStart`, and `chromEnd` may appear in any order
- [ ] add a section, before "UCSC track files", discussing information that is supplied out-of-band. This should include
  - [ ] which of the first 4-12 fields are standard BED fields and which are custom fields ([thanks](#issuecomment-857475922) @pdl)
  - [ ] genome assembly used
  - [ ] semantics of `score`, `itemRgb`, thick vs. thin positions, block vs. non-block positions
  - [ ] definitions of custom fields
  - [ ] whether tab is the only delimiter between fields ([thanks](#issuecomment-857475922) @pdl, [thanks](#issuecomment-857508028) @andrewyatz)
- [ ] add acknowledgments to any of the folks thanked in the above checklist, with full name and affiliation

pdl:
Thanks, @andrewyatz @michaelmhoffman @niujeffrey for indicating the required out-of-band information. All the changes proposed look great. I realise you don't want to innovate at this stage but in some version of the specification I hope it will be possible to standardise how to convey these things in-band (such as via a structured initial comment line) so that files can be self-contained.

michaelmhoffman:
Thanks @pdl, yes that is certainly something we have discussed. Having this specification makes it easier to identify both out-of-band information that we could formally specify and other warts and ambiguities that we could eliminate in a future specification. For example, one could imagine a "BEDX" specification which specifies files that are all valid BED v1, but also has structured comments with out-of-band info, single tab as the only allowed field separator, and other aspects where there are multiple choices now eliminated.

michaelmhoffman:
Thank you to all who have provided feedback. Getting this file to work on CircleCI is still in progress (see #576). I have used GitHub Releases to upload revised PDFs of:

- [BEDv1.pdf](https://github.com/michaelmhoffman/hts-specs/releases/download/20210628/BEDv1.pdf)
- [latexdiff from revision 6063151](https://github.com/michaelmhoffman/hts-specs/releases/download/20210628/BEDv1-diff-from-6063151.pdf), the version GA4GH sent out for public comment

latexdiff is not perfect and it has decided that all the tables have changed even where they haven't. Nonetheless the latexdiff should help show the differences from the last version.

michaelmhoffman:
On Twitter, @mschatz [suggests](https://twitter.com/mike_schatz/status/1409540127618211840) recommending a command for sorting. With recent versions of GNU coreutils, one could sort like this:

```
sort --key=1,1V --key=2,2n --key=3,3n
```

If one would prefer POSIX sort only, it would need to be

```
sort -k 1,1 -k 2,2n -k3,3n
```

which would lose the nice version sort behavior of GNU sort (which sorts `chr9` before `chr10` as humans might expect). Given that this is only a recommendation I would be fine with suggesting the more desirable behavior of GNU.

andrewyatz:
The [BigBED UCSC page](https://genome.ucsc.edu/goldenPath/help/bigBed.html) specifies `sort -k1,1 -k2,2n unsorted.bed > input.bed` but I don't know how important the order of sequence names is to BigBED or if it is just important to sort.

igordot:
For sorting, should locale be explicitly specified as well? One example of the issue that I am referring to: [Why doesn't sort sort the same on every machine?](https://stackoverflow.com/questions/28881/why-doesnt-sort-sort-the-same-on-every-machine).

According to the [`sort` man page](https://linux.die.net/man/1/sort): "Set `LC_ALL=C` to get the traditional sort order that uses native byte values".

mschatz:
That is a great suggestion- this has bitten me several times, plus setting
it to LC_ALL is often substantially faster

Mike

michaelmhoffman:
Good call. In that case:

```sh
LC_ALL=C sort --key=1,1V --key=2,3n
```

@andrewyatz In my experience, it is important for some parsing implementations that lines with the same `chrom` to be collected together, but the order can be arbitrary. Skimming the [bedToBigBed.c](https://github.com/ENCODE-DCC/kentUtils/blob/master/src/utils/bedToBigBed/bedToBigBed.c) seems to indicate the same there. So it shouldn't matter if it's `--key=1,1V --key=2,3n` or `--key=1,1 --key=2,2n`.

arq5x:
I would recommend:
    
    LC_ALL=C sort --key=1,1 --key=2,2n as this is what UCSC recommends and both `bedtools` and `bedops` expect as a default.

michaelmhoffman:
OK since we aren't using any GNU features anymore, might as well change to the POSIX

```sh
LC_ALL=C sort -k 1,1 -k 2,2n
```

simonbrent:
The current document states:
> field separator: One or more horizontal whitespace characters (space or tab). The field separator
must match the regex [ \\]+.

This is not a valid regex (i.e. `/[ \]+/` causes an error), and `/[ ]+/` does not match tab characters. I therefore conclude that you meant `[ \t]+`

simonbrent:
> Each data line contains between 3 and 12 fields delimited by a field separator. The first 3 fields
are mandatory, and the last 9 fields are optional. In optional fields, the order is binding—if 1 field
is filled, then all previous fields must also be filled. However, BED10 and BED11 are prohibited.

I suggest changing this to say "if a field is filled..." of "if one field is filled...", to avoid possible confusion between "1 field" and "field 1"

simonbrent:
> chrom: The name of the chromosome where the feature is present. Limiting only to word
characters only

Delete the either instance of "only"

simonbrent:
> name: String that describes the feature. name must be 1 to 255 non-tab characters. name
must not be empty or contain whitespace, unless the only field separator is a single tab.

This implies that the field separator can vary throughout the file, rather than, as I believe was the intention, that the field separator can be multiple whitespace characters.

I recommend:
1) In the field separator definition, state that the same field separator must be used throughout (like with line separator) 
2) Delete the word "only" in the definition for the name attribute

simonbrent:
In the strand definition:
> A parser should treat files that are not BED6+ as if strand were ..

could be better written as "Strand should be treated as . when parsing files that are not BED6+." in order to avoid the appearance of a typo due to a sentence ending ".."

simonbrent:
The itemRgb definition states that a single 0 is allowed for black. Is this a special case, or are other single-number values also valid (i.e. "123" is equivalent to "123,123,123")?

simonbrent:
> 3.5 Whitespace
Though data lines may use any kind of horizontal whitespace as a delimiter between fields, a
single tab (\t) should be used. This is because almost all tools support tabs while some tools do
not support other kinds of whitespace. Also, spaces within the name field may be used only if the
field delimiter is tab throughout the file.

"delimiter between fields" and "field delimiter" should be the defined term "field separator"

michaelmhoffman:
Thanks all for your additional comments! @niujeffrey can you please draft the following changes by making a pull request against michaelmhoffman/hts-specs called "Edits in resposne to public comments 2021-06-28 through 2021-06-29"

- [ ] Add recommended sort command `LC_ALL=C sort -k 1,1 -k 2,2n` ([thanks](https://twitter.com/mike_schatz/status/1409540127618211840) @mschatz, [thanks](#issuecomment-869824892) @andrewyatz, [thanks](#issuecomment-869845195) @igordot, [thanks](#issuecomment-869967147) @arq5x)
- [ ] fix the field separator regex ([thanks](#issuecomment-870356071) @simonbrent)
- [ ] define carriage return as `'\r'` ([thanks](#issuecomment-870356071) @simonbrent)
- [ ] change text to "if a field is filled" ([thanks](#issuecomment-870360052) @simonbrent)
- [ ] In "Limiting only to", delete "only" ([thanks](#issuecomment-870362311) @simonbrent)
- [ ] In the field separator definition, clarify that, unlike the line separator, the field separator can vary throughout the file, but that some capabilities of the BED format are available only if a single tab is always used as the field separator throughout the file, and this is recommended ([thanks](#issuecomment-870368101) @simonbrent)
- [ ] address issue with sentence ending ".." ([thanks](#issuecomment-870373198) @simonbrent)
- [ ] `itemRgb`: clarify 0 is a special case, the only case where a single-number value is valid ([thanks](#issuecomment-870376299) @simonbrent)
- [ ] replace "delimiter between field" and "field delimiter" with "field separator" ([thanks](#issuecomment-870383601) @simonbrent)
- [ ] add acknowledgments to any of the folks thanked in the above checklist, with full name and affiliation

pdl:
Having had another read through, I've spotted this:

> `name`: If a feature has no name, then a dot (`.`) should be used. Names should avoid using the
space character even if the only **field separator** is a single tab character, because parsers
may interpret a space as a **field separator**.

I **think** the intention is to say that "parsers might incorrectly interpret a space as a **field separator**", but a literal interpretation of the current wording (with the conventional meaning of 'may'), establishes the opposite, that it would be conformant behaviour for parsers to interpret a space as a field separator regardless of the out-of-band flag.

Appending "in the absence of out-of-band information about whether tab is the only **field separator** in the file" to the sentence would address this, I think.

alexpreynolds:
I have suggestions regarding this sort definition:

> **3.4 Sorting**
> BED files should be sorted by chrom, then by chromStart numerically, and finally by chromEnd
numerically. 

If this means that the BED file should be sorted on the end coordinate, where there are ties on the start coordinate, then the proposed a working `sort` command would perhaps instead be:

```
LC_ALL=C sort -k1,1 -k2,2n -k3,3n foo.bed
```

Consider the following input:

```
$ more foo.bed
chr1    1       2
chr4    7       8
chrX    100     101
chr11   9       100
chr11   9       99
chr20   11      12
chr2    3       4
chr3    5       6
```

Sorting only on the second column, as described in a previous comment, can result in the stop position being out of order (note `chr11`):

```
$ LC_ALL=C sort -k1,1 -k2,2n foo.bed
chr1	1	2
chr11	9	100
chr11	9	99
chr2	3	4
chr20	11	12
chr3	5	6
chr4	7	8
chrX	100	101
```

Sorting on the third column resolves this:

```
$ LC_ALL=C sort -k1,1 -k2,2n -k3,3n foo.bed
chr1	1	2
chr11	9	99
chr11	9	100
chr2	3	4
chr20	11	12
chr3	5	6
chr4	7	8
chrX	100	101
```

Or use tools like `sort-bed` that apply this sort:

```
$ sort-bed foo.bed
chr1	1	2
chr11	9	99
chr11	9	100
chr2	3	4
chr20	11	12
chr3	5	6
chr4	7	8
chrX	100	101
```

I would also suggest that the definition of sorting be made slightly more stringent, to deal with BED3+ or other four- or more column BED files.

Applying sort order on any additional fields in such a BED file is useful. Set and map operations on fully-sorted files will be deterministic. 

Without full sorting, running an operation on differently sorted files (or files with unknown sort order) could give two or more different answers, where there are matching intervals. The potential to get different answers can cause downstream problems in analysis, and applying a consistent and rigorous sort can be useful. The `sort-bed` tool applies this kind of ordering on input.

arq5x:
This is a nice catch, @alexpreynolds.  We should expand to:

    $ LC_ALL=C sort -k1,1 -k2,2n -k3,3n foo.bed

cmdcolin:
It might be worth noting for "Large BED files" that tabix is an option. I think both  tabix'ed BED files and bigBed are nice options for encoding large BED files. I am not sure if UCSC supports BED tabix files, but other browsers (e.g. JBrowse ) do support it.  Note: Using tabix requires using tab separators though.

Pros: BED tabix has the advantage that it's comparatively simpler to create also, since bigBed may require autoSql which is sort of an obscure. Furthermore, easier for analysis e.g. you can gunzip file.bed.gz and go rather than use bigBedToBed

Cons: BED tabix (and plaintext BED) itself are less self-describing than bigBed due to that same lack of autoSql, but users could make their own header format to try to help.

alexpreynolds:
More for raw compression than visualization is the Starch format, which re-encodes BED before compression. The compression efficiency for BED3, bedgraph-like, and BAM-like BED data is significantly improved over gzip/bgzip. The format offers random access useful for per-chromosome parallelization of tasks and can be used directly for set operations.

jmarshall:
§1.5 says that

```
0 <= chromStart < LEN(chrom)
chromStart <= chromEnd <= LEN(chrom)
```

and discusses the cases of `chromStart==chromEnd` (an insertion) and `chromStart==chromEnd==0` (a feature before the entire chromosome).

What about features after the entire chromosome? These would seem to be naturally represented as `chromStart==chromEnd==LEN(chrom)`, in which case I think the requirement on chromStart should be

> chromStart must be an integer […] less than **or equal to** the total number of bases of the chromosome to which it belongs.

michaelmhoffman:
> Applying sort order on any additional fields in such a BED file is useful. Set and map operations on fully-sorted files will be deterministic.
> 
> Without full sorting, running an operation on differently sorted files (or files with unknown sort order) could give two or more different answers, where there are matching intervals.

@alexpreynolds You have made an excellent case for sorting on all fields. We will change the recommended sort to include chromEnd again as that is a fairly simple fix that will address most cases. Since it is only a recommendation I do not want to make sorting as described here very complex. A future specification could require a specific full sort order, such as `sort-bed` does.

michaelmhoffman:
Thanks for your further comments! The following changes are planned for public comments 2021-06-29 through 2021-07-09:

- [ ] `chromStart`: add ability for it to be equal to the total number of bases to represent a feature after the entire chromosome ([thanks](#issuecomment-877313030) @jmarshall)
- [ ] `name`: add "in the absence of out-of-band information about whether tab is the only field separator in the file" ([thanks](#issuecomment-871179106) @pdl)
- [ ] sorting
  - [ ] change recommended command to `LC_ALL=C sort -k 1,1 -k 2,3n` ([thanks](#issuecomment-872356378) @alexpreynolds)
  - [ ] note after the sort recommendation that some downstream analyses will not produce identical results for files with the same data lines that are sorted differently. ([thanks](#issuecomment-872356378) @alexpreynolds)
- [ ] add alternate recommendation of tabix for large BED files ([thanks](#issuecomment-874255693) @cmdcolin)
- [ ] add acknowledgments to any of the folks thanked in the above checklist, with full name and affiliation

Also there will be some more changes thanks to comments from the GA4GH Peer Review Committee.

jmarshall:
* §1 says a BED file must have at least one line. While there is something to be said for disallowing empty files (especially as they are not self-identifying as BED files by peeking at their contents), it would be somewhat unfortunate if an empty set of features could not be represented as a BED file.

* §1.2 defines “0-start, half-open” but later sections use “0-based, half-open”. I think “0-based” is the more common term.

* Some of the within-tables footnotes are missing (is it too boring to recommend the old standby of separate `\footnotemark`/`\footnotetext` :smile:) and the numbering is haywire overall (looks like footnotes don't interact well with `\addstackgap`).

And some more minor typographical issues:

* §1.2 “line separator”: Fullstop missing after “followed by newline”; footnote defines newline as `'\n'` but that's vague (to a C programmer) — consider (additionally) having the footnote saying carriage return is `'\x0d'` and newline is `'\x0a'`.
* §1.4 I'm not sure that it explicitly says that the `Int` fields are written as text (rather than some encoded binary representation of an integer).
* §3.3 Similarly, consider “Text[ual] representation…” rather than “String representation…”; the footnotes here pointing to IEEE 754-1985 are a bit confusing as that standard IIRC defines primarily binary representations for floating point, not textual representations — which is not what's being used in these fields, no?
* §4 “The genome assembly that **defines**…”

Also combining `sort -k…n` options doesn't appear to work like that. I think that sort command needs to be
```
LC_ALL=C sort -k1,1 -k2,2n -k3,3n
```

Combining them as `-k2,3n` doesn't produce the desired ordering.

michaelmhoffman:
Thanks @jmarshall for noticing these issues. 

I agree that "0-based, half-open" is much more commonly used but "0-start" is what UCSC uses in their description of ["The UCSC Genome Browser Coordinate Counting Systems"](http://genome.ucsc.edu/blog/the-ucsc-genome-browser-coordinate-counting-systems/). I also think it is somewhat more intuitive for newcomers. That said, "0-based" is what is used by other specs in this repo that I have checked so let's change it here to match.

Could you please provide a short example where `-k2,2n -k3,3n` provides different results than `-k2,3n`?

Let's add these additional changes for public comments 2021-07-09:

- [ ] change "one or more lines" to "zero or more" ([thanks](#issuecomment-877456096) @jmarshall)
- [ ] change "0-start" to "0-based" ([thanks](#issuecomment-877456096) @jmarshall)
- [ ] remove `\addstackgap`. This will be unnecessary anyway as we plan to turn the tables into floats following the recommendation of the Peer Review Committee. ([thanks](#issuecomment-877456096) @jmarshall)
- [ ] line separator: add missing `.` to before footnote ([thanks](#issuecomment-877456096) @jmarshall)
- [ ] "line separator" definition
  - [ ] remove footnote ([thanks](#issuecomment-877456096) @jmarshall)
  - [ ] add "(`'\r'`, equivalent to `'\x0d'`)" ([thanks](#issuecomment-877456096) @jmarshall)
  - [ ] add "(`'\n'`, equivalent to `'\x0a'`)" ([thanks](#issuecomment-877456096) @jmarshall)
  - [ ] add "(`'\r\n'`, equivalent to `'\x0d\x0a'`)" ([thanks](#issuecomment-877456096) @jmarshall)
- [ ] make clear that `Int`s are decimal string representations ([thanks](#issuecomment-877456096) @jmarshall)
- [ ] change "string representation" to "decimal string representation", the term used in IEEE 754
- [ ] remove footnote to IEEE 754 for integers in user-defined fields; only have it for `Float` ([thanks](#issuecomment-877456096) @jmarshall)
- [ ] "genome assembly that define" -> "genome assembly that defines" ([thanks](#issuecomment-877456096) @jmarshall)

jmarshall:
> Could you please provide a short example where `-k2,2n -k3,3n` provides different results than `-k2,3n`?

_foo.bed_ from @alexpreynolds's https://github.com/samtools/hts-specs/pull/570#issuecomment-872356378 above, or a subset from it:

```
chr11   9       100
chr11   9       99
```

It appears that `-k2,3n` compares the pairs of columns as `9<tab>100` and `9<tab>99` — both of which are converted numerically as the value 9, so they compare equal according to `2,3n`. This follows [POSIX sort(1)](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sort.html)'s description of `-n` as “Restrict the sort key to an **initial** numeric string” (my emphasis), so I guess using `n` with more than one column at a time is basically always doomed to failure…

alexpreynolds:
Here is some test input:

```
echo -e "chrN\t9\t100\nchrN\t9\t99"                    
chrN	9	100
chrN	9	99
```

Here are two seemingly incorrect sort orders:

```
echo -e "chrN\t9\t100\nchrN\t9\t99" | sort -k1,1 -k2,2n  
chrN	9	100
chrN	9	99
```

```
echo -e "chrN\t9\t100\nchrN\t9\t99" | sort -k1,1 -k2,3n
chrN	9	100
chrN	9	99
```

Here are orderings that would be more expected, to me:

```
echo -e "chrN\t9\t100\nchrN\t9\t99" | sort -k1,1 -k2,2n -k3,3n
chrN	9	99
chrN	9	100
```

```
echo -e "chrN\t9\t100\nchrN\t9\t99" | sort-bed -
chrN	9	99
chrN	9	100
```

Here is a link that describes the Starch specification:

https://bedops.readthedocs.io/en/latest/content/reference/file-management/compression/starch-specification.html

It does a better job at compression than gzip/bgzip. For instance, starting with a set of `hg38` chromosomes:

```
fetchChromSizes hg38 | awk -v FS="\t" -v OFS="\t" '{ print $1, "0", $2; }' | grep -v ".*_.*" | sort-bed - > hg38.nuclearOnly.bed
```

We can make a bedgraph (BED3+1) file from this set, chopping it up into 1k intervals and adding some random signal in the fourth column to mimic a real-world dataset:

```
bedops --chop 1000 hg38.nuclearOnly.bed | awk -v FS="\t" -OFS="\t" -v seed=123 'BEGIN{srand(seed)}{ print $0, rand() }' > hg38.nuclearOnly.rand.bg
```

This file can be compressed a couple ways:

```
starch hg38.nuclearOnly.rand.bg > hg38.nuclearOnly.rand.bg.starch
bgzip -c hg38.nuclearOnly.rand.bg > hg38.nuclearOnly.rand.bg.gz
```

On my host, the bgzip file is 27.1% of the size of the original bedgraph file, while the Starch file is 9.2%. 

In other words, this is a 3x size reduction on typical work input, which is why I suggest this for archival or compression purposes.

michaelmhoffman:
Thanks, both. Have changed the sort command in the above checklist.

michaelmhoffman:
Thank you to all who have provided feedback, and additionally to the GA4GH Peer Review Committee for further feedback. I have used GitHub Releases to upload revised PDFs of:

- [BEDv1.pdf](https://github.com/michaelmhoffman/hts-specs/releases/download/20210830/BEDv1.pdf)
- [latexdiff from revision 6063151](https://github.com/michaelmhoffman/hts-specs/releases/download/20210830/BEDv1-diff-from-6063151.pdf), the version GA4GH sent out for public comment
- [latexdiff from revision 1caa7c6](https://github.com/michaelmhoffman/hts-specs/releases/download/20210830/BEDv1-diff-from-1caa7c6.pdf), the previous release posted 2021-06-28

pdl:
In section 3.3: 
- User-defined and custom seem to be used interchangeably but without explicitly being equivalent - I would replace the two instances of User-defined (3.3 title and 1.3 BEDn+m) with custom and also in 1.3 say "defined by the author of the file" in place of "defined by a user".
- No mention is made of a constraint that User-defined fields cannot contain spaces if the field separator is not `\t`, unlike the `name` column.
- "other than tab" in the definition for String and Character is now redundant, as .
- "a comma-separated list of values of any type other than String" permits a comma-separated list of Characters (which includes comma, so it's unclear how parsers with knowledge of such fields should react to `n,m,,,.` and `q,,e`). This seems to be asking for trouble or raising expectations unnecessarily when we could simply refer to these as strings. Does anyone produce files with comma-separated arrays of characters?
- Custom fields may be empty, whereas BED fields, mandatory or optional, must not be. The distinction I would therefore amend 1.5 to replace "field" with "BED field" in the section:
    > Any mandatory or optional field included on any data line in the file must not be empty on any other data line 

I've noticed that the definition of BEDn+ is not the same as the way it's used, e.g. in 3.4, The sentence beginning "For BED4+ files" refers to "any optional fields" (optional fields referring exclusively to BED fields 4-12), but BED4+ as defined in this spec has exactly one optional BED field, `name`, followed by custom fields. In this spec, BEDn+ actually seems to refer to any BED file where the first _p_ fields are BED fields, such that _p_ >= _n_, and where there may or may not be custom fields.

michaelmhoffman:
Thank you @pdl for these comments. I have addressed them in the new version here:

- [BEDv1.pdf](https://github.com/michaelmhoffman/hts-specs/releases/download/20210831/BEDv1.pdf) revision 2a5cd5d
- [latexdiff from revision 6063151](https://github.com/michaelmhoffman/hts-specs/releases/download/20210831/BEDv1-diff-from-6063151.pdf), the version GA4GH sent out for public comment
- [latexdiff from revision af943fe](https://github.com/michaelmhoffman/hts-specs/releases/download/20210831/BEDv1-diff-from-af943fe.pdf), the previous release posted 2021-08-30

pdl:
Great, thanks! One small fix: BEDn+m definition now reads "followed by _m_ custom field." - this should end "fields."

michaelmhoffman:
I noticed *comment line* and *blank line* are in boldface but they are not defined separately as terms.

brainstorm:
Just a minor observation on terminology: a few of the htslib-supported formats have the concept of "entries" as "records" (lines?) whereas this BED spec refers to those lines as "data lines"?

Would it make sense to substitute "data lines" for "records" or there's some historical reason for the naming that I'm missing here?

michaelmhoffman:
This is a line-based format, and for example has a line separator rather than a record separator. I think it is easier to understand the way it is.

arq5x:
> This is a line-based format, and for example has a line separator rather than a record separator. I think it is easier to understand the way it is.

Completely agree.

brainstorm:
> > This is a line-based format, and for example has a line separator rather than a record separator. I think it is easier to understand the way it is.
> 
> Completely agree.

It's alright, while implementing BED support based on this spec I observed that the rest of the API surface referred to "records" across almost all other bioinfo formats. But I understand that users and already existing implementations might be more used to "data lines" (generational and/or historical argument?). My argument was raised on the aim to have some more API term(s) uniformity, so better cognitively.

Personally I don't see a clear line between "record separator" and "line separator" pointed out by @michaelmhoffman since both terms seem quite abstract and interchangeable to me?

Anyways, data lines it is then ;)

michaelmhoffman:
I have addressed the remaining comments from this issue and the peer review committee.

- [BEDv1.pdf](https://github.com/michaelmhoffman/hts-specs/releases/download/20210923/BEDv1.pdf) revision 2481c79
- [latexdiff from revision 2a5cd5d](https://github.com/michaelmhoffman/hts-specs/releases/download/20210923/BEDv1-diff-from-2a5cd5d.pdf), the previous release posted 2021-08-31

michaelmhoffman:
When merging the `Makefile` changes in 3cfa33e, will need to set target-specific variable `LATEXMK_ENGINE=--lualatex` for `BEDv1.tex`.

Can take a similar approach with a target-specific variable `LATEXDIFF_ENGINE=--config LATEX=lualatex` for `diff/BEDv1.pdf`.

michaelmhoffman:
[The GA4GH Steering Committee has approved this specification.](https://twitter.com/GA4GH/status/1453825938421129218) Thank you to everyone who has contributed to this process.

Further work on this pull request will be on any technical changes required to get this pulled into the master branch of hts-specs. Please file separate issues for any substantive BED specification concerns.

michaelmhoffman:
This fails CircleCI build because the current code requires a later version of TeX Live (see #576). 

Additionally, CircleCI never gets this far, but locally I cannot build `diff/BEDv1.pdf` due to an error. I believe this error comes from a bad interaction between the `hyperref` package and the `--only-changes` option to `latexdiff`. After building a whole change document, `latexdiff` attempts to use `gs` to extract only changed pages, but then those pages have PDF bookmarks pointing to pages that no longer exist:

```
[...]
Output written on BEDv1.pdf (8 pages, 434584 bytes).
Transcript written on BEDv1.log.
Argument "9.53.3\n" isn't numeric in numeric ge (>=) at /bin/latexdiff-vc line 543.
Executing gs -sDEVICE=pdfwrite -dNOPAUSE -dBATCH -dSAFER -sPageList= -sOutputFile="BEDv1-changedpage.pdf" "BEDv1.pdf"GPL Ghostscript 9.53.3 (2020-10-01)
Copyright (C) 2020 Artifex Software, Inc.  All rights reserved.
This software is supplied under the GNU AGPLv3 and comes with NO WARRANTY:
see the file COPYING for details.
Processing pages .
GPL Ghostscript 9.53.3: ERROR: A pdfmark destination page 8 points beyond the last page 1.
```

@jmarshall What is the best way forward on these two issues? For me, the ideal situation would be that the CircleCI TeX Live be upgraded so the current code works as-is and that the diff issue be ignored. But you may feel otherwise :smile:

michaelmhoffman:
@jmarshall Building `BEDv1.pdf` disabled by default and build passes. This is ready to merge.

Description of `Makefile` changes in the message of 575e679278b37a6502bd1ef34727a70543313fdf. Use `make WITH_LUALATEX=1` to include `BEDv1.pdf` in building and cleaning. Or put `WITH_LUALATEX=1` in your `GNUmakefile`.

Can you please check in a `BEDv1.pdf`? You can use https://github.com/michaelmhoffman/hts-specs/releases/download/20210923/BEDv1.pdf but it sounds like your process is to build it yourself.

michaelmhoffman:
@jmarshall Is there a chance you could check in `BEDv1.pdf` soon? GA4GH wants to do some publicity for it and it would be helpful to have a more stable URL? Thanks!

jmarshall:
Apologies Michael, the day job has been heavy for a while. I'll finish it up hopefully tonight, or certainly over the weekend.

jmarshall:
This PR has now been rebased to after the change to `latexmk` on the repository's mainline, squashed more or less appropriately (while preserving commits corresponding to the various dated drafts along the way), and merged. I regenerated _BEDv1.pdf_ rather than use your PDF from September, as the regenerated one contains the correct commit hash as eventually merged.

I used a different method of temporarily preventing CircleCI from attempting to build _BEDv1.pdf_, so you won't need to define anything special to build it locally. (See 2a34fbfeeeada59c47dd6773ceeb17799905c04f's change to _Makefile_'s `PDFS` list.)

The BED specification is now available at <https://samtools.github.io/hts-specs/>, specifically as [BEDv1.pdf](https://samtools.github.io/hts-specs/BEDv1.pdf). Apologies for the delay.

PR #572, opened 2021-05-26
Specify that SAM header field tags must be unique and field ordering is immaterial

### LINKED ISSUE #571: Significance of the ordering of tagged fields in header lines
So I'd like to know if the header tags in some predefined order.
According to my understanding this would be a valid headerline.
`@HD VN:1.6 SO:coordinate`
But when I switch around the tags, would this still be valid?
`@HD  SO:coordinate VN:1.6`

jmarshall:
As mentioned in passing on samtools/htslib#1256, the spec currently doesn't mention that within a single header line, the field tags must all be distinct. HTSJDK has disallowed duplicate header line fields with different values for over a decade, so clarifying this should-be-obvious point shouldn't cause any significant problems.

Also clarify that ordering is insignificant for both header fields and alignment optional fields, which fixes #571.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -217,6 +217,8 @@ is a two-character string that defines the format and content of {\tt VALUE}.
 Thus header lines match {\tt
   /\char94@(HD|SQ|RG|PG)(\char92t[A-Za-z][A-Za-z0-9]:[
   -\char126]+)+\$/} or {\tt /\char94@CO\char92t.*/}.
+Within each (non-{\tt @CO}) header line, no field tag may appear more than
+once and the order in which the fields appear is not significant.
 
 The following table describes the header record types that may be used
 and their predefined tags.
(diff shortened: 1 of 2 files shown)


jkbonfield:
Thanks.  I approve of the additions.

Should we add a fail test for the duplicate header tag?  We have one already for duplicate aux tags in alignments (test/sam/failed/aux.fail-format4.sam).

jmarshall:
Good idea, I'll add one.

jmarshall:
@jkbonfield: If you're happy with the (single) test file I've added, I'll rebase and merge this.

jkbonfield:
> @jkbonfield: If you're happy with the (single) test file I've added, I'll rebase and merge this.

Looks good to me.

I don't know if we care enough to explicitly test the "but are allowed in `@CO` as it's not a parsed field" (which would be a passed test).  I doubt it's a problem as anyone applying parsing to the CO field would rapidly discover their error, duplicated field or otherwise. :-)

Issue #514, opened 2020-06-29
htsget additional constraint on fields parameter: SEQ, QUAL


jb-adams:
When requesting specific fields (using the `fields` parameter) from an htsget service, there may need to be additional language to clarify certain acceptable operations. In particular, emitting `QUAL` without `SEQ` will result in a `BAM/CRAM` that's unreadable by samtools, it will fail with the following message:

```
SEQ and QUAL are of different length
```

Should we add additional constraint(s) in the 'Field filtering' section of the specification text to identify this case, and others like these? Should an htsget service respond with an `InvalidInput` error if such a request would produce an unreadable alignment file?

jb-adams:
@mlin @jmarshall can we add the `htsget` label for this?

mlin:
Okayyy, long-delayed reply here, but: as a general pattern, htsget servers are allowed to return more results than requested, either with respect to the genomic range slice, or the fields and such. One specific reason for this is to give file-backed servers the option of shoveling byte ranges back to the client when feasible, as opposed to always having to parse, process, and recompress portions of the original file to produce the exact requested result set.

Therefore I think we could address this issue naturally by saying that servers should not return field subsets that would yield an unreadable result, and leave the details of what subsets can be returned implied by the respective format specifications. It's a bit of a cop-out, but a reasonable one, I feel.

jmarshall:
I think that's an excellent resolution to this problem.

It could be phrased as expanding the description of `fields` to say “the listed fields MUST be included; other fields MAY be omitted”, followed by using this as a example of when omittable fields need to be retained nonetheless.

(We might also take the opportunity to explain that “omitted” means “left as `*` or `0`” where appropriate. Not sure whether we're explicit about this anywhere in the spec.)

jkbonfield:
As I recall, the list of fields was discussed early on to permit e.g. efficient partial CRAM decoding (I wrote a proof of principle CRAM filter tool for this I).

However it's at least conceivable that a backend format could exist with different blocking levels for different types of data.  Eg sequence and qual being big are stored in blocks of 1000 records at a time, while POS being small has 10,000 at a time.  Clearly if we wanted to exploit this by byte shovelling we'd need to do some fakery to fill in the blanks, or a partial rewrite on some columns while not doing the others, but that's for the implementation to decide.

I too like the "cop-out" as it doesn't overly restrict us to existing ways of thinking and implementations.

Issue #386, opened 2019-02-21
htsget VCF fields specification


brainstorm:
As briefly discussed in https://github.com/samtools/hts-specs/pull/385, there seems to be a need to specify which VCF fields to include in the htsget spec, since only BAM fields are mentioned under the `/reads` API endpoint.

A prior [PR on htsget VCF support](https://github.com/samtools/hts-specs/pull/233) didn't cover the issue so that's why I'm rising it here, as prompted by @jmarshall. [As part of the OpenAPI3 htsget 1.1.1 spec, I've tentatively included those](https://github.com/samtools/hts-specs/pull/385/files#diff-3146fe385d670d50207a3f413041d145R169):

```
            enum:
              - "INFO"
              - "SAMPLE"
              - "FILTER"
              - "FORMAT"
              - "ALT"
```

But I'm pretty sure this attempt falls short and/or it doesn't satisfy common nor future use cases. Feedback welcome.

/cc @amilamanoj @jeromekelleher @ohofmann

mlin:
Wearing my opinionated hacker hat: I've personally been against htsget servers being responsible for filtering these individual fields because it leads to a need for complex scale-out backends in order to service large workloads. With just range slicing, it's possible to implement an htsget server that does not actually have to process or transport any of the real data in-band, if the files are reachable via HTTPS. This is what htsnexus tries to illustrate although in doing so, it takes it to an extremist extent.

However, as a spec maintainer of course I will approach discussion on this neutrally to the best of my ability :smile:

brainstorm:
+1 on your opinionated hat :)

On the plus side of removing those filtering bits (perhaps also the BAM ones?) from the spec is that it'll all be more like a plain simple http client/server, so less custom code to put on both ends.

jmarshall:
How `fields` is supposed to work is pretty underspecified in the spec. I assume it's supposed to be an interface to CRAM implementations' ability to avoid decoding specified column blocks.

When returning SAM/BAM this would look like the non-requested fields had decayed to `0`/`*`, and when returning CRAM perhaps those blocks could be omitted from the returned stream. Servers that didn't support it would just return the full SAM/BAM/CRAM and clients might see that the non-requested columns were not `0`/`*` but that would be okay because the client is ignoring them anyway.

[...]

PR #575, opened 2021-06-07
htsget field filtering: explicate that servers may return a superset of requested fields

mlin:
...and may in fact need to do so in order to produce a format-valid response. Per discussion #514

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -281,7 +281,7 @@ QUAL	| Base quality scores
 
 Example: `fields=QNAME,FLAG,POS`.
 
-As with range filtering, a server may include more fields than explicitly requested, a possibility which clients must be prepared to handle. The `fieldsParameterEffective` attribute in the GA4GH service-info metadata (see below) indicates whether a server supports filtering individual fields. In any case, servers must not generate any response payload that is invalid under the respective format specification (for example, if the request includes `QUAL`, then the response must provide `SEQ` as well).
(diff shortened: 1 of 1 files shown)


htsget.md
```diff
@@ -281,6 +281,8 @@ QUAL	| Base quality scores
 
 Example: `fields=QNAME,FLAG,POS`.
 
+As with range filtering, a server may include more fields than explicitly requested, a possibility which clients must be prepared to handle. The `fieldsParameterEffective` attribute in the GA4GH service-info metadata (see below) indicates whether a server supports filtering individual fields. In any case, servers must not generate any response payload that is invalid under the respective format specification (for example, if the request includes `QUAL`, then the response must provide `SEQ` as well).
```
jmarshall, line 284: I'm not 100% sure CRAM ties SEQ and QUAL together, and we haven't figured out how `fields` works for variants (cf #386) so I suggest making the example specifically reads/BAM.

```suggestion
As with range filtering, a server may include more fields than explicitly requested, a possibility which clients must be prepared to handle. The `fieldsParameterEffective` attribute in the GA4GH service-info metadata (see below) indicates whether a server supports filtering individual fields. In any case, servers must not generate any response payload that is invalid under the respective format specification (for example, if a request for reads in BAM format includes `QUAL`, then the response must provide `SEQ` as well).
```

jmarshall (approved):
One suggestion about the motivating example, but otherwise: looks good!

Issue #555, opened 2021-04-10
Persisting pdf artifacts?


cmdcolin:
Currently there is the nice @hts-specs-bot that uploads pdf samples from PRs but the artifacts appear to 404 and then it is hard to review (trying to follow MM/MP/ML thread just for context)

e.g. https://634-3666509-gh.circle-artifacts.com/0/root/project/diff/SAMtags.pdf

Is it part of the circle ci to purge artifacts? Any way to make them more persistent?

cmdcolin:
Managed to run the Makefile to build them locally, so technically I can read the pdf now :)

May be helpful to have a quick helper guide for how to do this for others. I used `sudo apt install texlive-science` (needed algorithms package) and then `make` on ubuntu 20

jmarshall:
There is such a building guide, and now there is a link to it in the README on the main repository page.

Circle CI artifacts [expire after a very reasonable 30 days](https://circleci.com/docs/2.0/artifacts/), which is fine for PRs in progress but less good for looking back at historical ones. I think in the CI PRs we discussed having the bot use the GitHub API to edit its previous comments and/or add the PDFs to its comments as attachments rather than link directly to the CircleCI artifacts, so if someone wanted to waste even more of GitHub's disk space they could look into automating that :smile:

cmdcolin:
Gotcha. Thanks for adding the note to the readme...maybe if they are expected to expire we can consider #556?

jmarshall:
Since PR #641 rewrote the PDF generation infrastructure, linked artifacts no longer expire for active PRs.

Instead they are checked into the his-specs repository[^1] and referenced from a `preview/NUM` Git reference (not a branch, but fetchable by adding to your remote fetch config if you wish) so they won't be garbage-collected while the PR is active.

(When a PR is merged or otherwise closed, the corresponding `preview/…` ref is deleted. So the PDFs and their commits will eventually go away some time after that.)

[^1]: Sadly, last time I looked, there was no API facility for working with issue/PR comment attachments. GitHub Support said this was to avoid algorithmic uploading of large amounts of data, which is fine™️ because I can't think of any other ways of uploading large amounts of data to GitHub…

PR #576, opened 2021-06-09
[TEST] Testing CI updates

jmarshall:


jmarshall:
It seems that Cheng Xu's image is missing both `make` and `git`, which complicates just moving to it…

lbergelson:
@jmarshall What's the advantage of the new image?  If you want to switch to it I'd recommend we bake a new custom image based on that with the tools we need added in already.

jmarshall:
The advantage of using someone else's image is that we don't have to maintain an image ourselves. In particular, [xu-cheng/latex-docker](https://github.com/xu-cheng/latex-docker) is automatically rebuilt periodically, so we'd go from an outdated image to one automatically kept up to date.

But this PR was really just me experimenting with the quick update to get this going again — so really I should get back to it if it's supposed to be a quick fix! :smile: If you want to build an updated image please feel free.

However this is a bit of a stopgap. I have also been looking at reimplementing this on GitHub Actions and using a different method to attach the PDFs to PR comments (thus solving #555) — see another stalled experiment in my [jmarshall/actions branch](https://github.com/jmarshall/hts-specs/tree/actions).

michaelmhoffman:
Are you still thinking about updating the TeX environment? I am still getting build failures on #570 because I am using features not available in the old TeX Live 2015 installation on CircleCI.

michaelmhoffman:
Could we please discuss on the 2021-10-14 meeting? I would tag Amreen but I can't figure out her GitHub username.

michaelmhoffman:
A recent enough TeX Live should be included in recent Ubuntu releases (21.04 or 21.10). Unfortunately, I don't think the last LTS release (20.04) is recent enough, and don't know how difficult it is to get a non-LTS Ubuntu image for CircleCI.

jmarshall:
Experiments continue on jmarshall/hts-specs#1, which is almost ready to land. In summary:

1. Uses GitHub Actions; we can retire the Circle CI integration.
2. Stores the PDFs as commits in the hts-specs repository (not on a branch, so that they are not fetched by default; on a ref, so they are not garbage collected), thus solving #555.
3. Currently comments as “**github-actions** (bot)”; we may wish to retire @hts-specs-bot (though the commits do need an author and email address, which currently uses hts-specs-bot's details).

 Happy to discuss the details at tomorrow's meeting, or a PR will be appearing in due course.

jmarshall:
This test PR is now superseded by those continuing experiments, which are now #641.

PR #579, opened 2021-06-16
Phred scale clarifications

### LINKED ISSUE #534: Phred scales vs qualities/scores
Having been confused by this just now and trying to find out which of 0 and 255 represents a likely event in VCF PL fields, I think we need to be more careful in our wording.  Specifically "scale" meaning the logarithmic scale, but not the broad sweeping view that high-phred is high-quality and low-phred is low-quality as defined in the phred program.  Our scales are sometimes reversed.

Going to source: (https://genetics.mgh.harvard.edu/goodman/doc/phrap.html)

> On the basis of the trace characteristics, phred computes a probability p of an error in the base call at each position, and converts this to a quality value q using the transformation q = -10 log_10(p).

Note here "probability p of an error".

In SAM we define:

> \item[Phred scale] Given a probability 0<p<=1, the phred scale of p equals -10 log_10(p), rounded to the closest integer.

So here we're saying "scale" meaning 10log10, but not defining at all what `p` is; specifically whether it's the probability of being true or of being false.  We're simply defining the concept of a logarithmic scale.   We then later on use the term "phred-scaled base QUALity", which may trip up the unwary reader, until they read the more detailed explanation a little later which then reverses the formula by explaining that p is the error probability.

Obviously I know how it works and in my brain "phred" always means P(error), but it should be spelt out earlier.  I think we'd be better off describing what we want from the outset given AFAIK we never use phred-scale in SAM for anything other than errors. Eg:

"Phred-scaled quality: Given a probability 0<p<=1 of an event being correct, the phred-scaled quality q is defined as q = -10 log_10(1-p) rounded to the closest integer."

In VCF we define:

>   \item QUAL --- quality: Phred-scaled quality score for the assertion made in ALT. i.e. -10log_10(prob(call in ALT is wrong)).
...
>   \item GQ (Integer): Conditional genotype quality, encoded as a phred quality -10log_10( p(genotype call is wrong, conditioned on the site's being variant).

Both good and matches the Phrap documentation and my suggested revision for SAM spec too.

However then also in VCF:

>   \item PL (Integer): The phred-scaled genotype likelihoods rounded to the closest integer, and otherwise defined precisely as the GL field.
>  \item PQ (Integer): Phasing quality, the phred-scaled probability that alleles are ordered incorrectly in a heterozygote (against all other members in the phase set).

An incredibly subtle distinction between those two lines:  "phred-scaled genotype likelihood" vs "phred-scaled probability that alleles are ordered incorrectly".  You have to read carefully to see that is p vs 1-p.

For consistency PQ maybe should be rephrased as "phred-scaled quality score" given it's 1-p and is the phrase we're using elsewhere.

Ultimately this is too subtle for people not versed in the art of decoding VCF to understand.  The only formula listed in VCF documenting the phred scale is for p(X is wrong), but whenever it uses the term "scale" it is referring to the unwritten formula of p(X is correct) unless explicitly stated otherwise.

For PL we really ought to be explicit here.  Saying "phred-scaled", while technically true, conjures up in my mind at least the standard of phred encoding errors, which is this case is wrong.  At the very least the sentence should be followed by the explicit formula, and maybe also explaining the relationship to GL verbatim (from @pd3 personal communication: "In other words, PL = int(-10*GL)".

jkbonfield:
This fixes #534.

We abuse the term Phred-scale, particularly in VCF.  This PR also has a minor wording change in SAM to be more explicit.

Sometimes VCF uses Phred as -10log_10(P(correct)) and sometimes -10log_10(P(incorrect)).  Having two completely opposite direction scales with bigger-is-better vs smaller-is-better is really unhelpful.  Plus sometimes it's not even probabilities at all, but likelihoods.  Sometimes they're normalised, and sometimes not.  This poor wording lead me up the garden path for a while and it wasn't until I checked source code that I understand what they meant, mainly due to preconceived (correct!) notions on what "Phred-scaled" means.  Phred explicitly defined the scale as P(incorrect), so we now only use the term only when that is true.

I see this was already done correctly for the CNL, CNP and CNQ VCF tags, so I applied the same logic to the PL, PP and PQ tags.


SAMv1.tex
```diff
@@ -167,8 +167,7 @@ \subsection{Terminologies and Concepts}
   between the 3rd and the 7th bases inclusive is $[2,7)$. The BAM, BCFv2, BED,
   and PSL formats are using the 0-based coordinate system.
 
-\item[Phred scale] Given a probability $0<p\le 1$, the phred scale of $p$
-  equals $-10\log_{10}p$, rounded to the closest integer.
+\item[Phred scale] Given a probability $0<p\le 1$ of an erroneous call, the phred scale of $p$ equals $-10\log_{10}p$, rounded to the closest integer.
```
lbergelson, line 170: It seems like the rest of the file is wrapped at 80 characters, do we want to keep that format style with this line?

VCFv4.3.tex
```diff
@@ -544,13 +544,14 @@ \subsection{VCF tag naming conventions}
 \begin{itemize}
     \item The `L' suffix means \emph{likelihood} as log-likelihood in the sampling distribution, $\log_{10} \Pr(\mathrm{Data}|\mathrm{Model})$.
     Likelihoods are represented as $\log_{10}$ scale, thus they are negative numbers (e.g.\ GL, CNL).
-    The likelihood can be also represented in some cases as phred-scale in a separate tag (e.g.\ PL).
+    The likelihood can be also represented in some cases as a phred-true scale ($-10 \log_{10}(probability\_of\_being\_correct)$) in a separate tag (e.g.\ PL).
```
lbergelson, line 547: Do we really want to introduce the term "phred-true scale"?  It seems like a bit of a strange one off.  It's not clear that it's contrasting probability of correct with probability of incorrect until you read the  'Q Suffix' box below.

lbergelson (commented):
This seems like a good idea.  Some small comments below.

SAMv1.tex
jkbonfield: Agreed.  Changing it to match the others.

VCFv4.3.tex
jkbonfield: Not really, but I'm unsure how best to describe it given the "P" prefix was almost certainly chosen for Phred.

Right now it states "Likelihoods are represented as $\log_{10}$ scale, thus they are negative numbers".  The intent of the phred-ish scale appears to be to negate that and have a positive scale, so maybe this is the correct way of describing it.  I'll have a stab at doing something more.

jkbonfield:
Rebased off current master, with the line-wrapping adjusted for SAMv1.tex.

The "phred-true-scale" sentence has been rewritten as left in as a separate commit so you can see the change, but should be squashed with the other VCF change when merging.

Note this doesn't change vcf 4.4 as I don't know what our policies are on versions to update.  Which versions need to be changed?  Do we have one commit per spec version so we can search git logs easier, or do we combine them together in one commit?

jkbonfield:
Applied the changes to VCF4.4 draft and, minimally to VCF4.2.  I also squashed the previous commit and this change into the main VCF doc updates.  For review purposes the changes applied to VCF 4.3 (bar a minor grammar fix) since previous review are still visible as https://github.com/samtools/hts-specs/commit/571f96c02259ba181f807786bc3c819f226582af

VCFv4.3.tex
```diff
@@ -427,9 +427,9 @@ \subsubsection{Genotype fields}
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
       MQ		& 1		& Integer	& RMS mapping quality \\
-      PL		& G		& Integer	& Phred-scaled genotype likelihoods rounded to the closest integer \\
-      PP		& G		& Integer	& Phred-scaled genotype posterior probabilities rounded to the closest integer \\
-      PQ		& 1		& Integer	& Phasing quality \\
+      PL		& G		& Integer	& $-10 log_{10}$ scaled genotype likelihoods rounded to the closest integer \\
```
jmarshall, line 430: Formatting nits: use `\log` to print a function in roman (and set off by thinspace from the -10) rather than badly spaced single-letter variables in italics; IMHO it looks better keeping the hyphen before `-scaled` (and it's typeset very differently from a minus sign so shouldn't be confusing).

There's already a bunch of `10 log 10` in _VCF*.tex_ that are missing the `\log`, but this PR adds a bunch more instances so should probably use `\log` in its additions.

```suggestion
      PL		& G		& Integer	& $-10\log_{10}$-scaled genotype likelihoods rounded to the closest integer \\
```

jkbonfield: Bah just noticed the "-scaled" bit too... I'll update again. :-)

jmarshall:
I think two commits would be appropriate: (1) fix existing `\log` instances, not really part of this PR; (2) add the new text, already with `\log` throughout. That way around, it's always consistent.

(I once had a draft of fixing all the existing `\log` and standardising on using `$ \Pr(foo|bar) $` for all the variously worded probabilities, but it appears I lost interest along the way… or maybe it got partly applied and then other `log` instances crept back in…)

d-cameron:
At the next file format meeting we should decide on a policy on how far we backport clarifications and how we want to balance specifications stability, maintenance burden, and specifications clarity (same issue with #652).

PR #583, opened 2021-07-21
Fix `##PEDIGREE` example typographical errors

jmarshall:
PR #413 both fixed some typographical errors in `##PEDIGREE` examples and added a new section describing previously undescribed `##PEDIGREE` usage. It has been bogged down for two years as the details of the latter were debated for a while and then ignored.

This PR pulls out the first commit of that, which fixes obvious typographical errors (or possibly transcription errors from when the original g1k wiki VCF text was converted to LaTeX) in the `##PEDIGREE` examples in one section of each spec, and updates it to also fix the errors duplicated into _VCFv4.4.draft.tex_.

This PR simply fixes existing typographical errors in the VCF specs and should be able to be merged without undue delay.

### DIFF
--- a/VCFv4.1.tex
+++ b/VCFv4.1.tex
@@ -867,13 +867,13 @@ Analogously, there might also be several normal genomes from the same patient in
 The general format of a PEDIGREE line describing asexual, clonal derivation is:
 
 \begin{verbatim}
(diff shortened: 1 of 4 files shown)


lbergelson (approved):
@jmarshall Looks good, thanks for cleaning this up.

PR #584, opened 2021-07-29
Remove extra Ml quality value.

jkbonfield:
A trivial fix to the example data.  There is one more element in Ml than Mm so the final 169 is not used.

Reported by Andrew Whitwham at
https://github.com/samtools/htslib/pull/1132#discussion_r678334228


jmarshall:
I concur that this is an error and the fix is correct, but I think there's another problem on the same line:

Don't both the final `C+m` offset and the final `C+76792` offset refer to the same C base, the one that is the second last base of the read? If so, 204 + 187 adds up to much more than 256 = 100%.

jmarshall:
Thoughts on the offsets? If my calculations are correct, then the htslib test file has the same problem.

jkbonfield:
Sorry, I'd forgotten about this.  Yes you're correct.  If we change the 187 to, say, 40, then it'll be valid without changing anything substantive in the tests.  I'll append a commit to do that to this PR.

jkbonfield:
Note the offsets do deliberately overlap at that point, to validate that we can describe multiple possibilities for the same position.  I just hacked the values though.  Thanks.

jmarshall:
If your htslib branch is adding up the `Ml` values, I'm a bit surprised it didn't notice :smile:

jkbonfield:
> 
> If your htslib branch is adding up the `Ml` values, I'm a bit surprised it didn't notice 😄

The API is supplying the data so it doesn't notice.  We could add more error checking, but it'd add overhead and I'm also a bit wary of it biting us in the bum at some stage later.

I do have a `samtools consensus` draft PR using this, so I'm not sure what that was doing with it, although maybe I didn't test it on that file (or maybe it'd just gloss over it in some way).

PR #586, opened 2021-07-30
Add htsget protocol version history appendix

### LINKED ISSUE #585: Providing information about old versions of htsget
We are now up to htsget 1.3.0, and <https://samtools.github.io/hts-specs/> does not (currently) provide direct access to earlier versions of the specification. Implementors may need to know what has changed between different versions of the protocol; for example, maintainers of a server written in the 1.2.0 era may need to know what capabilities were part of the protocol at that time and what is new that they should consider adding.

Such readers always have the option of spelunking around in the git history to find the exact specification text from earlier days, but this is not especially convenient. There are essentially two ways to provide convenient access to this information:

* Designate a commit for each older specification version to be considered “canonical for [e.g.] 1.2.0” and make them available at another URL with a version number in it.

* Maintain an appendix in the current specification listing the significant changes between versions. (See e.g. the respective appendices of [SAMv1.pdf](https://samtools.github.io/hts-specs/SAMv1.pdf) and [SAMtags.pdf](https://samtools.github.io/hts-specs/SAMtags.pdf).)

IMHO the latter is preferable: it provides all the information that most readers need, without multiplying the number of documents that we need to maintain or make available, and does not ask readers to study multiple documents and figure out the differences between them themselves.

As a first approximation, here are the significant additions in each of our htsget versions to date:

1.3.0 POST requests
1.2.1 service-info endpoint
1.2.0 header requests and `class` attributes
1.1.1 `referenceName=*` to request unmapped reads
1.1.0 variants endpoint
1.0.0 first promulgated version

jmarshall:
We've gradually added to htsget and not really realised that actually time has passed and we now have a number of previous published htsget protocol versions…

This PR adds an appendix listing the significant functionality introduced and changes made in each published version of the htsget protocol. Fixes #585.

~~It's a bit unfortunate that the “Possible future enhancements” section comes after this version history, but it needs to because the footnotes in “Existing clarification suggestions” are stuck at the end of the document in the web-rendered version. I seem to recall we've discussed creating issues to track each of these enhancement suggestions and removing these sections, which I think we should probably do while considering the text of this PR.~~ [Now done, in df96264bfc78487b0200a958f8783c3b1d893048.]

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -752,6 +752,44 @@ Example listing of htsget reads API and variants API registrations from a servic
 ]
 ```
 
+# Version history
+
+This appendix lists the significant functionality introduced and changes made in each published version of the htsget protocol.
+
+## 1.3.0 (March 2021)
+{:.no_toc}
+
+Added POST requests for both reads and variants endpoints, allowing data to be queried via the HTTP `POST` method, and defining a JSON request object to be used as the POST payload instead of the corresponding set of htsget query parameters.
+Defined a new `PayloadTooLarge` error type.
+
+<!-- The service-info addition (without POST) existed briefly as an unpublished htsget 1.2.1 version. -->
+Added a `service-info` endpoint and defined an `htsget` service-info response sub-object with `datatype`, `formats`, `fieldsParameterEffective`, and `tagsParametersEffective` optional fields.
+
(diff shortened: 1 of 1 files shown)


jmarshall:
Rendered version can be viewed at <http://jmarshall.github.io/hts-specs/htsget.html>

PR #599, opened 2021-09-14
Explicitly permit unrecognised data series in the record encoding map.

### LINKED ISSUE #598: test/cram: 1301_slice_aux.cram includes undefined CRAM 3.0 data series
In [`test/cram/3.0/passed/1301_slice_aux.cram`], the data series encoding map in the first data container's compression header includes two entries defined by neither the [CRAM 2.1] nor [CRAM 3.0] spec: `TC` and `TN`. [From htslib], these seem to be old CRAM 1.0 keys.

Are there any assumptions that should be made for unsupported data series?

[`test/cram/3.0/passed/1301_slice_aux.cram`]: https://github.com/samtools/hts-specs/blob/df96264bfc78487b0200a958f8783c3b1d893048/test/cram/3.0/passed/1301_slice_aux.cram
[CRAM 2.1]: https://github.com/samtools/hts-specs/blob/df96264bfc78487b0200a958f8783c3b1d893048/CRAMv2.1.tex#L591-L669
[CRAM 3.0]: https://github.com/samtools/hts-specs/blob/df96264bfc78487b0200a958f8783c3b1d893048/CRAMv3.tex#L720-L807
[From htslib]: https://github.com/samtools/htslib/blob/3eada2f748dddb7743bb336e326511eda77892b2/cram/cram_structs.h#L192

jkbonfield:
This is to accommodate htsjdk which emits empty CRAM 1.0 data-series in CRAM 3.0 files, even though they have no meaning in CRAM 3.0.  The TC and TN data series refer to the earlier tag encoding mechanism used in CRAM 1.0.  We changed this for CRAM 2.0 to permit a more compact data representation, but it appears htsjdk's CRAM 3.0 includes these data series in the record encoding map and even writing (empty) blocks for them.

Fixes #598

There is a tighter alternative which I may consider instead.  To ban unknown data series, but explicitly add TC and TN to the known data series list with "1.0 legacy: ignore" in the description.  This is more in keeping with how htslib has been implemented.  It does check the list of data series and emits a warning (not error) if it finds something it doesn't recognise, but it does recognise TC and TN so no warnings are issued for them even though they are not defined by CRAM 2.1 or 3.0.

Thoughts?


jmarshall:
What exactly does it mean to ignore a data series?

Suppose we invent a new data series TZ in CRAM 3.2, and support for it is added in htslib 1.20 / samtools 1.20. Someone makes a CRAM file with samtools 1.20 and makes it available.

I download that file and decode it with samtools 1.15, which uses htslib 1.15 which follows this “ignore unknown data series” proposal. So it prints out the contents, presumably omitting whatever was encoded into this TZ item.

What does this mean for what I see? Which of the following are possible?

1. `samtools-1.15 view` fails because parts of the data returned are incomplete and a resulting SAM record is inconsistent
2. `samtools-1.15 view` succeeds but some of the tagged fields (the ones that were encoded in the TZ item) are just omitted
3. It succeeds and some other fields also have different values from what someone using samtools-1.20 would see
4. It succeeds and produces the same output as samtools-1.20 (because somehow the contents of the TZ item were unnecessary, hence ignorable?)

jkbonfield:
Thanks for the comments.

By definition, a new data series means a new piece of information, so ignoring it is basically reporting the record without that new information.   If the record made sense without it (which it should do given SAM records right now are consistent) then it should be fine.  Ie your option 2. 

While this may permit some future proofing, realistically most new data elements are better done via auxiliary tags which are handled elsewhere.  The only real reason for additional non-tag data series is if we want to change how we break down the SAM columns into a new structure.  Eg CIGAR produces a whole plethora of data series, but it could produce something else.  That however is a major format bump and not the point.

The intention here isn't for future proofing the format with the expectation of adding more data series later, but rather to document how to behave when encountering the unexpected because, sadly, that unexpected situation is here already.  However I'm not really comfortable with this correction and after a day of reflection I'm thinking the second for explicit proposal is preferable and the appropriately small sticky-plaster.  I'll rework it that way.

jmarshall:
Certainly just grandfathering in the known legacy TC/TN items is a preferable conservative approach IMHO.

yash-puligundla:
Hi @jkbonfield, Looks like the above changes are not reflected in CRAMv3.pdf file

jmarshall:
@yash-puligundla: The PDF is updated now.

PR #606, opened 2021-11-11
Remove blank line from a vcf test file

jkbonfield:
passed_meta_pedigree.vcf ends on a blank line, which trips up bcftools (in a rather poor way).

Assuming that VCF is like SAM in which case lines mustn't be empty, this is an invalid VCF.  Although we may wish to add this case to the failed directory, if it's not already there.

Edit: also added fix for END < POS in a couple VCF test files.


jmarshall:
:+1: LGTM FWIW

test/vcf/4.3/passed/passed_body_info.vcf
```diff
@@ -32,7 +32,7 @@
 6	300	rs180734498	C	T	100	PASS	DB=1	GT:DS:GL	0|0:0.050:-0.13,-0.58,-3.62	0|1:1.000:-2.45,-0.00,-5.00
 7	100	rs180734498	C	T	100	PASS	DP=0	GT:DS:GL	0|0:0.050:-0.13,-0.58,-3.62	0|1:1.000:-2.45,-0.00,-5.00
 7	200	rs180734498	C	T	100	PASS	DP=2184	GT:DS:GL	0|0:0.050:-0.13,-0.58,-3.62	0|1:1.000:-2.45,-0.00,-5.00
-8	100	rs180734498	C	T	100	PASS	END=0	GT:DS:GL	0|0:0.050:-0.13,-0.58,-3.62	0|1:1.000:-2.45,-0.00,-5.00
+8	100	rs180734498	C	T	100	PASS	END=100	GT:DS:GL	0|0:0.050:-0.13,-0.58,-3.62	0|1:1.000:-2.45,-0.00,-5.00
```
tcezard, line 35: This would also apply to [test/vcf/4.2/passed/passed_body_info.vcf](https://github.com/samtools/hts-specs/blob/master/test/vcf/4.2/passed/passed_body_info.vcf) and [test/vcf/4.1/passed/passed_body_info.vcf](https://github.com/samtools/hts-specs/blob/master/test/vcf/4.1/passed/passed_body_info.vcf)

jkbonfield:
Updated with 4.1 and 4.2 changes too (a bit in both commits).  Also rebased.

PR #612, opened 2021-12-06
Test that BEDv1.pdf is suppressed on CI

jmarshall:


jmarshall:
Observed that _BEDv1.pdf_ has been skipped by CircleCI, as desired.

PR #614, opened 2021-12-09
Migrate base modifications SAM tags from draft to official name space

jkbonfield:


SAMtags.tex
```diff
@@ -662,6 +665,9 @@ \section{Tag History}
 \setlength{\parindent}{0pt}
 \newcommand*{\gap}{\vspace*{2ex}}
 
+\subsubsection*{December 2021}
+Migrate the draft Mm and Ml tags to the official MM and ML namespace.
```
jmarshall, line 669: I'm glad a history entry has been added, but I think it needs to be written to be more comprehensible and actionable for spec readers than this sentence. Something (in addition perhaps) that spells out “Base modifications should be written using the specification's MM and ML tags, rather than the previous draft Mm and Ml tags“, rather than using only standardese here.

(Also need to update to `\subsubsection*{January 2022}` or whenever this gets merged.)

jmarshall:
As discussed in the last meeting, I've taken the liberty of pushing suggestions for amended history entries.

This explains **why** the tag names are changing, and recommends that programs writing out tags use the MM/ML versions. It's left intentionally vague what programs should do on reading — whether to accept MM/ML only, or to accept Mm/Ml as well (perhaps temporarily).

The idea is also to advertise our new draft tag process. So the review period is mentioned, and I added a December entry (for Jared's changes) demonstrating that improvements occurred during the review period.

jmarshall (approved):
I regret that my quibbles about the history entry have been holding up this important change, with the MM/ML tags graduating at last to their production form. I've proposed some text for that trivia that I'm happy with, and explained on the PR why I think expanding this trivia is (a bit) important.

Back to you @jkbonfield: I'm happy for you to squash and merge as is, or to make further suggestions for another iteration.

jmarshall:
As applications start using the base modifications tags, this graduation away from the draft tag names is increasingly overdue.

I suggest we start a time limit: If there are no further comments, I will merge this in its current form on 2022-02-28. (Namely next Monday — this date selected as it avoids needing to update the history entry to say March.)

Or if the PR author and other SAM maintainer is happy with it, he should feel free to merge it as is immediately. :smile:

jkbonfield:
Sorry I've just had my head in the internals of bcftools recently and not had time to look.  I just checked and I'm happy with it.  I'll get it merged.

jmarshall:
Not a fun place to have your head :smile: — thanks.

Issue #593, opened 2021-08-31
[bcf spec] Empty variable length vector in Genotype encoding


h-2:
The VCF spec has the following genotype fields in the example records
```
GT:GQ:DP:HQ	0|0:48:1:51,51	1|0:48:8:51,51	1/1:43:5:.,.
GT:GQ:DP:HQ	0|0:49:3:58,50	0|1:3:5:65,3	0/0:41:3
[...]
GT:GQ:DP	0/1:35:4	0/2:17:2	1/1:40:3
```

For the last sample, HQ is missing from all three records, but this is encoded differently:
1. Two missing values
2. HQ is absent for this sample
3. HQ is absent for all samples

It is perfectly clear to me how BCF represents cases 1. and 3., but what about number 2?
I had assumed that "empty vector" would basically be encoded as a vector consisting of two END_OF_VECTOR values as the spec says:
```
In the situation when a genotype
field contain vector values of different lengths, these are represented in BCF2 by a vector of the maximum length
per sample, with all values in the each vector aligned to the left, and END OF VECTOR values assigned to all
values not present in the original vector. 
```
However, `bcftools convert` creates a vector of `[MISSING_VALUE, END_OF_VECTOR]`.

Which is correct? Does my implementation need to emulate bcftools to be compatible?

And could this maybe be made clearer in the spec? I think it is already confusing enough that there are so many different ways to encode the absence of something 🙄 

Thank you!

pd3:
The VCF specification allows to drop sample's trailing FORMAT fields (case 2), but it also allows to use a single MISSING_VALUE as an abbreviated way of expressing how many missing values there are in that field. HTSlib puts in the single missing value.

h-2:
> The VCF specification allows to drop sample's trailing FORMAT fields (case 2), but it also allows to use a single MISSING_VALUE as an abbreviated way of expressing how many missing values there are in that field.

Yes, this is correct for *VCF*, but *BCF* clearly mandates a fixed length vector and the spec also literally says  
```
END OF VECTOR values assigned to all values not present in the original vector.
```
There are no values present in the original vector, so shouldn't that result in `[END_OF_VECTOR, END_OF_VECTOR]`? 

I think this behaviour would be important to preserve the distinction in VCF when converting back and forth, so that

* VCF `.,.` -> BCF `[MISSING_VALUE, MISSING_VALUE]` -> VCF `.,.` (all values explicitly marked as missing)
[...]

PR #617, opened 2021-12-16
Clarification of empty vector representation

pd3:
as triggered by
    https://github.com/samtools/bcftools/issues/1622
    https://github.com/samtools/hts-specs/issues/593

States explicitly that missing vectors (`.`) in fields of variable length are in BCF represented as `MISSING,END_OF_VECTOR`, i.e. the leading `MISSING` value must be always present.


jmarshall:
#593 makes a case that `[END_OF_VECTOR, END_OF_VECTOR]` would also be a reasonable encoding of this situation.

What is the reason for requiring that “[e]mpty vectors (i.e. vectors with no data available)” be represented as `[MISSING, END_OF_VECTOR]` (which is seemingly less empty!)?

jkbonfield:
Fixes #593

Issue #608, opened 2021-11-11
Other misc. VCF test data queries.


jkbonfield:
Some of this is an issue for the data, but some also needs some major clarifications in the spec.

- Is contig "<1>" just a typo here:
```
@ seq4-head2[4.3/passed]; awk '!/^#/ {print $1,$2}' complexfile_passed_000.vcf
1 10583
<1> 10611
1 30923
1 46402
...
```

- Illegal? header lines in complexfile_passed_000.vcf (also passed_meta_pedigreedb.vcf)

The VCF spec carefully lists the syntax of specific line types.  These are what VCF calls "structured lines", but it's less clear on non-structured data.  The spec mentions structured lines, but I could find no definition for unstructured.  Is the opposite of structured "meta-information"?  If so the spec doesn't say much about them other than that they are key/value and must be "well formed", but sadly doesn't explain what that means.   That needs fixing.   I assume they must still follow the same quoting rules.

Which leads me on to meta-characters.  Some fields are quoted, and others not.  Eg from the spec:

```
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
##FORMAT=<ID=GQ,Number=1,Type=Integer,Description="Genotype Quality">
```

Here `Description` is quoted, presumably because it may contain spaces, or more importantly commas.  Does it *need* to be quoted?  Could `Description=Genotype` be used?  Could we say `Type="Integer"`?  There is no grammar or even a regexp for how to parse headers, which is a major problem.

So, on to the specifics which tripped up bcftools:

```
##AnalysisTitleBrackets=<"FINRISK: Whole-exome sequencing of Dietary, life style, and genetic determinants of obesity and metabolic syndrome (DILGOM)">
```
This one is entirely quoted, but it has no = in there at all.  Hence it also fails the requirement of being key=value.

```
##pedigreeDB=<ftp://user@host:8080/path/to/pedigreeDB?arg1=db1>
```

Now this line does have an equals sign, so technically matches the key=value requirement.  It clearly wasn't intended to be one though, but why is it failing?  I assume bcftools dislikes some characters in the key, eg : or ?.   What characters are permitted in the keys?  When would we need to quote it?  Can we even quote it if we desired to?  What characters *must* be quoted.  Obviously comma, and maybe <>, but = too?  I don't see why it would for a value, but clearly it would for the key and we'd probably not want to have differing quote requires for both.

[...]

PR #620, opened 2022-01-05
Clarify description of structured vs unstructured meta-information lines

jmarshall:
PR #88 introduced the term “**structured** meta-information line”, but was unclear about the relationship between those and other meta-information lines, and §1.4 remains unclear overall — see e.g. #608.

This PR introduces the term “**unstructured** meta-information line” (an alternative might be “**simple** meta-information line”), and rewords this section so it describes the two flavours of meta-information line clearly.

This PR does not attempt to tighten up the syntax around what should be quoted etc as advocated for in #608. OTOH it does newly specify that:

* unstructured values must be non-empty (as is claimed as a rule by some of the test files)
* unstructured values must not start with `<` (so that structured/unstructured are easily distinguished; this is already de facto enforced by the HTSlib implementation)

Hence it also removes `<>` from §5.4.11's unstructured `##pedigreeDB` example. As discussed in https://github.com/samtools/hts-specs/issues/608#issuecomment-966968626, PR #88 removed the `<>` from §1.4.9's instance of `##pedigreeDB=<url>` presumably on the grounds that they were merely metasyntactic variable notation and not intended to appear literally, but missed this instance. Adjust the test files accordingly.

The _test/vcf/4.[123]/passed/complexfile_passed_000.vcf_ test files still contain another unstructured meta-information line with angle bracket delimiters:

```
##AnalysisTitleQuotes="FINRISK: Whole-exome sequencing of Dietary, life style, and genetic determinants of obesity and metabolic syndrome (DILGOM)"
##AnalysisTitleBrackets=<"FINRISK: Whole-exome sequencing of Dietary, life style, and genetic determinants of obesity and metabolic syndrome (DILGOM)">
##AnalysisTitlePlain=FINRISK: Whole-exome sequencing of Dietary, life style, and genetic determinants of obesity and metabolic syndrome (DILGOM)
```

The maintainers should also update the `##AnalysisTitleBrackets` lines, either just removing them or perhaps rewriting them as `##AnalysisTitleStructured=<ID="FINRISK: … (DILGOM)">`.

The maintainers will likely wish to apply similar changes to _VCFv4.4.draft.tex_.

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -116,13 +116,14 @@ A \emph{structured} meta-information line is similar, but the value is itself a 
 \verb|##|\emph{key}\verb|=<|\emph{key}\verb|=|\emph{value}\verb|,|\emph{key}\verb|=|\emph{value}\verb|,|\emph{key}\verb|=|\emph{value}\verb|,|\ldots\verb|>|
 \end{quote}
 All structured lines require an ID which must be unique within their type, i.e., within all the meta-information lines with the same ``\verb|##|\emph{key}\verb|=|'' prefix.
-For all of the structured lines (\verb|##INFO|, \verb|##FORMAT|, \verb|##FILTER|, etc.), extra fields can be included after the default fields.
+For all of the structured lines (\verb|##INFO|, \verb|##FORMAT|, \verb|##FILTER|, etc.) described in this specification, extra fields can be included after the default fields.
 For example:
 \begin{verbatim}
 ##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="128">
 \end{verbatim}
 In the above example, the extra fields of ``Source'' and ``Version'' are provided.
 The values of optional fields must be written as quoted strings, even for numeric values.
+Other structured lines not defined by this specification may also be used; the only default field for such lines is the required \verb|ID| field.
 
 It is recommended in VCF and required in BCF that the header includes tags describing the reference and contigs backing the data contained in the file.
 These tags are based on the SQ field from the SAM spec; all tags are optional (see the VCF example above).
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -100,24 +100,37 @@ Flag, Character, and String.
 For the Integer type, the values from $-2^{31}$ to $-2^{31}+7$ cannot be stored in the binary version and therefore are disallowed in both VCF and BCF, see \ref{BcfTypeEncoding}.
 
 \subsection{Meta-information lines}
-File meta-information is included after the \#\# string and must be key=value pairs.
-Meta-information lines are optional, but if they are present then they must be completely well-formed.
-Note that BCF, the binary counterpart of VCF, requires that all entries are present.
-It is recommended to include meta-information lines describing the entries used in the body of the VCF file.
+File meta-information lines start with ``\verb|##|'' and must appear first in the VCF file, before the header line (section~\ref{header-line}) and data record lines (section~\ref{data-lines}).
+They may be either \emph{unstructured} or \emph{structured}.
 
-All structured lines that have their value enclosed within ``$<>$'' require an ID which must be unique within their type.
-For all of the structured lines (\#\#INFO, \#\#FORMAT, \#\#FILTER, etc.), extra fields can be included after the default fields.
+An \emph{unstructured} meta-information line consists of a~\emph{key} (denoting the type of meta-information recorded) and a~\emph{value} (which may not be empty and must not start with a `\verb|<|' character), separated by an `\verb|=|' character:
+\begin{quote}
+\verb|##|\emph{key}\verb|=|\emph{value}
+\end{quote}
+Several unstructured meta-information lines are defined in this specification, notably \verb|##fileformat|.
+Others not defined by this specification, e.g.\ \verb|##fileDate| and \verb|##source|, are commonly found in VCF files.
+These typically have meanings that are obvious, or they are immaterial for processing the file, or both.
+
+A \emph{structured} meta-information line is similar, but the value is itself a comma-separated list of key=value pairs, enclosed within `\verb|<|' and `\verb|>|' characters:
+\begin{quote}
+\verb|##|\emph{key}\verb|=<|\emph{key}\verb|=|\emph{value}\verb|,|\emph{key}\verb|=|\emph{value}\verb|,|\emph{key}\verb|=|\emph{value}\verb|,|\ldots\verb|>|
+\end{quote}
+All structured lines require an ID which must be unique within their type, i.e., within all the meta-information lines with the same ``\verb|##|\emph{key}\verb|=|'' prefix.
+For all of the structured lines (\verb|##INFO|, \verb|##FORMAT|, \verb|##FILTER|, etc.), extra fields can be included after the default fields.
 For example:
 \begin{verbatim}
-##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="description",Version="128">
+##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="128">
 \end{verbatim}
 In the above example, the extra fields of ``Source'' and ``Version'' are provided.
-Optional fields must be stored as strings even for numeric values.
+The values of optional fields must be written as quoted strings, even for numeric values.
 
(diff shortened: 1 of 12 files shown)


VCFv4.3.tex
```diff
@@ -100,24 +100,37 @@ \subsection{Data types}
...
+These typically have meanings that are obvious, or they are immaterial for processing the file, or both.
 
-All structured lines that have their value enclosed within ``$<>$'' require an ID which must be unique within their type.
-For all of the structured lines (\#\#INFO, \#\#FORMAT, \#\#FILTER, etc.), extra fields can be included after the default fields.
+A \emph{structured} meta-information line is similar, but the value is itself a comma-separated list of key=value pairs, enclosed within `\verb|<|' and `\verb|>|' characters:
+\begin{quote}
+\verb|##|\emph{key}\verb|=<|\emph{key}\verb|=|\emph{value}\verb|,|\emph{key}\verb|=|\emph{value}\verb|,|\emph{key}\verb|=|\emph{value}\verb|,|\ldots\verb|>|
+\end{quote}
+All structured lines require an ID which must be unique within their type, i.e., within all the meta-information lines with the same ``\verb|##|\emph{key}\verb|=|'' prefix.
+For all of the structured lines (\verb|##INFO|, \verb|##FORMAT|, \verb|##FILTER|, etc.), extra fields can be included after the default fields.
```
jkbonfield, line 119: I appreciate the clarification, and later it explains (partially) something that's always been a total mystery to me: why some strings are quoted in VCF and others are not.

However I'm unclear what "default fields" means.  Can we assume for the examples shown that everything in the example is classified as "default".  For example the definitions of `##INFO` always have ID, Number, Type and Description so it's clear.

However this isn't univerally true for all structured lines.  For example, the definition of `##contig` in section 1.4.7 is:

> The structured \texttt{contig} field must include the ID attribute and typically includes also sequence length, MD5 checksum, URL tag to indicate where the sequence can be found, etc.
> For example:
> \begin{verbatim}
> ##contig=<ID=ctg1,length=81195210,URL=ftp://somewhere.org/assembly.fa,...>
> \end{verbatim}

"Typically includes" isn't exactly a tight term for describing what the default fields are.  Why do our contig lines not quote all those values, given they're not explicitly defined and are simply examples of things that may occur.  Is the fact that our examples always quote "species" but not "md5" is because we view "species" as a non-default field and "md5" as a default field?  Or is it because in our examples the species is Homo sapiens and we decided to quote because of the space?

If I had to interpret this with no knowledge of what happens in practice, I'd say the ID, length and URL are default fields and the "..." indicates it can be followed by extra fields, but that's simply not true.

We later have an example here:

> ##contig=<ID=20,length=62435964,assembly=B36,md5=f126cdf8a6e0c7f379d618ff66beb2da,species="Homo sapiens",taxonomy=x>

This ends with `taxonomy=x` which isn't quoted, and therefore is considered to be one of the default fields, and by dint of being *after* species we can determine that is also a default field (using the rule than extra fields always follow the default ones).

Put simply, I appreciate your improvements and are minded to merge as-is, but the entire "default" vs "extra" field is clear as mud.  Maybe one of the original authors would care to explain this in a better way?

I think the error here though isn't with this PR, but with elsewhere in the document.  Specifically the lack of a proper definition for `##contig`.  (Or `PEDIGREE` with the "Name_1" .. "Name_N" being used in the example, without quoting.)

jkbonfield (commented):
A good improvement to the meta-data section.  I assume we need to apply this to VCF4.4.draft.tex too though.

I also noted in my comments that we have a number of other issues related to quoting and "default" vs "extra" fields.  I understand what it's trying to say, but it only really works if we actually define the default fields, which we don't!  That's a topic for another PR though probably.

VCFv4.3.tex
jmarshall: > Put simply, I appreciate your improvements and are minded to merge as-is

That is surely a determination to be made by the VCF maintainers…

jkbonfield: > > Put simply, I appreciate your improvements and are minded to merge as-is
> 
> That is surely a determination to be made by the VCF maintainers…

Agreed, I wasn't actually going to merge, but rather poorly adding my personal view.

As for the other points I made, I note that `##INFO` uses the terms "required" and "recommended" to define "default" and "extra" fields.

However they're not really the same thing.  "Default" may mean a field whose key is defined in the VCF spec, but it could be optional.  So probably `##INFO` needs to be more explicit too (along with every other structured definition).

jmarshall:
> A good improvement to the meta-data section. I assume we need to apply this to VCF4.4.draft.tex too though.

As noted in the introductory PR comment, “The maintainers will likely wish to apply similar changes to _VCFv4.4.draft.tex_”.

Most of your comments are IMHO issues for followup clarifications. I hope the VCF maintainers will consider this PR as is and look to merge it within a finite timespan, and consider questions around which fields are default fields etc to be issues they may wish to take up for separate future improvement.

jkbonfield:
One other thought here, we know we can add our own additional meta-information lines.  However as these are not formally defined, I assume *all* keys are therefore mandatory quoted (even if numbers) as they are all to be considered as "extra" and hence string only rather than the "default" keys with differing types?  Does that need to be made more explicit somewhere?

An example would be `##GATKCommandLine=<...>`, which does indeed quote everything, although in this case it's all strings anyway.  I've also seen `##Method=<...>` too, also quoted.

(Bcftools only seems to add unstructured lines.)

jmarshall:
Suggestion from the meeting: also mention that there can be structured lines not specified by the specification, and say something about default fields (only `ID` presumably) for those and that all others need to be quoted.

jmarshall:
As per the suggestion at last month's meeting, I've added a sentence mentioning the possibility of other structured meta-information lines beyond those defined in the spec and noting that their only default field is `ID`.

I did not add anything about quoting on these lines, because IMHO the existing text makes an unclear distinction between “extra fields” (which for these lines would be all fields other than ID) and “optional fields“ (which are the ones that require quoting). Going further would require a VCF maintainer to determine what is intended here.

VCFv4.3.tex
```diff
@@ -100,24 +100,38 @@ \subsection{Data types}
...
+These typically have meanings that are obvious, or they are immaterial for processing the file, or both.
 
-All structured lines that have their value enclosed within ``$<>$'' require an ID which must be unique within their type.
-For all of the structured lines (\#\#INFO, \#\#FORMAT, \#\#FILTER, etc.), extra fields can be included after the default fields.
+A \emph{structured} meta-information line is similar, but the value is itself a comma-separated list of key=value pairs, enclosed within `\verb|<|' and `\verb|>|' characters:
+\begin{quote}
+\verb|##|\emph{key}\verb|=<|\emph{key}\verb|=|\emph{value}\verb|,|\emph{key}\verb|=|\emph{value}\verb|,|\emph{key}\verb|=|\emph{value}\verb|,|\ldots\verb|>|
+\end{quote}
+All structured lines require an ID which must be unique within their type, i.e., within all the meta-information lines with the same ``\verb|##|\emph{key}\verb|=|'' prefix.
+For all of the structured lines (\verb|##INFO|, \verb|##FORMAT|, \verb|##FILTER|, etc.) described in this specification, extra fields can be included after the default fields.
```
tcezard, line 119:
```suggestion
For all of the structured lines (\verb|##INFO|, \verb|##FORMAT|, \verb|##FILTER|, etc.) described in this specification, extra fields can be included and are in some cases required, after the default fields.
```

tcezard (approved):

VCFv4.3.tex
jmarshall: Which ones are you suggesting are required?

tcezard: I was thinking of [Number, Type, and Description for INFO](https://github.com/samtools/hts-specs/blob/f7947e7554f289a9bf0027be93be4145b97243a2/VCFv4.3.tex#L133)

jmarshall: To my mind, ID+Number/Type/Description are the “default fields” for `##INFO`.

IMHO the problem here is that §1.4 starts talking about details of INFO etc that should be in §1.4.2 etc. Disentangling that (and revisiting the mysterious default/extra/optional fields terminology) I think should be a followup to this PR.

tcezard: Fair enough.

Issue #404, opened 2019-04-29
include also full diff (not only changed pages) 


yfarjoun:
to circumvent the problem regarding largetable in latexdiff

PR #621, opened 2022-01-05
Tidy up longtable captions and labels [minor]

### LINKED ISSUE #605: `table:reserved-genotypes` multiply defined in `VCFv4.3.tex` and `VCFv4.4.draft.tex`
Building `VCFv4.3.tex` and `VCFv4.4.draft.tex` files leads to this warning:

```
LaTeX Warning: Label `table:reserved-genotypes' multiply defined.
```

This is probably because of the way the `\label` is applied in a longtable `\endhead` block rather than `\endfirsthead` so it appears only once.

I found this because my personal `.latexmkrc` includes `$warnings_as_errors = 1;`. If you want these sorts of warnings to lead to errors (which I suggest), you can add `-Werror` to `LATEXMK_FLAGS`. If you have this personal setting in `.latexmkrc` but want this to build anyway, you can workaround with

```
make LATEXMK_FLAGS="-e '\$\$warnings_as_errors=0;'"
```

jmarshall:
No changes to the text or even to the formatting. Removes a LaTeX warning.

Move `\label` to `\endlastfoot` so it is emitted once even for a multi-page table. Fixes #605.

We don't currently use `\listoftables`; nonetheless use `\caption[]` and `\caption[TOC_text]{…}` to emit only one contents entry per table.

### DIFF
--- a/VCFv4.3.tex
+++ b/VCFv4.3.tex
@@ -348,14 +348,19 @@ Fixed fields are:
 \begin{longtable}[c]{ | p{2.5cm} | p{1.5cm} | p{1.5cm} | p{10.3cm} | }
 	\hline
 	Key		& Number	& Type		& Description \\ \hline
+  \endfirsthead
+	\multicolumn{4}{l}{\small\emph{\ldots Continued from previous page}} \\[0.7ex]
+	\hline
+	Key		& Number	& Type		& Description \\ \hline
   \endhead
-    \hline
-    \multicolumn{4}{l}{} \\
-    \caption{\label{table:reserved-info}Reserved INFO keys (continued on next page)}
+	\hline
+	\multicolumn{4}{r}{\small\emph{Continued on next page\ldots}} \\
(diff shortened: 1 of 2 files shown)


jkbonfield:
Sadly I see latexdiff is failing.  I've had this recently too in other projects. 

I really struggled to get it to not produce buggy output within a few specific tables, and in the end took the approach of producing a new table environment specifically to avoid latexdiff from exploding.  Ie:

```
% To enable avoidance of latexdiff on a few selected tables.
\newenvironment{tablex}{\table}{\endtable}
```

That rather sucks though!

jmarshall:
I suspect Yossi (#404) and Michael (https://github.com/samtools/hts-specs/pull/570#issuecomment-958131494) are correct in their theory that these latexdiff failures are down to cross-references to filtered-out unchanged pages. I consider that a limitation within latexdiff that we should hope they fix in due course, and the revamped CI infrastructure will ignore these latexdiff failures rather than failing the whole thing.

PR #623, opened 2022-01-05
BEDv1.tex: remove extraneous backslash in LC_ALL

### LINKED ISSUE #613: BEDv1.tex: \verb|LC\_ALL=C| should be \verb|LC_ALL=C|
Hi- thanks for putting together the bed format specs! I think there is and extra `\` in BEDv1.tex at https://github.com/samtools/hts-specs/blob/402856123e47b5cbf20911a19be43906b46e9d01/BEDv1.tex#L429

`\verb|LC\_ALL=C|` should be `\verb|LC_ALL=C|` so that it renders as `LC_ALL` without backslash?

michaelmhoffman:
Fixes #613.


michaelmhoffman:
Manually generated [`diff/BEDv1.pdf`](https://github.com/samtools/hts-specs/files/7817501/BEDv1.pdf) since BED disabled for CircleCI.

michaelmhoffman:
@jmarshall Looks like I messed this up by including the PDF as a second commit in the pull request and then doing rebase+merge. I think that is what I messed up, anyway.

The hash referred to in the PDF only matches a commit in my fork and not in master. Should I now regenerate `BEDv1.pdf` with the correct hash and push it without using a pull request?

jmarshall:
Yes, that's what happened: the GItHub UI rebase+merge button has caused the CommitDates to change, which has changed the two commits' hashes.

Regenerating the PDF from the commit as pushed to **master** and pushing that directly would indeed be the way to fix this. You are just about within the time limit where you could first force-push **master** back to 9ddbc52d633f3c48a0a913b56314fee0bfacc568 if you wanted to… :smile:

michaelmhoffman:
Force-pushed back to 9ddbc52 and new PDF generated and pushed at 1ecad17. Thank you!

jmarshall:
:+1: All looks good now.

PR #630, opened 2022-02-02
CRAM: Relax alignment start and alignment span language.

jkbonfield:
These were previously stated to be both 0 for unmapped data. Unfortunately htslib and scramble have been writing span=1 for
unmapped data since inception. (And have issues when reading span=0, see samtools/htslib#1387)

Given the querying bug with span=0, fixing htslib to be spec compliant may lead to more problems.  So we relax the specification to permit span=1.  However the recommendation hasn't changed from the previously documented values.  We also now document that these values essentially have no meaning during decoding if the data is unmapped or multi-reference (ref id < 0).

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -547,13 +547,9 @@ itf8 & reference sequence id & reference sequence identifier  or\linebreak{}
 -2 for multiple reference sequences.\linebreak{}
 All slices in this container must have a reference sequence id matching this value.\tabularnewline
 \hline
-itf8 & starting position on the reference & the alignment start position or\linebreak{}
-0 if the container is multiple-reference
-or contains unmapped unplaced reads\tabularnewline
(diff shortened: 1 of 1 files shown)


jkbonfield:
Note the alternative approach would be to state the recommended value should be start=0, span=1, matching the htslib current implementation, but not the htsjdk one.  This would mean, if following recommendations, older copies of htslib, samtools, etc would still be able to do range queries on the unmapped data (`samtools view in.cram "*"`).

We did ponder this briefly, but decided the smallest change of keeping the same values as before was best.  (Htslib is obviously now fixed too and no longer uses these values during queries, but it's an issue for existing copies.)

Comments welcomed naturally.

cmnbroad:
Looks good to me - and it seems to agree nicely now with the CRAI section, which already has similar language.

PR #638, opened 2022-03-28
Clarify CRAM language on empty files (PR #638).

### LINKED ISSUE #636: cram: Required number of data containers
This is regarding [_CRAM format specification (version 3.0)_](https://github.com/samtools/hts-specs/blob/ef58a8c058a06585c81bb33470b1dcaabd5a8f91/CRAMv3.pdf) (2021-10-15).

§ 5 "File structure" states that a CRAM file has "one or more data containers":

> A CRAM file consists of a fixed length file definition, followed by a CRAM header container, then one or more data containers, and finally a special end-of-file container.

Should this be _zero_ or more data containers? Otherwise, this definition invalidates `test/cram/3.0/passed/{0001_empty_eof,0100_header1,0101_header2}.cram`, which have no data containers.

jkbonfield:
Fixes #636 reported by Michael Macias.

I chose to add "optionally" as IMO it reads better than "zero or more".

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -338,7 +338,7 @@ The overall CRAM file structure is described in this section. Please refer to ot
 sections of this document for more detailed information. 
 
 A CRAM file consists of a fixed length file definition, followed by a CRAM header container,
-then one or more data containers, and finally a special end-of-file container. 
+then zero or more data containers, and finally a special end-of-file container.
 
 \begin{center}
 \begin{tikzpicture}[


cmnbroad:
👍 LGTM.

Tangentially related question, since it was mentioned in  https://github.com/samtools/hts-specs/issues/636: shouldn't [0001_empty_eof.cram](https://github.com/samtools/hts-specs/blob/master/test/cram/3.0/passed/0001_empty_eof.cram), which has no eof sentinel container, be in the failed folder, since by this definition its not a valid cram file ?

daviesrob:
I think it should be an error on CRAM, given that the EOF block has always been part of the specification.  The only reason it couldn't be for BAM was because the EOF block was added later, and allowance had to be made for the earlier files.

Of course you could note that implementations may not always check EOF presence, e.g. when doing random access.

jkbonfield:
The reason we don't enforce EOF as an error is simply the history of the code.  EOF didn't exist in CRAM 2.0.  When it got added, we needed code that could spot EOF, but also spot the absence of it and warn.  That's basically still there, despite no one ever using 2.x any more and EOF being mandatory now.

We should update our code really, but either way I agree with the idea that this should be moved to failures.

PR #641, opened 2022-04-26
Build PDFs and comment on PRs using GitHub Actions

jmarshall:
Use GitHub's recommended two-part approach[^1][^2] to separate building the PDFs (which runs in the PR branch's context and has only a read-only GH token) from commenting on the PR (which runs on the master branch and has write access to the repository).

[^1]: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
[^2]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow

See jmarshall/hts-specs#2 and jmarshall/hts-specs#3 for examples of this running in that repository.

### DIFF
--- a/.circleci/README.md
+++ b/.circleci/README.md
@@ -1,3 +1,5 @@
+[The CircleCI web hook has been deactivated in favour of GitHub Actions. This legacy infrastructure will be removed once that has bedded in.]
+
 We use [circle-ci](https://circleci.com/gh/samtools/hts-specs) to build the pdfs and highlight the differences for each pull-request. This is done at several steps.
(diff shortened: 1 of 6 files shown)


jmarshall:
The CircleCI integration has not been working for a while — it runs successfully and comments on PRs, but the links mostly lead to 404s rather than usable PDFs — so I have now deactivated it.

To be precise, I have removed the CircleCI web hook and PAT from the hts-specs repository; I don't know if anything ought to be deactivated on the CircleCI side too.

PR #652, opened 2022-06-12
Fix minor typo in VCF specs

kojix2:
Dear hts-spec team. Hello.

I was reading hts-spec's `5.4 Specifying complex rearrangements with breakends` today, and I am reporting that I have found what appears to be a minor typo.

Have a nice day.

### DIFF
--- a/VCFv4.1.tex
+++ b/VCFv4.1.tex
@@ -525,7 +525,7 @@ The example shows in order:
 
 An arbitrary rearrangement event can be summarized as a set of novel \textbf{adjacencies}. Each adjacency ties together $2$ \textbf{breakends}. The two breakends at either end of a novel adjacency are called \textbf{mates}.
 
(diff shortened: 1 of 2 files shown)


d-cameron:
This typo was fixed in VCF 4.3. We will discuss whether we should be fixing typos in old versions of the specifications in the next GA4GH File Formats meeting (The next one is 2022-06-21 2100 GMT. They public meetings and you are welcome to join and contribute). Generally speaking, we don't backport clarification to all earlier versions of the specifications and leave published specifications as-is.

kojix2:
Thank you @d-cameron. 

I recently used VirusBreakend with docker, and the Kraken2 database uses genotype D HBV sequences by default, but since genotype B and genotype C are more common in Asia, accurate breakpoints may not be detectable, and I wanted to replace the HBV sequences, but I did not know how to do it.

I didn't notice v4.3 because v4.2 is the first thing that comes up in a Google search.
If it is fixed in the latest version, then it seems to be fine.
Should I close this issue?

PR #659, opened 2022-06-22
Tighten CRAM container language.

### LINKED ISSUE #645: cram: Define container header fields for the header container
This is regarding [_CRAM format specification (version 3.0)_](https://github.com/samtools/hts-specs/blob/3a17aaf64515b51023b29267fb69833f31adfaee/CRAMv3.pdf) (2022-04-12).

Recently, an extra description was added to § 7 "Container header structure" describing the alignment start and span fields depending on the reference sequence ID (https://github.com/samtools/hts-specs/commit/132abaf2e011d15b42975bde377bbf37feb285b0). This focuses on data containers but omits discussion about the CRAM header container. The following table shows that different writers emit different combinations of values for the header container.

| library                 | reference sequence ID | alignment start | alignment span | example               |
|-------------------------|----------------------:|----------------:|---------------:|-----------------------|
| [HTSJDK < 2.22.0]       |                     0 |              -1 |              0 | [1301_slice_aux.cram] |
| [HTSJDK >= 2.22.0]      |                    -1 |               0 |              0 | N/A                   |
| [HTSlib] (as of 1.15.1) |                     0 |               0 |              0 | [0100_header1.cram]   |

Like the EOF container, the header container is a special case, but unlike the EOF container, the header container's header is undefined (and unintuitive, given the three cases above). There should be a more thorough description and a set of values defined to avoid this ambiguity.

[HTSJDK < 2.22.0]: https://github.com/samtools/htsjdk/blob/5445b9081a057253b66ed15566bba1e1854202e8/src/main/java/htsjdk/samtools/cram/build/CramIO.java#L272
[1301_slice_aux.cram]: https://github.com/samtools/hts-specs/blob/3a17aaf64515b51023b29267fb69833f31adfaee/test/cram/3.0/passed/1301_slice_aux.cram
[HTSJDK >= 2.22.0]: https://github.com/samtools/htsjdk/blob/f461401e38fe95362a6d3c5afd8b592964b4bd29/src/main/java/htsjdk/samtools/cram/structure/ContainerHeader.java#L153-L154
[HTSlib]: https://github.com/samtools/htslib/blob/3c44c0b9eabcb1f6e111ec1b94e287155085e523/cram/cram_io.c#L4884
[0100_header1.cram]: https://github.com/samtools/hts-specs/blob/3a17aaf64515b51023b29267fb69833f31adfaee/test/cram/3.0/passed/0100_header1.cram

jkbonfield:
The reference sequence ID, alignment start and alignment span have no
meaning in the initial CRAM header container.  We make this explicit.

Fixes #645

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -561,7 +561,7 @@ itf8 & number of blocks & the total number of blocks in this container\tabularne
 \hline
 array\texttt{<}itf8\texttt{>} & landmarks & the locations of slices in this container as byte offsets from the end of
 this container header, used for random access indexing.
-The landmark count must equal the slice count.\linebreak{}
+For sequence data containers, the landmark count must equal the slice count.\linebreak{}
 Since the block before the first slice is the compression header,
 landmarks[0] is equal to the byte length of the compression header.\tabularnewline
 \hline
@@ -571,11 +571,18 @@ byte[ ] & blocks & The blocks contained within the container.\tabularnewline
 \hline
 \end{tabular}
 
-The alignment start and alignment span values should only be utilised
-during decoding if the container has mapped data aligned to a single
-reference (reference sequence id $>= 0$).  For multi-reference
-containers or those with unmapped data, it is recommended to fill
-these fields with value 0.
+In the initial CRAM header container, the reference sequence id,
+starting position on the reference, and alignment span fields must be
+ignored when reading. The landmarks array is optional for the CRAM
+header, but if it exists it should point to block offsets instead of
+slices, with the first block containing the textual header.
+
+In data containers specifying unmapped reads or multiple reference
+sequences (i.e. reference sequence id $< 0$), the starting position on
+the reference and alignment span fields must be ignored when
+reading. When writing, it is recommended to set each of these ignored
+fields to the value 0.
+
 
 \subsection{\textbf{CRAM header container}}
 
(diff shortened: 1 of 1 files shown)


zaeleus:
The appropriateness of a field is subjective, given they are allowed to be overloaded.

For example, in the header container header, htslib uses the `landmarks` field for _block_ offsets, despite the field being defined as the "location of slices" and "landmark count must equal the slice count". The header container has no slices, so it contradicts the specification. To some implementations, this may be an appropriate field, i.e., it should be 0. However, to htslib, this is an inappropriate field (as slice offsets) and is not used as intended.

jkbonfield:
I think that's maybe a different issue, but perhaps one we can address in another commit to this PR.  Are you happy with the initial change?

zaeleus:
#645 encompassed the entire header container header, but yes, this PR deals with the three fields explicitly discussed in that issue.

A more straightforward alternative (as a suggestion):

> In the CRAM header container header, the _reference sequence id_, _starting position on the reference_, and _alignment span_ fields must be ignored when reading. In data container headers specifying unmapped reads or multiple reference sequences (i.e., _reference sequence id_ < 0), the _starting position on the reference_ and _alignment span_ fields must be ignored when reading. When writing, it is recommended to set each of these ignored fields to the value 0.

jkbonfield:
I've now further updated this PR to include information about the `landmarks` field, indicating that it is optional for the SAM header container (htsjdk doesn't use it) and if specified it is block offsets instead of slice offsets.

Issue #654, opened 2022-06-16
Clarification on absence of ? or . in MM tags


jrobinso:
This issue arose implementing support for MM in IGV. There is an apparent discrepancy between the spec and current practice when interpreting MM tags which do not contain a "?" or "." flag.   The spec says that when this flag is not present it should be assumed the modifications at skipped bases are present with low probability (i.e. equivalent to ".").   However in files produced to date without this flag the opposite is true,  i.e. nothing is known about skipped bases.   Consequently this is how IGV interprets skipped bases.   

So the request would be to change the spec to conform to current practice, or maybe more specifically existing files from past practice since the major vendors are updating their tools to always include either ? or .

@amwenger @marcus1487 jump in if I've misunderstood.

marcus1487:
I support this clarification in the specification. This should only effect legacy versions of the tag, as I would hope current/new tags would include the ? or . specifier. And at Nanopore we have just recently started to output all-context 5mC calls. So all old Nanopore files will default to the ? tag version since calls are not made at any non-CG contexts.

amwenger:
@jrobinso, your explanation is correct for PacBio's methylation caller, and I would support modifying the spec so that `?` mode is default.

The currently released version does not specify a mode but uses `?` in practice, only providing calls at CpG sites.  The next release will be be explicit about `?` mode.

jkbonfield:
I'm not convinced in changing the spec as using `?` mode means explicitly specifying the non-called bases as having confidence 0, which doesn't happen as far as I can see.

At the time of writing the spec I think I had a grand total of one real called read (it was somewhat chicken and egg), but that was called end to end and non-methylated bases were simply absent.

Looking at one of the PacBio called data sets, I do indeed see it's not specifying `.` or `?`, but has explicitly added methylation data about all bases with confidence 0.   However this is an issue with their caller and not the specification.  The spec came first as it's only recently that PB have started producing this data as far as I'm aware.  If we start changing to spec to fit the incorrect software then we'll probably then get a complaint from another vendor saying that by changing the defaults we've altered the meaning of their data which doesn't write out those runs of 0s.

jrobinso:
I was just pointing out that there is no known software or existing data files that produces files with the "." default assumption,  on the contrary all known software and files in existence use the "?" default assumption.   

[...]

PR #660, opened 2022-06-22
Recommend always using the MM ? . encoding.

jkbonfield:
The '.' code is the default interpretation, but historically tools omitting '?' and '.' have used both styles.  An explicit definition in
the MM string removes any ambiguity. 

See #654 comments for background.

I haven't updated the tag history section as this isn't doesn't technically change anything, other than adding some stronger encouragement to data producers.


jkbonfield:
I went with your simpler suggestion.

jmarshall:
Those paragraphs are a definite improvement!

PR #651, opened 2022-06-07
Add ELEMENT to PL header tag

ryankelley-lm:
This is a request to add ELEMENT to the controlled vocabulary for the PL header tag

Issue #648, opened 2022-05-31
Adding Ultima Genomics to the list of allowed SAM platforms


ilyasoifer:
Hello, 
We would like to add the Ultima Genomics platform to the list of allowed platforms for the PL tag in SAM. 
Reference [here](https://www.biorxiv.org/content/10.1101/2022.05.29.493900v1). 
I would appreciate an advice on how to best approach this

Thank you very much in advance

Ilya Soifer (ilyasoifer)
Director, Bioinformatics
Ultima Genomics

jmarshall:
Raising an issue here is the right way to get the process started. The last time we added to this list can be seen in #178 and PR #454, which have more background on what I'll summarise below — but there's also quite a lot of discussion there that goes off into the weeds so I don't recommend reading all of it!

The current PL list is

> CAPILLARY, DNBSEQ (MGI/BGI), HELICOS, ILLUMINA, IONTORRENT, LS454, ONT (Oxford Nanopore), PACBIO (Pacific Biosciences), and SOLID.

As these have accreted piecemeal over the years, there is a mixture of names of manufacturers and of the underlying technologies. As discussed on the PR mentioned, we've moved to preferring to have the names of the technologies or platforms in the PL list. The idea is that the reason for tools to consult this field is to make inferences about the error model etc, which is a property of the platform rather than of the manufacturer. (In some cases, the two are the same; in other cases, the manufacturer may have changed names several times already or there may be a different name for the platform in common use.)

As Ultima Genomics has only decloaked today :smile: you are best placed to suggest an appropriate keyword to add to the list. Probably `ULTIMA` or `UG`? Or something else that better represents your mostly natural sequencing-by-synthesis technology?

jkbonfield:
With @yfarjoun in the author list on the paper I would expect nothing less than following the correct procedures for specification updates, but a big thank you too for getting this in early.  I'd be happy with `ULTIMA` too.  If the company moves on to another unrelated technology, then it can present a new more-specific name.

As an aside, you may also wish to fill out `PM` (platform model) too as it can be useful for data consumers to distinguish between the different updates as the instrument progresses.  Unlike `PL` this is not a controlled vocabulary, so you can decide what granularity of information you deem appropriate there and it doesn't need a specification update.

ilyasoifer:
@jkbonfield, @jmarshall  - thanks for the help. 
I think we will go with the UG keyword. 
What happens next? Should I just updated the Spec tex document? 
[...]

PR #662, opened 2022-07-20
adding ultima genomics platform

yfarjoun:
....to the list of allowed values for the PL tag.


jkbonfield:
Looks like a no brainer.  Thanks.

@jmarshall Were you considering doing this at the same time as #651?  This misses a history section update, but 651 has it so we could apply that first and then amend that line at the time of merging this and let both arrive in master at the same time.

I can do the work if you wish, or assign yourself if you wish to do the leg work.

jmarshall:
Sure, I'll do the leg work on both.

On issue #648 @yfarjoun asked for a final sign off on this (and checked if any other tags were desired) from @ilyasoifer et al, and I was hoping for that confirmation before going ahead and merging this one.

Issue #664, opened 2022-08-02
SAM header encoding


d-cameron:
The SAM specifications does not specify the encoding of the header.

S1 states `Unless explicitly specified elsewhere, all fields are encoded using 7-bit US-ASCII`

S1 states that alignment lines have fields, and for headers, midway through S1.3 is says `each data field follows a format`, which implies there are parts of the header which are not "fields". For example, `header line begins with the character ‘@’`, is not described as a "field".

I recommend changing S1 to: `Unless explicitly specified elsewhere, all content is 7-bit US-ASCII encoded`.

Newline definition is also missing. Is that a can of worms that I don't want to reopen?

jmarshall:
I'm glad all other problems are solved, so that we can focus on quibbling about the text describing something for which the intention is IMHO clear… :smile:

The intention is that SAM files are text files whose entire contents are US-ASCII encoded in the obvious (8-bit) way, apart from several particular fields that are individually specified as allowing UTF-8 (i.e., Unicode characters encoded as UTF-8) in their values. Or equivalently, that SAM files are UTF-8-encoded text files without BOM but characters beyond ASCII are only allowed in certain particular field values as individually specified.

Certainly this could be expressed better — e.g., I have never been happy with the “Unless explicitly specified elsewhere” clause that seems to suggest anything could happen in some parts of the file, and would prefer something like “Apart from certain field values explicitly specified as allowing UTF-8, …”. And the bit about locales in that sentence needs some copy editing and explicitness about what it's trying to tell us.

Do you have examples where implementors have been confused as to character encoding issues in SAM files? Or examples were users have had difficulties due to such issues? (Note that in [this tabix report](https://www.biostars.org/p/9530170/) the issue was that the user did not realise his input file was UTF-16-encoded, not that he had any expectation that these tools would accept a UTF-16-encoded file.)

So IMHO we should rephrase this somewhat along the lines of my second paragraph above, but I think the intention is relatively clear to most and I don't believe character encoding issues are causing any particular confusion in SAM.

jkbonfield:
Agreed.  While it could almost certainly be tightened up, that's more from a strictly-speaking stance rather than a lack of clarity in what the intention is for encodings.

[...]

PR #670, opened 2022-09-13
Clarify SAM file encoding (mostly ASCII, some UTF-8 parts)

jmarshall:
Draft text clarifying the paragraph around file encoding, locales, and regular expressions.
Addresses #664.

I would like to spell it out in the first sentence a bit more but haven't come up with a great wording yet; something like

> SAM files are encoded in UTF-8 using mainly ASCII characters only.

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -67,8 +67,16 @@ This specification is for version 1.6 of the SAM and BAM formats.  Each SAM and
 BAM file may optionally specify the version being used via the
 {\tt @HD VN} tag. For full version history see Appendix~\ref{sec:history}. 
 
-Unless explicitly specified elsewhere, all fields are encoded using 7-bit US-ASCII \footnote{Charset ANSI\_X3.4-1968 as defined in RFC1345.} in using the POSIX / C locale.
-Regular expressions listed use the POSIX / IEEE Std 1003.1 extended syntax.
+SAM file contents are 7-bit US-ASCII, except for certain field values as individually specified which may contain other Unicode characters encoded in UTF-8.
+Alternatively and equivalently, SAM files are encoded in UTF-8 but non-ASCII characters are permitted only within certain field values as explicitly specified in the descriptions of those fields.%
+\footnote{Hence in particular SAM files must not begin with a byte order mark~(BOM) and lines of text are delimited by ASCII line terminator characters only.
+% Unicode identifies VT and FF as line break characters as well, but no one uses them in SAM.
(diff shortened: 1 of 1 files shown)


jkbonfield:
I'm happy with the new text, although I note it's draft still.

PS.  The rendering of the diff PDF is visible in Opera and MS Edge on Windows, but firefox shows the hex codes for unknown font symbol.  I'm not sure what's going on there, but maintainers only need 1 working browser for validation and the final PDF is fine. :)

jmarshall:
I was less happy with the new text, I think because it said it was UTF-8 first and then adjusted it to mostly-ASCII second. I've turned that around again, and moved most of the text out of footnotes, so I think it's clearer and with a more accurate emphasis now.

Unicode adds several more line termination characters, so I added a footnote pointing out that only ASCII line terminators are valid.

And I've proposed a second commit using @jkbonfield's suggestion on #719 of using `[:print:]` in the regex. This might be a good idea, or it might be better to leave the regex as is and add a footnote saying “uh, for UTF-8 field values we actually mean `[ -0xFFFE]` or something”.

jmarshall:
The page breaking is now horrible, splitting the table between VN and SO. I have a followup PR that rearranges the first section a bit to bring both figures in the SAM example back to the first page, which restores the previous better page breaks in the header section.

jkbonfield:
> And I've proposed a second commit using @jkbonfield's suggestion on #719 of using `[:print:]` in the regex. This might be a good idea, or it might be better to leave the regex as is and add a footnote saying “uh, for UTF-8 field values we actually mean `[ -0xFFFE]` or something”.

I'd be wary of attempting to specify unicode with a hex range.  We could, but I'm not 100% sure what "printable" even means in unicode.  Is the block replacement character (FFFD) printable?  It's there for when decoding fails I think.  Also why to FFFE?  There are more beyond that.  𐘬  :-)

I just think it's a hiding to nothing and frankly `[ -0xFFFE]` does nothing to clarify the *intention* of the regular expression.  There is no easy solution here though.

jmarshall:
I've split the second commit adjusting the header regex off as a separate PR, #726.

This PR is now a fairly straightforward clarification to a single paragraph. I'd like to get it merged so that I can prepare the followup formatting improvement PR that affects the same sections. So let's start a one week timer for this one: if there are no objections before then, I will merge this on Friday May 19th.

@jkbonfield: Feel free to review and merge this before then if you wish.

jkbonfield:
Looks good to me, thanks.  I'll merge it.

PR #676, opened 2022-10-20
VCF tandem repeat support

d-cameron:
This PR adds tandem repeat support to the upcoming VCF version 4.4 specifications.

Current support for STRs and VNTRs in VCF is poor. Although it is technically possible to do so, it only really works when the exact sequence of the sample haplotype is known and even then, it's very difficult to compare the calls made by different callers as there are many possible sets of variants that encode the differences. Left-aligning works in theory but, in practice, a single SNV difference between two almost identical haplotypes will frequently result in a completely different set of variants being called.

The design goals of this PR are as follows:
- Backward compatibility with VCFv4.3
  - That is, tools that that process earlier version of VCF should be able to parse a VCF4.4 TR VCF (even if much information is ignored)
- A unified representation that supports both STRs and VNTRs
- Relatively straight-forward translation of the custom ouputs of existing tools into VCF
- Support for non-reference repeat motifs
-  Support for multiple repeat motifs in a given locus
- Support for novel STRs not present in the reference
- Support for both haplotype-resolve allele-specific repeats and overall total repeat length
- Reduce the complexity in comparing/benchmarking different callers
- Support for complex nested repeat such as VNTRs containing STRs.

I believe this PR satifies all of the above goals except the final one in which I have favoured simplicity.

The essence of this PR is the introduction of the `<CNV:TR>` subtype and a set of TR INFO fields. The `<CNV:TR>` is defined over the entire reference repeat sequence interval, and is treated as a CNV. Repeat expansions are considered copy number gain, and repeat contractions are copy number losses. This corresponds fairly closely to how read depth based TR callers actually operate. For sequence-based TR callers, the `<CNV:TR>` provides an opportunity to report summary information in addition to the exact sequences. Although this format does involve some annoying CN calculations on behalf of the TR tool authors purely so `<CNV:TR>` can be transparently treated as `<CNV>` records by non-TR-aware tools, my opinion is that this is an acceptable trade-off to maintaining backward compatibility.

Judging from the output of TR callers, what TR consumers are typically interested in is a repeat-unit centric view/summary of each loci. Here, I have encoded this in the following new INFO fields:

```
##INFO=<ID=RN,Number=A,Type=Integer,Description="Number of repeat sequences in this allele.">
##INFO=<ID=RS,Number=.,Type=String,Description="Repeat sequence">
##INFO=<ID=RL,Number=.,Type=Integer,Description="Repeat length">
##INFO=<ID=RC,Number=.,Type=Float,Description="Count of repeats in multiples of RS/RL">
##INFO=<ID=RB,Number=.,Type=Integer,Description="Repeat bases. Total allelic length of the RS/RL repeat motif length in 
##INFO=<ID=CIRC,Number=.,Type=Float,Description="Confidence interval around RC">
##INFO=<ID=CIRB,Number=.,Type=Integer,Description="Confidence interval around RB">
##INFO=<ID=RUL,Number=.,Type=Integer,Description="Length of each repeat unit">
```

In the model in this PR, each `<CNV:TR>` allele consists of a sequence of repeat units (repeat_unit_1)count_1(repeat_unit_2)count_2. Since VCF doesn't support list-of-lists, the lists are flattened and the RN field is used to encode the length of the lists for each allele.

A few of the fields are essentially different representations of the same information, but I've kept both as different tools prefer different formations. RS is suited to STRs, whereas RL makes sense for VNTRs. Callers are free to choose RC or RB depending on the implementation. 

The most questionable inclusion is the RUL field. This is messy as it's essentially a list-of-lists-of-lists. It's designed to provide basic support for the complex repeats such as VNTRs containing STRs being assembled/called from long read based callers. It's not design for exact sequence reconstruction (VCF already encodes exacty sequence differences from the reference), but to provide a minimal amount of usable information for downstream interpretation (namely, the ability to encode the start/end/length of each VNTR).

Note that all these INFO fields are designed purely as an imprecise summary of the TR. If the actual sequence is CAGCACAGCAG then a TR caller is free to encode this either as CAG(1)CA(1)CAG(2) or just as CAG(4). If a tool knows the actual sequence then just report it using the usual VCF records and phase them with each other and the summary `<CNV:TR>` record. This allows downstream to use either the summary information (e.g. is this a pathogenic repeat?), or reconstruct the exact haplotype sequence if needed.


VCFv4.4.draft.tex
```diff
@@ -1622,6 +1693,64 @@ \subsection{Representing unspecified alleles and REF-only blocks (gVCF)}
 \end{flushleft}
 \normalsize
 
+\pagebreak
+\subsection{Representing tandem repeats}
+\label{tandem-repeats}
+
+The repetative nature of both Short Tandem Repeats (STRs) and Variable Number Tandem Repeats (VNTRs) makes identifying variants in these regions difficult.
```
cmdcolin, line 1700: picky spelling: repetitive. great effort here though

hdashnow:
This is great! Thank you for taking this on!
I'd love to see examples of all of these to make sure I'm fully understanding. For example, I'm unsure if I'm understanding RUL correctly from the description.
You refer to repeat unit and repeat sequence, which I'm assuming are the same thing. I'd pick one. Repeat unit or motif are the two I see used most frequently.
CIRB should be a range as min-max is used by some tools, and is not symmetrical around the estimate. There isn't an example, so I'm unsure if it's already set up this way.

d-cameron:
> CIRB should be a range as min-max is used by some tools, and is not symmetrical around the estimate. There isn't an example, so I'm unsure if it's already set up this way.

It's set up the same way as CIPOS so the CI values are relative to the nominal value. So if your confidence interval is [5, 20] and your best estimate is 10 then RB=10 and CIRB=-5,10.

I'll add example usage for all fields to section 5.6.

hdashnow:
Here is an example of real output from TRGT on PacBio HiFi that includes confidence intervals, motif etc. May be helpful in this discussion.

```
##fileformat=VCFv4.2
...
##INFO=<ID=TRID,Number=1,Type=String,Description="Tandem repeat ID">
##INFO=<ID=MOTIFS,Number=1,Type=String,Description="Motifs that the tandem repeat is composed of">
##INFO=<ID=STRUC,Number=1,Type=String,Description="Structure of the region">
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
##FORMAT=<ID=AL,Number=.,Type=Integer,Description="Length of each allele">
##FORMAT=<ID=ALCI,Number=.,Type=String,Description="Length confidence interval per allele">
##FORMAT=<ID=SD,Number=.,Type=Integer,Description="Number of spanning reads supporting per allele">
##FORMAT=<ID=MC,Number=.,Type=String,Description="Motif counts per allele">
##FORMAT=<ID=MS,Number=.,Type=String,Description="Motif spans per allele">
##FORMAT=<ID=AP,Number=.,Type=Float,Description="Allele purity per allele">
##FORMAT=<ID=AM,Number=.,Type=Integer,Description="Mean methylation level per allele">
#CHROM  POS ID  REF ALT QUAL    FILTER  INFO    FORMAT  cmh003372-04.GRCh38.deepvariant.haplotagged
chrX    147912051   .   CGGCGGCGGCGGCGGCGGCGGCGGCGGCGGAGGCGGCGGCGGCGGCGGCGGCGGCGGCGG    CGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCTGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGG,CGGCGGCGGCGGCGGCGGCGGCGGCGGCGGAGGCGGCGCGGGCGGCGGCGGCGGCGGCGGCGGAGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGGGCGGCGGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCTGCGGCGGCGGCGGCGGCGGCTGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGCCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCTGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCTGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCTGGCGGCGGCGGCGGCGGCTGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCTGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGTGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGGCGG 0   .   TRID=FMR1;END=147912110;MOTIFS=CGG;STRUC=(CGG)n GT:AL:ALCI:SD:MC:MS:AP:AM   1/2:656,1198:656-659,1080-1301:2,3:219,399:0(0-656),0(0-1198):0.993902,0.969115:.,.
```

gymreklab:
Thanks so much for taking this on! 

Our tools currently use sequence-based representations of alleles but as you suggest I think we could add these as extra fields.

We might also want to think about how to support these in trtools: https://github.com/gymreklab/trtools
CCing @LiterallyUniqueLogin

I have some initial comments:
* The tags that have info for each repeat unit in each allele (e.g. RS, RB) are a little confusing since they concatenate info from different alleles in a single comma-separated list. I wonder if it would be better to have those as `Number=A` and have some delimited between them to denote the boundaries between info for each allele? e.g. rather than CAG,CAG,CA,CAG for the example in section 5.6 this would be CAG;CAG,CA,CAG.

* The description of the RB field seems to be cut off at the end. I think it might want to end with "in bp"? e.g.

```
##INFO=<ID=RB,Number=.,Type=Integer,Description="Repeat bases. Total allelic length of the RS/RL repeat motif length in bp.
```
* Some descriptions in the header use "repeat sequence" and others use "repeat unit" making it occasionally unclear if they refer to the entire repeat track or the repeat unit itself. Possible rewording (which might be wrong if I misunderstand some of the fields):

```
##INFO=<ID=RN,Number=A,Type=Integer,Description="Number of distinct repeat unit sequences in this allele.">
##INFO=<ID=RS,Number=.,Type=String,Description="Repeat unit sequence">
##INFO=<ID=RUL,Number=.,Type=Integer,Description="Length of each repeat unit">
##INFO=<ID=RL,Number=.,Type=Integer,Description="Repeat unit length">
##INFO=<ID=RC,Number=.,Type=Float,Description="Count of repeat units in multiples of RS/RL">
```

Agree with the others that more examples will be helpful to see.

d-cameron:
> Some descriptions in the header use "repeat sequence" and others use "repeat unit" making it occasionally unclear if they refer to the entire repeat track or the repeat unit itself. Possible rewording (which might be wrong if I misunderstand some of the fields):

The terminology used in indeed inconsistent and needs to be tightened up. Given the sequence composition at a loci of (CAG)10(CCT)15, are there widely accepted terms for the following?

- (CAG)
  - this PR uses a mix of "repeat unit" and "repeat sequence". Should be "repeat unit" and the sequence of it be "repeat unit sequence"
- (CAG)15
  - this PR uses "repeat sequence"
- (CAG)10(CCT)15
  - this PR doesn't directly refer to the full sequence composition of the ALT allele. Is there a widely used term for this?
  
"repeat sequence" is a particular bad term as it can be interpreted as refering the any of these three concepts. If there's commonly accepted terminology in the TR community for these, I'll update the PR and the INFO field names to reflect this (e.g. RS->RUS)

@ahwagner @reece : does VRS have terminology for these? If so, I can align this VCF terminology to match.

egor-dolzhenko:
Thank you very much for putting this together! The proposal looks great to me overall and I agree with the comments above.

As @gymreklab already mentioned, I think it would be useful to add an optional field to store the entire repeat allele sequence. This way, RS / RN fields can encode an imprecise, high level summary of the TR structure while the new field can encode the actual TR sequence when it is available.

This will be especially useful for long & messy VNTRs that would otherwise correspond to hundreds of variants per allele like `precise_alt1` / `precise_alt1` in the provided example.

d-cameron:
> * The tags that have info for each repeat unit in each allele (e.g. RS, RB) are a little confusing since they concatenate info from different alleles in a single comma-separated list. I wonder if it would be better to have those as `Number=A` and have some delimited between them to denote the boundaries between info for each allele? e.g. rather than CAG,CAG,CA,CAG for the example in section 5.6 this would be CAG;CAG,CA,CAG.

This list-of-lists problem has come up before in VCF (e.g. #133 back in 2016 - the issue in question was resolved by redefining the field to not need a list-of-lists) and unfortunately is out of scope for VCF 4.4. 4.4 is intended to be a 'small' update that is primarily a clean-up the the many issues the VCF format has with the representation of SVs/genomic rearrangements.

One issue with proper list-of-list support is that there aren't any reserved/disallowed characters left that could be used as separators. Semicolon is already the INFO field separator so your example of `RS=CAG;CAG,CA,CAG` actually gets parsed as `RS=CAG` `CAG,CA,CAG` which results in a parsing error because `CAG,CA,CAG` isn't of the form `key=value(,value)*`. Quote escaping doesn't work either because VCF already defines escaping to use percent encoding so, while we could encode the semicolon `RS=CAG%3BCAG,CA,CAG` this presents a significant barrier to adoption as, as far as I'm aware, many widely used VCF parser implementations don't actually implement percent encoding of INFO fields (e.g. htsjdk is a VCFv4.2 parser and percent encoding was introduced in VCFv4.3).

The final hurdle is the `RUL` field which is actually a list of list of lists so to encode that, we'd a second separator character. ','/%2C would be a decent candidate but overall, the implementation burden of using a helper length field for flattening/rehydrating does indeed exist, but is isolated to TR tools with minimal scope creep into the rest of the VCF ecosystem (it's for this same reads that I've used `<CNV:TR>` instead of `<TR>` or `<STR>`/`<VNTR>`.

In the simple case of RN=1 for all ALTs, then the fields can be interpreted in the simple `Number=A` fashion. It only gets complicated when you have multiple different repeat units in the same locus.

The other design option is to use a string encoding. Something like `TR=(CAG)5(CA)1(CAG)4,(CAG)30` would work and does not involve any nested lists. My concern that that is that it moves the parsing burden downsteam and is more vulnerable to misuse. I went with multiple fields as it enforces the conceptual model and prevents an implementation writing something like `((CAG)5(CA)1)5` as there is no way to represent such a nested definition in the proposed model.

d-cameron:
`RB` could be a `Number=A` field. Is it useful to have a breakdown of the total length of each constituent repeat motif, or would a single overall length per loci be sufficient?

d-cameron:
> I think it would be useful to add an optional field to store the entire repeat allele sequence
> This will be especially useful for long & messy VNTRs that would otherwise correspond to hundreds of variants per allele like `precise_alt1` / `precise_alt1` in the provided example.

This is already supported by VCF. There are no restrictions on redundant information in the representation - just write out the REF as the full sequence over which the `<CNV:TR>` is defined, and write your halpotypes as your ALTs. In the spec example this would be:

```
#CHROM POS ID REF ALT QUAL FILTER INFO FORMAT sample
chr1 100 cnv_notation T <CNV:TR>,<CNV:TR>
chr1 100 full_sequence_in_single_record TCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAG TCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAGCAG,TCAGCAGCAGCAGCAGCACAGCAGCAGCAG
```

lfearnley:
@d-cameron: there's very little that I'd add at this point - it's looking great, and thank you for taking on formalising this!

I'd suggest that using "repeat motif" instead of "repeat unit" might remove some ambiguity, but realise that this could also result in other nomenclature issues downstream.

gringer:
A few questions, based on my own observations of tandem repeats in DNA:
* What should be done for cryptic repeats that have variants in one or more repeated units (e.g. `ACGACGATGACGACG`), i.e. when the repeat unit sequence cannot be perfectly determined?
* What should be done for sub-repeats (e.g. `[(ACCGGC)(ACCGGC)(ACCAGT)(ACCCGC)(ACCAGT)]3−5`), i.e. when there are multiple answers for "What is repeated?"
* What should be done when the repeat unit length varies within a repetitive region (e.g. see [here](https://twitter.com/gringene_bio/status/1479289673440628736))?
* What should be done when multiple repeat elements with different repeat unit lengths are present in the same region (e.g. see [here](https://twitter.com/gringene_bio/status/1264056123046326274))?
* Is there any desire to encourage [left normalisation](https://genome.sph.umich.edu/wiki/Variant_Normalization) (or similar) in the repeat representations?

I presume from your last `CAGCACAGCAG` example that the best guess at the fundamental repetitive element would be used.

d-cameron:
> What should be done for cryptic repeats that have variants in one or more repeated units (e.g. ACGACGATGACGACG), i.e. when the repeat unit sequence cannot be perfectly determined?

That's up to the caller. That said, I'll update the PR so that RS must contain only IUPAC DNA bases (with ambiguous bases allowed).

> What should be done for sub-repeats (e.g. [(ACCGGC)(ACCGGC)(ACCAGT)(ACCCGC)(ACCAGT)]3−5), i.e. when there are multiple answers for "What is repeated?"

Again, up to the caller whether they want to have a 6bp repeat unit, or a 30bp one.

> What should be done when the repeat unit length varies within a repetitive region (e.g. see [here](https://twitter.com/gringene_bio/status/1479289673440628736))?

Report the length of each individual repeat using the RUL field.

> What should be done when multiple repeat elements with different repeat unit lengths are present in the same region (e.g. see [here](https://twitter.com/gringene_bio/status/1264056123046326274))?

That's already supported and repeat sequences such as (CAG)6(AAT)20(ACTGGG)15 can be encoded without issue. What the PR doesn't support is complex multi-level nesting of repeats (such as your [(ACCGGC)(ACCGGC)(ACCAGT)(ACCCGC)(ACCAGT)]3−5 example) and I've left that out as a) they're not supported by many tools, b) the tools that do support them do not do so in a consistent manner, c) supporting them complicate the specifications considerably.

I expect that tools that do support them would write their own custom INFO fields in addition to the standard ones (e.g a repeat description field such as `RD=[(ACCGGC)(ACCGGC)(ACCAGT)(ACCCGC)(ACCAGT)]4`).

> Is there any desire to encourage [left normalisation](https://genome.sph.umich.edu/wiki/Variant_Normalization) (or similar) in the repeat representations?

Not really. Left normalisation doesn't achieve it's desired goal of standardising variant representation when applied to repetitive regions. A single base pair difference between otherwise identical haplotypes can completely change the nominal variants. Furthermore, many (most/all?) variant normalisation algorithms don't respect phasing and treat each individual variant independently. Left-normalising each variant independently can change the order of variants thus resulting in a different haplotype post-normalisation - something that everyone should agree is a very bad outcome.

reece:
@d-cameron Nice work. Very clear goals and descriptions.

See https://vrs.ga4gh.org/en/stable/terms_and_model.html#repeatedsequenceexpression for VRS terms.

I agree with your decision to use "repeat unit", which corresponds to `seq_expr` in the VRS text. An additional wrinkle/annoyance for VRS was the desire to handle large sequences by reference (DerivedSequenceExpression). 

VRS allows for count expressions (e.g., CAG(15..20)), rather than firm counts (CAG(15)).  Do you imagine that count ranges would be permitted in VCF?

ahwagner:
adding to @reece's comments above re: VRS, I wanted to follow-up on my comments from the VRS/VCF alignment call.

In particular, I want to draw attention to the distinction we make between CNV as a [Systemic Variation](https://vrs.ga4gh.org/en/latest/terms_and_model.html#systemic-variation) concept, and Repeated Sequence Expression Alleles (RSEs) as a [Molecular Variation](https://vrs.ga4gh.org/en/latest/terms_and_model.html#molecular-variation) concept. Disambiguating these concepts was, IMO, a key advancement of VRS 1.2+. This relates to my only notable concern about this PR: the categorization of STRs and VNTRs as subclasses of CNVs. I think it is important that the terms used to describe TRs / RSEs are kept distinct from the terms used to describe CNVs, to help with the semantic distinction between these concepts as single molecule vs. multi-molecule / systemic.

d-cameron:
> to help with the semantic distinction between these concepts as single molecule vs. multi-molecule / systemic

These concepts are already conflated in VCF. Depending on the INFO fields, a CNV can be a single molecule change (i.e. an allele-specific copy number change), or systemic. VCFv4.4 has provided some clarification in that the Number=A INFO CN fields are ASCN fields, and the Number=1 FORMAT CN fields are overall CN.

d-cameron:
> VRS allows for count expressions (e.g., CAG(15..20)), rather than firm counts (CAG(15)). Do you imagine that count ranges would be permitted in VCF?

The CIRC and CIRB fields are used to support such ranges.

hdashnow:
The figure is very helpful!

Jong-hun-Park:
The figure is helpful! I think there are minor typos in the second example (CACACA, (CA)3) in the figure and the corresponding text Figure 11 (page 31). The second example is probably meant to be RUS: CA, and RUB: 2, 2, 2.

PR #684, opened 2022-11-09
SAM: add a sentence on case-insensitivity of RG PL (PR #684)

### LINKED ISSUE #679: sam: Clarify case-sensitivity of read group platform values
This is regarding _Sequence Alignment/Map Format Specification_ (2022-08-22).

§ 1.3 "The header section" lists valid values for read group platforms:

> Valid values: `CAPILLARY`, `DNBSEQ` (MGI/BGI), `ELEMENT`, `HELICOS`, `ILLUMINA`, `IONTORRENT`, `LS454`, `ONT` (Oxford Nanopore), `PACBIO` (Pacific Biosciences), `SOLID`, and `ULTIMA`.

Given the typewriter (`tt`)/monospace font used for these values, I interpret these are to be parsed as is, i.e., they are case-sensitive. However, there seems to be many files and datasets that failed to follow the spec in this way. For example, the [1000 Genomes Project](https://www.internationalgenome.org/) uses the lowercase form of `ILLUMINA`:

```
$ samtools head https://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase3/data/HG00096/high_coverage_alignment/HG00096.wgs.ILLUMINA.bwa.GBR.high_cov_pcr_free.20140203.bam | grep "^@RG"
@RG ID:H7AGF.1	PL:illumina	PU:H7AGFADXX131213.1	LB:Solexa-206008	SM:HG00096	CN:BI
@RG ID:H7AGF.2	PL:illumina	PU:H7AGFADXX131213.2	LB:Solexa-206008	SM:HG00096	CN:BI
```

Are read group platform values supposed to be case-_insensitive_, or are headers using the lowercase form not spec-compliant?

jkbonfield:
This is not changing what is valid / permitted, and indeed this hopefully clarifies it further.  However the practicality of dealing with wide-spread non-compliant data with lowercase PL values is that tools may wish to be lenient and use case-insensitive matching.

Fixes #679

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -330,7 +330,8 @@ The {\tt TP} field is often omitted, which implies linear.}\\\cline{2-3}
   & {\tt PI} & Predicted median insert size, rounded to the nearest integer.\\\cline{2-3}
   & {\tt PL} & Platform/technology used to produce the reads. \emph{Valid values}:
   {\tt CAPILLARY}, {\tt DNBSEQ} (MGI/BGI), {\tt ELEMENT}, {\tt HELICOS}, {\tt ILLUMINA}, {\tt IONTORRENT}, {\tt LS454}, {\tt ONT} (Oxford Nanopore), {\tt PACBIO} (Pacific Biosciences), {\tt SINGULAR}, {\tt SOLID}, and {\tt ULTIMA}.
-  This field should be omitted when the technology is not in this list (though the {\tt PM} field may still be present in this case) or is unknown.\\\cline{2-3}
+  This field should be omitted when the technology is not in this list (though the {\tt PM} field may still be present in this case) or is unknown.
+  The values should be written as described in uppercase, however due to the existance of public data with lowercase values tools should also accept lowercase when decoding.\\\cline{2-3}
   & {\tt PM} & Platform model. Free-form text providing further details of the platform/technology used.\\\cline{2-3}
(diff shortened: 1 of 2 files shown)


jkbonfield:
Updated wording.

jmarshall:
Also need to update _test/sam/failed/hdr.RG6.sam_, which lists this file as failed due to `PL:illumina`.

jkbonfield:
This raises an interesting question on the use of the test data.  From above:

> This is not changing what is valid / permitted, ...

I'm hesitant to move test/sam/failed/hdr.RG6.sam to the test/sam/passed directory, as that implies we consider it to be valid.  We are clearly still stating that it is not strictly valid.  However we are saying that tools need to be lenient and should consider being pragmatic and implement checks in a case-insensitive manner due to widespread non-adherance to the specification.

I feel if we wish to have lowercase in the passed folder, then we're pretty much also stating that we should just make the rule case insensitive and drop the statement that it should be in uppercase.

So I've simply removed the test file, as a middle ground.

Ideally we'd have passed, failed, and warn, but personally I don't have the bandwidth to do that justice.  I think it's more in the grey area that Daniel is exploring with StrictSAM, and there are many severity layers of warning too (some of which in my personal view come under the "just write your tool to be more robust" category).

jmarshall:
I've rebased this on current **master** to avoid a conflict due to the subsequent addition of SINGULAR. And added another commit suggesting changing this to a footnote. Can't remember if we have discussed footnote vs text in the entry, but IMHO a footnote is less cluttered and better as this sentence is not really normative.

Issue #674, opened 2022-09-30
<*> scope


d-cameron:
The unknown alternative allele `<*>`  is defined against all possible alleles but, in practice, non-haplotype-assembly based callers only consider a subset of variants. The actual claims being made by gVCF callers is "no SNVs/small indels in this region", not "there are no variants in this region". Is it worth the complication of adding additional field(s) to define the scope of `<*>`?

d-cameron:
No, it's not worth the additional complexity.

PR #688, opened 2022-11-27
Added copy number section with total CN and ASCN examples

d-cameron:


d-cameron:
@lbergelson please review this PR

VCFv4.4.draft.tex
```diff
@@ -1701,6 +1706,40 @@ \subsection{Representing unspecified alleles and REF-only blocks (gVCF)}
...
+For example, a region on chr1 from position 101 to 300 (both inclusive) with allele-specific copy numbers of 1 and 2 can be represented as follows:
+
+\footnotesize
+\begin{verbatim}
+	chr1 100 . T <CNV>,<CNV> . . END=130;SVLEN=30,30;CN=1,2 GT:CN 1/2:3
+\end{verbatim}
+\normalsize
+
+All $<$CNV$>$ alleles in the same VCF record should have the same SVLEN.
+To eliminate ambiguity in variant interpretation when GT=0, all copy number haplotypes should be explicitly specified with $<$CNV$>$ alleles, even when copy number is 1.
```
maxrossi91, line 1728:
```suggestion
For example, a region on chr1 from position 101 to 130 (both inclusive) with allele-specific copy numbers of 1 and 2 can be represented as follows:

\footnotesize
\begin{verbatim}
	chr1 100 . T <CNV> . . END=130;SVLEN=30;CN=2 GT:CN 0/1:3
\end{verbatim}
\normalsize

To eliminate ambiguity in variant interpretation when GT=0, all copy number haplotypes should be explicitly specified with $<$CNV$>$ alleles, even when copy number is 1, as follows:

\footnotesize
\begin{verbatim}
	chr1 100 . T <CNV>,<CNV> . . END=130;SVLEN=30,30;CN=1,2 GT:CN 1/2:3
\end{verbatim}
\normalsize

All $<$CNV$>$ alleles in the same VCF record should have the same SVLEN.
```

There is a typo in the first sentence: the region on chr1 spans from 101 to 130 and not 300. I would also add one example using `GT=0` with the REF copy number meaning, and the disambiguation option with the two `<CNV>` ALT alleles.

VCFv4.4.draft.tex
```diff
@@ -987,8 +991,9 @@ \section{FORMAT keys used for structural variants}
 ##FORMAT=<ID=AHAP,Number=1,Type=Integer,Description="Unique identifier of ancestral haplotype">
 \end{verbatim}
 \normalsize
-These keys are analogous to GT/GQ/GL/GP and are provided for genotyping imprecise variants by copy number (either because there is an unknown number of alternate alleles or because the haplotypes cannot be determined).
-CN specifies the integer copy number of the variant in this sample.
+CN specifies the total copy number over the region defined in the $<$CNV$>$, $<$DEL$>$, $<$DUP$>$ alleles.
+When CN is present, all $<$CNV$>$, $<$DEL$>$ and $<$DUP$>$ alleles must have the same SVLEN.
```
eroller, line 995: more specifically: `When FORMAT CN is present`

VCFv4.4.draft.tex
```diff
@@ -1701,6 +1706,40 @@ \subsection{Representing unspecified alleles and REF-only blocks (gVCF)}
...
 \normalsize
 
+
+\pagebreak
+\subsection{Representing copy number variation}
+\label{cnv}
+
+To encode copy number variation, VCF uses $<$CNV$>$, $<$DEL$>$ and $<$DUP$>$ symbolic structural variant alleles, CN INFO and FORMAT fields.
+
+Allele specific copy number is specified through a $<$CNV$>$ ALT allele for each distinct allelic copy number.
```
eroller, line 1716: Can `<DEL>` and `<DUP>` also be used to specify the allele specific copy number?

VCFv4.4.draft.tex
```diff
@@ -1701,6 +1706,40 @@ \subsection{Representing unspecified alleles and REF-only blocks (gVCF)}
...
+
+All $<$CNV$>$ alleles in the same VCF record should have the same SVLEN.
+To eliminate ambiguity in variant interpretation when GT=0, all copy number haplotypes should be explicitly specified with $<$CNV$>$ alleles, even when copy number is 1.
+
+If only total copy number is known, the copy number of the segment should be defined with a single $<$CNV$>$ ALT allele with a missing INFO CN field.
+In the above example this corresponds to the following:
+
+\footnotesize
+\begin{verbatim}
+	chr1 100 . T <CNV> . . END=130;SVLEN=30 GT:CN ./.:3
```
eroller, line 1735: Could `<DUP>` be used here instead of `<CNV>` since we know there must be a duplication (i.e. CN>1) haplotype given that the total CN for the diploid region is 3.

Could we then refer to the `<DUP>` haplotype in the GT field as `./1`? This would give a more clear indication that the sample is variant and not just reporting an unknown genotype?

VCFv4.4.draft.tex
```diff
@@ -523,6 +523,9 @@ \subsubsection{Genotype fields}
       \item $\mid$ : allele is phased (according to the phase-set indicated in {\tt PS} or {\tt PSL})
     \end{itemize}
 
+    For symbolic structural variant alleles, GT=0 indicates the absence of any of the ALT symbolic structural variants defined in the record.
```
eroller, line 526: I think GT=0 should actually mean the absence of any structural variants for the region, even those no explicitly listed in the ALT column?

eroller: It would be useful to define an unspecified structural variant ALT allele similar to the `<*>` that is used for gVCF ref blocks. This could be used both for copy neutral CNV records and for partially overlapping CNVs.

VCFv4.4.draft.tex
eroller, line 1728: It doesn't seem like there will be any ambiguity for the interpretation of GT=0 when there is a symbolic SV ALT allele. It may be better to just stick with GT=0 for representing the copy neutral haplotype rather than introducing an alternative representation that would have the same meaning.

VCFv4.4.draft.tex
eroller: @d-cameron One problem with defining GT=0 for SV events like this, is that it requires that ALT cannot be ".". If ALT is not "." then the definition of QUAL is changed, which is undesirable. It seems like using some kind of INFO field rather than the ALT column to inform interpretation of GT=0 would be preferable. I would say we should use SVTYPE for that, but it has been removed in VCF 4.4 right? Maybe this is justification to bring it back?

VCFv4.4.draft.tex
d-cameron: DEL implies a CN<1, and DUP implies CN>2 since CN is now float, CN<1 does not necessarily mean CN=0.

That is DEL implies CICN=[0,1) and DUP CICN=(1,.] (with the slight complication that the actual CICN field is inclusive of both endpoints so the notation that I just used isn't allowed in an actual VCF).

VCFv4.4.draft.tex
d-cameron: It's already in the FORMAT CN definition, but yes, I can make that explicit.

VCFv4.4.draft.tex
d-cameron: SVTYPE was always redundant with ALT so there's no additional information conveyed by it.

Overall GT=0 is just a mess for SVs. If we define GT=0 as no other SVs (or SVs of the symbolic SV types defined in ALT)

The intent with going with the definition that I did was to match the existing semantics of SV callers. When an SV GT=0/1 claim they're claiming that the SV is heterozygous and, generally speaking, not making a claim about any other variant. This is slightly different to a CNV caller making a GT=0/1 claim as it's an implicit <CNV> CN=1 claim over SVLEN.

Complications for GT=0 are:
- There could be an earlier SV overlapping this position. An earlier <DUP> won't be listed in ALT for our position but it's definitely an SV that overlaps the interval
- Multiple SVLEN can be defined. It's unclear over what interval REF GT=0 is defined.
- BND claims have no interval - they're pointwise claims. GT=0 for these event types have a meaning closest to the usual REF GT meaning (although I don't know of any that actually validate that there's not a SNV/indel at POS).
- As defined, `<*>` means REF-only calls over the interval in which `<*>` is defined. SVs are not excluded so, as written, `<*>` technically excludes any symbolic SV calls starting in the interval and any breakpoints ending in the interval. I raised this a few months ago (#674) but there didn't see to be a need to clarify this at the time. This might need to be revisited.

VCFv4.4.draft.tex
d-cameron: `<DUP>` `./1` only works if you make a haploid or diploid assumption. When dealing with genomic rearrangements, these assumptions are implicitly violated. In general, my preference is for the VCF to match the claim being made as closely as possible. A CN=3 claim is subtly different than a `<DUP> `./1` claim (e.g. the latter could be based on a single molecule containing multiple copies of a given region - (e.g. from ONT, optical mapping or flow-sorting individual chromosomes)).

VCFv4.4.draft.tex
d-cameron: GT=0 is only unambiguous if all ALT records are copy number records. I've updated the wording to allow GT=0 as CN=1 if all the ALT records are copy number claims (and made the stronger claim that copy number ALTs shouldn't be mixed with non-copy number ALT).

PR #703, opened 2023-02-09
Rewrite the CRAM Base Substitution (BS) code explanation

### LINKED ISSUE #408: CRAM spec read feature clarifications

jkbonfield:
Fixes #408

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -1600,55 +1600,95 @@ Hard clip & H (0x48) & int & HC & number of hard clipped bases, SAM operator H\t
 \hline
 \end{tabular}
 
+Note for compatibility with BAM, all base comparisons should be done
+in a case-insensitive manner, and all bases written to SC, IN and BA
+data series should be in upper-case.
+
 \subsubsection*{Base substitution codes (BS data series)}
 
-A base substitution is defined as a change from one nucleotide base (reference base) to
-another (read base), including N as an unknown or missing base. There are 5 possible reference
-bases (ACGTN), with 4 possible substitutions for each base, and 20 substitutions in total.
-The codes for all possible substitutions are stored in a substitution matrix. To restore a
-base, one would use the reference base and the substitution code, resolving the base via lookup
-in the substitution matrix.
+A base substitution is defined as a change from one nucleotide base
+(reference base) to another (read base), including N as an unknown or
+missing base. There are 5 supported reference bases (ACGTN), with 4
+possible substitutions for each base.  Any other base type, such as an
+ambiguity code, must be written verbatim using the BA data series.
+
+The codes for all possible substitutions are stored in a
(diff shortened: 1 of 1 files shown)


jkbonfield:
@jmarshall This fails checks because latexdiff fails, and not because building the PDF from the main TeX fails.

The error details above show the command as:

```
make new/CRAMv3.pdf  && make -k OLD=$mergebase_sha NEW=HEAD diff/CRAMv3.pdf 
```

I'm wondering if we should replace it in pr-pds.yaml with

```
run: make ${{ env.pdfs }} && (make -k OLD=$mergebase_sha NEW=HEAD ${{ env.diffs }} || true)
```

so even when the diff fails, provided the primary document builds correctly then we're still OK.

What do you think?  I wouldn't mind if it weren't for latexdiff being such a buggy mess.  I've had numerous tables break with it sadly.   Sometimes I've rewritten them until it works, but it feels a bit like tail wagging the dog.  The preview is nice to have, but should never cause a CI failure IMO.

jmarshall:
That's what `make -k` is there for (to ignore the error if one of several diff PDFs spuriously fails), but of course it doesn't actually stop the exit status from being non-zero. We could use what you're suggested or about equivalently `make -i` to really ignore errors, but I've long been vaguely dissatisfied by the `&&` joining what are really separate commands, so I've made _run.sh_ more fancy again instead.

Rebase onto **master** again to get c044c045f617aeeb64edd6d3ebbd7a61c3eb8de9 and this should finally go away.

jkbonfield:
Thanks @jmarshall

PR #714, opened 2023-03-29
Add an MN:i tag (number of SEQ bases at time of modification tag update).

### LINKED ISSUE #646: MM tag and hard clipped alignments
The `MM` tag describes the modification string relative to the sequencing read as generated by the instrument. For instance, consider this read, where only the 3rd C is modified:

```
read_a    ACCTCGCCA
```

If the read is aligned end-to-end, the CIGAR, SEQ and modification tag would be `9M`, `ACCTCGCCA` and `MM:Z:C+m,2`.  If the mapper has hard clipped the first two bases of the read however then the modification tag is the same but the CIGAR and SEQ will be `2H7M`, `CTCGCCA`. A naive interpretation of the `MM` tag will indicate that the wrong base is modified (in this case, the 4th C from the original read). 

This situation is not currently handled by the MM tag specification. I would suggest discouraging the use of hard clips (e.g. suggest the `-Y` flag is used for minimap2) and saying something like "MM tags are invalid when the alignment contains hard clips".

jkbonfield:
This is used as a sanity check on the validity of the MM and ML tags. It holds the length of SEQ at the time MM and ML were produced and/or updated.  The intention is to provide a mechanism to detect hard-clipping has been performed with a tool that is not MM/ML aware.

Fixes #646

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -96,6 +96,7 @@ or
   {\tt MI} & Z & Molecular identifier; a string that uniquely identifies the molecule from which the record was derived \\
   {\tt ML} & B,C & Base modification probabilities \\
   {\tt MM} & Z & Base modifications / methylation  \\
+  {\tt MN} & i & Length of sequence at the time {\tt MM} and {\tt ML} were produced \\
   {\tt MQ} & i & Mapping quality of the mate/next segment \\
   {\tt NH} & i & Number of reported alignments that contain the query in the current record \\
   {\tt NM} & i & Edit distance to the reference \\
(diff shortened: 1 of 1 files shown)


SAMtags.tex
```diff
@@ -621,6 +622,16 @@ \subsection{Base modifications}
 {\tt ML} values for ambiguity codes give the probability that the modification is one of the possible codes compatible with that ambiguity code.
 For example {\tt MM:Z:C+C,10; ML:B:C,229} indicates a C call with a probability of 90\% of having some form of unspecified modification.
 
+\item[MZ:i:\tagvalue{length}]
+\hfill\\
+Tools may edit the {\sf SEQ} sequence data, such as modifying the alignment with hard-clipping.
+If the sequence is shrunk in this manner then the base offsets in {\tt MM} and {\tt MM} become invalid unless they are also updated accordingly.
```
zaeleus, line 628:
```suggestion
If the sequence is shrunk in this manner then the base offsets in {\tt MM} and {\tt ML} become invalid unless they are also updated accordingly.
```

jkbonfield: Thanks.  Fixed this now

jkbonfield:
Agreed, it's still obviously up for anyone to propose something better.  Starting with M is reasonable as it's then with the others.  MN may be better as it's them ML, MM, MN in a clump together for the sorted list of tags?  N for Number of bases?  Z was tentatively just siZe in my head I guess.

jkbonfield:
As agreed during the last conference call, I renamed MZ to MN.  (Mnenomic could be Modification Number of bases?)
I'll update htslib soon too so it's in sync before the next release happens.  (The changes there haven't appeared in any release yet.)

jmarshall (approved):

jmarshall:
I've pushed my rearrangement of the text (to put the description of MN before the motivation of what it's good for) that I did after the discussion in the last meeting.

Back to you @jkbonfield — If you're happy with this text, I'm happy for the PR to be squashed 'n' merged.

jkbonfield:
Thanks John.  Will squash and merge.

PR #723, opened 2023-05-02
Fix spurious errors in test/sam/failed/aux.fail-f1.sam et al

### LINKED ISSUE #722: sam: Irrelevant test failure in `aux.fail-f1.sam`
[`test/sam/failed/aux.fail-f1.sam`](https://github.com/samtools/hts-specs/blob/144e32acb582b414a281bf9dc06223b43609a489/test/sam/failed/aux.fail-f1.sam) tests for single-precision floating-point limits, but it also fails by having duplicate `F1` tags, which is irrelevant to the test.

jmarshall:
[I almost just committed this to **master** under the “sufficiently obvious” rule, but as it's day in @jkbonfield's timezone now here is a PR. James, you should probably just merge this if you're happy.]

Fix #722 (duplicate F1 tags) by renaming to F0/F1/F2/F3 as several other records with four aux fields in the test suite are named thus. Also fix _test/sam/passed/aux.pass-f.sam_ which has the same problem.

The F1 fields just repeat the F0 fields, while F2/F3 exercise positive and negative. It looks like F0/F1 was supposed to do the same thing, so negate the F1 values.

In _aux.fail-f1.sam_, F2/F3 are values just past `FLT_MAX` so are outside `f`'s expressible range, hence failing. F0/F1 however are just on the expressible side of `FLT_MIN` so are valid. Change these to use values smaller than the minimum [~~denormalised~~ subnormal](https://genomic.social/@johnm/110298375858348774) value representable so they underflow to 0 when evaluated as float, hence fail too.


jkbonfield:
Thanks John, Michael.

PR #724, opened 2023-05-02
Document the ambiguity of 1-base long seq with QUAL *

### LINKED ISSUE #715: sam: Quality score ambiguity when sequence is a single base
This in regard to _Sequence Alignment/Map Format Specification_ (2022-08-22) § 1.4 "The alignment section: mandatory fields".

In the following SAM record, the quality scores field (`QUAL`) is ambiguous.

```
*	4	*	0	255	*	*	0	0	A	*
```

Since there is a singe base in the sequence, the quality scores field can either be unavailable (`*`) or represent `[9]`.

jkbonfield:
This is an extreme edge case likely to never occur, but nevertheless tool implementors still need to know how to handle it.  Given it *may* be QUAL 9 or it *may* be QUAL "unknown", we treat it as always unknown.

Fixes #715

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -636,13 +636,14 @@ Each bit is explained in the following table:
 \item {\sf QUAL}: ASCII of base QUALity plus 33 (same as the quality
   string in the Sanger FASTQ format). A base quality is the phred-scaled
   base error probability which equals $-10\log_{10}\Pr\{\mbox{base is
-    wrong}\}$.
-  This field can be a `*' when quality is not stored.\footnote{Note an
-  ambiguity exists for the unlikely case of a non-`*' sequence of length 1
-  with base quality 9 (ASCII `*').  Given the ambiguity, the quality
-  should always deemed to be unavailable in this scenario.}
-  If not a `*', {\sf SEQ} must not be a `*' and the length of the quality
-  string ought to equal the length of {\sf SEQ}.
+  wrong}\}$. This field can be a `*' when quality is not stored.\footnote{In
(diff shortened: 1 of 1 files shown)


jmarshall:
I have finally got around to finishing up my draft of adjusting this footnote. (The `diff` PDF preview has gone slightly mad — the actual change is the new footnote 17 on page 9.) It's been so long that I can't remember the details, but this is a tidied-up version of the text lying around in my specs directory after discussions on this one in meetings a few months ago.

The changes are basically:
* Avoid referring to this as ambiguous — now that the spec provides a ruling (albeit in a footnote), it's not ambiguous.
* I think we discussed offering advice on what tools should do if they have a single-base QUAL of 9 to write out. The second sentence does so, and says do it for all of SAM/BAM/CRAM to avoid issues later when BAM/CRAM files are viewed as SAM. But I'd also be happy if you preferred to drop that and just have the first sentence for succinctness.

jmarshall:
On the merged commit, @colinhercus comments (https://github.com/samtools/hts-specs/commit/f7809caf3db2e1cf531c5c87e9238c56f98f0e32#commitcomment-162178132):
> In this situation Novoalign has always reduced the base quality to 8. I don't plan to change it.

(The PR recommends adjusting the base quality to 10.)

I wish you'd noticed this updated PR in the month before we merged it… :smile: Or in the two years since the issue was first noted as an issue on this repository… :smile: :smile:

I thought of looking to see if any of the base quality binning strategies used either 8 or 10 or a nearby value, but in the end we picked 10 arbitrarily. Do you recall whether you picked 8 arbitrarily or was there some reasoning behind the choice?

Issue #18, opened 2014-02-24
String Encoding


zorgiepoo:
VCF should either have an option in the header to choose a specific string encoding (e.g: utf-8, latin-1, ascii) with a default option set, or it should be documented which encoding VCF should be in.

pd3:
Addressed by https://github.com/samtools/hts-specs/commit/a09e56feac09c831a490d97b423aac5a78960650

PR #726, opened 2023-05-12
Allow for UTF-8 field values in header regular expression

### LINKED ISSUE #719: sam: Clarify the usage of UTF-8 characters in header
This is in regard to _Sequence Alignment/Map Format Specification_ (2022-08-22).

§ 1.3 "The header section" defines patterns for header lines:

> Thus header lines match `/^@(HD|SQ|RG|PG)(\t[A-Za-z][A-Za-z0-9]:[ -~]+)+$/` or `/^@CO\t.*/`.

This invalidates the following test examples:

* [test/sam/passed/hdr.PG3.sam](https://github.com/samtools/hts-specs/blob/144e32acb582b414a281bf9dc06223b43609a489/test/sam/passed/hdr.PG3.sam)
  > `@PG	ID:bwa-1	CL:echo "🐟"`

* [test/sam/passed/hdr.PG5.sam](https://github.com/samtools/hts-specs/blob/144e32acb582b414a281bf9dc06223b43609a489/test/sam/passed/hdr.PG5.sam)
  > `@PG	ID:x	DS:☢ description ☢`

* [test/sam/passed/hdr.SQ6.sam](https://github.com/samtools/hts-specs/blob/144e32acb582b414a281bf9dc06223b43609a489/test/sam/passed/hdr.SQ6.sam)
  > `@SQ	SN:tea	LN:99	DS:☕`

The text "UTF-8 encoding may be used" for the `CL` and `DS` fields does not remove the character set constraint. It also remains arbitrary as to why only some fields have this definition.

jmarshall:
Use `[:print:]` in the header regex and note that for ASCII it is equivalent to `[ -~]` and that the aim is to forbid control characters. Fixes #719.

To be honest, I'm tempted to add the extra `[]` to the `\cclass` definition and waste a bit of space each time this appears rather than add the “For brevity” sentence.

An alternative to this PR might be to just leave the regex as `[ -~]` and add a footnote explaining that this is an oversimplification for fields that allow Unicode values.


zaeleus:
Note that [UTS #18: Annex C](https://unicode.org/reports/tr18/#Compatibility_Properties) suggests `:print:` character class compatibility for Unicode as `\p{graph}\p{blank}--\p{cntrl}`, which is likely not the appropriate definition here since it includes `:blank:`. It may be better to note a property matcher instead, e.g., `\P{Other}`.

jmarshall:
Which specific `:blank:` characters are you worried about inadvertently including? (The obvious worry is TAB, but presumably that is removed by `\p{cntrl}`.)

zaeleus:
Ah, yes, you're right. I had a set operation wrong when I tested with an example. `:print:` seems sufficient for this change.

jkbonfield:
I see I was assigned this in the last meeting.

Personally my preference is `[[:print:]]` over `[:print:]` as it's a standard and ironically the extra couple of characters we add a few times is less text than the "for brevity" statement.  Not a hard "must be so" line, but if in agreement I'd prefer that before merging.  Otherwise I'm happy with it.

jmarshall:
The ball was indeed with me. See the new preview for how this uglifies `[[:rname:^*=]][[:rname:]]*`…

PR #729, opened 2023-06-20
Resolve #707: remove remaining *ADJ references

### LINKED ISSUE #707: vcf: Clarify statuses of DPADJ, CNADJ, and CICNADJ INFO keys
In the [changelog for VCF 4.4](https://samtools.github.io/hts-specs/VCFv4.4.pdf#subsection.7.1), it says that the `DPADJ` and `CNADJ` INFO keys are deprecated. However, they are still used in [§ 5.4.10 "Sample mixtures"](https://samtools.github.io/hts-specs/VCFv4.4.pdf#subsubsection.5.4.10) without definition.

Similarly, `CICNADJ` is marked as deprecated but is no longer defined. In this case, I would expect it to be marked as removed rather than deprecated.

d-cameron:
*ADJ was defined which adjacent location and/or segment the ADJ actually referred to.


daviesrob:
Fixes #707

PR #731, opened 2023-06-27
Revert defn of CRAM container size to be sum of block sizes (PR#731)

jkbonfield:
This was added as clarification in #398 after discussion in #396, but this was in error.  In our attempts to clarify and nail down these corner cases, we failed to recall that the SAM header is permitted to be padded out by non-block allocated space.

History on this decision dates back to 2013 and is show in Samtools issue samtools/samtools#1852.

There are good reasons for changing away from the decision of padding via a second block, as changing block sizes can also change block structure size (if we're using a generic shared piece of code, due to ITF8 being a variable length integer), and this in turn makes it cumbersome to handle every possible change in SAM header size.  It is far easier and simpler to just have unallocated space after the block and before the end of the container.  This is how htslib works since CRAM 3.0 and I believe how CRAMtools.jar works.

Fixes samtools/samtools#1852.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -361,6 +361,13 @@ Figure 1: A CRAM file consists of a file definition, followed by a header contai
 
 Containers consist of one or more blocks. The first container, called the CRAM header container,
 is used to store a textual header as described in the SAM specification (see the section 7.1).
+This container may have additional padding bytes present for purposes
+of permitting inline rewriting of the SAM header with small changes in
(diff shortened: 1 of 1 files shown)


jkbonfield:
TODO: Does this apply only to header container, or all containers?

jkbonfield:
Minimal update made to explicitly state the additional padding bytes are for the CRAM header container only.  If we find a compelling reason to later we can always relax this limitation while keeping backwards compaibility.

PR #740, opened 2023-08-22
CRAM 3.1 test data

jkbonfield:
This comes in two flavours

- Whole files, as CRAM 3.1, CRAM 3.0 and BAM format.  The data is enough to be a reasonable test, but not overly larger.  They comprise around 6MB worth of test files combined.
- Raw CRAM data streams encoded in multiple codecs with different codec options used.  Combined these are around 8MB.  It's possible some of this could be culled, in particular the name tokeniser.

Overall this bumps the hts-specs repo from 51MB to 68MB (most of whicj is in the .git directory).  We could cull some of the tests perhaps, but it's a starting point.


test/cram/3.1/README.md
```diff
@@ -0,0 +1,48 @@
+This directory contains a set of test files in the CRAM 3.1 format.
+
+As CRAM 3.1 is identical to CRAM 3.0 except the version number and
+choice of codecs available, this directory does not contain a large
+number of files.  It is recommended that CRAM 3.1 implementations
+first validate themselves against the CRAM 3.0 data sets, along with
+validatig the specific codecs in isolation.
```
zaeleus, line 7:
```suggestion
validating the specific codecs in isolation.
```

zaeleus (commented):
I'm able to (re)validate all the inputs using noodles' implementations.

When I initially wrote the CRAM 3.1 decoders, I used htscodecs' test set that's copied here. It's incredibly helpful to have a fairly comprehensive set to start with, so it's nice to see them included next to the specs.

PR #743, opened 2023-10-19
Add CIGAR =/X entry to version history appendix

jmarshall:
The `=` and `X` CIGAR operations were added to SAM in 07dc1c67a717 in July 2010, which was the initial addition of the TeX specification; they were not present in the previous Pages document. It's so long ago as to be barely relevant now, but it's worth adding them to the history appendix and identifying them as requiring VN:1.3 rather than VN:1.0.

See the comments on [this biostars posting](https://www.biostars.org/p/9577441/) for the motivating context.

In the SAM regexp in §1.4, write the operations in the familiar canonical order (though it doesn't affect the meaning of the regexp). Define `\cigarops{...}` to improve the formatting of lists of CIGAR operations like "M/I/D" by making the slashes non-`\tt`, and also use this in _SAMtags.tex_.

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -9,6 +9,10 @@
 \newcommand{\tagregex}[1]{{\tt #1}}
 \newcommand{\metavar}[1]{{\rm\emph{#1}}}
 
+% Use as, e.g., \cigarops{MID} to produce M/I/D with the operators in \tt
+\newcommand*{\cigarops}[1]{\cigaropsAux#1*}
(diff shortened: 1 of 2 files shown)


zaeleus:
Alternatively, `\cigarops{MIS=X}` can be referred to as "query consuming operations"; and `\cigarops{M=XDN}`, "reference consuming operations". This would remove the need for a new command and might make it more natural to read.

jmarshall:
My preference would be to keep it explicit.

It's obvious what "reference-consuming" etc mean if you read the CIGAR section. But one of the first things we learn from editing specs or maintaining software like samtools is that People Don't Read…

jkbonfield (approved):
Thanks.  Looks good to go.

PR #745, opened 2023-11-13
Local Alleles (PR #434) applied to VCFv4.4

jmarshall:
This draft PR applies #434's Local Allele changes to the _VCFv4.4_ document on top of current **master**. It is not intended to split the conversation from the existing PR, but to provide a preview PDF and demonstration of how this applies to current v4.4.

If the PR author and VCF maintainers agree, and wish to apply the local alleles additions to v4.4 instead of v4.3 (or perhaps to both), then we can force push #434's branch to this commit, close this PR, and continue the conversation on the original PR.

### DIFF
--- a/VCFv4.4.tex
+++ b/VCFv4.4.tex
@@ -482,9 +482,9 @@ See also Section~\ref{sv-format-keys} for a list of genotype keys reserved for s
       GQ		& 1		& Integer	& Conditional genotype quality \\
       GT		& 1		& String	& Genotype \\
       HQ		& 2		& Integer	& Haplotype quality \\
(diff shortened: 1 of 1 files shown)


jmarshall:
Incidentally, diff PDF previews include all pages rather than only the changed pages due to a [ghostscript bug](https://bugs.ghostscript.com/show_bug.cgi?id=707137) that was fixed two months ago but has not yet appeared in a release. So that's fun…

jmarshall:
@d-cameron: Please comment (or just do it!) on whether you wish to convert PR #434 to apply to the VCFv4.4 document by force-pushing this PR's HEAD to Yossi's PR's branch.

jmarshall:
PR #434 now contains this update.

PR #752, opened 2024-01-29
Recommend 1bp SEQ with QUAL="*" as quality-unavailable.

jkbonfield:
The change is simply a footnote indicating the presence of the ambiguity and a mild recommendation, however it's not likely to make any practical difference in real-world data where even if 1bp reads exist their quality values aren't likely to be a deciding factor over their validity.

Also in attempting to build the SAM spec my latex blew up.  This turns out to be due to #670 which added use of `\>` for non-breaking space.  Our systems here are pretty old, and latex couldn't cope with that.  However in the interests of keeping the spec as easy to build as possible, I did a trivial replacement in its own commit.


jmarshall:
IMHO PR #724 has the footnote in the better position, so I'll close this PR, cherry-pick the non-breaking space commit, and coalesce the best of both footnote texts over there.

PR #753, opened 2024-01-29
Add SAM recommendation of MAPQ=0 for unmapped reads

### LINKED ISSUE #727: sam: Recommend using 0 for mapping quality when record is unmapped
As suggested by @jkbonfield in https://github.com/samtools/hts-specs/issues/715#issuecomment-1508097708:

> as an aside, I note you're also using MAPQ of 255 for "unavailable". Commendable, but my experience is that everyone just uses 0 with unmapped data. I think this is because when FLAG 4 is set the specification states no assumption can be made about MAPQ, so it just feels cleaner to zero it out as all other fields have been.

jkbonfield:
We already have a recommendation that MAPQ of 255 should not be used. Expand on this to recommend zero is used when unmapped.

This is purely a recommendation, made for maximum compatibility, and not a specification requirement.

Fixes #727

### DIFF
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -737,6 +737,7 @@ specific software package for it to function properly.
   \item Unmapped reads should be stored in the orientation in which they came
     off the sequencing machine and have their {\sf reverse} flag bit~(0x10)
     correspondingly unset.
+  \item Unmapped reads should have mapping quality zero.
   \end{enumerate}
 \item Multiple mapping
   \begin{enumerate}[label=\arabic*]
--- a/SAMv1.tex
+++ b/SAMv1.tex
@@ -737,7 +737,20 @@ specific software package for it to function properly.
   \item Unmapped reads should be stored in the orientation in which they came
     off the sequencing machine and have their {\sf reverse} flag bit~(0x10)
     correspondingly unset.
-  \item Unmapped reads should have mapping quality zero.
+  \item Fields stated to have no assumptions on their content if {\sf FLAG}
+      bit 0x4 is set (unmapped) should have values `{\tt *}' or `{\tt 0}' as
+      appropriate.
+      \footnote{This is desirable as CRAM will not store some or all of these
+        fields for unmapped data, so round-trips should yield these default
+        values for maximum compatibility.}
+
+      Specifically this includes {\sf RNAME} (`{\tt *}', but see caveat),
+      {\sf POS} (`{\tt 0}', see caveat), {\sf CIGAR} (`{\tt *}'),
+      {\sf MAPQ} ('{\tt 0}') and {\sf FLAG} bits 0x2, 0x100 and 0x800
+      (all unset).
+
+      One caveat to this is with paired sequencing protocols where one record
+      is aligned and one unaligned within a pair.  See item 1 above.
   \end{enumerate}
 \item Multiple mapping
   \begin{enumerate}[label=\arabic*]


zaeleus:
Should this be further constrained to unmapped records with _missing_ mapping qualities? That is, recommend using 0 instead of 255. Otherwise, is this recommending that all unmapped records have their mapping qualities reset to 0?

jkbonfield:
Is there any value in another mapping quality?  By definition, if it's not mapped, the mapping quality doesn't make sense.

d-cameron:
> Is there any value in another mapping quality? By definition, if it's not mapped, the mapping quality doesn't make sense.

The use case for non-zero mapq is for an aligner that can confidently map to a sequence that is not included in the reference genome.

The primary use case for this is graph genome aligners. Graph genome aligners can have a mapq w.r.t. to graph but when projected onto a linear reference, that that alignment position is fully contained within an insertion. How should these be represented? The two approaches I've seen in the wild are:

1) Map with an insertion CIGAR to the insertion location (Seven Bridges aligner approach)
- Problematic since the alignment position is not defined since there is no align base and that's how POS is defined
- No issue with mapq representation

2) Unmapped read placed at the insertion location (DRAGEN aligner approach)
- Problematic since `samtools fixmate` moves the location
- This mapq0 change breaks this representation notation

Making this mapq change will cause an existing implementation in widespread usage to break/become non-compliant. Ok to do, but it would be good to have a) some time to fix the implementation, and b) spec-defined guideance on how non-reference alignment should be handled (both those with INS liftovers and those that do not map to reference contig (e.g. graph decoy sequence). For example, DRAGEN retains information about originating alignment via a SA-style custom SAM tag. This will be a lossy operation if mapq is forced to 0.

jkbonfield:
I disagree.  The specification already states "If 0x4 is set, no assumptions can be made about RNAME, POS, CIGAR, MAPQ, and bits 0x2, 0x100, and 0x800" so adding a recommendation to use zero values for such cases cannot break existing implementations any more than they are already broken.  Spec compliant software should ignore those fields.  All this PR is doing is trying to encourage people to stop using them in such situations.

My recommendation is people should use additional aux fields if they need a way to represent something which is currently outside of the specification, or to raise a PR here requesting things are changed.

Mind you, that sentence above also rules out the well established (since day one) strategy of "placed" reads that are unmapped by POSitioned along side their mate pair, which is rather unfortunate.

jkbonfield:
On the related issue of POS having no meaning for reads with no aligned bases, I hit this many years ago when implementing Gap5.  It's worse than it looks too as it's still problematic even when you do have aligned bases.

Insertions basically appear before the next base, so a read that starts with an insertion could have CIGAR of  `99I10M` and a POS for that `M`.   As we get shorter and shorter reads all with the same insertion but less mapped data at the end, POS doesn't change, and `99I1M` is the same location.  Although it's undefined, logically, so is `99I` (and 0M).  So POS being the base after does make some sort of sense here.

Fundamentally though this breaks indexing.  Even the normal case of `90I10M` means our insertions are only known about after they've been and gone.  Consider:

```
POS      12345      6789                                                        
Ref      AGCTA      AGCT                                                        
Sample   AGCTAGGTTGGAGCT                                                        
Seq0     AGCTAGGTTGGAGCT   Pos 1, CIGAR 5M6I4M
Seq1     AGCTAGGTT         Pos 1, CIGAR 5M4I                                    
Seq2            TTGGAGCT   Pos 6, CIGAR (2P)4I4M                                
```

A query for chr:5-5 wouldn't give us Seq2.  (And a query for chr:6-6 wouldn't give us Seq1.)  We need more nuance here.

Processing each read iteratively in a pileup orientation also doesn't work and would give us a set of reads with an insertion at position 5 that doesn't include Seq2, as it hasn't yet started.  Then when we get to position 6 we're told there was additional data we should have been using for computing the insertion consensus.  Doh!

In Gap5 I replaced every CIGAR with 1D<cigar> and decremented POS by 1, and then during processing removed the initial deleted base.  VERY ugly, but it made the pileup iterator work as our position coordinates are now the base before an insertion rather than after the insertion.  It's a miss-design of the spec IMO, but we are where we are and I don't really know how to retrofit it without resorting to such hacks.

jmarshall:
@jkbonfield to replace this with text to specify that RNAME/POS/MAPQ/CIGAR/etc/etc should all be zero / `*` as the default/placeholder values in this situation. Add to Recommended Practice section rule 4 “Unmapped reads” being careful to remain consistent with the other rules in that section.

Possibly add a footnote (or new rationale section):

> In particular, CRAM does not preserve MAPQ on unaligned reads so it would always be read back as zero.

PR #754, opened 2024-01-29
Document CRAM FP is delta-encoded

### LINKED ISSUE #747: CRAM: Need to improve feature positions description
The CRAM `FP` data series is described as "position of the read feature" in the CRAMv3 spec. [here](https://github.com/samtools/hts-specs/blob/346a94a99/CRAMv3.tex#L805) and [here](https://github.com/samtools/hts-specs/blob/346a94a99/CRAMv3.tex#L1583).  This seems to imply an absolute position, but in fact it's relative to the previous feature, as can be seen in the encoders for [htslib](https://github.com/samtools/htslib/blob/d1aae521/cram/cram_encode.c#L680), [htsjdk](https://github.com/samtools/htsjdk/blob/3964abef/src/main/java/htsjdk/samtools/cram/encoding/writer/CramRecordWriter.java#L232) and [noodles](https://github.com/zaeleus/noodles/blob/dc00c91cd/noodles-cram/src/writer/record.rs#L473C1-L473C1).  To make this clear without having to read existing implementations, the "Read feature records" section could do with a line or two explaining how the `FP` data series works.

jkbonfield:
Fixes #747

(Also updated the languishing CRAM4 PR with the same changes.)

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -803,7 +803,7 @@ features in each record\tabularnewline
 FC & encoding\texttt{<}byte\texttt{>} & read features codes & see separate section\tabularnewline
 \hline
 FP & encoding\texttt{<}int\texttt{>} & in-read positions & positions of the read
-features\tabularnewline
+features; a positive delta to the last position (starting with zero)\tabularnewline
 \hline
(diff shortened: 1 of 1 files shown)


daviesrob:
Looks OK to me.

PR #768, opened 2024-04-22
Add a LICENSE.md file for the various test files.

jkbonfield:
These are my own work on behalf of GRL.  SAM / VCF may take more digging to correctly attribute ownership in order to categorically obtain a license statement, so they have been omitted in this commit.

WIP: will consider SAM next, but likely its own commit.


jmarshall:
First trivial comment is that, in this repo, I'd rather see these as plain text in files named `LICENSE`.

jkbonfield:
I'm curious why.  We already have README.md and MAINTAINERS.md.  Markdown is, by definition, both plain text to people doing a `git clone` and viewing with e.g. `less`, as well as a bit better formatted for web browsing.

I could switch to just plain text, but I'd like to know the logic behind it first.

jkbonfield:
Updated a few small things.

- renamed to LICENSE

- Expanded on examples/vcf/simple.vcf.  I still don't know the origin.  It's in VCF early specs as an example, but it's also in VCFtools in various forms (by different authors as it progresses through the spec versions).  No idea if VCF example in the spec is copied from VCFtools or vice versa, but they have differing licenses (including LGPL) to make matters more confusing.

d-cameron:
Happy with the same license as everything else.

PR #773, opened 2024-06-18
Add ID subfield to VCF v4.3 test file's structured header

### LINKED ISSUE #760: vcf: Handling structured header records with missing IDs in VCF 4.1/4.2
In VCF 4.3, the concept of structured and unstructured records was introduced, which requires all structured records to include an ID field (§ 1.4 "[Meta-information lines](https://samtools.github.io/hts-specs/VCFv4.3.pdf#subsection.1.4)" (2022-11-27)). Was this a clarification or a change from previous versions?

VCF 4.1 and 4.2 don't define what header lines are. How are they expected to be parsed? Specifically, I'm interested in how structured header records with missing IDs are supposed to be handled/interpreted.

For example,

```
$ curl --silent https://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh38/archive_2.0/2017/clinvar_20170516.vcf.gz | gzip -d | head | grep --extended-regexp "^##(fileformat|ID)"
##fileformat=VCFv4.1
##ID=<Description="ClinVar Variation ID">
```

jmarshall:
In VCF v4.3 structured meta-information lines must have an ID subfield. Split the value into identifier-like ID and text Description subfields to demonstrate the structured benefit, though for a non-INFO/FORMAT/etc line there's probably nothing requiring ID's value to be identifier-like. Fixes #760.

This leaves the similar 4.1 and 4.2 files as is, as it is less clear whether ID-less structured headers are merely discouraged there and/or is unspecified and largely left up to implementations — cf #760.


zaeleus (approved):
The fix looks good to me.

> This leaves the similar 4.1 and 4.2 files as is, as it is less clear whether ID-less structured headers are merely discouraged there and/or is unspecified and largely left up to implementations — cf #760.

It not up to implementations if an authoritative example is expected to pass. Leaving them forces parsers to accept such header records, despite the behavior being undefined in those versions.

PR #776, opened 2024-07-15
Update base modification test files to use the final MM/ML tags

jmarshall:
Change the test files so they use `MM:Z` and `ML:B:C` in their final capitalised forms.

Also add HTSlib's _MM-explicit.sam_, which exercises the skipped bases modification state indicator character, and add an example `MN:i` field to _MM-multi.sam_, to keep it in sync with HTSlib's copy.

This is fairly trivial so should be able to be merged quickly.


jkbonfield:
Looks good to me.  Thanks John

PR #799, opened 2024-10-28
Explicitly permit eg N+m in MM tag (PR#799)

### LINKED ISSUE #785: MM tag preferred format for TAPS data
Hello,

TAPS is a methyl-seq method where mC are converted into T.

In SAM spec, most of the mC examples highlight “C+m”, however C+m does not work off the shelf with TAPS data since the mC are represented in the SEQ as Ts and not Cs.

Alternatives to represent the mC mod could include T+m or N+m however these are not officially listed in the table of combinations. Is there any recommendation on which format to settle on?

Thanks,
James

jkbonfield:
The text already states that an unmodified base of N means we count any base type, but base N code N in the table is a little misleading as to the intention.  It was intended to mean any unspecified modification, in the same way C+C is any unspecified C mod, but in this case it's against all bases rather than a specific base type.

However that doesn't solve the issue of whether we can record specific mods against any "fundamental" source base.  Clarified this by adding an extra line to the table and some text.  (However note this doesn't necessarily imply downstream processing tools will not do any compatibility assessment and reject N+m when the SEQ base is a T.)

Fixes #785

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -532,6 +532,7 @@ For example `{\tt C+76792,57;}' is the same as `{\tt C+h,57;}'.
 
 An unmodified base of `{\tt N}' means count any base in {\sf SEQ}, not only those of `{\tt N}'.
 Thus `{\tt N+n,100;}' means the 101st base is Xanthosine (n), irrespective of the sequence composition.
+A fundamental base of `{\tt N}' may also be used with a base-specific modification code to force the counting to be applied per base rather than per base-type.
 
 The standard code types and their associated ChEBI values are listed
 below, taken from Viner {\it et al.}%
@@ -567,6 +568,7 @@ G & G &       & Ambiguity code; any G mod & \\
 \hline
 N & n & Xao   & Xanthosine              & 18107 \\
 N & N &       & Ambiguity code; any mod & \\
+N & any &     & Mod applied to any base & \\
 \end{tabular}
 \end{center}
 


jmarshall:
Previous sentence says “An unmodified base of ‘N’”; this new sentence says “A fundamental base of ‘N’”. Is there a distinction between _unmodified_ and _fundamental_ (if so, let's describe it)? Otherwise, let's use one or the other.

jkbonfield:
I reviewed the change again and decided "fundamental" base is the clearer term.

I note the table below this has unmodified base, but I'm unsure whether that should be changed or not.  For the language "C+hm,<list>" it is the "C" here we refer to as "fundamental", given unmodified is a confusing term to use when describing a base that is infact modified.  However for the table of from-this to-that unmodified may be easier to understand and we're not explicitly referring to a string or serialisation syntax for that table.

I'm not overly happy with the first paragraph of "_from the primary unmodified sequence as originally reported by the sequencing instrument_", given the sequencing instrument is likely the thing stating this is modified.  However typically it does that in stages, recording the unmodified base type in FASTA/FASTQ and a modification as a side-channel.  I want to say as recorded in the SAM SEQ column as that forbids things like h, m, etc, but then it gets mired in reverse-complementing confusions.

I think for now it's better, but not perfect.  Ideas welcomed, although maybe best dealt with in another PR if it's a big rewrite.

PR #803, opened 2025-01-07
Clarify the name tokeniser uncomp_len calculation (PR #803)

### LINKED ISSUE #802: Name tokenizer codec clarification
I'm trying to finish up Yash's CRAM 3.1 codecs for htsjdk, and I came across an ambiguity in the name tokenizer spec that could use clarification. It currently states "The serialised data stream starts with two unsigned little endian 32-bit integers holding the total size of uncompressed name buffer and the number of read names".

Based on the values I see when inspecting samtools CRAMs/code, the uncompressed name buffer size assumes that the uncompressed names are formatted into a buffer the way htslib formats them, i.e., on decode, it would match the length of the read name data that is reconstructed from the stream(s), PLUS one byte for a separator for each read name, including a terminal separator. I don't think that's stated anywhere in the spec, so if correct, it would be useful to state explicitly.

jkbonfield:
This includes all visible read name bytes plus 1 termination byte per name (e.g. '\0').

Fixes #802

### DIFF
--- a/CRAMcodecs.tex
+++ b/CRAMcodecs.tex
@@ -2450,11 +2450,18 @@ can lead to many separate data streams.  The name tokeniser codec is
 a format within a format, as the multiple byte streams $B_{pos,type}$
 are serialised into a single byte stream.
 
-The serialised data stream starts with two unsigned little endiand 32-bit
-integers holding the total size of uncompressed name buffer and the
-number of read names.  This is followed the array elements
-themselves.
+The serialised data stream starts with two unsigned little endian
+32-bit integers holding the total size of uncompressed name buffer and
+the number of read names, and a flag byte indicating whether data is
+compressed with arithmetic coding or rANS Nx16.
+Note the uncompressed size is calculated as the sum of
+all name lengths including a termination byte per name (e.g. the nul
+char).  This is irrespective of whether the implementation produces
+data in this form or whether it returns separate name and name-length
+arrays.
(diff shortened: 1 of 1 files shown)


CRAMcodecs.tex
```diff
@@ -2450,10 +2450,14 @@ \section{Name tokenisation codec}
...
 are serialised into a single byte stream.
 
-The serialised data stream starts with two unsigned little endiand 32-bit
-integers holding the total size of uncompressed name buffer and the
-number of read names.  This is followed the array elements
-themselves.
+The serialised data stream starts with two unsigned little endian
+32-bit integers holding the total size of uncompressed name buffer and
+the number of read names.  This is followed the array elements
+themselves.  Note the uncompressed size is calculated as the sum of
```
zaeleus, line 2456: > The serialised data stream starts with two unsigned little endian 32-bit integers... This is followed the array elements themselves.

This is unrelated to the length calculation, but note there is also a 1 byte flag between the 2 integers and the data stream:

| Bytes | Type   | Name            |
| ----: | ------ | --------------- |
|     4 | uint32 | `uncomp_length` |
|     4 | uint32 | `num_reads`     |
|     1 | uint8  | `use_arith`     |

jkbonfield: Agreed.  It also had a poor and nebulous "array elements" term which I expanded on.

PR #804, opened 2025-01-07
Copy the tag type text from SAM spec to SAMtags (PR #804)

### LINKED ISSUE #798: Somewhat unclear references to types in SAMtags
"Optional fields are usually displayed as TAG:TYPE:VALUE; the type may be one of A (character), B (general
array), f (real number), H (hexadecimal array), i (integer), or Z (string)."

however there is C,I,S, and the notion that B is combined with the types is not immediately obvious from the sentence. somewhat unclear what H is also.

digging deeper into cross references from BAM/CRAM might reveal more clarity, but perhaps even explicitly linking to those docs from SAMtags could help

jkbonfield:
SAM section 1.5 clearly defines the standard SAM tag types along with the expanded codes used in the B byte-array type.  This text has been copied into the SAMtags document to remove a rather woolly definition there.  The text describing lower-case tags has been removed, because this was already discussed in further detail at the end of the SAMtags document (and is itself somewhat woolly when it comes to half-upper half-lower combinations due to the addition of draft tags).

Fixes #798

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -42,17 +42,41 @@ conventions around creating new tags.
 
 \section{Standard tags}
 
+All optional fields follow the {\tt TAG:TYPE:VALUE} format
+where {\tt TAG} is a two-character string that matches {\tt /[A-Za-z][A-Za-z0-9]/}.
+In an optional field, {\tt TYPE} is a single case-sensitive letter which
+defines the format of {\tt VALUE}:
+\begin{center}\small
+\begin{tabular}{cll}
+\hline
+{\bf Type} & {\bf Regexp matching {\tt VALUE}} & {\bf Description} \\
+\hline
+A & {\tt [!-\char126]} & Printable character \\
+i & {\tt [-+]?[0-9]+} & Signed integer\footnotemark\\
+f & {\tt [-+]?[0-9]*\char92.?[0-9]+([eE][-+]?[0-9]+)?} & Single-precision floating number \\
+Z & {\tt [\,\,\,!-\char126]*} & Printable string, including space\\
+H & {\tt ([0-9A-F][0-9A-F])*} & Byte array in the Hex format\footnotemark\\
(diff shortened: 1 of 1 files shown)


zaeleus:
Why not link to § 1.5 "[The alignment section: optional fields](https://samtools.github.io/hts-specs/SAMv1.pdf#subsection.1.5)", rather than duplicating the serialization format here? The introduction notes that this is a companion document, which associates the two together anyway.

Alternatively, you could replace the type codes with generalized types, e.g., `i` => `integer`, `Z` => `string`, `B,I` => `uint32_t[]`, etc. This would remove the need to describe the format.

jmarshall:
I agree with @zaeleus that it'd be better to have a less woolly summary of the types that are actually used in the table and a pointer to _SAMv1.pdf_ “for the complete details” rather than duplicate the complete details here.

I don't think we should use informal types such as `integer`, `string`, `uint32_t[]`. The reason the types are listed in the quick-reference table is that these fields need to be of **exactly the specified type**, so we do need to be precise.

jkbonfield:
If this is woolly then the SAM specification is woolly too as this is literally just replacing the poor text originally here with the SAM specification text.  Or maybe I just misunderstand and you're saying this is better than the old text?  I simply tightened up what was there before as it had very vague terms ("character" could permit UTF-8, and "hexadecimal array" could imply comma-separated lists of hex values). 

We do need to be careful about sizing though. SAM is *not* specifically tied to 32-bit (signed or otherwise), and we've exploited this in the past when we've gone from 32-bit coordinates to 64-bit coordinates.  The BAM serialisation has issues, but the SAM encoding does not and we shouldn't needlessly paint ourselves into corners by over specifying as it restricts migration to better encodings.  (Sadly CRAM made those same mistakes though and I regret not fixing them for CRAM 3.0)

Arguably however, and maybe contentiously, if we wish to only have one canonical source for the meaning of the type fields then it perhaps should be SAMtags.  The reason it was split in the first place was that the tag encoding is shared between SAM, BAM and CRAM (and now other things like extended FASTQs and GFA).  Hence keeping it in its own document.

Issue #581, opened 2021-07-12
Standardize htsget:// scheme on spec


brainstorm:
Briefly mentioned on https://github.com/igvteam/igv/pull/850#issuecomment-707727642, discussed on some of the htsget meetings, GA4GH's Slack `#fasp` channel (as of today) and possibly on other issues in this repo (please refer to them if so), it is unclear if the `htsget://` scheme should officially appear on the spec?

From a client perspective it'd be advantageous to discern the protocol right away but this might have other (unforeseen?) side effects?

/cc @jb-adams @ohofmann @jrobinso
/cc @CastilloDel @mmalenic

andrewpatto:
Hi, this was discussed today in the FASP call - and we thought we'd put some notes for further discussion:

The call agreed there was probably a need for a `htsget://` scheme and that DRS could serve these up
i.e there does not appear to be a natural mechanism in DRS to expose the richness of the htsget protocol by returning (for example) `https://` URLs of htsget endpoint - so something calling it out as a special protocol scheme seems needed

If DRS was to serve up some form of `htsget://` url, we think the client would need it (the url) to contain
- the base URL (hostname?) of a htsget service
- the ID of the object in question

Is there other info the url would need to contain? Does the `htsget://` url also contain paths to /reads or /variants or is the concept of it specifying the htsget [base] a good one?

(there was then a discussion about the naming of the /reads and /variants endpoints - and if these are custom then how  could they be discovered - at the moment even the  service-info endpoints are not at a 'known' spot if custom paths are used)

jb-adams:
Looping in @mlin @jmarshall @daviesrob . Please add in anyone else who should be on this thread

jrobinso:
I'm not really in this loop, and am a bit confused on the status of htsget:// URIs.  These are already used in the 2.24.1 (and probably earlier) release of the htsjdk, but I am inferring from the existence of this thread that they aren't standardized yet.    The htsjdk recognizes the following as an endpoint to an alignment file, by recognizes I mean accepts

```
htsget://htsget.ga4gh.org/reads/giab.NA12878.NIST7086.1
```

jmarshall:
[...]

PR #808, opened 2025-02-03
Add encryptionScheme parameter to htsget

brainstorm:
This is a non-breaking, optional change for htsget.

This is already supported as `experimental` feature in htsget-rs since https://github.com/umccr/htsget-rs/pull/298. It specifically allows for Crypt4GH requests for each underlying bioinformatics format.

There's also the corresponding work in progress in htsget-compliance testsuite to support this extra parameter/schemes in https://github.com/ga4gh/htsget-compliance/pull/11.

/cc @ohofmann @mmalenic @mlin @jppay

### DIFF
--- a/htsget.md
+++ b/htsget.md
@@ -212,7 +212,7 @@ _optional_
 </td><td>
 Request data encrypted with a particular encryption scheme.
 
-If `encryptionScheme` is specified a 4 letter code for the encryption standard MUST be returned. When `c4gh` is specified, Crypt4GH payload for a particular object will be returned (if available).
+If `encryptionScheme` is specified a 4 letter value for the encryption standard MUST be provided. When `c4gh` is specified, Crypt4GH payload for a particular object will be returned (if available).
 
 The server SHOULD reply with a `NotFound` error if the requested (encrypted) id does not exist.
 </td></tr>
--- a/htsget.md
+++ b/htsget.md
@@ -19,6 +19,7 @@ This data retrieval API bridges from existing genomics bulk data transfers to a 
 
 * Incumbent data formats (BAM, CRAM) are preferred initially, with a future path to others.
 * Multiple server implementations are supported, including those that do format transcoding on the fly, and those that return essentially unaltered filesystem data.
+* Multiple encryption schemes on the files are supported, including but not limited to [Crypt4GH](https://samtools.github.io/hts-specs/crypt4gh.pdf).
 * Multiple use cases are supported, including access to small subsets of genomic data (e.g. for browsing a given region) and to full genomes (e.g. for calling variants).
 * Clients can provide hints of the information to be retrieved; servers can respond with more information than requested but not less.
 * We use the following conventions:
@@ -203,6 +204,20 @@ Request the SAM/CRAM/VCF headers only. The ticket response represents read or va
 The server SHOULD respond with an `InvalidInput` error if any other htsget query parameters other than `format` are specified at the same time as `class=header`.
 </td></tr>
 </table>
+</td></tr>
+<tr markdown="block"><td>
+
+`encryptionScheme`
+_optional_
+</td><td>
+
+Request data encrypted with a particular encryption scheme.
+
+
+
+If `encryptionScheme` is specified a 4 letter code for the encryption standard MUST be returned. When `c4gh` is specified, Crypt4GH payload for a particular object will be returned (if available).
+
+The server SHOULD reply with a `NotFound` error if the requested reference does not exist.
 </td></tr>
 <tr markdown="block"><td>
 
@@ -760,6 +775,11 @@ Example listing of htsget reads API and variants API registrations from a servic
 # Version history
 
 This appendix lists the significant functionality introduced and changes made in each published version of the htsget protocol.
+
+## 1.3.1 (February 2025)
+{:.no_toc}
+
+Added an `encryptionScheme` query parameter for requesting encrypted payload of any object (any underlying bioinformatics format).
 
 ## 1.3.0 (March 2021)
 {:.no_toc}
--- a/htsget.md
+++ b/htsget.md
@@ -210,14 +210,11 @@ The server SHOULD respond with an `InvalidInput` error if any other htsget query
 `encryptionScheme`
 _optional_
 </td><td>
-
 Request data encrypted with a particular encryption scheme.
-
-
 
 If `encryptionScheme` is specified a 4 letter code for the encryption standard MUST be returned. When `c4gh` is specified, Crypt4GH payload for a particular object will be returned (if available).
 
-The server SHOULD reply with a `NotFound` error if the requested reference does not exist.
+The server SHOULD reply with a `NotFound` error if the requested (encrypted) id does not exist.
 </td></tr>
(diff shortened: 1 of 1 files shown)


daviesrob:
How exactly are you implementing crypt4gh in `htsget-rs`?

There's basically two ways you could do it - either encrypt each segment linked in the htsget response separately, or have them all join together to make a single encrypted file.  HTSlib has historically supported the latter, and also looks for the magic number so it can tell an encrypted payload apart from a plain one without having to make any changes to the htsget specification.

mmalenic:
> How exactly are you implementing crypt4gh in htsget-rs?

It's implemented by joining segments together into a single encrypted file. It calculates the bytes to return to the client, and then re-encrypts the Crypt4GH header and adds an edit list packet using a configured public/private key pair.

> and also looks for the magic number so it can tell an encrypted payload apart from a plain one without having to make any changes to the htsget specification.

This works, however the `encryptionScheme` parameter is non-breaking and optional, and I think it would clearer and more practical for the client.

The advantage of adding an additional parameter is that it allows the client to directly request encrypted or non-encrypted files without having to inspect the response header type. I think this lines up with the logic of specifying CRAM/BAM or BCF/VCF in the `format` parameter. It also allows the client to request encrypted or non-encrypted data for the same `<id>`.

Similar discussion to https://github.com/samtools/hts-specs/issues/581, where clients can give hints on the data they are expecting to make their logic easier.

brainstorm:
@daviesrob If needed, shall we discuss this over a the 4-weekly htsget APAC call?:

```
GA4GH htsget videoconference taking place at an Asia-Pacific friendly time of 10pm BST / 2pm PDT / 5pm EDT and next day 7am Melbourne AEST.
```

jmarshall:
I agree with @daviesrob that detecting what you got back should be the primary way of selecting how to decode the response.

However it does seem like there would be some use in enabling the client to say it wants encryption and having a way for the client and server to negotiate that. Somewhat unclear that that should be in the initial request URL rather than something negotiated via the ticket… 🤔

jmarshall:
> If `encryptionScheme` is specified a 4 letter code […] When `c4gh` is specified […]

Where do these 4-character codes come from? Are they in common with some other protocol, or have they been invented for this PR? Are letters really in such short supply that this has to be specifically 4 characters?

htsget.md
```diff
@@ -206,6 +207,20 @@ The server SHOULD respond with an `InvalidInput` error if any other htsget query
...
+_optional_
+</td><td>
+
+Request data encrypted with a particular encryption scheme.
+
+
+
+If `encryptionScheme` is specified a 4 letter code for the encryption standard MUST be returned. When `c4gh` is specified, Crypt4GH payload for a particular object will be returned (if available).
+
+The server SHOULD reply with a `NotFound` error if the requested reference does not exist.
```
jmarshall, line 220: This doesn't make sense and appears to have been copy/pasted from the `referenceName` section. Do you mean if the server does not recognise a requested encryption code?

htsget.md
```diff
@@ -206,6 +207,20 @@ The server SHOULD respond with an `InvalidInput` error if any other htsget query
...
 
+`encryptionScheme`
+_optional_
+</td><td>
+
+Request data encrypted with a particular encryption scheme.
+
+
+
+If `encryptionScheme` is specified a 4 letter code for the encryption standard MUST be returned. When `c4gh` is specified, Crypt4GH payload for a particular object will be returned (if available).
```
jmarshall, line 218: Do you mean a 4-character code must be specified as `encryptionScheme`'s **value**? I don't understand why such a code would be being **returned** by the server.

htsget.md
brainstorm: Good catch, I meant (encrypted) object Id, fixed in https://github.com/samtools/hts-specs/pull/808/commits/4e6fa1a36658d81bbf16c86867ca15f7cccfcf28

brainstorm:
> > If `encryptionScheme` is specified a 4 letter code […] When `c4gh` is specified […]
> 
> Where do these 4-character codes come from? Are they in common with some other protocol, or have they been invented for this PR? Are letters really in such short supply that this has to be specifically 4 characters?

I was hesitant to go for free-form string and I borrowed the idea of restricting format names to (at most) a 4 letter enum: BAM, CRAM, VCF, BCF.

Admittedly a lot of encryption schemes might require more than 4 letters to declare, I wish there was a standard that defined this (IETF? W3C? OpenSSL source code?) that enumerates the most common encryption scheme strings. Anyways, very good point, what would you suggest @jmarshall?

htsget.md
brainstorm: Agreed, fixed in https://github.com/samtools/hts-specs/pull/808/commits/016f5957d4626ee4ef99916c98e2d62141c3149e

brainstorm:
@jmarshall:

> I agree with @daviesrob that detecting what you got back is and should be the primary way of selecting how to decode the response.

Why?

htsget.md
jmarshall: If this is about the resource's `<id>`, i.e., what forms the bulk of the path part of the URL, then it should be in the `URL parameters` section (in fact, it seems this has always been missing there!). That would be a separate PR.

If this is about the 4-letter code designating the encryption method, then it should be here and use the same `code` terminology.

jmarshall:
Discussion from the September meeting:

* Let's spell out `encryptionScheme=crypt4gh` rather than abbreviating as `c4gh`.
* Further consider this proposal vs requesting via `Accept` headers (probably this proposal is more specific and bespoke so appropriate in this case) and what interplay there is with adding fields in the ticket to represent what encryption is available.
* Consider also adding an `encryptionScheme=none|unencrypted|plaintext` (pick one!) value for the client to request specifically unencrypted data.
* What should the server return when `encryptionScheme=crypt4gh` is requested but the server only has unencrypted data available? Perhaps a plain NotFound indistinguishable from when it has no data at all for that sample/etc; or a new NotAvailableEncrypted error; or it just returns the plaintext version despite the client's request (almost certainly not).

We also clarified the motivation for this functionality:
1. Clients may want to say “I definitely want the transport used for data coming to me to be crypt4gh”.
2. Provides a way to select which edition to provide if a server has both plain and encrypted copies of the same data.

jmarshall:
Often it'll be the **server** that's motivated to not send unencrypted data, and it can already just return an error instead. But still some utility in this explicitness.

jmarshall:
Discussion from the May meeting:

* Probably no need for `encryptionScheme=none`, just leave the optional parameter off.
* Need to add server's encryption capabilities to the service-info response, e.g., `"encryption": ["crypt4gh"]`.
* Possibly have a way to say "I support encryption for CRAM and VCF but not BAM" but probably unnecessary.
* Use a new `NotAvailableEncrypted` error, as anything revealed could probably be figured out by making additional requests anyway.

jmarshall:
Discussion from the June/July meeting:

* Service-info response should have a way of saying “I will only return cryp4gh-encrypted data, never unencrypted“
* Client should be able to specify its client public key that the server should use for the encryption
* Specify exactly how htsget+crypt4gh works! (i.e. what the interplay of the two ideas of blocks looks like)

PR #813, opened 2025-02-16
feat: add tags related to optical duplicates.

yfarjoun:
feat: add tags related to optical duplicates.

Picard has long been able to add Sam Tags that provide information about duplicate templates: their type (Library or Sequencing), the size of the sets they belong to, and the query name of the representative template within the set that is _not_ a duplicate. As I look to use these tags in another repository (fgbio), I want to ensure they are generally accepted and well-defined.

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -77,6 +77,9 @@ or
   {\tt CS} & Z & Color read sequence \\
   {\tt CT} & Z & Complete read annotation tag, used for consensus annotation dummy features \\
   {\tt CY} & Z & Phred quality of the cellular barcode sequence in the {\tt CR} tag \\
+  {\tt DI} & Z & Duplicate Identity, for identifying the queryname that this read is a duplicate of\\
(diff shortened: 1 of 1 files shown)


yfarjoun:
I wasn't sure if the 1.1 section was required for all tags or only for those for which there's something interesting to say....but OK... :-)

SAMtags.tex
```diff
@@ -158,6 +161,21 @@ \subsection{Additional Template and Mapping data}
...
+\item[DS:i:\tagvalue{count}]
+Size of the duplicate set that the template is part of.
+
+\item[DT:Z:\tagvalue{str}]
+(For duplicate templates) Either LB or SQ indicating the source of the duplication.
+Use LB if the duplication occurred during library-construction (e.g., via PCR).
+Use SQ if the duplication occurred during sequencing (e.g., due to imaging error, aka "optical duplicates", or due to
+over-aggressive bridge-amp).
+
+\item[GL:f:\tagvalue{score}]
```
jmarshall, line 177: What is GL? Was this line accidentally committed?

PR #816, opened 2025-02-19
Movement of refget from hts-specs to GA4GH refget

andrewyatz:
As part of approval for refget sequence collections, we have created a new repository for refget under the [GA4GH GitHub space](https://github.com/ga4gh/refget). This PR denotes the current supported location is in the new location but retains the v2.0.0 spec here for reference.

### DIFF
--- a/refget.md
+++ b/refget.md
@@ -4,6 +4,8 @@ title: refget protocol
 suppress_footer: true
 ---
 
+<strong>Refget's specification is no longer hosted here and has moved to the central <a href="https://ga4gh.github.io/refget/sequences/">GA4GH Refget repository</a>. Please update your bookmarks</strong>
+
 # Refget API Specification v2.0.0
 {:.no_toc}
 


jkbonfield:
We should probably remove it from the README file (or point it elsewhere) and similarly the http://samtools.github.io/hts-specs/ site too.

jmarshall:
If refget wishes to move, then it should do so properly IMHO and remove the soon-to-be-outdated text. I have activated [jekyll-redirect-from](https://github.com/jekyll/jekyll-redirect-from) in this repository, so (after rebasing on current **master**) you can replace all of _refget.md_ with simply

```
---
redirect_to: https://ga4gh.github.io/refget/sequences/
---
```

You should also either remove the refget paragraph in _README.md_ or (better) change the link and add text saying that it is maintained elsewhere; remove the link in _index.md_; delete _pub/refget-openapi.yaml_. You may also wish to remove the refget entry from _MAINTAINERS.md_.

PR #819, opened 2025-03-24
Fix the rANS 4x8 order-1 frequency encoding example.

### LINKED ISSUE #817: cram-codecs: rANS 4x8 order-1 frequency table example does not follow description
This is in regard to _[CRAM codec specification (version 3.1)](https://github.com/samtools/hts-specs/blob/00a2ed45e75679c0d2550bf553f40c022286f2d7/CRAMcodecs.tex)_ (2023-03-15).

The example in § 2.1 "[Frequency table: Order-1 encoding](https://samtools.github.io/hts-specs/CRAMcodecs.pdf#subsection.2.1)" splits the following input

```
abracadabraabracadabraabracadabraabracadabr
```

into

```
abracadabra abracadabra abracadabra abracadabr
```

It's not obvious why the example is split this way. The note directs you to § 2.2.2 "[rANS entropy encoding: Interleaving](https://samtools.github.io/hts-specs/CRAMcodecs.pdf#subsection.2.2)", which says

>  We therefore split the input data into 4 approximately equal sized fragments<sup>5</sup> starting at 0, ⌊_len_/4⌋, ⌊_len_/4⌋ × 2 and ⌊_len_/4⌋ × 3.

I.e.,

```
abracadabr aabracadab raabracada braabracad abr
```

This method invalidates the observed frequencies for the example and footnote 5.

> <sup>5</sup>This was why the `\0` → `a` context in the example above had a frequency of 4 instead of 1.

jkbonfield:
The text erroneously divided 43 into 11,11,11,10 when it would be 10,10,10,13.

Instead of 4x11 minus 1 I chose a new example of 4x11 plus 1, so the word boundaries ended at the same points.

Fixes #817

### DIFF
--- a/CRAMcodecs.tex
+++ b/CRAMcodecs.tex
@@ -412,11 +412,10 @@ data stream (in fact for 4 equally spaced starting points, see
 the starting context for each interleaved rANS state and so need to
 consider this in our frequency table.
 
-Consider \texttt{abracadabraabracadabraabracadabraabracadabr} as
-example input.  Note for the last ``a'' was omitted from this string
-in order to demonstrate how the method works when the data is not a
-multiple of 4 long.  This can be broken into 4 approximate equal
-portions \texttt{abracadabra abracadabra abracadabra abracadabr}.  We
+Consider \texttt{abracadabraabracadabraabracadabraabracadabrad} as
+example input.  Note for the additional trailing ``d'' giving us 45
+characters instead of 44.  This can be broken into 4 approximate equal
+portions \texttt{abracadabra abracadabra abracadabra abracadabrad}.  We
 operate one independent rANS stream per portion, providing us the
 opportunity to exploit CPU data parallelism.
 
@@ -432,7 +431,7 @@ Context & Symbol & Frequency\\
 a & a & 3 \\
   & b & 8 \\
   & c & 4 \\
-  & d & 4 \\
+  & d & 5 \\
 \hline
 b & r & 8 \\
 \hline
@@ -440,7 +439,7 @@ c & a & 4 \\
 \hline
 d & a & 4 \\
 \hline
-r & a & 7 \\
+r & a & 8 \\
 \hline
 \end{tabular}
 \end{minipage}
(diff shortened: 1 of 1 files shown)


CRAMcodecs.tex
```diff
@@ -453,10 +452,10 @@ \subsubsection*{Order-1 encoding}
...
 \textbackslash0 & a & 4095 \\
 \hline
-a & a &  646 \\
-  & b & 1725 \\
-  & c &  862 \\
-  & d &  862 \\
+a & a &  614 \\
+  & b & 1638 \\
+  & c &  819 \\
+  & d & 1024 \\
```
zaeleus, line 458:
```suggestion
a & a &  614 \\
  & b & 1639 \\
  & c &  819 \\
  & d & 1023 \\
```

I know the spec doesn't define the adjustment to the normalized sum,[^1] but consider being consistent with the (implicit) method used in the order-0 example, i.e., the symbol with the highest frequency is adjusted.

[^1]: § 2.1 "[Frequency table](https://samtools.github.io/hts-specs/CRAMcodecs.pdf#subsection.2.1)" (2023-03-15): "Given rounding differences when renormalising to a fixed sum, it is up to the encoder to decide how to distribute any remainder or remove excess frequencies."

CRAMcodecs.tex
```diff
@@ -491,10 +490,10 @@ \subsubsection*{Order-1 encoding}
...
 
 0x61                 # `a' context
-0x61      0x82 0x86  # a            <646>
-0x62 0x02 0x86 0xbd  # b <+2: c,d> <1725>
-          0x83 0x5e  # c (implicit) <862>
-          0x83 0x5e  # d (implicit) <862>
+0x61      0x82 0x66  # a             <614>
+0x62 0x02 0x86 0x66  # b <+2: c,d>  <1638>
+          0x83 0x33  # c (implicit)  <819>
+          0x84 0x00  # d (implicit) <1024>
```
zaeleus, line 496:
```suggestion
0x61      0x82 0x66  # a             <614>
0x62 0x02 0x86 0x67  # b <+2: c,d>  <1639>
          0x83 0x33  # c (implicit)  <819>
          0x83 0xff  # d (implicit) <1023>
```

PR #820, opened 2025-03-25
Rework the CRAM bit encoding example and clarify text

### LINKED ISSUE #812: cram: Define bit stream value
This is in regard to [_CRAM format specification (version 3.1)_](https://github.com/samtools/hts-specs/blob/4127441a0d18cba74ea6ef7b8852a65274408207/CRAMv3.tex) (2024-09-04).

§ 2.2 "[Writing bits to a bit stream](https://samtools.github.io/hts-specs/CRAMv3.pdf#subsection.2.2)" does not define what a _value_ is. This is required to know how to interpret the bits and the maximum number of bits that can be written to the stream from said value.

A single value type can be generally defined for the bit stream (e.g., `int32_t` >= 0), or the encodings that use the bit stream (e.g., [beta](https://samtools.github.io/hts-specs/CRAMv3.pdf#subsection.13.5)) can describe their own appropriate value types.

jkbonfield:
Given it's working in bits, the example is much clearer if we describe the bits instead of hex, especially distinguishing bits set/unset from bits-yet-to-use.

I also reworked the note at the end of the section as it was quite hard to follow.  I gave it real examples of BETA and HUFFMAN to clarify what is meant by the decoder needing to know the number of bits to consume.

Fixes #812.

### DIFF
--- a/CRAMv3.tex
+++ b/CRAMv3.tex
@@ -124,13 +124,13 @@ An array of any logical data type: array\texttt{<}type\texttt{>}
 % \textbf{Array}   & An array of any logical data type: \texttt{<}type\texttt{>}[ ] \\
 % \end{tabular}
 
-\subsection{\textbf{Writing bits to a bit stream}}
+\subsection{\textbf{Reading and writing bits in a bit stream}}
 
-A bit stream consists of a sequence of 1s and 0s. The bits are written most significant 
-bit first where new bits are stacked to the right and full bytes on the left are 
-written out. In a bit stream the last byte will be incomplete if less than 8 bits 
-have been written to it. In this case the bits in the last byte are shifted to 
-the left.
+The CORE block supports bit-based encoding methods.
+A bit stream consists of a sequence of 1s and 0s.
+The bits are written most significant bit first where new bits are stacked to the right and full bytes on the left are written out.
+In a bit stream the last byte will be incomplete if less than 8 bits have been written to it.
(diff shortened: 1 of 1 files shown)


CRAMv3.tex
```diff
@@ -166,26 +166,29 @@ \subsubsection*{Example of writing to bit stream}
...
-from a bit stream the number of bits must be known in advance. In case of prefix 
-codes (e.g. Huffman) all possible bit combinations are either known in advance 
-or it is possible to calculate how many bits will follow based on the first few 
-bits. Alternatively, two codes can be combined, where the first contains the number 
-of bits to read. 
+When reading and writing to a bit stream our numeric values are
+typically held in a byte oriented data type, such as an 8-bit or
+32-bit integer.
+The bit stream itself does not explicitly store the number of bits
+per value, and it will vary by context, so we must know this by other means.
```
zaeleus, line 177: This answers #812.

The examples that follow are useful but would be better defined with their respective codecs. E.g., § 13.5 "[Beta encoding: codec ID 6](https://samtools.github.io/hts-specs/CRAMv3.pdf#subsection.13.5)" should define the resulting value from the bit stream is a 32-bit integer, meaning the _length_ value is limited to 32 (or 31 if the value is always supposed be non-negative?). This is the context required to meaningfully read from the bit stream.

zaeleus (approved):

PR #841, opened 2025-08-12
Minor CRAM notation suggestion: Use IEC-style MiB notation for powers of two, rather than MB

jmarshall:
CRAM parts split out from PR #839 — see the conversation there:

* _CRAMv2.1.tex_ and _CRAMv3.tex_ correctly distinguish `MB` (megabytes, **though could use `MiB`**) and `kb`/`Mb` (kilobases and megabases, correctly decimal, and the correct abbreviation for “bases” in a bioinformatics context).

### DIFF
--- a/CRAMv2.1.tex
+++ b/CRAMv2.1.tex
@@ -1668,11 +1668,11 @@ $\bullet$ Streaming is more convenient for small containers
 
 $\bullet$ Applications typically buffer containers into memory
 
-We recommend 1MB containers. They are small enough to provide good random access 
+We recommend 1MiB containers. They are small enough to provide good random access
 and streaming performance while being large enough to provide good compression. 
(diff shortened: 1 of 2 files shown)


jkbonfield:
Thank you for the MB vs MiB changes.

I added an extra commit to completely rewrite the worked examples of container sizes, as they were wrong in the maths and plain misleading anyway as technology has changed with quantised quality values in NovaSeq and unquantised many-more quality values in ONT.  It's best to just stick to the basic principles and leave the specifics unsaid.

I realise now that this means the MB vs MiB change is completely invisible as it's all removed again!  However I don't see that as necessarily a reason for removing the commit as it does accurately reflect the history of edits.

Back to @jmarshall I think to review the wording.  Thanks

Issue #732, opened 2023-06-27
sam-tags: `MM` field value pattern does not allow ambiguity codes


zaeleus:
This is in regard to _Sequence Alignment/Map Optional Fields Specification_ (2022-08-17).

The base modifications (`MM`) field allows modifications to be either short codes or an ChEBI ID. Short codes are constrained to `[a-z]+` (i.e., lowercase letters) but the table of "standard common types" lists ambiguity codes that do not match this (i.e., uppercase letters).

> | Unmodified base | Code | Abbreviation | Name                      | ChEBI |
> | --------------- | ---- | ------------ | ------------------------- | ----- |
> | C               | C    |              | Ambiguity code; any C mod |       |
> | T               | T    |              | Ambiguity code; any T mod |       |
> | U               | U    |              | Ambiguity code; any U mod |       |
> | A               | A    |              | Ambiguity code; any A mod |       |
> | G               | G    |              | Ambiguity code; any G mod |       |
> | N               | N    |              | Ambiguity code; any mod   |       |

jkbonfield:
The short codes are modified bases, so "m" and "h" being 5mC and 5hmC.  It doesn't make any sense to have a base modification from nucleotide to ambiguity code, so I'm not sure I follow this.

We don't support ambiguity codes in the unmodified base component, so we couldn't do `MM:Z:Y+h,4;` for example as it wouldn't may sense.  "N" covers this case anyway with the different counting regime.

jkbonfield:
Oh wow I'd totally forgotten about that!

Yes, the regexp should be `([a-zACGTUN]+|[0-9]+)` for the code portion.  Good spot.

PR #859, opened 2026-07-07
Relax MM tag regexp

### LINKED ISSUE #854: Update MM tag regex to include upper case "Ambiguity codes"
The regex for the MM tag (section 1.7):
```
  MM:Z:([ACGTUN][-+]([a-z]+|[0-9]+)[.?]?(,[0-9]+)*;)*
```
  specifies `([a-z]+|[0-9]+)` for the modification codes which is lower case only but this does not therefore parse the ambiguity codes like "C - Ambiguity code; any C mod"

the spec may want to update it to
```
([a-z]+|[A-Z]|[0-9]+)
```

jkbonfield:
We permit e.g. "C" as a base modification code to mean "any unspecified C modification".  This means C+C,1,2,3 or even C+mhC,1,2,3 to show 5mC and 5hmC specifically, and a joint "any C including m, h plus other unspecified ones".

Fixes #854

### DIFF
--- a/SAMtags.tex
+++ b/SAMtags.tex
@@ -488,7 +488,7 @@ Each modified base prediction listed also has a quality value associated with it
 Given the unmodified base already has a phred likelihood, this base modification quality should be interpreted as the likelihood of this modification being correct given an assumption the original call is correct.
 
 \begin{description}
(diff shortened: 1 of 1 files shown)


zaeleus:
This also fixes https://github.com/samtools/hts-specs/issues/732, which is the same issue and resolution.

jkbonfield:
Thanks!

